PR #2707, opened 2022-01-23
feat(connect-form): add username/password auth options COMPASS-5435

Anemy:
COMPASS-5435

This adds auth inputs for the Username/Password authentication tab.
Drive by cleanup - removed the `directConnection` setting handler and made the toggle use the general search param one we've added.
How do y'all feel on the behavior around setting a username empty or setting a password when there's no username?

Here's how it looks:

https://user-images.githubusercontent.com/1791149/150684239-d3e4eb5f-ed4e-4a27-a395-2de64bc5f4d2.mp4


packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-tab.tsx
```diff
@@ -69,27 +81,65 @@ const contentStyles = css({
...
 
   const optionSelected = useCallback(
     (event: ChangeEvent<HTMLInputElement>) => {
       event.preventDefault();
-      updateConnectionFormField({
-        type: 'update-search-param',
-        currentKey: 'authMechanism',
-        value: event.target.value,
+
+      if (event.target.value === 'AUTH_NONE') {
```
mcasimir, line 134: just a super silly thing, probably this can be done with less code:

``` js
return updateConnectionFormField({
   type: 'update-auth-mechanism',
   authMechanism: event.target.value === 'AUTH_NONE' ? null : event.target.value as AuthMechanism,
});
```

packages/connect-form/src/utils/authentication-handler.ts
```diff
@@ -0,0 +1,162 @@
...
+  connectionOptions: ConnectionOptions;
+  errors?: ConnectionFormError[];
+} {
+  try {
+    const updatedConnectionString = connectionStringUrl.clone();
+
+    updatedConnectionString.username = action.username;
+
+    // This will throw if the connection string is invalid
+    new ConnectionStringUrl(updatedConnectionString.toString());
```
mcasimir, line 91: probably better if we extract the `tryParseConnectionString` from the other file instead of repeating the code. Are there other places where this is necessary where we could use it?

packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-default.tsx
```diff
@@ -1,9 +1,181 @@
...
+  const selectedAuthTab =
+    defaultAuthMechanismOptions.find(
+      ({ value }) => value === selectedAuthMechanism
+    ) ?? defaultAuthMechanismOptions[0];
+
+  const onAuthMechanismSelected = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      event.preventDefault();
+      updateConnectionFormField({
+        type: 'update-search-param',
```
mcasimir, line 71: Currently this action also sets empty values, ie. it would set `&authMechanism=&` for a null value, could we also add `updatedSearchParams.delete(action.currentKey);` in case the value is empty here https://github.com/mongodb-js/compass/blob/main/packages/connect-form/src/hooks/use-connect-form.ts#L386-L393?

Anemy: I see how that helper could be useful here, but I think we want to allow empty values to be set since we use this in the advanced tab to help populate the advanced options. In that case setting something to empty is useful as part of the process of entering those options:
Screen Shot 2022-01-24 at 1 43 41 PM

mcasimir: I'm not sure that is so useful though, there is no search param that makes sense as empty, and I don't know why someone would prefer to populate the value in the connection string rather than in the form after they selected the key there. I believe is a good feature if we produce a cleaner connection string without random empty search params. 

Do you have an alternative to at least delete `authMechanism` and the other params if they are empty?

packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-default.tsx
```diff
@@ -1,9 +1,181 @@
...
+function AuthenticationDefault({
+  errors,
+  connectionStringUrl,
+  updateConnectionFormField,
+}: {
+  connectionStringUrl: ConnectionStringUrl;
+  errors: ConnectionFormError[];
+  updateConnectionFormField: UpdateConnectionFormField;
+}): React.ReactElement {
+  const password = connectionStringUrl.password;
```
mcasimir, line 57: Is this relevant here? https://jira.mongodb.org/browse/COMPASS-5097

can we at least add a test or 2 with the weird password there?

packages/connect-form/src/utils/authentication-handler.ts
Anemy: Updated to share one 👍

packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-default.tsx
Anemy: On this page the `authMechanism` will always be set since it's a radio with 3 options and unsetting isn't one.
Yup we can use the `delete-search-param` action.
I see what you mean. I'm cool with either in that case. Not sure this is the pr for that change.

packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-default.tsx
Anemy: Could we do that work when we do that ticket? This work is definitely related, but since we have a ticket for it, that might be a better place to make a pr, test all of the cases and make those changes.

mcasimir (approved):

PR #2710, opened 2022-01-24
style(connect-form): fix padding with overflow-y: scroll

mcasimir:
Remove the additional space added to the right by the `overflow-y: scroll`

Main:
Screenshot 2022-01-24 at 13 15 37
Screenshot 2022-01-24 at 13 16 11

PR:
Screenshot 2022-01-24 at 16 33 56
Screenshot 2022-01-24 at 16 34 02

### DIFF
--- a/packages/connect-form/src/components/connect-form.tsx
+++ b/packages/connect-form/src/components/connect-form.tsx
@@ -48,10 +48,18 @@ const descriptionStyles = css({
   marginTop: spacing[2],
 });
 
+const scrollbarWidth = spacing[2];
 const formContentContainerStyles = css({
   padding: spacing[4],
-  overflow: 'scroll',
+  paddingRight: spacing[4] - scrollbarWidth,
+  overflowY: 'scroll',
   position: 'relative',
+  scrollPadding: 0,
+  scrollbarGutter: 'unset',
+  scrollbarWidth: 'none',
+  '::-webkit-scrollbar': {
+    width: scrollbarWidth,
+  },
 });
 
 const formFooterStyles = css({
--- a/packages/connect-form/src/components/connect-form.tsx
+++ b/packages/connect-form/src/components/connect-form.tsx
@@ -48,18 +48,10 @@ const descriptionStyles = css({
   marginTop: spacing[2],
 });
 
-const scrollbarWidth = spacing[2];
 const formContentContainerStyles = css({
   padding: spacing[4],
-  paddingRight: spacing[4] - scrollbarWidth,
-  overflowY: 'scroll',
+  overflowY: 'auto',
   position: 'relative',
-  scrollPadding: 0,
-  scrollbarGutter: 'unset',
-  scrollbarWidth: 'none',
-  '::-webkit-scrollbar': {
-    width: scrollbarWidth,
-  },
 });
 
 const formFooterStyles = css({


packages/connect-form/src/components/connect-form.tsx
```diff
@@ -48,10 +48,18 @@ const descriptionStyles = css({
...
+const scrollbarWidth = spacing[2];
 const formContentContainerStyles = css({
   padding: spacing[4],
-  overflow: 'scroll',
+  paddingRight: spacing[4] - scrollbarWidth,
+  overflowY: 'scroll',
   position: 'relative',
+  scrollPadding: 0,
+  scrollbarGutter: 'unset',
+  scrollbarWidth: 'none',
```
gribnoysup, line 59: These are not applied in the app due to the electron version we use:

image

```suggestion
```

packages/connect-form/src/components/connect-form.tsx
```diff
@@ -48,10 +48,18 @@ const descriptionStyles = css({
...
 });
 
+const scrollbarWidth = spacing[2];
 const formContentContainerStyles = css({
   padding: spacing[4],
-  overflow: 'scroll',
+  paddingRight: spacing[4] - scrollbarWidth,
+  overflowY: 'scroll',
   position: 'relative',
+  scrollPadding: 0,
```
gribnoysup, line 57: This is not scrollbar padding, it's a padding that browser considers a viewbox for the scollable content when scroll snap is used (it also almost always 0 by default anyway), we can probably remove that, unless I'm missing something
```suggestion
```

packages/connect-form/src/components/connect-form.tsx
```diff
@@ -48,10 +48,18 @@ const descriptionStyles = css({
   marginTop: spacing[2],
 });
 
+const scrollbarWidth = spacing[2];
 const formContentContainerStyles = css({
   padding: spacing[4],
-  overflow: 'scroll',
+  paddingRight: spacing[4] - scrollbarWidth,
+  overflowY: 'scroll',
```
gribnoysup, line 55: Maybe we could leave it at `auto` and avoid overriding the scrollbar size and paddings? This seems like a pretty standard behavior where the scrollbar will appear when needed and move the content to the side a bit when appearing

```suggestion
  overflowY: 'auto',
```

packages/connect-form/src/components/connect-form.tsx
```diff
@@ -48,10 +48,18 @@ const descriptionStyles = css({
...
   position: 'relative',
+  scrollPadding: 0,
+  scrollbarGutter: 'unset',
+  scrollbarWidth: 'none',
+  '::-webkit-scrollbar': {
+    width: scrollbarWidth,
+  },
```
gribnoysup, line 62: We now have scrollbars with two different widths on the screen at the same time which looks a bit odd, I think we might want to either set this value globally, or (as I mentioned in the comment above) not change it alltogether. What do you think?

image

mcasimir: I'd love the `overflow-y: auto` if it works!

mcasimir: if not we could just have an <Scroll> component or something shared in compass-component as variable?

packages/connect-form/src/components/connect-form.tsx
mcasimir: this is leftover from some brute-force attempts that ended in a git stash by mistake :)

packages/connect-form/src/components/connect-form.tsx
gribnoysup: I mean, depends on what you mean by "works" 😅

`auto` will only show the scrollbar (and shift the content with additional) when the content overflow the container. Content can overflow (and scrollbar can show up) in like a million places, so I'm not even sure what exactly we can put into a shareable component here if we consider this behavior a problem. A `Box` component that we would need to wrap around almost everything seems like a bit too much work for something like this

mcasimir: Ok `auto` worked, I think that is good enough!

gribnoysup (approved):

PR #2711, opened 2022-01-24
test(e2e): unskip SRV resolution test on Windows

addaleax:
~~Debugging PR because I couldn’t get this to fail locally~~

Now with an actual fix! Mostly just https://github.com/mongodb-js/os-dns-native/pull/3

### DIFF
--- a/packages/compass-e2e-tests/tests/connection.test.js
+++ b/packages/compass-e2e-tests/tests/connection.test.js
@@ -74,11 +74,6 @@ describe('Connection screen', function () {
 
 describe('SRV connectivity', function () {
   it('resolves SRV connection string using OS DNS APIs', async function () {
-    if (process.platform === 'win32') {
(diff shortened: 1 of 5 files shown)


packages/compass-e2e-tests/tests/connection.test.js
```diff
@@ -136,8 +131,5 @@ describe('SRV connectivity', function () {
       error: true,
       wasNativelyLookedUp: false, // Due to fallback to Node.js API after resolution error
     });
-
-    // The connection attempt was aborted early.
-    expect(logs.filter((log) => log.id === 1_000_000_036)).to.have.lengthOf(1);
```
addaleax, line 141: This is because we accept either success or failure, we already adjusted that above but did not remove this check here

lerouxb (approved):

PR #2712, opened 2022-01-24
chore: add telemetry for socks5 COMPASS-5449

addaleax:
Track whether Socks5 was used as a tunnel, and expose the target
topology type in telemetry (useful as a way of checking whether
multi-host connections are being used or not).

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/connections/src/modules/telemetry.ts
+++ b/packages/connections/src/modules/telemetry.ts
@@ -3,7 +3,7 @@ import { createLoggerAndTelemetry } from '@mongodb-js/compass-logging';
 import { isLocalhost, isDigitalOcean, isAtlas } from 'mongodb-build-info';
 import { getCloudInfo } from 'mongodb-cloud-info';
 import ConnectionString from 'mongodb-connection-string-url';
-import { MongoServerError } from 'mongodb';
+import { MongoServerError, MongoClientOptions } from 'mongodb';
 
 const { track, debug } = createLoggerAndTelemetry('COMPASS-CONNECT-UI');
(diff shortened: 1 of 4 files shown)


packages/connections/src/modules/telemetry.ts
```diff
@@ -117,6 +120,7 @@ export function trackNewConnectionEvent(
         server_version: build.version,
         server_arch: host.arch,
         server_os_family: host.os_family,
+        topology_type: dataService.currentTopologyType(),
```
addaleax, line 123: @mmarcon Can you verify that this is okay?

`topology_type: Single | ReplicaSetNoPrimary | ReplicaSetWithPrimary | Sharded | LoadBalanced | Unknown`

In the context of Socks5, this is helpful because it lets us verify that users can use Socks5 or SSH to connect to replica sets, but I figured it might be nicer to expose this the way we get it from the driver rather than using a boolean replicaset/not-replicaset flag?

mmarcon: As long as `Single | ReplicaSetNoPrimary | ReplicaSetWithPrimary | Sharded | LoadBalanced | Unknown` are all strings, that looks good to me. We can easily filter and group by them in Amplitude.

Anemy (approved):
👍

Anemy (approved):

PR #2713, opened 2022-01-24
feat(sidebar): use new favorite connection modal in sidebar COMPASS-5277

Anemy:
COMPASS-5277

This PR makes the `compass-sidebar` package use the favorite connection editing modal from the new `connect-form` package. 
It also adds a focus ring to the button and makes it a button not `a` for some more accessibility.
This also fixes the issue where, after updating a favorite connection's title we wouldn't update the whole page title (moused over in the demo vid).

These changes take the approach of passing the connectionInfo as a prop to the compass-sidebar component. I'm thinking this might simplify things down the line if many places need connectionInfo as it can be used as a prop and isn't something that we set with the instance model updates. Right now, `connectionInfo` is passed once on connect with the `data-service-connected` event, but since at least the favorite of the connection info subject to change, we probably want a cleaner way to update it and notify things that it's updated. I think we could either add the connectionInfo as something available on the instance model, pass it as a prop to components which is what this pr is currently proposing, or maybe something else entirely?

https://user-images.githubusercontent.com/1791149/150876561-fe79f287-44e1-4cae-a63b-a96b09c98516.mp4


gribnoysup (commented):
> if many places need connectionInfo as it can be used as a prop and isn't something that we set with the instance model updates

Looking at the usage of this event, only sidebar is using the `connectionInfo` right now, so I don't think it's something we should be concerned about. More than that, the way we pass whole models from the redux stores to the react components is a big issue in Compass right now that causes unnecessary rerenders almost everywhere in the application affecting app performance. Passing whole models around as props just replicates the same behavior without using redux.

It also feels pretty cumbersome that we have to manually pass this state and a modal component, and wire everything up to make the editing work. What if we need to add this modal to another place in the app? We would need to repeat this whole thing again, potentially degrading rendering performance for parts of the app that don't even care about this state for the reasons mentioned above. What if component interface changes? We would need to introduce changes in multiple places in the app even though nothing really changes, which hints to me that this leaks too much implementation details to the places of the app that shouldn't be aware of it.

I said this a couple of times already and I still believe that it's not a good idea to introduce new approaches on top of existing ones without a strategy, I get that we already did it for compass-home, but I would at least try to avoid mixing approaches in plugins that already exist.

It's ultimately your call if you want to change anything here, I don't want to block this work as it seems like we are trying to wrap this up as soon as possible, but I genuinely don't think it's a good idea that the whole UI tree needs to be aware of the `connectionInfo` state (or any other truly global state, like `instance` model one for example) passed down manually from the very root of the application.

packages/compass-sidebar/src/components/favorite-button/favorite-button.jsx
```diff
@@ -0,0 +1,52 @@
...
+} from '@mongodb-js/compass-components';
+
+import styles from './favorite-button.module.less';
+
+const FavoriteButton = ({
+  favoriteOptions,
+  toggleIsFavoriteModalVisible
+}) => {
+  const isFavorite = !!favoriteOptions;
+  const fontAwesomeName = isFavorite ? 'star' : 'star-o';
```
gribnoysup, line 16: Don't we have this in `compass-components` now?

Anemy: Ah good thinking did just add that. Updated

packages/compass-sidebar/src/components/sidebar-instance/sidebar-instance.jsx
```diff
@@ -1,94 +1,79 @@
...
+        toggleIsFavoriteModalVisible={() => setIsFavoriteModalVisible(
+          !isFavoriteModalVisible
+        )}
+      />
+      <SaveConnectionModal
+        initialFavoriteInfo={connectionInfo.favorite}
+        open={isFavoriteModalVisible}
+        onCancelClicked={() => setIsFavoriteModalVisible(false)}
+        onSaveClicked={(favoriteInfo) => onClickSaveFavorite(favoriteInfo)}
+      />
```
gribnoysup, line 54: So this new form is not behind any feature flag and allows user to choose color, which doesn't change anything if you are not using new connection form (and all the other things that come with it), is this expected?

packages/compass-sidebar/src/components/favorite-button/favorite-button.jsx
```diff
@@ -0,0 +1,57 @@
...
+  toggleIsFavoriteModalVisible
+}) => {
+  const isFavorite = !!favoriteOptions;
+  const hex = favoriteOptions?.color;
+  const style = {
+    backgroundColor: hex || '#243642',
+    color: isFavorite ? '#ffffff' : '#88989a'
+  };
+
+  const focusRingProps = useFocusRing();
```
gribnoysup, line 22: This focus ring was not really built for this use-case and looks pretty weird because of that:

image

You can override focus ring styles by adding some styles on top of that, but it needs to be done using `css`, because `less` styles will not take precedence over (due to the style insertion order). Something like this should do the trick:

```js
css({
    // -1 to compensate for the weird transparent border this button has
    top: -1,
    right: -1,
    bottom: -1,
    left: -1,
    borderRadius: 14,
})
```

gribnoysup (approved):
Thanks for changing sidebar plugin back to using redux for global state. When we inevitably will need to clean up how this plugin passes state to UI later down the road, I'm sure it will be much easier to figure our where global state is coming from as this will only be one place.

Overall looks good to me, left a few small comments for things that caught my eye

packages/compass-sidebar/src/components/favorite-button/favorite-button.jsx
Anemy: Removed the transparent border since it wasn't doing much but mess up these styles 👍 
Screen Shot 2022-01-27 at 10 13 52 AM

packages/compass-sidebar/src/components/sidebar-instance/sidebar-instance.jsx
gribnoysup: Ah, got it, thanks for the context! Yeah, totally if you want to do it that way! FWIW the project seems to be wrapping up pretty soon and it's such a small thing that if you feel like merging it already I think it's totally fine, but I trust your judgement here

PR #2721, opened 2022-01-25
COMPASS-5409 redux store tests 

mabaasit:
~~This PR fixes the tests for `compass-saved-aggregations-queries` plugin.~~

~~The plugin imports `FavoriteQueryCollection` model from `compass-query-history` plugin with throws a compile-time error as it imports electron. And in order to avoid this, we are using env var.~~

~~I tried to use `mocha`'s global setUp and teardown to configure the env var, but again the plugin threw compile-time error because mocha executes all the hooks after it discovers the test files.~~

~~So I ended up creating `test/setup.ts` to set the env var and export cleanUp function from it, which should be called after test run.~~

Tested `readPipelinesFromStorage` in its own module (`@mongodb-js/compass-aggregations`), added `FavoriteQueryStorage` in `@mongodb-js/compass-query-history` that exposes `loadAll`,  `updateAttributes` and `delete`.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-saved-aggregations-queries/src/utlis/queries.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { expect } from 'chai';
+import { cleanUp } from '../test/setup';
+
+import { FavoriteQueryCollection } from '@mongodb-js/compass-query-history';
```
gribnoysup, line 4: Looking at how slow it is to run due to the amount of code we need to import from the external library (two whole plugins, CI seems to be currently failing with a timeout because of that), the amount of code changes in other packages that are required to run this in node env, and the amount of setup we had to do here, I would maybe consider just mocking both packages instead of doing all this pre-setup. It feels to me like we are testing these external deps more than we are testing this package

gribnoysup (commented):
Great to see the tests for this plugin working in node environment now, but I do have some concerns with this particular approach

packages/compass-query-history/src/models/recent-query.js
```diff
@@ -10,7 +10,7 @@ const RecentQuery = Query.extend(storageMixin, {
   namespace: 'RecentQueries',
   storage: {
     backend: 'disk',
-    basepath: remote.app.getPath('userData')
+    basepath: process.env.MONGODB_COMPASS_STORAGE_MIXIN_TEST_BASE_PATH || remote.app.getPath('userData')
```
mcasimir, line 13: this basepath can't be overriden like this, has to be done in storage-mixin

packages/compass-saved-aggregations-queries/src/utlis/queries.spec.ts
mcasimir: We could do that too yeah, to be fair in the new connection form we mock the connection storage (that is tested somewhere else) and tests became more stable.

Even with this approach, I don't think the reason it times out is cause it takes too long. My guess is that it fails due to something that just doesn't return though.

gribnoysup: This suite fails for me locally with a timeout, I did some changes to the logic to time how much the require takes in the most basic test in `index.test.ts` and it's 30s (aggregations plugin is almost 1mb packaged, query history is 600kb, both will be imported and ts-node will need to process them):

image

packages/compass-saved-aggregations-queries/src/index.spec.ts
```diff
@@ -1,3 +1,7 @@
+// this is to avoid BigInteger is not defined error, which is thrown in ipv6/lib/node/bigint.js
+// import chain: compass-query-history -> mongodb-data-service -> ssh-tunnel -> socksv5 -> ipv6 (which is pretty old and deprecated).
+globalThis.BigInteger = undefined;
```
gribnoysup, line 3: I feel like some sort of setup file would be a better place for this instead of just one of the test files

gribnoysup (approved):

packages/compass-aggregations/src/modules/saved-pipeline.js
```diff
@@ -73,7 +73,7 @@ export const getSavedPipelines = () => {
 export const getDirectory = () => {
   const { remote } = require('electron');
   const path = require('path');
-  const userDataDir = remote.app.getPath('userData');
+  const userDataDir = remote ? remote.app.getPath('userData') : process.env.MONGODB_COMPASS_AGGREGATE_TEST_BASE_PATH;
```
mcasimir, line 76: A really silly thing, feel free to merge without changing it, but could we align the name of the variable with the plugin name?

```suggestion
  const userDataDir = remote ? remote.app.getPath('userData') : process.env.MONGODB_COMPASS_AGGREGATIONS_TEST_BASE_PATH;
```

packages/compass-aggregations/src/modules/saved-pipeline.spec.js
```diff
@@ -0,0 +1,63 @@
+/* eslint-disable no-sync */
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+
+import { readPipelinesFromStorage } from './saved-pipeline';
+
+const initialAggregateValue = process.env.MONGODB_COMPASS_AGGREGATE_TEST_BASE_PATH;
```
mcasimir, line 8:
```suggestion
const initialAggregateValue = process.env.MONGODB_COMPASS_AGGREGATIONS_TEST_BASE_PATH;
```

packages/compass-aggregations/src/modules/saved-pipeline.spec.js
```diff
@@ -0,0 +1,63 @@
...
+    JSON.stringify(data)
+  );
+};
+
+describe('saved-pipeline', function() {
+  let tmpDir;
+  beforeEach(function() {
+    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'SavedPipelines'));
+    fs.mkdirSync(path.join(tmpDir, 'SavedPipelines'));
+    process.env.MONGODB_COMPASS_AGGREGATE_TEST_BASE_PATH = tmpDir;
```
mcasimir, line 22:
```suggestion
    process.env.MONGODB_COMPASS_AGGREGATIONS_TEST_BASE_PATH = tmpDir;
```

packages/compass-aggregations/src/modules/saved-pipeline.spec.js
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+  afterEach(function() {
+    fs.rmdirSync(tmpDir, { recursive: true });
+    process.env.MONGODB_COMPASS_AGGREGATE_TEST_BASE_PATH = initialAggregateValue;
```
mcasimir, line 27:
```suggestion
    process.env.MONGODB_COMPASS_AGGREGATIONS_TEST_BASE_PATH = initialAggregateValue;
```

packages/compass-query-history/src/utils/query-storage.js
```diff
@@ -0,0 +1,59 @@
+import { promisifyAmpersandMethod } from 'mongodb-data-service';
+import { FavoriteQueryCollection, FavoriteQuery } from '../models';
+
+export class QueryStorage {
```
mcasimir, line 4: suggestion: If this is only for favorites should we name it `FavoriteQueryStorage`?

Alternatively, if we feel we would have a `QueryStorage` for both favorites and history, we may want to rename the methods to `loadAllFavoriteQueries`, `updateFavoriteQueryAttributes` and `deleteFavoriteQuery`.

mcasimir (approved):

packages/compass-query-history/src/utils/query-storage.js
mabaasit: renamed class to `FavoriteQueryStorage`

PR #2726, opened 2022-01-26
feat(connect-form): add support for MONGODB_AWS COMPASS-5439

mcasimir:
Screenshot 2022-01-26 at 17 55 33

### DIFF
--- a/packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-gssapi.tsx
+++ b/packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-gssapi.tsx
@@ -13,6 +13,11 @@ import {
   parseAuthMechanismProperties,
 } from '../../../utils/connection-string-helpers';
 
+export const GSSAPI_PRINCIPAL_NAME_LABEL = "Principal";
+export const GSSAPI_SERVICE_NAME_LABEL = "Service Name";
+export const GSSAPI_CANONICALIZE_HOST_NAME_LABEL = "Canonicalize Host Name";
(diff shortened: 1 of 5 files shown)


addaleax (approved):

packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-gssapi.tsx
```diff
@@ -13,6 +13,11 @@ import {
   parseAuthMechanismProperties,
 } from '../../../utils/connection-string-helpers';
 
+export const GSSAPI_PRINCIPAL_NAME_LABEL = 'Principal';
+export const GSSAPI_SERVICE_NAME_LABEL = 'Service Name';
+export const GSSAPI_CANONICALIZE_HOST_NAME_LABEL = 'Canonicalize Host Name';
+export const GSSAPI_SERVICE_REALM_LABEL = 'Service Realm';
```
Anemy, line 19: This might be a totally unnecessary thing I do, but with things like label names I like to not export them for tests if possible and instead make them hardcoded strings. It does mean that every time a label is updated we need to update the test, but I think that helps make the test a bit more solid and regression safe if the label accidentally has a typo added or something.

Anemy (approved):
lgtm!

packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-gssapi.tsx
mcasimir: that's a good point, i'll revert everything in the next one :)

PR #2728, opened 2022-01-26
chore(e2e): TypeScript COMPASS-5248

lerouxb:
Update to everything written below: With some amazing help from @gribnoysup we get to have our cake and eat it too. CompassBrowser type that automatically knows about all our custom commands.

---

If you try and use a namespace you get the lint error "ES2015 module syntax is preferred over custom TypeScript modules and namespaces.eslint@typescript-eslint/no-namespace" See https://github.com/typescript-eslint/typescript-eslint/blob/v4.28.4/packages/eslint-plugin/docs/rules/no-namespace.md

I suppose we can disable that, but even then I realised that if we follow [webdriverio's documentation](https://webdriver.io/docs/typescript#adding-custom-commands) it would mean typing every browser command twice (once when setting up the namespace and once when adding the command, ie writing the body of the code. Apart from being inelegant I worry that that would be error prone because we'd maintain two copies of the type signatures. But maybe I'm missing something?

Furthermore I'd ideally like to keep the "one command per file" structure because some are quite big, there are already quite a few and it grows fairly linearly as we add more and more tests. So I looked into ways to achieve the [mixin](https://www.typescriptlang.org/docs/handbook/mixins.html)  pattern in typescript and none of them work particularly well. 

The one listed under [Alternative Pattern](https://www.typescriptlang.org/docs/handbook/mixins.html#alternative-pattern) where you use the `applyMixins()` helper looked promising, but I realised we'd have to type all the properties we use on each one (ie. if browser is `this`, then things like `this.click`).

I considered just making a CompassBrowser class that extends Browser, but again it has the problem where we either put all the commands in that giant class or we have to specify an interface or something with the types and then duplicate it in the implementation elsewhere. So I'd have to loosen my "1000 or 2000 lines in a code file is a lot" standard to something like 10000 :).

Another thing to consider is.. is the browser custom command pattern even a good idea? It puts all our code into the same namespace as webdriverio.

In the end I just kept all the commands in files as before and helpers/commands/index.ts  just re-exports them all. Then just use them as `await Commands.clickVisible(browser, selector);`. If anyone feels strongly about it we can move all the code into one file and just extend the Browser type or make our own or type it twice or enable namespaces.. At this stage it would largely be a find/replace to fix it all up since everything is Commands.something(browser.

Really I'm more of a "composition over inheritance, leave classes/types/interfaces not under your own control alone" person so this fits my personal style, but let me know what you think.

What's interesting is that chai-as-promised has the same "plugin" or "mixin" problem and when I installed its types I had a look at how it is achieved: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/chai-as-promised/index.d.ts Not sure how that would map from a `.d.ts` to a `.ts`, but we'd have to allow namespaces either way.

Ideas / suggestions welcome!


packages/compass-e2e-tests/helpers/commands/click-visible.ts
```diff
@@ -0,0 +1,12 @@
+import { Browser } from 'webdriverio';
```
addaleax, line 1: I think this can usually be 

```suggestion
import type { Browser } from 'webdriverio';
```

correct? That’s often a bit nicer because it differentiates a type-only dependency from a dependency on the code (and helps understand the structure of the code that way)

packages/compass-e2e-tests/helpers/commands/connect-with-connection-string.ts
```diff
@@ -0,0 +1,19 @@
...
+import * as Commands from '../commands';
+import * as Selectors from '../selectors';
+
+const defaultTimeoutMS = 30_000;
+
+export async function connectWithConnectionString(
+  browser: Browser<'async'>,
+  connectionString: string,
+  timeout = defaultTimeoutMS,
+  connectionStatus = 'success'
```
addaleax, line 11:
```suggestion
  connectionStatus: 'success' | 'failure' | 'either' = 'success'
```

packages/compass-e2e-tests/helpers/commands/do-connect.ts
```diff
@@ -0,0 +1,26 @@
+import { Browser } from 'webdriverio';
+import * as Commands from '../commands';
+import * as Selectors from '../selectors';
+
+export async function doConnect(
+  browser: Browser<'async'>,
+  timeout: number,
+  connectionStatus = 'success'
```
addaleax, line 8:
```suggestion
  connectionStatus: 'success' | 'failure' | 'either' = 'success'
```

packages/compass-e2e-tests/.npmrc
```diff
@@ -0,0 +1,4 @@
+engine-strict=true
```
lerouxb, line 1: What? I removed this..

packages/compass-e2e-tests/helpers/commands/connect-with-connection-form.ts
```diff
@@ -0,0 +1,212 @@
...
+  sshTunnelUsername?: string;
+  sshTunnelPassword?: string;
+  sshTunnelIdentityFile?: string;
+};
+
+export async function connectWithConnectionForm(
+  browser: Browser<'async'>,
+  options: ConnectOptions,
+  timeout = defaultTimeoutMS,
+  connectionStatus = 'success'
```
addaleax, line 31:
```suggestion
  connectionStatus: 'success' | 'failure' | 'either' = 'success'
```

packages/compass-e2e-tests/helpers/result-logger.ts
```diff
@@ -58,25 +60,77 @@ const ENV_VARS = [
...
+  Pass = 'pass',
+  Fail = 'fail',
+  SilentFail = 'silentfail',
+}
+
+enum CI {
+  Evergreen = 'evergreen',
+  GithubActions = 'github-actions',
+  Unknown = 'unknown',
+}
```
addaleax, line 80: Do we need enums here or would `type ResultStatus = 'start' | 'pass' | 'fail' | 'silentfail';` also do the trick? Almost all of the time enums only provide an additional layer of indirection without any additional value

packages/compass-e2e-tests/helpers/result-logger.ts
```diff
@@ -85,51 +139,53 @@ class ResultLogger {
...
-      process.env.EVERGREEN_REVISION || process.env.GITHUB_SHA || 'unknown';
-
-    this.context.url = process.env.EVERGREEN
-      ? process.env.EVERGREEN_TASK_URL
-      : process.env.GITHUB_ACTIONS
-      ? githubWorkflowRunUrl()
-      : 'unknown';
+    this.context = {
+      // copy known env vars as-is if they are set
+      env: ENV_VARS.reduce((obj: any, name) => {
```
addaleax, line 144:
```suggestion
      env: ENV_VARS.reduce((obj: Env, name: keyof Env) => {
```

packages/compass-e2e-tests/helpers/telemetry.ts
```diff
@@ -32,25 +45,21 @@ async function startTelemetryServer() {
     delete process.env.HADRON_METRICS_SEGMENT_HOST_OVERRIDE;
   }
 
-  function events() {
+  function events(): Array<any> {
     return requests.flatMap((req) => req.body.batch);
   }
 
-  function screens() {
+  function screens(): Array<any> {
```
addaleax, line 52:
```suggestion
  function screens(): string[] {
```

packages/compass-e2e-tests/helpers/telemetry.ts
```diff
@@ -32,25 +45,21 @@ async function startTelemetryServer() {
     delete process.env.HADRON_METRICS_SEGMENT_HOST_OVERRIDE;
   }
 
-  function events() {
+  function events(): Array<any> {
```
addaleax, line 48:
```suggestion
  function events(): TelemetryEvent[] {
```

packages/compass-e2e-tests/helpers/telemetry.ts
```diff
@@ -1,9 +1,18 @@
...
-async function startTelemetryServer() {
-  let requests = [];
+// TODO: lots of any here
+export type Telemetry = {
+  requests: Array<any>;
+  stop: () => Promise<void>;
+  events: () => Array<any>;
+  key: string;
+  screens: () => Array<any>;
+};
```
addaleax, line 12:
```suggestion
// TODO: lots of any here
export interface TelemetryEvent {
  event: string;
  properties: Record<string, any>;
}

export interface Telemetry {
  requests: Array<any>;
  stop(): Promise<void>;
  events(): TelemetryEvent[];
  key: string;
  screens(): string[];
};
```

packages/compass-e2e-tests/helpers/commands/listen-for-telemetry-events.ts
```diff
@@ -0,0 +1,35 @@
+import { Browser } from 'webdriverio';
+import { Telemetry } from '../telemetry';
```
addaleax, line 2:
```suggestion
import { Telemetry, TelemetryEvent } from '../telemetry';
```

packages/compass-e2e-tests/helpers/commands/listen-for-telemetry-events.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function listenForTelemetryEvents(
+  browser: Browser<'async'>,
+  telemetry: Telemetry
+): (eventName: string) => Promise<void> {
+  const existingEventCount = telemetry.events().length;
+
+  function lookupNewEvent(eventName: string) {
```
addaleax, line 10:
```suggestion
  function lookupNewEvent(eventName: string): TelemetryEvent {
```

packages/compass-e2e-tests/index.ts
```diff
@@ -1,22 +1,25 @@
...
+} from './helpers/compass';
+import { createUnlockedKeychain } from './helpers/keychain';
+import ResultLogger from './helpers/result-logger';
 
+const debug = Debug('compass-e2e-tests');
 const keychain = createUnlockedKeychain();
 
-let metricsClient;
+// We can't import mongodb here yet because native modules will be recompiled
+let metricsClient: any;
```
addaleax, line 22: You should still be able to do something like this if you want to:

```suggestion
import type { MongoClient } from 'mongodb';
let metricsClient: MongoClient;
```

packages/compass-e2e-tests/package.json
```diff
@@ -6,32 +6,38 @@
   "scripts": {
     "clean": "node -e \"try { fs.rmdirSync('.mongodb', { recursive: true }); } catch (e) {}\" && node -e \"try { fs.rmdirSync('.log', { recursive: true }); } catch (e) {}\"",
     "eslint": "eslint",
-    "prettier": "prettier",
+    "prettier": "rm -f fixtures/*.json && prettier",
```
addaleax, line 9:
```suggestion
    "prettier": "rimraf fixtures/*.json && prettier",
```

(`rm` is not part of the Windows default environment)

packages/compass-e2e-tests/package.json
```diff
@@ -6,32 +6,38 @@
...
     "check": "npm run lint && npm run depcheck",
     "pretest": "npm run unzip-fixtures",
-    "test": "xvfb-maybe --auto-servernum --server-args=\"-screen 0 1024x768x24\" -- node index.js",
+    "test": "xvfb-maybe --auto-servernum --server-args=\"-screen 0 1024x768x24\" -- ts-node index.ts",
     "test-ci": "npm run test",
     "posttest-ci": "node ../../scripts/killall-mongo.js",
     "test-packaged": "npm run test -- -- --test-packaged-app",
     "test-packaged-ci": "npm run test-ci -- -- --test-packaged-app",
-    "reformat": "npm run prettier -- --write .",
+    "reformat": "rm -f fixtures/*.json && npm run prettier -- --write .",
```
addaleax, line 19:
```suggestion
    "reformat": "rimraf fixtures/*.json && npm run prettier -- --write .",
```

packages/compass-e2e-tests/tests/logging.test.ts
```diff
@@ -1,26 +1,31 @@
...
+import { expect } from 'chai';
+import { beforeTests, afterTests, Compass } from '../helpers/compass';
+import { startTelemetryServer, Telemetry } from '../helpers/telemetry';
+import * as Commands from '../helpers/commands';
 
 describe('Logging and Telemetry integration', function () {
   describe('after running an example path through Compass', function () {
-    let logs;
-    let telemetry;
+    let logs: Array<any>;
```
addaleax, line 8:
```suggestion
    let logs: MongoLogEntry[];
```

packages/compass-e2e-tests/tests/logging.test.ts
```diff
@@ -1,26 +1,31 @@
-const { expect } = require('chai');
-const { beforeTests, afterTests } = require('../helpers/compass');
-const { startTelemetryServer } = require('../helpers/telemetry');
+import { expect } from 'chai';
+import { beforeTests, afterTests, Compass } from '../helpers/compass';
+import { startTelemetryServer, Telemetry } from '../helpers/telemetry';
+import * as Commands from '../helpers/commands';
```
addaleax, line 4:
```suggestion
import * as Commands from '../helpers/commands';
import type { MongoLogEntry } from 'mongodb-log-writer';
```

packages/compass-e2e-tests/tests/logging.test.ts
```diff
@@ -309,7 +314,7 @@ describe('Logging and Telemetry integration', function () {
         },
       ];
 
-      let criticalPathActualLogs;
+      let criticalPathActualLogs: Array<any>;
```
addaleax, line 317:
```suggestion
      let criticalPathActualLogs: MongoLogEntry[];
```

addaleax (approved):
nice work! so glad to see this finally moving over to TS :)

packages/compass-e2e-tests/helpers/commands/click-visible.ts
lerouxb: Makes sense. I just didn't know about import type. Will change.

packages/compass-e2e-tests/helpers/result-logger.ts
lerouxb: You're right. Will change it.

packages/compass-e2e-tests/package.json
lerouxb: That's supposed to be gone (can do the same by just using prettier's ignore file) but somehow it came back due to squash and merge me having branched off the webdriver port branch..

packages/compass-e2e-tests/tests/logging.test.ts
```diff
@@ -340,7 +342,8 @@ describe('Logging and Telemetry integration', function () {
         // resulting in duplicate test names.
         it(`logs "${expected.msg}" (${i})`, function () {
           const actualLogIndex = criticalPathActualLogs.findIndex(
-            ({ id }, index) => id === expected.id && !testedIndexes.has(index)
+            ({ id }, index) =>
+              id === mongoLogId(expected.id) && !testedIndexes.has(index)
```
addaleax, line 346: I don’t think the first condition can ever be true … maybe we need to modify the types a bit, or use `{ id }: any` here

lerouxb: It complained that it could never work because one's a mongo id and the other's a number :)

addaleax: Right, but this now creates a new object and compares reference equality of that new object to another object, which is never true in JS

addaleax: (Plus, `id` actually *is* a number in the actuaĺ logs after processing)

lerouxb: is this correct, then? https://github.com/mongodb-js/mongodb-log-writer/blob/2422aab82eb64fca45d8278ee99f84bb726916c2/src/index.ts#L34

addaleax: It is correct for the writing side, yes.

packages/compass-e2e-tests/helpers/retry-with-backoff.ts
```diff
@@ -1,10 +1,12 @@
...
-async function retryWithBackoff(
-  fn = async () => {},
+export default async function retryWithBackoff(
+  fn = async () => {
+    return Promise.resolve();
+  },
   retries = 3,
   backoffStep = 200
-) {
+): Promise<any> {
```
gribnoysup, line 9: Something like this maybe:

```ts
export default async function retryWithBackoff<T>(
  fn: () => Promise<T>,
  retries = 3,
  backoffStep = 200
): ReturnType<typeof fn> {
```

In hindsight doesn't make sense to have a default value for `fn` anyway and allows to correctly type return type instead of `any`

packages/compass-e2e-tests/index.ts
```diff
@@ -119,14 +122,15 @@ async function main() {
...
+  tests.forEach((testPath: string) => {
     mocha.addFile(path.join(__dirname, testPath));
   });
 
   const metricsConnection = process.env.E2E_TESTS_METRICS_URI;
   if (metricsConnection) {
     debug('Connecting to E2E_TESTS_METRICS_URI');
     // only require it down here because it gets rebuilt up top
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
     const { MongoClient } = require('mongodb');
```
gribnoysup, line 134:
```suggestion
    const { MongoClient } = await import('mongodb');
```

lerouxb: Oh yeah I forgot about await import. Thanks.

packages/compass-e2e-tests/helpers/retry-with-backoff.ts
lerouxb: Cool. Lots of typescript I still have to learn. Or at least internalise.

gribnoysup (approved):
> Another thing to consider is.. is the browser custom command pattern even a good idea? It puts all our code into the same namespace as webdriverio.

I don't really think it's bad one, this is a common pattern across most e2e testing frameworks and it has it's benefits: allows you to avoid mixing code patterns inside the suites (we inevitably will things like `browser.doOneThing()` followed by `Commands.doAnotherThing(browser)` in the same test case and I find it harder to dechipher at a glance), and makes commands discovery easier as they are all part of the same object that you have to create anyway.

I don't mind this `Commands` approach though. I might've consider it a bigger deal if we would aim for something like 100% test coverage with e2e tests, but as we are keeping this relatively scaled down, I think it's totally fine like that

lerouxb:
Using @gribnoysup 's typescript magic the commands now get added as browser commands and the types are automatically added to a CompassBrowser type which means that at least in the tests we can do browser.something(blah) rather than Commands.something(browser, blah).

I tried to have the commands themselves do the same thing but it gets a bit hit and miss whether it works. My guess is because there are some complicated circular "references" that way. Chick and egg problems ;) So I left the commands themselves alone for now - they specify browser as Browser<'async'> (from webdriverio) and use the commands from Commands.*.

We can change it down the line if we find a way around it, but I think this isn't the end of the world.

packages/compass-e2e-tests/index.ts
lerouxb: The problem there was with importing mongodb before it gets rebuilt. Caused really obscure failures that took me days to find.

addaleax: Right, but `import type` really is a type-only import, i.e. would not be affected by rebuilds at all because it doesn’t actually load the module at runtime

packages/compass-e2e-tests/helpers/commands/listen-for-telemetry-events.ts
lerouxb: Actually TelemetryEvent | undefined.

packages/compass-e2e-tests/helpers/telemetry.ts
lerouxb: This is actually not right. it is request.body.batch which is something like:

```
{
      userId: '90b7a520-82fd-475e-8258-a0d8b2d451d0',
      traits: [Object],
      type: 'identify',
      context: [Object],
      _metadata: [Object],
      timestamp: '2022-01-27T09:51:24.591Z',
      messageId: 'node-14885ba35ccc8e962b88fb092d3e75f4-c5262f1f-9bfd-4c0c-9be4-74da65dbaaf2'
    }
```

lerouxb: Or I'm confused..

lerouxb:
This is new
Screenshot 2022-01-27 at 15 15 05

PR #2729, opened 2022-01-26
feat(connect-form): Add PLAIN and x509 auth mechanisms COMPASS-5436, COMPASS-5437

mcasimir:
Screenshot 2022-01-26 at 18 44 02
Screenshot 2022-01-26 at 18 51 14

### DIFF
--- a/packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-plain.tsx
+++ b/packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-plain.tsx
@@ -1,9 +1,70 @@
 import React from 'react';
+import { TextInput } from '@mongodb-js/compass-components';
 
-function AuthenticationPlain(): React.ReactElement {
+import ConnectionStringUrl from 'mongodb-connection-string-url';
+import { UpdateConnectionFormField } from '../../../hooks/use-connect-form';
+import FormFieldContainer from '../../form-field-container';
+import {
+  ConnectionFormError,
+  errorMessageByFieldName,
+} from '../../../utils/validation';
+import {
+  getConnectionStringPassword,
+  getConnectionStringUsername,
+} from '../../../utils/connection-string-helpers';
+
+export const PLAIN_USERNAME_LABEL = 'Username';
+export const PLAIN_PASSWORD_LABEL = 'Password';
+
+function AuthenticationPlain({
+  connectionStringUrl,
+  updateConnectionFormField,
+  errors,
+}: {
+  connectionStringUrl: ConnectionStringUrl;
+  errors: ConnectionFormError[];
+  updateConnectionFormField: UpdateConnectionFormField;
+}): React.ReactElement {
+  const username = getConnectionStringUsername(connectionStringUrl);
+  const password = getConnectionStringPassword(connectionStringUrl);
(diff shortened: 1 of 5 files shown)


packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-plain.tsx
```diff
@@ -1,9 +1,70 @@
...
+        <TextInput
+          onChange={({
+            target: { value },
+          }: React.ChangeEvent<HTMLInputElement>) => {
+            updateConnectionFormField({
+              type: 'update-username',
+              username: value,
+            });
+          }}
+          label={PLAIN_USERNAME_LABEL}
```
Anemy, line 45: Similar question here to the last one, do we want to add an info link like we currently have in Compass?
Screen Shot 2022-01-26 at 12 53 16 PM
links to https://docs.mongodb.com/manual/core/security-ldap/

packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-x509.tsx
```diff
@@ -1,9 +1,12 @@
+import { Banner, BannerVariant } from '@mongodb-js/compass-components';
 import React from 'react';
 
 function AuthenticationX509(): React.ReactElement {
   return (
     <>
-      <p>X.509</p>
+      <Banner variant={BannerVariant.Info}>
+        X.509 Authentication type requires a Client Certificate to work.
```
Anemy, line 8: Maybe we can guide the user a bit more here, I think we've chatted in the past on having links to the TLS/SSL tab, which is something we might want to do. For the time being maybe we can mention the tab until we add that feature?
`X.509 Authentication type requires a Client Certificate to work. This can be set in the TLS/SSL tab.`

packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-plain.tsx
mcasimir: I was chatting with Claudia about these, considering that it would require a "custom component" I would remove all of them. Possibly replacing it with a more generic help for the whole auth method.

Although I would argue that our goal should not be to teach our users what LDAP is and how it works but only make them know how to connect to MongoDB with LDAP.

If you look into that page, i do not believe that a user would get any real help from that, that's all the "client" part says about it: https://docs.mongodb.com/manual/core/security-ldap/#connect-to-a-mongodb-server-via-ldap-authentication

packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-x509.tsx
mcasimir: Good idea, updated!

packages/connect-form/src/components/advanced-options-tabs/authentication-tab/authentication-plain.tsx
Anemy: We use the info icon on a text input for the authentication database already:
Screen Shot 2022-01-26 at 1 11 07 PM
I see what you mean, the current link probably doesn't help people out much. I'm cool with removing it 👍

Anemy (approved):
:shipit:

PR #2730, opened 2022-01-27
chore(connections): Add react error boundary to compass-components and wrap new connect form COMPASS-5075

Anemy:
COMPASS-5075

This adds a [react rendering error boundary ](https://reactjs.org/docs/error-boundaries.html) around the Connect Form. This means if an error occurs when rendering the connect form, which really never should happen, we display the error and allow the user to keep using Compass without showing a blank screen. (screenshot below).
This error boundary helps to keep that experience for an unforeseen bug not fully broken, and gives the user something to report when that does happen (since they can now see the error message).

In Compass we currently have an error boundary in `hadron-react-components` called `UnsafeComponent`: https://github.com/mongodb-js/compass/blob/main/packages/hadron-react-components/src/unsafe-component.jsx
I think we'll want to replace this with this new ErrorBoundary eventually, but that work is not in this/planned yet.

Screen Shot 2022-01-27 at 12 54 43 PM

### DIFF
--- a/packages/compass-components/src/components/error-boundary.spec.tsx
+++ b/packages/compass-components/src/components/error-boundary.spec.tsx
@@ -1,0 +1,69 @@
+import React from 'react';
+import { render, screen, cleanup } from '@testing-library/react';
+import { expect } from 'chai';
+import sinon from 'sinon';
+
+import { ErrorBoundary } from './error-boundary';
+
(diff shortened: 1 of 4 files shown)


mcasimir (approved):
nice!

lerouxb (approved):
👏

PR #2735, opened 2022-01-28
chore(connect-form): Set tls on when an option is set, remove description

Anemy:
This PR: 
- Removes the description for “Default / On / Off” in the TLS tab
- Keeps all the tls fields enabled when tls is set to Default (they were disabled before)
- Sets tls=true if the user fills in a field when Default is selected

https://user-images.githubusercontent.com/1791149/151633905-776aefd2-7836-42cb-9d86-b13f9f0d10fd.mp4


mcasimir (approved):
❤️

PR #2737, opened 2022-01-31
chore(e2e): Instance sidebar e2e COMPASS-5457

lerouxb:
New tests:

* contains a dbs/collections tree view
* can search for a collection
* add database and drop it
* add collection and drop it

We'll probably have more add database/collection tests soon to test the other options on those modals and to make sure the other add/drop buttons work so I pre-emptively pulled these out into commands. Also trying to keep smoke.test.js manageable.

This still leaves the following functionality in the instance sidebar untested that we may or may not add in future:

* clicking local goes back to instance
* refresh
* edit favourite
* open collection in new tab


packages/compass-e2e-tests/tests/smoke.test.ts
```diff
@@ -45,7 +45,7 @@ describe('Smoke tests', function () {
     await afterTest(compass, this.currentTest);
   });
 
-  describe('Sidebar', function () {
```
Anemy, line 48: Since this file is getting more tests and as you mentioned more will be coming, should we break it into multiple files? How could that look? Maybe each of these suites gets a file in the smoke tests folder?

Anemy (approved):
lgtm! One question on breaking down smoke.test.js

packages/compass-e2e-tests/tests/smoke.test.ts
lerouxb: I was just thinking about that because making a bunch of small PRs against the same file that are in review at the same time makes for annoying merges.

I should probably make each describe its own file as that's the most obvious way forward. The downside is that we'll be re-launching the app and connecting again every time so things will be a bit slower. Probably not the end of the world.

lerouxb: I have split it up now.

PR #2742, opened 2022-01-31
fix(connect-form): Remove app name from copied connection string COMPASS-5473

Anemy:
COMPASS-5473
Removes `appName` when using the Copy Connection String action:
Screen Shot 2022-01-31 at 3 31 08 PM

The main reason we're doing this is that older connections have appName saved in the connection string, and when they're copied we want to ensure a user doesn't bring the Compass appName into their application or place where they are using the connection string. 
We could change this check just check for the appName matching Compass, open to thoughts/suggestions.


Anemy:
Closing this, I think probably in our conversions from old model to connectionInfo we just remove appName from these connection strings when it's matching the compass appName and allow users to copy whatever appName is in their connection string incase they want to customize it.

PR #2743, opened 2022-01-31
chore(compass-aggregations): use leafygreen toggle in compass-aggregations

Anemy:
COMPASS-5470 (part 1 of 2)

This updates our toggles in compass-aggregations to use the leafygreen-ui toggles (wrapped in compass-components). It removes the `react-ios-switch` package as we're no longer using it to show any switches/toggles.

Updated the label of these toggles to be actual labels for a bit more usability/accessibility.

The ticket involves making the toggles in the new connect form visible in dark mode since currently they aren't. I opened this pr so that the fix we do there also fixes these toggles (which suffer from the same lack of visibility in dark mode).

The toggles are now blue and used to be green so I'll run those colors by Claudia before merging and check if we want to keep the green or go with leafygreen's color.
*Before:*
Screen Shot 2022-01-31 at 5 41 13 PM
*After:*
Screen Shot 2022-01-31 at 5 40 44 PM
*Toggled off:*
Screen Shot 2022-01-31 at 5 42 53 PM

*Before*
Screen Shot 2022-01-31 at 6 16 25 PM

*After*
Screen Shot 2022-01-31 at 6 16 08 PM

### DIFF
--- a/packages/compass-aggregations/package.json
+++ b/packages/compass-aggregations/package.json
@@ -168,7 +168,6 @@
     "react-bootstrap": "^0.32.4",
     "react-dnd": "^10.0.2",
     "react-dnd-html5-backend": "^10.0.2",
-    "react-ios-switch": "^0.1.19",
     "react-redux": "^5.0.6",
     "react-select-plus": "^1.2.0",
     "redux": "^4.1.2",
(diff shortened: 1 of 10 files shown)


mcasimir (approved):
Aren't those going to disappear in dark mode too? or is in preparation to fix the toggles everywhere?

Anemy:
@mcasimir this is prep to fix them everywhere, currently these already disappear in dark mode

PR #2746, opened 2022-02-01
chore(e2e): create a db and drop it from the instance databases tab COMPASS-5458

lerouxb:
I started adding some more tests for the list and grid view like changing the view type, changing the sort by and changing the sort order but kinda shelved that for now because writing selectors for those components is a bit tricky and leafygreen doesn't always make it easy to write selectors for things.

So for now I'm going to just stick with this.

### DIFF
--- a/packages/compass-e2e-tests/helpers/commands/click-visible.ts
+++ b/packages/compass-e2e-tests/helpers/commands/click-visible.ts
@@ -3,24 +3,8 @@ import type { CompassBrowser } from '../compass-browser';
 export async function clickVisible(
   browser: CompassBrowser,
   selector: string,
-  scrollIntoView = true
 ): Promise<void> {
   const element = await browser.$(selector);
-
-  // Allow opting out of scrolling the item into view because that can be quite
-  // finicky for things that only display on hover.
-  if (scrollIntoView) {
-    // You can't scroll a thing that doesn't exist yet
-    await element.waitForExist();
-
-    // The element can be displayed, but not on screen and then click will silently fail.
-    // Furthermore, the default scrollIntoView() behaviour is to align the element
-    // to the top of the screen which can leave the item visible, yet apparently
-    // not enough because click can still silently do nothing.  Aligning with the
-    // bottom of the screen seems to be more reliable. My guess is because of fixed
-    // position things at the top.
-    await element.scrollIntoView(false);
-  }
 
   await element.waitForDisplayed();
 
--- a/packages/compass-e2e-tests/helpers/commands/click-visible.ts
+++ b/packages/compass-e2e-tests/helpers/commands/click-visible.ts
@@ -2,11 +2,30 @@ import type { CompassBrowser } from '../compass-browser';
 
 export async function clickVisible(
   browser: CompassBrowser,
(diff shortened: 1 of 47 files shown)


packages/compass-e2e-tests/tests/instance-databases-tab.test.ts
```diff
@@ -40,5 +40,32 @@ describe('Instance databases tab', function () {
...
+    await browser.addDatabase(dbName, collectionName);
+
+    const databaseCard = await browser.$(Selectors.databaseCard(dbName));
+    await databaseCard.waitForDisplayed();
+
+    await databaseCard.scrollIntoView(false);
+
+    await browser.waitUntil(async () => {
+      // open the drop database modal from the database card
+      await browser.hover(Selectors.databaseCard(dbName));
```
lerouxb, line 59: Ended up writing a loop because hover didn't always work the first time.

packages/databases-collections-list/src/items-grid.tsx
```diff
@@ -108,7 +108,7 @@ const GridControls = () => {
 
   return (
     <>
-      {createControls && <div className={control}>{createControls}</div>}
+      {createControls && <div className={control} data-testid="create-controls">{createControls}</div>}
```
lerouxb, line 111: Putting data-testid on containers like these was the best I could come up with because I can't see how to add anything specific to leafygreen buttons.

packages/compass-e2e-tests/helpers/commands/click-visible.ts
```diff
@@ -2,11 +2,30 @@ import type { CompassBrowser } from '../compass-browser';
...
-  selector: string
+  selector: string,
+  scrollIntoView = false
 ): Promise<void> {
   const element = await browser.$(selector);
-  await element.scrollIntoView();
+
+  // Allow opting out of scrolling the item into view because that can be quite
+  // finicky for things that only display on hover.
+  if (scrollIntoView) {
```
addaleax, line 12: Maybe just to make sure I’m really not missing anything – this seems to be opt-in rather than opt-out?

addaleax (approved):
nice!

packages/compass-e2e-tests/helpers/commands/click-visible.ts
lerouxb: aaargh. default's supposed to be false. Well spotted ;)

packages/compass-e2e-tests/tests/instance-databases-tab.test.ts
```diff
@@ -40,5 +40,50 @@ describe('Instance databases tab', function () {
     }
   });
 
-  it('can create a database from the databases tab');
+  it('links database cards to the database collections tab', async function () {
+    // Click on the db name text inside the card specifically to try and have
+    // tighter control over where it clicks, because clicking in the center of
+    // the last card if all cards don't fit on screen can silently do nothing
+    // even after scrolling it into view.
+    const selector = `${Selectors.databaseCard('test')} [title="test"]`;
```
lerouxb, line 48: webdriver will make me cry

PR #2747, opened 2022-02-01
chore(eslint): enable consistent type imports rule

mcasimir:
We recently realized that importing types via `import` may have side effects. For example, this caused issues while trying to use the new `connect-form` in VSCode.

This PR enables a strict check for types to be imported as `types` only. (and fixes the resulting errors .. which caused tons of changes)

cc: @alenakhineika


addaleax (approved):
Big :+1: for this!

PR #2749, opened 2022-02-02
feat(connect-form): Adds Save and Connect button in connect-form COMPASS-5300

leorossi:
## Description
Adds "Save" and "Connect" button to connection form.

"Save" will be visible only when the current active connection is a Favorite
It will be enabled only when the form fields/connection string changes.

When clicking "New connection" only "connect" will be shown.

## Motivation and Context

- [ ] Bugfix
- [X] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/connect-form/src/components/connect-form-actions.tsx
+++ b/packages/connect-form/src/components/connect-form-actions.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import {
   Button,
   ButtonVariant,
@@ -11,11 +11,20 @@ import type {
   ConnectionFormError,
   ConnectionFormWarning,
 } from '../utils/validation';
+import { ConnectionInfo } from 'mongodb-data-service';
 
 const formActionStyles = css({
   borderTop: `1px solid ${uiColors.gray.light2}`,
   paddingLeft: spacing[4],
   paddingRight: spacing[4],
+});
+
+const formActionButtonStyles = css({
+  display: 'flex',
+  justifyContent: 'flex-end',
+  '& div:first-child': {
+    marginRight: spacing[2],
+  },
 });
 
 const formButtonsStyles = css({
@@ -25,23 +34,53 @@ const formButtonsStyles = css({
 });
 
 function ConnectFormActions({
+  initialConnectionInfo,
   errors,
   warnings,
   onConnectClicked,
+  onSaveClicked,
+  saveDisabled,
 }: {
-  onConnectClicked: () => void;
+  initialConnectionInfo: ConnectionInfo;
   errors: ConnectionFormError[];
   warnings: ConnectionFormWarning[];
+  onConnectClicked: () => void;
+  onSaveClicked: () => void;
+  saveDisabled: boolean;
 }): React.ReactElement {
   return (
     <div className={formActionStyles}>
       {warnings.length ? <WarningSummary warnings={warnings} /> : ''}
       {errors.length ? <ErrorSummary errors={errors} /> : ''}
-
-      <div className={formButtonsStyles}>
-        <Button variant={ButtonVariant.Primary} onClick={onConnectClicked}>
-          Connect
-        </Button>
+      <div className={formActionButtonStyles}>
+        {!initialConnectionInfo.favorite && (
+          <div className={formButtonsStyles}>
+            <Button variant={ButtonVariant.Primary} onClick={onConnectClicked}>
+              Connect
+            </Button>
+          </div>
+        )}
+        {!!initialConnectionInfo.favorite && (
+          <Fragment>
+            <div className={formButtonsStyles}>
+              <Button
+                variant={ButtonVariant.Default}
+                disabled={saveDisabled}
+                onClick={onSaveClicked}
+              >
+                Save
+              </Button>
+            </div>
+            <div className={formButtonsStyles}>
+              <Button
+                variant={ButtonVariant.Primary}
+                onClick={onConnectClicked}
+              >
+                Connect
+              </Button>
+            </div>
+          </Fragment>
+        )}
       </div>
     </div>
   );
--- a/packages/connect-form/src/components/connect-form-actions.tsx
+++ b/packages/connect-form/src/components/connect-form-actions.tsx
@@ -22,12 +22,8 @@ const formActionButtonsStyles = css({
   display: 'flex',
   justifyContent: 'flex-end',
   gap: spacing[2],
-});
-
-const formButtonsStyles = css({
   paddingTop: spacing[3],
   paddingBottom: spacing[3],
-  textAlign: 'right',
 });
 
 function ConnectFormActions({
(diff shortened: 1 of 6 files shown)


packages/connect-form/src/components/connect-form-actions.tsx
```diff
@@ -11,37 +11,76 @@ import {
   ConnectionFormError,
   ConnectionFormWarning,
 } from '../utils/validation';
+import { ConnectionInfo } from 'mongodb-data-service';
```
mcasimir, line 14:
```suggestion
import type { ConnectionInfo } from 'mongodb-data-service';
```

packages/connect-form/src/components/connect-form-actions.tsx
```diff
@@ -11,37 +11,76 @@ import {
...
 function ConnectFormActions({
+  initialConnectionInfo,
   errors,
   warnings,
   onConnectClicked,
+  onSaveClicked,
+  saveDisabled,
 }: {
-  onConnectClicked: () => void;
+  initialConnectionInfo: ConnectionInfo;
```
mcasimir, line 44: the `initialSomething` is used more for components that keep track of the state of "something" and then change it, for example, if there would be an "initial connection" and one that is edited from a `useState` hook. This component is stateless, I'd just call this parameter `connectionInfo`.

packages/connect-form/src/components/connect-form-actions.tsx
```diff
@@ -11,37 +11,76 @@ import {
...
 const formActionStyles = css({
   borderTop: `1px solid ${uiColors.gray.light2}`,
   paddingLeft: spacing[4],
   paddingRight: spacing[4],
 });
 
+const formActionButtonStyles = css({
+  display: 'flex',
+  justifyContent: 'flex-end',
+  '& div:first-child': {
```
mcasimir, line 25: with flex you don't need to do this, you can specify a gap: https://css-tricks.com/almanac/properties/g/gap/

I'm not an expert and maybe there are other properties and tricks, but definitely adding a margin with `div:first-child` should not be needed.

Note that even keeping this solution, the selector: `'& div:first-child':` is technically a hack and not really a clean solution, maybe a bit better with a "direct child" selector `> ..` but you should really use a class to clarify the intention and make sure that future changes won't mess the style: `.action-button:first-child { ... }` and `<div className={cx('action-button', actionButtonStyles)}>`.

Take a look at the `gap` property though, we can use flex and avoid these weird selectors entirely

packages/connect-form/src/components/connect-form-actions.tsx
```diff
@@ -11,37 +11,76 @@ import {
...
 } from '../utils/validation';
+import type { ConnectionInfo } from 'mongodb-data-service';
 
 const formActionStyles = css({
   borderTop: `1px solid ${uiColors.gray.light2}`,
   paddingLeft: spacing[4],
   paddingRight: spacing[4],
 });
 
+const formActionButtonStyles = css({
```
mcasimir, line 22: suggestion: this is a container of multiple buttons and the name of the component is "form-actions", so maybe for consistency could be better to use the plural here for both:

```suggestion
const formActionsButtonsStyles = css({
```

packages/connect-form/src/components/connect-form-actions.tsx
```diff
@@ -11,37 +11,76 @@ import {
...
       {warnings.length ? <WarningSummary warnings={warnings} /> : ''}
       {errors.length ? <ErrorSummary errors={errors} /> : ''}
-
-      <div className={formButtonsStyles}>
-        <Button variant={ButtonVariant.Primary} onClick={onConnectClicked}>
-          Connect
-        </Button>
+      <div className={formActionButtonStyles}>
+        {!initialConnectionInfo.favorite && (
+          <div className={formButtonsStyles}>
```
mcasimir, line 57: this wrapper is not necessary anymore, we can move the `padding` to  `formActionButtonStyles` and drop this, so we don't bloat the layout.

packages/connect-form/src/hooks/use-connect-form.ts
```diff
@@ -557,6 +558,9 @@ export function useConnectForm(
               : validateConnectionOptionsWarnings(
                   updatedState.connectionOptions
                 ),
+          isDirty:
```
mcasimir, line 561: this is not enough, we need to compare the entire `connectionOptions` object

packages/connect-form/src/components/connect-form.tsx
```diff
@@ -187,8 +193,18 @@ function ConnectForm({
           </div>
           <div className={formFooterStyles}>
             <ConnectFormActions
+              initialConnectionInfo={initialConnectionInfo}
               errors={connectionStringInvalidError ? [] : errors}
               warnings={connectionStringInvalidError ? [] : warnings}
+              saveDisabled={!isDirty}
```
mcasimir, line 199: just a thought: is a bit strange to derive whether to show the save button inside the  `ConnectFormActions` component but pass if it should be disabled directly. Not sure if this would improve it, but maybe something like `saveButton: 'enabled' | 'disabled' | 'hidden'`?

mcasimir (commented):
Looks good already, I suggested a few changes, mostly few random minor comments. I like the idea of keeping track of the "dirty" state in one central place, makes our life easier. Before merging we should at least make sure we compare all of the connection options and add some tests.

packages/connect-form/src/components/connect-form-actions.tsx
leorossi: I think we need the wrapper because it's the flex container that will place the buttons horizontally (before them there is the warning/errors banner that should stack vertically instead)

packages/connect-form/src/components/connect-form.tsx
leorossi: change

packages/connect-form/src/components/connect-form-actions.tsx
mcasimir: What I mean is that you need 1 wrapper but not 2. right now we have:

```
<div className={formActionButtonStyles}>
   <div className={formButtonsStyles}><button></div>
   <div className={formButtonsStyles}><button></div>
</div> 
```

we can do just:

```
<div className={formActionButtonStyles}>
   <button>
   <button>
</div> 
```

packages/connect-form/src/components/connect-form-actions.tsx
```diff
@@ -11,37 +11,76 @@ import {
...
+              <Button
+                variant={ButtonVariant.Default}
+                disabled={saveDisabled}
+                onClick={onSaveClicked}
+              >
+                Save
+              </Button>
+            </div>
+            <div className={formButtonsStyles}>
+              <Button
```
mcasimir: other nit, is not necessary to copy paste the connect button here, you can have the whole thing like this:

```
{initialConnectionInfo.favorite ?? <button>Save</button>}
<button>Connect</button>
```

packages/connect-form/src/components/connect-form.tsx
mcasimir: Looks good to me, if you feel this is weird we could also just pass everything to the component as is (`isDirty` and `favorite`) and do the "math" inside where you need it.

packages/connect-form/src/components/connect-form-actions.tsx
leorossi: I removed the `connectionInfo` parameter from this component, because it was used only for that check. Let me know what you suggest to do

mcasimir (approved):
👍

packages/connect-form/src/components/connect-form-actions.tsx
leorossi: got it, fixed!

PR #2751, opened 2022-02-02
fix(compass-components): improve dark theme toggle appearance COMPASS-5470

Anemy:
COMPASS-5470
This PR adds a theme context to the react tree that the connections page is in. This theme context is used in the toggle to render the leafygreen dark theme when dark theme is turned on.

This does not yet fix the toggles in the aggregations view as those are currently in a different react tree and do not have access to the theme context. Work to make them use one react root will happen in COMPASS-5481 . Once that is done, these changes will fix those toggles as well.

I think we could initialize the theme preference, listen to updates in the theme from the main compass process, and propagate that state to the react renderer in a more concise, generalized way which would hopefully free up adding more settings to Compass. I'll leave a comment where this is in the code with this text also. Since the theme is the only state of Compass which is currently managed between multiple windows, I think it's alright to have the solution in this pr for now to unblock the connect form and give the theme to the react context. We can improve and generalize the initialization and synchronization of the shared state in future work when there's more to sync. What do y'all think? Should we do more now?

Before:
Screen Shot 2022-02-02 at 1 44 11 PM

After:
Screen Shot 2022-02-02 at 1 43 48 PM


packages/compass/src/app/index.js
```diff
@@ -69,10 +69,18 @@ window.addEventListener('error', (event) => {
 
 const darkreaderOptions = { brightness: 100, contrast: 90, sepia: 10 };
 function enableDarkTheme() {
+  // Compass-home initializes the theme and listens to these events
+  // to update the theme in the react context.
+  global.hadronApp.theme = 'DARK';
+  global.hadronApp.appRegistry?.emit('darkmode-enable');
```
Anemy, line 75: I think we could initialize the theme preference, listen to updates in the theme from the main compass process, and propagate that state to the react renderer in a more concise, generalized way which would hopefully free up adding more settings to Compass. Since the theme is the only state of Compass which is currently managed between multiple windows, I think it's alright to have the solution in this pr for now to unblock the connect form and give the theme to the react context. We can improve and generalize the initialization and synchronization of the shared state in future work when there's more to sync. What do y'all think? Should we do more now?

Anemy:
Hmm seeing two different build ci e2e tests failing for timeout after merging main. Could be a rendering slowdown from these changes or unrelated. Looking into

packages/compass/src/app/index.js
mcasimir: I think is fine for this PR, we can do that work when we will add user settings.

One improvement I'd really love to see is to extract all of this "theme" related code from this already bloated `index.js`, including the initialization: https://github.com/mongodb-js/compass/blob/dad2a58c23b888f2b832bb6458e155b1f6ccff5c/packages/compass/src/app/index.js#L391

We did something similar for the logging: https://github.com/mongodb-js/compass/blob/ca7f24fca54d7249696821edb1417301253aad3e/packages/compass/src/main/logging.ts

Among just cleaning up that file a bit would also enable us to write some tests for it at some point

mcasimir (approved):
This is really cool

packages/compass-home/src/components/home.tsx
```diff
@@ -246,7 +273,9 @@ function Home({ appName }: { appName: string }): React.ReactElement | null {
     return (
       <div className={homeViewStyles} data-test-id="home-view">
         <div className={homePageStyles}>
-          <Connections onConnected={onConnected} />
+          <ThemeProvider theme={theme}>
```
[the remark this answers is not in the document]
Anemy: It was because it wasn't being used elsewhere. I'll move it up now that https://github.com/mongodb-js/compass/pull/2754 is landing 👍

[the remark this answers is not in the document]
Anemy: updated

packages/compass-components/src/hooks/use-theme.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { createContext, useContext } from 'react';
+
+export type Theme = 'LIGHT' | 'DARK';
```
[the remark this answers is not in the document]
Anemy: updated 👍

packages/compass-components/src/components/toggle.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import LeafyGreenToggle from '@leafygreen-ui/toggle';
+
+import { Theme, useTheme } from '../hooks/use-theme';
+
+function Toggle(
+  props: React.ComponentProps<typeof LeafyGreenToggle>
```
gribnoysup, line 7:
```suggestion
  props: Omit<React.ComponentProps<typeof LeafyGreenToggle>, 'darkMode'>
```

packages/compass-components/src/components/toggle.tsx
```diff
@@ -0,0 +1,14 @@
...
+): ReturnType<typeof LeafyGreenToggle> {
+  const theme = useTheme();
+
+  return <LeafyGreenToggle darkMode={theme?.theme === Theme.Dark} {...props} />;
+}
```
gribnoysup, line 12: Raised this with leafygreen, if that's the only way to set the darkMode for their components, we are at the risk of wrapping every single leafygreen component that we export here which sounds like a giant maintenance headache

mcasimir: yes it would be massively better if they'd have a provider we can use, I'd expect LG to also provide themes and guidelines for the palette, which is something we'd have to do on our side otherwise:

```
const darkTheme = {
    theme: Theme.Dark
    colors: {
        accent: palette.green.light1,
        // .. etc
    }
}
```

mcasimir: We could do something like this before re-exporting the components on our side:

```
const LGComponentWithTheme<T>(Wrappee: T) {
   const theme = useTheme()
   return (props) => Wrappee(...props, {...darkmode: theme?.theme})
}
```

```
const Toggle = LGComponentWithTheme(Toggle)
export Toggle;
```

gribnoysup: I mean, yeah, we totally can do something like that, HOC is indeed a good pattern to use here. It's just weird that we have even though we all agree that we want to do as little changes to leafygreen as possible

gribnoysup: lg team [said](https://mongodb.slack.com/archives/CFFA5BS79/p1643962820401349) this will be addressed when they are eventually working on a dark mode support

Anemy: What should we do for now? I think we probably won't need to do many fixes like this until dark theme/the new provider is released in leafygreen so having one component listening to the theme is probably alright?

packages/compass-components/src/components/toggle.tsx
Anemy: Good question. I think we probably want it to be overrideable, yes, but I'm alright with either approach. If it's a component which always is displayed in a darkMode or always needs to be light mode I think we'll want to customize it and override the provider. I think we could also add more providers in that situation, but if it's just one component I think it's simpler to just override this. I'm cool with either here.

packages/compass-components/src/components/toggle.tsx
mcasimir: I would merge this as it is

packages/compass-components/src/components/toggle.tsx
gribnoysup: Yeah, this makes sense! Let's keep it like that for now 👍

PR #2753, opened 2022-02-02
chore(connect-form): Update tls input label

Anemy:
Updates the tls cert input texts. We still need a way to surface the actual connection string param these inputs correspond to. That will happen in another ticket.

Before:
Screen Shot 2022-02-02 at 3 11 23 PM

After:
Screen Shot 2022-02-02 at 3 15 15 PM

### DIFF
--- a/packages/connect-form/src/components/advanced-options-tabs/tls-ssl-tab/tls-client-certificate.tsx
+++ b/packages/connect-form/src/components/advanced-options-tabs/tls-ssl-tab/tls-client-certificate.tsx
@@ -36,7 +36,7 @@ function TLSClientCertificate({
           description={'Learn More'}
           disabled={disabled}
           id="tlsCertificateKeyFile"
-          label="Client Certificate Bundle (.pem)"
+          label="Client Certificate and Key (.pem)"
           dataTestId="tlsCertificateKeyFile-input"
           link={
             'https://docs.mongodb.com/manual/reference/connection-string/#mongodb-urioption-urioption.tlsCertificateKeyFile'
--- a/packages/connect-form/src/components/advanced-options-tabs/tls-ssl-tab/tls-client-certificate.tsx
+++ b/packages/connect-form/src/components/advanced-options-tabs/tls-ssl-tab/tls-client-certificate.tsx
@@ -36,7 +36,7 @@ function TLSClientCertificate({
           description={'Learn More'}
           disabled={disabled}
           id="tlsCertificateKeyFile"
-          label="Client Certificate (.pem)"
+          label="Client Certificate Bundle (.pem)"
           dataTestId="tlsCertificateKeyFile-input"
           link={
             'https://docs.mongodb.com/manual/reference/connection-string/#mongodb-urioption-urioption.tlsCertificateKeyFile'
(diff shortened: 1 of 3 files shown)


packages/connect-form/src/components/advanced-options-tabs/tls-ssl-tab/tls-client-certificate.tsx
```diff
@@ -36,7 +36,7 @@ function TLSClientCertificate({
           description={'Learn More'}
           disabled={disabled}
           id="tlsCertificateKeyFile"
-          label="Client Certificate (.pem)"
+          label="Client Certificate Bundle (.pem)"
```
mcasimir, line 39: What about aligning this name with the param (`tlsCertificateKeyFile`) right away? 
```suggestion
          label="Client Certificate and Key (.pem)"
```

I believe the confusion here would be more that is not clear that a key can be put there, the fact that is a bundle is kinda also in the `.pem` extension already.

packages/connect-form/src/components/advanced-options-tabs/tls-ssl-tab/tls-ssl-tab.spec.tsx
```diff
@@ -39,7 +39,8 @@ describe('SchemaInput', function () {
 
     it('should render the client cert and CA file labels', function () {
       expect(screen.getByText('Certificate Authority (.pem)')).to.be.visible;
-      expect(screen.getByText('Client Certificate (.pem)')).to.be.visible;
+      expect(screen.getByText('Client Certificate Bundle (.pem)')).to.be
```
mcasimir, line 42:
```suggestion
      expect(screen.getByText('Client Certificate and Key (.pem)')).to.be
```

packages/connect-form/src/components/advanced-options-tabs/tls-ssl-tab/tls-client-certificate.tsx
mcasimir: @addaleax do you have any suggestions here?

packages/connect-form/src/components/advanced-options-tabs/tls-ssl-tab/tls-client-certificate.tsx
```diff
@@ -49,7 +49,7 @@ function TLSClientCertificate({
           }}
           showFileOnNewLine
           optional
-          optionalMessage="Optional (only required with X.509)"
+          optionalMessage="Optional (required with X.509)"
```
addaleax, line 52: I’d suggest

```suggestion
          optionalMessage="Optional (required with X.509 auth)"
```

or

```suggestion
          optionalMessage="Optional (required with X.509 authentication)"
```

if we have the space for displaying that (since X.509 usage != X.509 authentication)

packages/connect-form/src/components/advanced-options-tabs/tls-ssl-tab/tls-client-certificate.tsx
addaleax: I think `and Key` sounds okay, yeah. It’s probably not possible to explain that they should be concatenated in the limited space here – can/should we do something like tooltips on hover?

mcasimir: DS wasn't super happy when we proposed that, but yeah technically we could

mcasimir (approved):

addaleax (approved):

PR #2756, opened 2022-02-03
feat(connect-form): Replace hardcoded hex codes with color codes COMPASS-5475

mcasimir:
- Replace hex colors with color codes (color1, color2 ...) that can be mapped dynamically to different colors depending on the theme.
- Fix a bug that allows to save favorites without a name
- Implement new design for the color indicators in the sidebar
Screenshot 2022-02-03 at 13 53 02
Screenshot 2022-02-03 at 13 53 26
Screenshot 2022-02-03 at 13 53 43
Screenshot 2022-02-03 at 13 54 06

### DIFF
--- a/packages/connect-form/src/hooks/use-connection-color.ts
+++ b/packages/connect-form/src/hooks/use-connection-color.ts
@@ -1,16 +1,6 @@
 import { useCallback } from 'react';
 
-type ColorCode =
-  | 'color1'
-  | 'color2'
-  | 'color3'
-  | 'color4'
-  | 'color5'
-  | 'color6'
-  | 'color7'
-  | 'color8'
-  | 'color9'
-  | 'color10';
+type ColorCode = `color${1|2|3|4|5|6|7|8|9|10}`;
 
 const COLOR_CODES_TO_UI_COLORS_DARK_THEME_MAP: Record<ColorCode, string> = {
   color1: '#00A35C', // green.dark1
--- a/packages/connect-form/src/hooks/use-connection-color.ts
+++ b/packages/connect-form/src/hooks/use-connection-color.ts
@@ -32,10 +32,8 @@ export const CONNECTION_COLOR_CODES = Object.keys(
   COLOR_CODES_TO_UI_COLORS_DARK_THEME_MAP
 ) as ColorCode[];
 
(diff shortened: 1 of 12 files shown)


packages/connect-form/src/hooks/use-connection-color.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  if (isColorCode(hexOrColorCode)) {
+    return hexOrColorCode as ColorCode;
+  }
+
+  return LEGACY_COLORS_TO_COLOR_CODE_MAP[hexOrColorCode];
+}
+
+export function useConnectionColor(): {
```
mcasimir, line 65: this is a `hook` so we can call `useTheme` here and change the set of colors based on the current theme.

packages/connect-form/src/hooks/use-connection-color.ts
```diff
@@ -0,0 +1,87 @@
...
+  | 'color1'
+  | 'color2'
+  | 'color3'
+  | 'color4'
+  | 'color5'
+  | 'color6'
+  | 'color7'
+  | 'color8'
+  | 'color9'
+  | 'color10';
```
addaleax, line 13: Obviously completely optional suggestion, but you can also do:

```suggestion
type ColorCode = `color${1|2|3|4|5|6|7|8|9|10}`;
```

packages/connect-form/src/hooks/use-connection-color.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const CONNECTION_COLOR_CODES = Object.keys(
+  COLOR_CODES_TO_UI_COLORS_DARK_THEME_MAP
+) as ColorCode[];
+
+function isColorCode(hexOrColorCode: string | undefined) {
+  return hexOrColorCode
+    ? (CONNECTION_COLOR_CODES as string[]).includes(hexOrColorCode)
+    : undefined;
+}
```
addaleax, line 49:
```suggestion
function isColorCode(hexOrColorCode: string | undefined): hexOrColorCode is ColorCode {
  return (CONNECTION_COLOR_CODES as string[]).includes(hexOrColorCode);
}
```

(renders the cast in the next function unnecessary)

addaleax (approved):

PR #2758, opened 2022-02-03
fix db/collection card flakyness

lerouxb:
Clicking on database or collection cards that don't start off on screen is flaky. The workaround is to click on the green card title text.

### DIFF
--- a/packages/compass-e2e-tests/tests/database-collections-tab.test.ts
+++ b/packages/compass-e2e-tests/tests/database-collections-tab.test.ts
@@ -1,10 +1,7 @@
-import chai from 'chai';
 import type { CompassBrowser } from '../helpers/compass-browser';
 import { beforeTests, afterTests, afterTest } from '../helpers/compass';
 import type { Compass } from '../helpers/compass';
 import * as Selectors from '../helpers/selectors';
-
-const { expect } = chai;
 
 describe('Database collections tab', function () {
   let compass: Compass;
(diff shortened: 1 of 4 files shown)


packages/compass-e2e-tests/tests/database-collections-tab.test.ts
```diff
@@ -28,9 +25,38 @@ describe('Database collections tab', function () {
...
+    await collectionsGrid.waitForDisplayed();
+
+    const collectionSelectors = ['json-array', 'json-file', 'numbers'].map(
+      (collectionName) => Selectors.collectionCard('test', collectionName)
+    );
+
+    for (const collectionSelector of collectionSelectors) {
+      const collectionElement = await browser.$(collectionSelector);
+      await collectionElement.waitForExist();
+    }
```
addaleax, line 38: Does it make a difference whether this is for-loop or Promise.all() (do `.waitForExist()` calls resolve immediate if the element already exists)?

mcasimir (approved):

addaleax (approved):

packages/compass-e2e-tests/tests/database-collections-tab.test.ts
lerouxb: Pretty much immediately unless you want to count the next tick or whatever happens when awaiting a promise.

lerouxb: I'm basically using waitForExist as an implicit expect. You can't immediately check if they already exist because the code will execute before the page updated. So in reality there will be a short wait before the first one as the page transitions and renders. There are lots of ways to write this but this feels simple and compact to me. Also accomplishes both the waiting and the assertions in one go.

PR #2760, opened 2022-02-04
save a screenshot when a before hook fails

lerouxb:
If an `it()` fails, then the `afterEach` hook calls `afterTest(compass, this.currentTest)` which can check the test object to see if it failed and then store a screenshot named after the test.

But if a before/after/beforeEach/afterEach fails we don't currently take a screenshot.

We used to take a screenshot in every `afterTests(compass)` which tends to be called in an after hook, but that just timestamped the screenshot making it possible but often a bit of a pain to map back to the thing that failed if anything. It also took screenshots indiscriminately and sometimes too late (ie. after something else already happened on screen).

The before hook is actually a very common place for tests to fail because that's where we usually open compass, connect and browse to the area that's interesting for the tests in that file.

So this attempts to set a reasonable screenshot name and pass it down to afterTests if the test failed which will be very handy.

I tried to come up with something that will automatically figure out if the hook failed and come up with a reasonable filename automatically, but it is surprisingly tricky. It is usually the before hook for the first test or something and no link to the describe block and its name and whether it failed as far as I can tell. ResultLogger does very similar things but that uses events to keep track of where it is. ResultLogger might actually be a good place to do the screenshotting but then there's no obvious link back to the browser object.. So this felt like the simplest, least messy way to achieve this.

Let me know if you can think of anything cleaner.

What's especially nice is if you use `npm run testfast` (which will bail on the first error) and it fails, you can just `open ./log/s<tab><enter>` and see the screenshot. And if you download the artifacts from github actions or evergreen and just one test failed then it will be the only screenshot in there. If there are multiple fails, then the filenames will help you quickly tell which is which.

### DIFF
--- a/packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
+++ b/packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
@@ -10,18 +10,22 @@ const { expect } = chai;
 describe('Collection aggregations tab', function () {
   let compass: Compass;
   let browser: CompassBrowser;
+  let screenshot: string | undefined;
 
   before(async function () {
+    screenshot = 'collection-aggregations-tab-before';
     compass = await beforeTests();
     browser = compass.browser;
 
     await browser.connectWithConnectionString('mongodb://localhost:27018/test');
 
     await browser.navigateToCollectionTab('test', 'numbers', 'Aggregations');
+
+    screenshot = undefined;
   });
 
   after(async function () {
-    await afterTests(compass);
+    await afterTests(compass, screenshot);
   });
 
   afterEach(async function () {
--- a/packages/compass-e2e-tests/tests/connection.test.ts
+++ b/packages/compass-e2e-tests/tests/connection.test.ts
@@ -24,14 +24,17 @@ async function disconnect(browser: CompassBrowser) {
 describe('Connection screen', function () {
   let compass: Compass;
   let browser: CompassBrowser;
+  let screenshot: string | undefined;
 
   before(async function () {
+    screenshot = 'connection-screen-before';
     compass = await beforeTests();
     browser = compass.browser;
+    screenshot = undefined;
   });
 
   after(function () {
-    return afterTests(compass);
+    return afterTests(compass, screenshot);
   });
 
   afterEach(async function () {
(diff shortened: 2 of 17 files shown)


packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -88,10 +93,10 @@ describe('SRV connectivity', function () {
         undefined,
         'either'
       );
+      screenshot = undefined;
     } finally {
-      await disconnect(browser);
```
lerouxb, line 92: No need to disconnect if we're going to close the app and disconnecting in the finally means the screen changes before we can take the screenshot.

packages/compass-e2e-tests/tests/logging.test.ts
```diff
@@ -383,23 +386,24 @@ describe('Logging and Telemetry integration', function () {
...
     });
 
     after(async function () {
-      if (compass) {
-        // cleanup outside of the test so that the time it takes to run does not
-        // get added to the time it took to run the first query
-        await afterTests(compass);
+      // clean up if it failed during the before hook
+      if (screenshot) {
+        await afterTests(compass, screenshot);
```
lerouxb, line 406: Note that before also calls afterTests, but only if it gets to the end.

packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
```diff
@@ -10,18 +10,22 @@ const { expect } = chai;
...
   before(async function () {
+    screenshot = 'collection-aggregations-tab-before';
     compass = await beforeTests();
     browser = compass.browser;
 
     await browser.connectWithConnectionString('mongodb://localhost:27018/test');
 
     await browser.navigateToCollectionTab('test', 'numbers', 'Aggregations');
+
+    screenshot = undefined;
```
addaleax, line 24: I don’t necessarily mind this bookkeeping, but I’m wondering if there’s a better way. For one, I wouldn’t necessarily consider it bad if we just always captured a screenshot on failure, and maybe we could base the screenshot name on the test name? That information should be accessible, e.g.:

https://github.com/mongodb-js/mongosh/blob/5afb80349eab42e026c5c8219cf378c7933b0d66/packages/cli-repl/test/e2e.spec.ts#L19

https://github.com/mongodb-js/mongosh/blob/5afb80349eab42e026c5c8219cf378c7933b0d66/packages/cli-repl/test/test-shell.ts#L79

lerouxb: you can do this.currentTest.failed (which we do in the afterTest handler for afterEach), but this.currentTest.title is not very helpful. it will be `the describe test / the next test that would have run` at best. Which I suppose I could try and use. ie. cut off after the `/` and run some regex on it all to try and turn into a filename..

lerouxb:
Closing because I think https://github.com/mongodb-js/compass/compare/screenshot-afterTests2?expand=1 is nicer.

PR #2763, opened 2022-02-04
fix(connection-storage): disable legacy validation on save COMPASS-5476

mcasimir:
Allows saving connections that would have been invalid with the old connection model. Still attempts a best effort to preserve backward compatibility as much as possible to allow downgrades.

Screenshot 2022-02-04 at 20 36 59


Anemy (approved):
nice

Anemy (approved):
nice

PR #2764, opened 2022-02-06
feat(connect-form): set appName on connect and don't save it COMPASS-5473

mcasimir:
Noticed we were not setting appName with the new connection form since that feature was done in the old UI

### DIFF
--- a/packages/connections/src/components/connections.spec.tsx
+++ b/packages/connections/src/components/connections.spec.tsx
@@ -73,7 +73,7 @@ describe('Connections Component', function () {
         <Connections
           onConnected={onConnectedSpy}
           connectionStorage={mockStorage}
-          appName={'Test App Name'}
+          appName='Test App Name'
         />
       );
     });
--- a/packages/connections/src/components/connections.spec.tsx
+++ b/packages/connections/src/components/connections.spec.tsx
@@ -73,7 +73,7 @@ describe('Connections Component', function () {
         <Connections
           onConnected={onConnectedSpy}
           connectionStorage={mockStorage}
-          appName='Test App Name'
+          appName="Test App Name"
         />
       );
     });
--- a/packages/connections/src/components/connections.spec.tsx
+++ b/packages/connections/src/components/connections.spec.tsx
@@ -52,13 +52,9 @@ describe('Connections Component', function () {
 
   beforeEach(function () {
     onConnectedSpy = sinon.spy();
-    this.clock = sinon.useFakeTimers({
-      now: 1483228800000,
-    });
   });
 
   afterEach(function () {
-    this.clock.restore();
     sinon.restore();
     cleanup();
   });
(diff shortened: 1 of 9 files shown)


packages/connections/src/stores/connections-store.ts
```diff
@@ -26,13 +28,34 @@ export function createNewConnectionInfo(): ConnectionInfo {
...
 
+function tryParseConnectionString(
+  connectionString: string
+): ConnectionString | undefined {
+  try {
+    return new ConnectionString(connectionString);
+  } catch (e) {
+    //
+  }
+}
```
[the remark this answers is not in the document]
mcasimir: easy to change :)

mcasimir:
> I am confused. I thought we had agreed in [MONGOSH-1015](https://jira.mongodb.org/projects/MONGOSH/issues/MONGOSH-1015) and [COMPASS-4901](https://jira.mongodb.org/projects/COMPASS/issues/COMPASS-4901) that we _do_ want users to be able to specify their own appName and we _do_ want it to be persisted in that case?

Thanks for bringing this up, had no clue it was already implemented in the old model:

```
  if (typeof connectionModel.appname === 'undefined') {
    connectionModel.appname = electron.remote.app.getName();
  }
```

Make sense to do this here too. Then we can remove it only if it starts with `Mongodb+Compass` or something in the `convertConnectionModelToInfo`

packages/connections/src/components/connections.tsx
```diff
@@ -54,16 +54,29 @@ const formContainerStyles = css({
...
+    return electron?.remote?.app?.getName() || defaultAppName;
+  } catch (e) {
+    return defaultAppName;
+  }
+}
+
 function Connections({
   onConnected,
   connectionStorage = new ConnectionStorage(),
+  appName = getElectronAppName(),
```
Anemy, line 71: In `compass-home` we have appName and can pass it as a prop to `Connections`: https://github.com/mongodb-js/compass/blob/main/packages/compass-home/src/components/home.tsx#L257 
Maybe we can remove this defaulting code?

packages/connections/src/components/connections.spec.tsx
```diff
@@ -113,50 +115,62 @@ describe('Connections Component', function () {
     });
   });
 
-  describe('when rendered with saved connections in storage', function () {
+  describe.only('when rendered with saved connections in storage', function () {
```
addaleax, line 118: :)

packages/connections/src/stores/connections-store.ts
```diff
@@ -26,13 +28,35 @@ export function createNewConnectionInfo(): ConnectionInfo {
...
+  } catch (e) {
+    //
+  }
+
+  if (!connectionStringUrl) {
+    return connectionString;
+  }
+
+  if (!connectionStringUrl.searchParams.has('appName')) {
+    connectionStringUrl.searchParams.set('appName', appName);
```
addaleax, line 54: totally optional, but since we have started being fairly consistent about this elsewhere:

```suggestion
  const searchParams = connectionStringUrl.typedSearchParams<MongoClientOptions>();
  if (!searchParams.has('appName')) {
    searchParams.set('appName', appName);
```

(with `import type { MongoClientOptions } from 'mongodb';` above)

addaleax (approved):

PR #2765, opened 2022-02-07
feat(compass-components): add useToast COMPASS-5473

mcasimir:
Replace state-based toasts with a global provider + hooks

https://user-images.githubusercontent.com/334881/152865201-f066056e-b46c-4cb2-bb43-0b65f5778b52.mp4

### DIFF
--- a/packages/compass-components/src/hooks/use-toast.tsx
+++ b/packages/compass-components/src/hooks/use-toast.tsx
@@ -1,0 +1,161 @@
+import React, { createContext, useCallback, useContext, useState } from 'react';
+import type { ToastVariant } from '..';
+import { Toast } from '..';
+
+type ToastProperties = {
+  title?: React.ReactNode;
+  body: React.ReactNode;
+  variant: ToastVariant;
+  progress?: number;
+  timeout?: number;
+};
+
+type ToastState = ToastProperties & {
+  timeoutRef?: ReturnType<typeof setTimeout>;
+};
+
+type ToastId = string;
+
+type ToastContextState = {
+  toasts: Record<ToastId, ToastState>;
+  setToasts: (toasts: Record<ToastId, ToastState>) => void;
+};
+
+interface ToastActions {
+  openToast: (id: ToastId, toastProperties: ToastProperties) => void;
+  closeToast: (id: ToastId) => void;
+}
+
+const ToastContext = createContext<ToastContextState>({
+  toasts: {},
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  setToasts: (t: Record<ToastId, ToastProperties>) => {
+    return;
+  },
+});
+
+const ToastPile = (): React.ReactElement => {
+  const { toasts } = useContext(ToastContext);
+  const { closeToast } = useGlobalToast();
+
+  return (
+    <>
+      {Object.entries(toasts).map(
+        ([id, { title, body, variant, progress }]) => (
+          <Toast
+            key={id}
+            title={title}
+            body={body}
+            variant={variant}
+            progress={progress}
+            open={true}
+            close={() => closeToast(id)}
+          />
+        )
+      )}
+    </>
+  );
+};
+
+/**
+ * @example
+ *
+ * ```
+ * const MyButton = () => {
+ *   const { openToast } = useToast();
+ *   return <button onClick={() => openToast(
+ *      'myToast1', {title: 'This is a notification'})} />
+ * };
+ *
+ * <ToastArea><MyButton/><ToastArea>
+ * ```
+ *
+ * @returns
+ */
+export const ToastArea = ({
+  children,
+}: {
+  children: React.ReactChildren;
+  theme: ToastActions;
+}): React.ReactElement => {
+  const [toasts, setToasts] = useState({});
+
+  return (
+    <ToastContext.Provider value={{ toasts, setToasts }}>
+      <>{children}</>
+      <ToastPile />
+    </ToastContext.Provider>
+  );
+};
+
+function useGlobalToast(): ToastActions {
+  const { toasts, setToasts } = useContext(ToastContext);
+
+  const closeToast = useCallback(
+    (toastId: ToastId): void => {
+      const { timeoutRef } = toasts[toastId] || {};
+      if (timeoutRef) {
+        clearTimeout(timeoutRef);
+      }
+
+      const newToasts = { ...toasts };
+      delete newToasts[toastId];
+      setToasts(newToasts);
+    },
+    [toasts, setToasts]
+  );
+
+  const openToast = useCallback(
+    (toastId: ToastId, toastProperties: ToastProperties): void => {
+      // if updating clear timeouts first
+      const { timeoutRef } = toasts[toastId] || {};
+      if (timeoutRef) {
+        clearTimeout(timeoutRef);
+      }
+
+      setToasts({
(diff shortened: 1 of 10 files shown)


packages/compass-components/src/hooks/use-toast.tsx
```diff
@@ -0,0 +1,168 @@
...
+        ...toasts,
+        [toastId]: {
+          ...toastProperties,
+          timeoutRef: toastProperties.timeout
+            ? setTimeout(() => {
+                closeToast(toastId);
+              }, toastProperties.timeout)
+            : undefined,
+        },
+      });
```
gribnoysup, line 133: Because `setState` is not a sync operation, running `setToasts` like that will lead to concurrently opened toasts swallowing each other (you can test it by calling `openToast` in sequence anywhere in the code), what you want here is:

```suggestion
      setToasts(currentState => ({
        ...currentState,
        [toastId]: {
          ...toastProperties,
          timeoutRef: toastProperties.timeout
            ? setTimeout(() => {
                closeToast(toastId);
              }, toastProperties.timeout)
            : undefined,
        }
      }));
```

or you can move this logic into a React reducer with explicit actions for put / delete action types so that you ensure that you are always deriving state from the previous one, not overriding it.

We were just chatting with Basit about this, my rule of thumb is that if you need to put an complex data structure to `useState`, you usually either want to just split it to multiple primitive values that are controlled by `useState`, or use `useReducer` instead

packages/compass-components/src/hooks/use-toast.tsx
```diff
@@ -0,0 +1,168 @@
...
+          <Toast
+            className={toastStyles}
+            key={id}
+            title={title}
+            body={body}
+            variant={variant}
+            progress={progress}
+            open={true}
+            close={() => closeToast(id)}
+          />
```
gribnoysup, line 61: Seems like by default toast component renders itself with position fixed which means that multiple toasts are just stacked on top of each other and you can't see them:

This is how three toasts look in UI:

image

And this is the DOM (just to prove that there are multiple of them):

image

I think we either need to ask leafygreen team to implement toast stack or work around that by having our own fixed container here + adding some styles to the toast to override its position (we probably would need to do both because I don't see leafygreen implementing this change any time soom)

packages/compass-components/src/hooks/use-toast.tsx
```diff
@@ -0,0 +1,168 @@
...
+  return (
+    <ToastContext.Provider value={{ toasts, setToasts }}>
+      <>{children}</>
+      <ToastPile />
+    </ToastContext.Provider>
+  );
+};
+
+function useGlobalToast(): ToastActions {
+  const { toasts, setToasts } = useContext(ToastContext);
```
gribnoysup, line 99: It's generally fine as it is right now, but as both methods exported from `useGlobalToast` need to be defined only once, they would be better candidates to get passed to the provider, instead of just `useState` return values. I would probably go about it like this. The `useGlobalToast` hook defines the toast state instead of getting it from context ...

```suggestion
function useGlobalToast(): ToastActions {
  const [toasts, setToasts] = useState({});
```

... and is used inside the `ToastArea` component and then open/close methods are the things that are passed in the context and accessed in the `useToast` hook

gribnoysup (commented):
Overall I think it looks good, but there are a few issues that I spotted that I think we should address before this can be merged

packages/compass-components/src/hooks/use-toast.tsx
mcasimir: > fixed container here + adding some styles

That's probably easy enough and probably we could rely on the fact that `position: fixed` should always be on the Toast element itself.

We decided not to make this change here, but worth a try

packages/compass-components/src/hooks/use-toast.tsx
mcasimir: Nice, will change it

packages/compass-components/src/hooks/use-toast.tsx
mcasimir: That's my noobness revealing itself :), nice to know.

packages/compass-components/src/hooks/use-toast.tsx
```diff
@@ -0,0 +1,149 @@
...
+        clearTimeout(timeoutRef);
+      }
+
+      setToasts((prevToasts) => ({
+        ...prevToasts,
+        [toastId]: {
+          ...toastProperties,
+          timeoutRef: toastProperties.timeout
+            ? setTimeout(() => {
+                closeToast(toastId);
```
mcasimir, line 93: is this ok? do we need to do something to cancel this timeout on unmount?

gribnoysup: Oh, yeah, that's actually a good point. You do want to clean up the timeouts on unmount. I think if you want to go full galaxy brain / idiomatic hooks usage on this, you would probably even want this timeout management be out of this state and done by something separately, something with an interface like this:

```js
const timeoutManager = useTimeoutManager((id) => {
  // This will be some sort of onTimeout callback
  closeToast(id);
});

const openToast = useCallback(() => {
  ...
  // This will keep a reference to the timeout in some sort of internal timeout manager map or something
  timeoutManager.startTimeout(id, timeout)
})
```

And then the code inside `useTimeoutManager` can do the clean-up on unmount. This will make a robust, single purpose hook that's easily reusable somewhere else.

... or tbh you can just have something like this

```js
useEffect(() => {
  return () => {
    Object.values(toasts).forEach(({timeoutRef}) => timeoutRef && clearTimeout(timeoutRef))
  }
}, [])
```

and call it a day. Eslint will scream at you because the dependencies list is not matching, but you would want it like that so that the clean-up callback is only called on unmount and not every time that toasts state is changing

mcasimir: updated this one, you may want to take a look :)

packages/compass-components/src/hooks/use-toast.tsx
gribnoysup: Ah yeah, I see now that they render them in a portal (on top of having fixed position, why?) and in this case don't allow to override the portal node (or disable the portal similar to other components). I will ask them how are we even supposed to handle this. If you want to merge this as-is that's fine, but basically with how leafygreen works right now we can't make multiple toasts work on the same screen and so maybe we shouldn't even "allow" this on the hook level, because we are basically allowing broken behavior that way

mcasimir: At hook level is partially handled, I think is best if LG fixes it, but since you have to pass an id, creating multiple toasts has to be intentional. Ie. the `copy-connection-string` will not create multiple toasts, but the next one will replace the previous one if is there.

gribnoysup (approved):
Looks good! I'm concerned about this https://github.com/mongodb-js/compass/pull/2765#discussion_r803415095 and would prefer to decide how we should handle this before this PR lands, but also don't want to block it on this

packages/compass-components/src/hooks/use-toast.tsx
gribnoysup: Yeah, we definitely handle it on our side! What I mean is what happens when I need to add a toast with another id somewhere else in the app? It will result in a broken UI with overlapping toasts, do we just agree that this is okay for now and wait for leafygreen team to provide the solution in the future?

PR #2771, opened 2022-02-07
chore(e2e): One test for the performance tab for now COMPASS-5459

lerouxb:


packages/compass-e2e-tests/tests/instance-performance-tab.test.ts
```diff
@@ -21,5 +24,8 @@ describe('Instance performance tab', function () {
     await afterTest(compass, this.currentTest);
   });
 
-  it('can kill slow queries via the performance tab');
+  it('loads up without issue', async function () {
+    const stats = await browser.$(Selectors.ServerStats);
+    await stats.waitForDisplayed();
```
addaleax, line 29: I guess there’s not much point here in going deeper and attempting to make sure that the individual stats are being displayed?

addaleax (approved):

PR #2777, opened 2022-02-09
chore(compass-crud): Long field names in document view are not truncated anymore COMPASS-5361

leorossi:
## Description
Removes ellipsis that truncates long field names in document view
Screenshot_2022-02-09_11-14-23

## Motivation and Context

- [X] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-crud/src/components/editable-element-field.less
+++ b/packages/compass-crud/src/components/editable-element-field.less
@@ -3,12 +3,14 @@
   border: none;
   padding-left: 1px;
   -webkit-user-select: text;
+  max-width: 15%;
+  text-overflow: ellipsis;
   white-space: nowrap;
   overflow: hidden;
 
   &-is-editing {
     border: 1px solid #807f7f;
-    box-shadow: 0px 2px 4px 0px rgba(0, 0, 0, 0.2);
+    box-shadow: 0px 2px 4px 0px rgba(0,0,0,0.2);
     z-index: 200;
     margin-top: -1px;
     margin-bottom: -1px;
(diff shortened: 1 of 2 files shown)


packages/compass-crud/src/components/editable-element-field.less
```diff
@@ -3,14 +3,12 @@
   border: none;
   padding-left: 1px;
   -webkit-user-select: text;
-  max-width: 15%;
-  text-overflow: ellipsis;
```
Anemy, line 7: Do we still want some ellipses when the field name and value are more than the width of the window?
Screen Shot 2022-02-09 at 1 40 33 PM

packages/compass-crud/src/components/insert-document-dialog.less
```diff
@@ -38,114 +38,115 @@
...
-        .editable-element-types {
-          .btn {
-            background-color: #dee0e3;
-            color: #313030;
-          }
-          .caret {
-            visibility: visible;
+      .editable-element {
+        &:hover {
+          background-color: #f5f6f7;
```
Anemy, line 85: I think we're losing the hover over field effect we used to have - is that an intentional change?
Before 
Screen Shot 2022-02-09 at 1 44 38 PM

After
Screen Shot 2022-02-09 at 1 44 44 PM

packages/compass-crud/src/components/editable-element-field.less
Anemy: Would increasing the `max-width` to something like `70%` and re-adding `text-overflow: ellipsis;` work?

packages/compass-crud/src/components/insert-document-dialog.less
Anemy: Sure thing, whichever works, mostly interested in preserving the existing functionality if possible.

mcasimir (approved):

PR #2779, opened 2022-02-10
chore(e2e): Create an index and drop it COMPASS-5469

lerouxb:
Just this one basic test for indexes for now. We can probably test out the options and multi-key indexes when we redo that screen.


packages/compass-e2e-tests/helpers/commands/add-collection.ts
```diff
@@ -98,6 +98,7 @@ export async function addCollection(
     );
     await menu.waitForDisplayed();
     const span = await menu.$(`span=${options.timeseries.granularity}`);
+    await span.waitForDisplayed();
```
lerouxb, line 101: Drive-by for something I missed in the previous PR.

addaleax (approved):

lerouxb:
got the windows npm install flake. will have to re-run check and test.

PR #2784, opened 2022-02-11
bug(compass-import-export): Include manually added fields in CSV output COMPASS-4971

leorossi:
## Description
When exporting to CSV, if the user add a field in the dialog box, it is not included in the output.
## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-import-export/src/modules/export.js
+++ b/packages/compass-import-export/src/modules/export.js
@@ -478,7 +478,7 @@ export const startExport = () => {
     // Pick the columns that are going to be matched by the projection,
     // where some prefix the field (e.g. ['a', 'a.b', 'a.b.c'] for 'a.b.c')
     // has an entry in the projection object.
-    const columns = Object.keys(exportData.allFields)
+    const columns = Object.keys(exportData.fields)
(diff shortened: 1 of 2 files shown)


Anemy (approved):
lgtm, tried locally and works well. If possible maybe we can add a test to prevent regressions.
Looks like the query projection is ignored for the results, but I think that's a separate issue.

packages/compass-import-export/src/modules/export.js
```diff
@@ -478,7 +478,7 @@ export const startExport = () => {
     // Pick the columns that are going to be matched by the projection,
     // where some prefix the field (e.g. ['a', 'a.b', 'a.b.c'] for 'a.b.c')
     // has an entry in the projection object.
-    const columns = Object.keys(exportData.allFields)
+    const columns = Object.keys(exportData.fields)
```
mcasimir, line 481: Any chance we could have a test that captures this behavior for the future?

mcasimir (approved):
👍

PR #2792, opened 2022-02-15
feat(compass-saved-aggregations-queries): update query/aggregation COMPASS-5506

mabaasit:
feat(compass-saved-aggregations-queries): update query/aggregation COMPASS-5506

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-saved-aggregations-queries/src/components/edit-item-modal.tsx
```diff
@@ -0,0 +1,91 @@
...
+          label="Name"
+          name="name"
+          placeholder="Name"
+          onChange={(event) => {
+            setName(event.target.value);
+          }}
+          value={name}
+        />
+      </form>
+    </ConfirmationModal>
```
gribnoysup, line 68: Asked leafygreen about that yesterday and they suggest not to use confirmation modal for forms (see [this thread](https://mongodb.slack.com/archives/CFFA5BS79/p1644944835907879?thread_ts=1644942475.749649&cid=CFFA5BS79)) and instead build it ourselves from other leafygreen components

packages/compass-saved-aggregations-queries/src/stores/edit-item.ts
```diff
@@ -0,0 +1,102 @@
...
+      const pipelineStorage = new PipelineStorage();
+      await pipelineStorage.updateAttributes(id, {
+        name: attributes.name,
+      });
+    }
+
+    dispatch({
+      type: ActionTypes.EditItemUpdated,
+    });
+    dispatch(fetchItems());
```
gribnoysup, line 99: I think we talked about this one too before, instead of dispatching a fetch, the way redux recommends handling cases like this is to make saved items reducer react to the `EditItemUpdated` action and update stored items (similar to how delete is handled). See [this](https://redux.js.org/style-guide/style-guide#allow-many-reducers-to-respond-to-the-same-action) and [this](https://redux.js.org/style-guide/style-guide#avoid-dispatching-many-actions-sequentially) for some additional details

packages/compass-saved-aggregations-queries/src/components/edit-item-modal.tsx
```diff
@@ -0,0 +1,125 @@
...
+  onSubmit,
+  onCancel,
+}) => {
+  const [name, setName] = useState(item?.name ?? '');
+  useEffect(() => {
+    setName(item?.name ?? '');
+  }, [item]);
+
+  if (!item) {
+    return <></>;
```
gribnoysup, line 50: I think a more conventional way of handling this is to return `null` here and not a fragment. IIRC the reason for this was that `null`s were handled as a sorta special empty slot case in React reconciler (there is a [reference](https://reactjs.org/docs/implementation-notes.html#what-we-left-out) to this in the docs, but not much detail). It's probably not important in that particular case, just FYI

packages/compass-saved-aggregations-queries/src/components/edit-item-modal.tsx
```diff
@@ -0,0 +1,125 @@
...
+          >
+            Cancel
+          </Button>
+          <Button
+            disabled={isSubmitDisabled()}
+            variant="primary"
+            onClick={(event) => {
+              event.preventDefault();
+              onSubmitForm();
+            }}
```
gribnoysup, line 95: IIRC even without the onClick handler and preventing the default it should just trigger a onSubmit on a form and so we don't need to call it here additionally. I'm guessing though leafygreen doesn't allow to omit the onClick prop on the component?

gribnoysup (commented):
Overall I think it looks pretty good, there are a bunch of things that caught my eye, but let me know if these don't make too much sense to you

packages/compass-saved-aggregations-queries/src/components/edit-item-modal.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { cancelEditItem, updateItem } from '../stores/edit-item';
+
+type EditItemModalProps = {
+  isModalOpen: boolean;
+  item?: Pick<Item, 'id' | 'name' | 'type'>;
+  onSubmit(id: string, attributes: UpdateItemAttributes): void;
+  onCancel: () => void;
+};
+
+const formTitleStyles = css({
```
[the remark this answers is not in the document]
mabaasit: thanks. made it close to this, except for `margin-bottom`. Sticking to a number provided by `spacing` that fits.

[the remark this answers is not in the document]
gribnoysup: Sorry, I should've been more clear, I meant like all the modal styles in this file I linked to: we are missing the divider between controls and confirm etc., but it's fine like that too, I have a ticket for creating a modal form component for compass-components, so we can pick it up later (there are a few other places in compass where it needs to be used anyway)

packages/compass-saved-aggregations-queries/src/components/saved-item-card.tsx
```diff
@@ -19,7 +19,7 @@ import {
 import type { Item } from '../stores/aggregations-queries-items';
 import { formatDate } from '../utlis/format-date';
 
-export type Action = 'open' | 'delete' | 'copy' | 'rename';
+export type Action = 'open' | 'delete' | 'copy' | 'edit';
```
gribnoysup, line 22: I'd double-check the name with product just in case. IIRC we wanted it to be "Rename" because "Edit" reads as if you can edit the whole thing, but you can only rename it

packages/compass-saved-aggregations-queries/src/stores/aggregations-queries-items.ts
```diff
@@ -72,6 +68,20 @@ const reducer: Reducer<State, Actions | DeleteItemActions> = (
...
+      if (!item) {
+        return state;
+      }
+      const updatedItem =
+        item.type === 'query'
+          ? mapQueryToItem(action.payload as Query)
+          : mapAggregationToItem(action.payload as Aggregation);
+      return {
+        ...state,
+        items: [...state.items.filter((x) => x.id !== action.id), updatedItem],
```
gribnoysup, line 82: Total nit, but can be done in one go (instead two iterations)

```suggestion
        items: state.items.map((item) => {
          if (item.id === action.id) {
            return updatedItem;
          } else {
            return item
          }
        }),
```

gribnoysup (approved):
Looks good, a few small comments, but feel free to ignore

PR #2798, opened 2022-02-16
chore: scope bootstrap out of connection form COMPASS-5482

Anemy:
COMPASS-5482

This PR removes the influence of [bootstrap](https://github.com/twbs/bootstrap) from the new connection form and adds a reset css. This is a first step in the process of removing bootstrap from Compass. Bootstrap is currently required globally and thus the global styles it applies as well as its normalize css impacts all of the views of Compass. This PR scopes bootstrap and the reset css so that we can overtime isolate and remove bootstrap from the codebase.

This pr also has styling fixes for the indexes table, and the schema view. I'll leave a comment where those are at, sorry for the extra things to review - couldn't resist drive by.

I had a hard time finding a way to cleanly scope components rendered at the same time using emotion style caching. The solution I ended up at was copying out all of the compiled bootstrap, extracting the items which globally influence Compass, and applying a wrapping class with those globals. The items which are already namespaced with classes we keep in a few files now in the `deprecated-bootstrap-styles` folder. (Open to other names) We use that bootstrap globals class where we currently have bootstrap and still need to remove it.

Bootstrap is still a dependency as we require it's js in a few places: https://github.com/mongodb-js/compass/blob/main/packages/compass/src/app/index.js#L25

Here's how the connection form looks now (will do a follow up pr updating the styles and probably tweaking the reset css a bit more):
Screen Shot 2022-02-14 at 3 32 16 PM

I'd appreciate if the reviewer checks out the branch and clicks around Compass a bit to see if there are any styling differences in the views of Compass resulting from these changes I may have missed.
I think we could also make this reset css a lot better. It's pretty basic right now, and maybe does too much.


packages/compass-schema/src/components/compass-schema/compass-schema.module.less
```diff
@@ -365,6 +369,8 @@
             -moz-user-select: none;
             -webkit-user-select: none;
             -ms-user-select: none;
+            padding: 2px 4px;
+            border-radius: 4px;
```
Anemy, line 373: Not sure when these started breaking, but these changes fix them.
Before
Screen Shot 2022-02-16 at 7 07 52 PM

After
Screen Shot 2022-02-16 at 7 07 42 PM

packages/compass-indexes/src/components/index-header/index-header.module.less
```diff
@@ -1,21 +1,5 @@
 .index-header {
-  table tr th {
-    cursor: pointer;
-    user-select: none;
-    -webkit-user-select: none;
+  th {
+    padding: 12px 24px;
```
Anemy, line 3: Before:
Screen Shot 2022-02-16 at 7 11 45 PM

After:
Screen Shot 2022-02-16 at 7 22 18 PM


Also not sure when these styles broke, I have a version of 1.21 and they're looking alright in that.

packages/compass-explain-plan/src/components/index-definition-type/index-definition-type.module.less
```diff
@@ -3,6 +3,11 @@
 .index-definition {
   position: relative;
 
+  p {
+    margin: 0;
+    margin-top: 3px;
+  }
```
Anemy, line 9: before
Screen Shot 2022-02-16 at 7 54 20 PM


after
Screen Shot 2022-02-16 at 7 56 20 PM

packages/compass/src/app/styles/reset.less
```diff
@@ -0,0 +1,69 @@
+body {
```
mcasimir, line 1: Is this what LG recommend / uses in the showcase?

packages/compass/src/app/styles/index.less
```diff
@@ -33,6 +38,9 @@ body {
   height: 100vh;
```
mcasimir, line 38: We are going to move these to css-in-js with some `Global` stuff in a different PR?

mcasimir (commented):
This is amazing! I thought we would have needed an amount of small PRs over time to chip away at the global styles, you got rid of all of them in one go!

As you suggested I will try it out to see if something looks off before approving, I'm super hyped now :)

mcasimir (commented):
The labels don't have a margin now, should we add one to `<Label>` by default or we keep that as it is since comes from LG and we try something else?

Screenshot 2022-02-17 at 09 46 40
Screenshot 2022-02-17 at 09 49 35
Screenshot 2022-02-17 at 09 49 17

This link, and possibly others also looks a bit odd, maybe that's the way it should be:

Screenshot 2022-02-17 at 09 56 31

There are also the descriptions on the Create Database modal, we may want to have a quick fix here, and a separate ticket to follow up and fix all of these with some nice factored components for our forms:

Screenshot 2022-02-17 at 10 02 06

packages/compass-query-bar/src/components/query-option/query-option.jsx
```diff
@@ -125,7 +125,7 @@ class QueryOption extends Component {
         className={_className}
         data-test-id="query-bar-option">
         <div
-          className={classnames(styles.label)}
```
alenakhineika, line 128: For what this classnames was used here before?

Anemy: It wasn't doing anything, there are a lot of occurrences in Compass of this, I try to remove them when making changes in the area. The places where `classnames` has a use are when classes need to be concatenated or resolved, places where it's wrapping a single class don't do anything (maybe a super slight perf upgrade by removing them).

packages/compass/src/app/styles/reset.less
Anemy: [I asked leafygreen](https://mongodb.slack.com/archives/CFFA5BS79/p1645111726152939),  they just require one rool. This reset css is more aggressive than we need I think. Ultimately it's up to us, going to see if there's anything we should remove. We can also tweak things overtime and reduce what we have here.

Anemy:
@mcasimir Pushed a fix for the create database modal styles.
Screen Shot 2022-02-17 at 4 24 46 PM

I think in those places where we have labels and need to re-add margin we should add a wrapper for those components that need the labels I think. Is it cool if I do that in the follow up pr that addresses the new connect form styles?

mcasimir:
> @mcasimir Pushed a fix for the create database modal styles.

Considering that we are enabling the form I'd say that we need to make it look decent at least right away. 

We have 2 strategies:
1. just "proxy" the `<Label>` component to add a margin everywhere, fix the other minor issues, and merge this one. Follow up without pressure with a proper fix as part of the LG project.
2. wait after the next beta release Monday to merge this and then right away provide a fix.

I vote for option 1, so we don't need to rush with the "proper" approach to the form components and also we don't need to hold back the release of the form in beta.

PR #2800, opened 2022-02-17
fix(compass-import-export): Picks only specified columns in JSON export COMPASS-4256

leorossi:
## Description
Previously all columns where serialized in JSON (not only `_id`).
This PR fixes this adding a `{ columns }` parameter to the JSON formatter. If empty it will take all fields.
### Checklist
- [X] New tests and/or benchmarks are included

## Motivation and Context

- [X] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-import-export/src/modules/export.spec.js
```diff
@@ -50,24 +45,65 @@ describe('export [module]', () => {
...
+        beforeEach(() => {
+          tempFile = path.join(
+            os.tmpdir(),
+            `test-${Date.now()}.csv`
+          );
+        });
+        afterEach((afterEachDone) => {
+          rimraf(tempFile, afterEachDone);
+        });
+        it('should set the correct fields to CSV export', (done) => {
```
mcasimir, line 59: suggestion: we are trying to standardize towards `async/await` for async code, I know this is a test, so is not that big of a deal, but would it be possible to `promisify` or use Promise based api here instead of `done`?

packages/compass-import-export/src/utils/formatters.js
```diff
@@ -7,23 +7,24 @@ import { EJSON } from 'bson';
 import { serialize as flatten } from './bson-csv';
 import { Transform } from 'stream';
 import { EOL } from 'os';
-
+import pick from 'lodash.pick';
 /**
  * @returns {Stream.Transform}
  */
-export const createJSONFormatter = function({ brackets = true } = {}) {
+export const createJSONFormatter = function({ brackets = true, columns = [] } = {}) {
```
mcasimir, line 14: I think all the filtered fields are selected with a projection, shouldn't get here

PR #2801, opened 2022-02-17
chore(e2e): More E2E tests for aggregations COMPASS-5462

lerouxb:
This also includes a fix for the "click on a database/collection card and nothing happens" flakes - had to manually scroll those into view. click normally scrolls if it has to but in that case for some reason it doesn't help.

TODOs still left outstanding for the future:

* Toggling sample preview more on and off. That's a bit broken anyway and we're going to remove it soon.
* Drag and Drop of stages. Leaving that for someone adventurous in future :)
* save a pipeline, close compass, re-open compass, load the pipeline.
* test Collapse/Expand all stages button (currently broken)

Out of these I think the save a pipeline and re-open it one would probably be most useful. Interestingly I think the state will automatically stick around because we don't clear it out when we open compass in the tests, but we would have to clear it out at some point (or come up with unique names) so that the tests are idempotent.


packages/compass-e2e-tests/tests/instance-databases-tab.test.ts
```diff
@@ -38,12 +42,15 @@ describe('Instance databases tab', function () {
     }
   });
 
-  it('links database cards to the database collections tab', async function () {
+  it.skip('links database cards to the database collections tab', async function () {
```
lerouxb, line 45: This fails almost every time in CI. Here is the screen inside clickVisible at the moment before it actually clicks the card:

database-card
 
I'm skipping it for now until we can figure out what is going on.

packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
```diff
@@ -114,33 +137,275 @@ describe('Collection aggregations tab', function () {
...
+    await browser.waitUntil(async function () {
+      const textElement = await browser.$(
+        Selectors.stagePreviewToolbarTooltip(0)
+      );
+      const text = await textElement.getText();
+      return text === '(Sample of 100 documents)';
+    });
+
+    // save as a view
+    // TODO: This is currently broken, so will have to test at a later stage
```
lerouxb, line 239: We can do this now or after this PR. Actually creating a view works, it is the browsing to the view that was broken.

packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
```diff
@@ -114,33 +137,275 @@ describe('Collection aggregations tab', function () {
...
 
-  // stages can be deleted and the preview is refreshed after disabling them
-  it('allows stages to be deleted');
+    // run the $out stage
+    await browser.clickVisible(Selectors.stageOutSaveButton(0));
+
+    // go to the new collection
+    const linkElement = await browser.$(Selectors.stageOutCollectionLink(0));
+    await linkElement.waitForDisplayed();
+    // TODO: clicking this button crashes at the moment
```
lerouxb, line 321: same as above

packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
```diff
@@ -114,33 +137,275 @@ describe('Collection aggregations tab', function () {
...
-  it('supports creating a view');
+    // run the $out stage
+    await browser.clickVisible(Selectors.stageMergeSaveButton(0));
 
-  // different languages, with and without imports, with and without driver usage
-  it('can export to language');
+    // go to the new collection
+    const linkElement = await browser.$(Selectors.stageMergeCollectionLink(0));
+    await linkElement.waitForDisplayed();
+    // TODO: clicking this button crashes at the moment
```
lerouxb, line 358: Also here

packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
```diff
@@ -114,33 +137,275 @@ describe('Collection aggregations tab', function () {
...
+ * query: The query in MQL.
+ */
+{
+  query
+}`);
+
+    //change $match
+    await browser.setAceValue(Selectors.stageEditor(0), '{ i: { $gt: 5 } }');
+
+    // TODO: click collapse and then expand again
```
lerouxb, line 169: Are we keeping this? Is it worth testing?

packages/compass-e2e-tests/helpers/commands/click-visible.ts
```diff
@@ -1,16 +1,30 @@
...
   await browser.waitForAnimations(selector);
 
+  if (options?.scroll) {
+    const scrollElement = await browser.$(selector);
+    await scrollElement.scrollIntoView();
+    await browser.pause(1000);
+  }
+
   const clickElement = await browser.$(selector);
+  if (options?.screenshot) {
```
lerouxb, line 26: This turns out to be super handy for testing what's actually going on when clicking. Which seems to be where all the bugs materialise, so I think it is worth keeping around.

packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
```diff
@@ -75,7 +98,7 @@ describe('Collection aggregations tab', function () {
     ]);
   });
 
-  // the aggregation runs and the preview is shown
+  // TODO: we can probably remove this one now that there is a more advanced one. or merge that into here?
```
lerouxb, line 101: I'm in two minds about keeping this simple case around or removing it given that the more complicated one does all the same things and more.

gribnoysup (approved):

packages/compass-e2e-tests/tests/instance-databases-tab.test.ts
lerouxb: https://jira.mongodb.org/browse/COMPASS-5529

PR #2809, opened 2022-02-18
feat(export-to-language) add Rust support as a target language for export RUST-1155

abr-egn:
## Description
This adds a Rust target for the language transpilation, enabling export for find and aggregate pipelines.
### Checklist

## Motivation and Context
Rust is a relatively new language rapidly gaining traction in industry usage, and has an officially supported mongodb driver; this provides an easy path to working with that driver for Compass users.
- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions
Rust does not support some constructs referenced in the export machinery (DBRefs, Decimal128), I couldn't tell if there was any better way to flag that than not implementing the relevant templates.

## Dependents
None.

## Types of changes
- [ ] *Backport Needed*


addaleax:
Thank you @abr-egn! :slightly_smiling_face: I’ve opened COMPASS-5527 to track reviewing this on our side, we’ll hopefully get to that soon, and kicked off CI here :+1:

lerouxb:
Please just have a look here: https://github.com/mongodb-js/compass/blob/main/packages/compass-export-to-language/src/components/editor/editor.jsx#L5-L11

That got missed recently with the ruby fix. Throws some runtime error if you select a language that's not loaded there.

abr-egn:
I've updated the editor imports - wouldn't have found that, thanks!

I'm afraid I couldn't make much out of the test failures; they didn't seem at all related to this PR?

addaleax:
> I'm afraid I couldn't make much out of the test failures; they didn't seem at all related to this PR?

Yeah, that might happen sometimes. If test failures are looking unrelated, you can ignore them and see if they stick out to us, since we’ll probably take a look before merging anyway :)

packages/bson-transpilers/test/yaml/bson-methods.yaml
```diff
@@ -35,6 +35,10 @@ tests:
                 BSON::CodeWithScope.new('test code', {
                   'x' => 1
                 }).scope
+            rust: |-
+                JavaScriptCodeWithScope { code: "test code", scope: doc! {
```
patrickfreed, line 39: I think we need a `.to_string()` here after `"test code"`

packages/bson-transpilers/test/yaml/bson-methods.yaml
```diff
@@ -47,6 +51,7 @@ tests:
             shell: new Code('test code').code
             object: "'test code'"
             ruby: BSON::Code.new('test code').javascript
+            # Rust doesn't have a code attribute
```
patrickfreed, line 54: We could use `Bson::JavaScriptCode` here, though I'm not sure if that matches the expected type/usage.

packages/bson-transpilers/test/yaml/driver-syntax.yaml
```diff
@@ -366,6 +390,28 @@ tests:
...
+                      "name": 1
+                    })
+                        .skip(5)
+                        .limit(10)
+                        .max_time(std::time::Duration::from_millis(100))
+                        .collation(doc! {
+                      "locale": "zh"
+                    })
+                        .build();
+                    let result = client.database("db").collection("collection").find(doc! {
```
patrickfreed, line 412: we'll probably want to use `Document` as a generic parameter to `collection` here since I don't think the compiler can infer the type otherwise in this example.

packages/bson-transpilers/test/yaml/native.yaml
```diff
@@ -859,6 +935,7 @@ tests:
           csharp: DateTime.Now
           object: "new Date()"
           ruby: Time.new.utc
+          rust: Utc::now()
```
patrickfreed, line 938: I think we'll need to use the `bson::DateTime::now()` here, since this example requires `chrono`, which we're trying to move away from.

packages/bson-transpilers/test/yaml/native.yaml
```diff
@@ -963,6 +1048,7 @@ tests:
           csharp: new DateTime(1995, 12, 17, 0, 0, 0).ToString("ddd MMM dd yyyy HH':'mm':'ss UTC")
           object: Date(1995, 11, 17)
           ruby: Time.utc(1995, 12, 17, 0, 0, 0).strftime('%a %b %d %Y %H:%M:%S %Z')
+          rust: Utc.ymd(1995, 12, 17).and_hms(0, 0, 0).to_rfc2822()
```
patrickfreed, line 1051: Unfortunately we don't currently have a way to do this with `bson::DateTime`, though we probably should. I filed https://jira.mongodb.org/browse/RUST-1198 to cover the work for that, maybe in the meantime we just omit these examples?

packages/bson-transpilers/test/yaml/native.yaml
```diff
@@ -1016,6 +1106,7 @@ tests:
           csharp: "new Regex(\"(?:)\")"
           object: "RegExp('')"
           ruby: "/(?:)/"
+          rust: "Regex { pattern: \"(?:)\", flags: \"\" }"
```
patrickfreed, line 1109: need `.to_string()` for both of these. ditto for the other regex examples

packages/bson-transpilers/test/yaml/partial.yaml
```diff
@@ -87,6 +92,7 @@ tests:
           csharp: BsonJavaScript
           shell: Code
           ruby: BSON::Code
+          rust: ''
```
patrickfreed, line 95: can we use `Bson::JavaScriptCode` here?

packages/bson-transpilers/test/yaml/partial.yaml
```diff
@@ -175,6 +188,7 @@ tests:
           csharp: DateTime
           shell: Date
           ruby: Time
+          rust: Utc
```
patrickfreed, line 191: ditto using `bson::DateTime` instead of a chrono type

packages/bson-transpilers/test/yaml/partial.yaml
```diff
@@ -204,6 +218,7 @@ tests:
           csharp: new ObjectId().ToString()
           shell: new ObjectId().toString()
           ruby: BSON::ObjectId.new.to_s
+          rust: ObjectId::new().bytes()
```
patrickfreed, line 221: I think we want `to_hex` here

packages/bson-transpilers/test/yaml/bson-methods.yaml
abr-egn: Done, and similarly for uses of `Bson::JavaScriptCode()`.

packages/bson-transpilers/test/yaml/bson-methods.yaml
abr-egn: This is testing the ability to take a code construction (i.e. `Bson::JavaScriptCode("test code".to_string())`) and extract the code string by applying an attribute/accessor, but because we rely on destructuring here there isn't one to apply.

packages/bson-transpilers/test/yaml/driver-syntax.yaml
abr-egn: Good catch, done.

packages/bson-transpilers/test/yaml/native.yaml
abr-egn: Updated.

packages/bson-transpilers/test/yaml/native.yaml
abr-egn: Actually, we can pull the same trick as js - emit the date as a string and parse from that.  More direct constructors would still certainly be a good to have.

packages/bson-transpilers/test/yaml/partial.yaml
abr-egn: This is testing "partial" constructions, i.e. transpilations of fragmentary input.  In this case, it's testing translating the code "symbol" without any arguments.  Because for Rust we have to switch between `JavaScriptCode` and `JavaScriptCodeWithScope` based on whether the scope argument is present, the symbol itself has to be empty.  I don't think this matters much because it's just testing nonsense.

packages/bson-transpilers/test/yaml/partial.yaml
abr-egn: Done.

packages/bson-transpilers/test/yaml/partial.yaml
abr-egn: Done.

mcasimir:
@abr-egn @patrickfreed can you ping us when this PR will be ready for review by DevTools? Thanks!

packages/bson-transpilers/test/yaml/driver-syntax.yaml
```diff
@@ -199,6 +220,9 @@ tests:
                     } = require('mongodb');
                 shell: ''
                 ruby: require 'mongo'
+                rust: |-
+                    use bson::doc;
```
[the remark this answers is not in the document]
abr-egn: Done.

packages/bson-transpilers/symbols/rust/templates.yaml
```diff
@@ -0,0 +1,854 @@
...
+    #                                           #
+    # They follow the same pattern with the
+    # *Template/*ArgsTemplates: usually no args #
+    # to the Template and lhs plus any original #
+    # arguments to the ArgsTemplate.            #
+    #                                           #
+    #############################################
+    CodeCodeTemplate: &CodeCodeTemplate null
+    CodeCodeArgsTemplate: &CodeCodeArgsTemplate null
+    CodeScopeTemplate: &CodeScopeTemplate !!js/function >
```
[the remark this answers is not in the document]
abr-egn: This name is how it's used; I agree that `CodeScopeScopeTemplate` would have made more sense with the naming conventions.  There doesn't seem to be any support for a `CodeScopeCodeTemplate`.

packages/bson-transpilers/symbols/rust/templates.yaml
```diff
@@ -0,0 +1,854 @@
...
+    BinaryValueTemplate: &BinaryValueTemplate !!js/function >
+        (arg) => `${arg}.bytes`
+    BinaryValueArgsTemplate: &BinaryValueArgsTemplate !!js/function >
+        () => ''
+    BinaryLengthTemplate: &BinaryLengthTemplate !!js/function >
+        (arg) => `${arg}.bytes.len()`
+    BinaryLengthArgsTemplate: &BinaryLengthArgsTemplate !!js/function >
+        () => ''
+    BinaryToStringTemplate: &BinaryToStringTemplate !!js/function >
+        (arg) => `std::string::from_utf8(${arg})?`
```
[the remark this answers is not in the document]
abr-egn: I ... honestly have no idea what past-me was thinking here.  Updated to your suggestion.

packages/bson-transpilers/symbols/rust/templates.yaml
```diff
@@ -0,0 +1,854 @@
...
+    TimestampGetHighBitsTemplate: &TimestampGetHighBitsTemplate !!js/function >
+        (arg) => `${arg}.increment`
+    TimestampGetHighBitsArgsTemplate: &TimestampGetHighBitsArgsTemplate !!js/function >
+        () => ''
+    TimestampTTemplate: &TimestampTTemplate !!js/function >
+        (arg) => `${arg}.time`
+    TimestampITemplate: &TimestampITemplate !!js/function >
+        (arg) => `${arg}.increment`
+    TimestampAsDateTemplate: &TimestampAsDateTemplate !!js/function >
+        (arg) => `Utc.timestamp(${arg}.time, 0)`
```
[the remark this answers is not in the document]
abr-egn: Whoops, missed that one.  Fixed.

packages/bson-transpilers/symbols/rust/templates.yaml
```diff
@@ -0,0 +1,854 @@
...
+                    return literal;
+                case 'O':
+                case '0':
+                    return literal.charAt(0) + 'o' + literal.substring(2);
+                default:
+                    return literal.charAt(0) + 'o' + literal.substring(1);
+            }
+        }
+    NumericTypeTemplate: &NumericTypeTemplate null # args: literal, argType
+    ArrayTypeTemplate: &ArrayTypeTemplate !!js/function >
```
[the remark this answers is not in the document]
abr-egn: Looks much nicer, done.

packages/bson-transpilers/symbols/rust/templates.yaml
```diff
@@ -0,0 +1,854 @@
...
+            if (
+                (bytes.charAt(0) === '\'' && bytes.charAt(bytes.length - 1) === '\'') ||
+                (bytes.charAt(0) === '"' && bytes.charAt(bytes.length - 1) === '"')) {
+                newBytes = bytes.substr(1, bytes.length - 2);
+            }
+            newBytes = `"${newBytes.replace(/\\([\s\S])|(")/g, '\\$1$2')}"`;
+            if (type === null) {
+                type = 'BinarySubtype::Generic';
+            }
+            return ` { subtype: ${type}, bytes: ${bytes}.into_bytes() }`
```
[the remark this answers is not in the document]
abr-egn: I don't think it's base64, but there's inconsistency in how the other languages handle it - javascript treats it as a buffer object and converts *that* to base64, csharp treats it as an ASCII string, and java treats it as a UTF-8 string.  Given all that I think you're right that it's best to hold off for now; removed what was there.

patrickfreed (approved):
LGTM from the Rust end!

alenakhineika (approved):

PR #2813, opened 2022-02-20
chore(connection-form): Add html form tag to connection form, handle on textarea enter submit

Anemy:
This pr adds the `form` html tag to the connection form. It now connects when a user hits `enter` in the connection string input.

### DIFF
--- a/packages/connection-form/src/components/connect-form.tsx
+++ b/packages/connection-form/src/components/connect-form.tsx
@@ -48,6 +48,10 @@ const descriptionStyles = css({
   marginTop: spacing[2],
 });
 
+const formStyles = css({
+  display: 'contents'
+});
+
 const formContentContainerStyles = css({
   padding: spacing[4],
   overflowY: 'auto',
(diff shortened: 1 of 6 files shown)


packages/connection-form/src/components/connect-form.tsx
```diff
@@ -128,104 +132,117 @@ function ConnectForm({
...
 
+  const onSubmitForm = useCallback(() => {
+    const updatedConnectionOptions = cloneDeep(connectionOptions);
+    const formErrors = validateConnectionOptionsErrors(
+      updatedConnectionOptions
+    );
+    if (formErrors.length) {
+      setErrors(formErrors);
+      return;
+    }
```
addaleax, line 143: I’m sorry I didn’t see this before, can we maybe merge this a bit with the code that’s now in https://github.com/mongodb-js/compass/blob/4daa8ef4bd75abce4748c4844eebb04748328d93/packages/connection-form/src/components/connect-form.tsx#L135-L142?

addaleax (approved):

packages/connection-form/src/components/connect-form.tsx
Anemy: Merged in main 👍

PR #2816, opened 2022-02-22
fix(connections): hide sslPass in ampersand COMPASS-5239

mabaasit:
fix(connections): hide ssl tunnel password in ampersand

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/connection-model/lib/extended-model.js
+++ b/packages/connection-model/lib/extended-model.js
@@ -32,7 +32,7 @@ const ExtendedConnection = Connection.extend(storageMixin, {
     namespace: 'Connections',
     basepath,
     appName, // Not to be confused with `props.appname` that is being sent to driver
-    secureCondition: (val, key) => key.match(/(password|passphrase|secrets)/i)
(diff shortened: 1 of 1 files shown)


mcasimir (approved):
What would happen if someone used to have sslPass saved on disk? Will it be loaded once from there and then saved to the keychain from the next time?

mabaasit:
> What would happen if someone used to have sslPass saved on disk? Will it be loaded once from there and then saved to the keychain from the next time?

Yes, the `sslPass` will be loaded from the disk and once the user connects, it will be saved in the keychain immediately. It's because we update the `lastUsed` property of the connection.

mcasimir:
> > What would happen if someone used to have sslPass saved on disk? Will it be loaded once from there and then saved to the keychain from the next time?
> 
> Yes, the `sslPass` will be loaded from the disk and once the user connects, it will be saved in the keychain immediately. It's because we update the `lastUsed` property of the connection.

Nice! Let's merge it then!

PR #2820, opened 2022-02-22
chore: Add check for LeafyGreen dependencies outside compass-components package COMPASS-5421

Anemy:
COMPASS-5421

This adds a check to ensure we don't introduce [LeafyGreen](https://github.com/mongodb/leafygreen-ui) dependencies outside of `compass-components`. 

Here's how it looks when this check fails:
Screen Shot 2022-02-22 at 3 38 02 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "scripts": {
     "bootstrap": "npm install && lerna run bootstrap --stream",
     "bootstrap-ci": "npm ci && lerna run bootstrap",
-    "precheck": "npm run depcheck && npm run check-logids",
+    "precheck": "npm run depcheck && npm run check-logids && npm run check-leafygreen-dependency-usage",
     "changed": "node ./scripts/changed.js",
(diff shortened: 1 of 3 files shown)


mcasimir (approved):
👏

gribnoysup:
I have a [branch](https://github.com/mongodb-js/compass/tree/compass-components-eslint) that does this with eslint allowing to see the errors right away and autofix them, it needs a bit of polish I did a poc, but didn't have time to finish, but as we are working on this right now, maybe you want to pick it up as part of this too? It looks like this in the editor when applied:

image

Anemy:
Synced w/ @gribnoysup - we'll keep what's in this PR, and down the line we can look into custom eslint rules for Compass with a check similar to this in it. This check also does some things eslint doesn't since it checks the dependencies in `package.json`, so maybe we'll have both checks at once later. Updated the PR to also check for `emotion` dependencies. Will merge when CI green.

PR #2821, opened 2022-02-22
chore(compass-query-bar): Make export to language a button on query bar instead of dropdown item

Anemy:
This PR gives the export to language an icon button instead of a solo menu option.
Was going to move the query history button to the left of the query input but decided to keep the changes here minimal for now.

Here's how this looks:
Screen Shot 2022-03-01 at 12 24 15 PM


packages/compass-e2e-tests/tests/collection-documents-tab.test.ts
```diff
@@ -252,18 +252,11 @@ describe('Collection documents tab', function () {
...
-    const queryBarActionsMenu = await browser.$(
-      Selectors.queryBarActionsMenu('Documents')
-    );
-    const exportToLanguageButton = await queryBarActionsMenu.$(
-      'a=Export To Language'
-    );
-    await exportToLanguageButton.waitForDisplayed();
-    await exportToLanguageButton.click();
+    await queryBarExportToLanguageButton.waitForDisplayed();
+    await queryBarExportToLanguageButton.click();
```
lerouxb, line 259: You can probably just `await browser.clickVisible(Selectors.queryBarExportToLanguageButton('Documents'));`

Anemy:
This pr adds a dependency on `@mongodb-js/compass-components` to this package. In [compass-components](https://github.com/mongodb-js/compass/blob/main/packages/compass-components/package.json#L37) we have a number of `@leafygreen-ui` packages imported. I'm thinking this might not bundle nicely on cloud, will put this pr on hold for now till we sort that, this was just a drive by fix so doesn't need to be merged anytime soon.
cc @johnjackweir

johnjackweir:
> I'm thinking this [number of @leafygreen-ui packages] might not bundle nicely on cloud

Thanks for keeping InTel's usage of this component in mind! I've checked the leafygreen dependency list and confirmed that MMS already imports most of these dependencies. To incorporate this change as is, MMS would just need to add new major/minor versions of `@leafygreen-ui/logo` and `@leafygreen-ui/toggle` and introduce a new dependency on `@leafygreen-ui/segmented-control`. This _seems_ doable but let me know if there seems to be any complexity I'm missing.

Anemy:
Closing this, we'll address this with the toolbars epic in more detail.

PR #2822, opened 2022-02-23
chore(compass-shell): use LeafyGreen in compass-shell info modal, remove less usage

Anemy:
This PR makes the info modal in `compass-shell` use [LeafyGreen](https://www.mongodb.design/) components. It also removes the use of `less` for styling in `compass-shell`, so now we use [emotion](https://emotion.sh/docs/introduction) from `compass-components` with css in js. Might do a follow up pr a bit later to move this package over to typescript, but I figure these changes make that swap a bit simpler. Should I leave in the `less` webpack style loading webpack rules? This pr removes them since they're now unused, but if y'all think they could be useful there we can keep them.

Video (everything should look the same except the `loading in progress` now has its left margin fixed, and the info modal looks different):

https://user-images.githubusercontent.com/1791149/155265236-b7418c50-c555-4438-ac63-b4bffdae3472.mp4

Before:
Screen Shot 2022-02-22 at 11 29 47 PM

After:
Screen Shot 2022-02-22 at 11 29 31 PM

I'll sync on the styles with Claudia before merging.


packages/compass-shell/src/components/shell-header/shell-header.jsx
```diff
@@ -43,45 +90,41 @@ export class ShellHeader extends Component {
...
+              <span className={operationInProgressStyles}>
                 <ShellLoader
-                  className={styles['compass-shell-header-loader-icon']}
                   size="12px"
-                />
-                <span
-                  className={styles['compass-shell-header-operation-in-progress']}
-                >Command in progress...</span>
-              </>
+                />&nbsp;Command in progress&hellip;
```
addaleax, line 107: Loving these small cleanups here :)

addaleax (approved):

PR #2828, opened 2022-02-23
chore(connection-form): Use label inside checkbox label, update accordion styles

Anemy:
Before:
Screen Shot 2022-02-23 at 4 06 49 PM

After:
Screen Shot 2022-02-23 at 11 54 34 AM

Before:
Screen Shot 2022-02-23 at 4 06 35 PM

After:
Screen Shot 2022-02-23 at 3 52 58 PM


packages/compass-components/src/components/accordion.tsx
```diff
@@ -5,13 +5,9 @@ import Icon from '@leafygreen-ui/icon';
...
 
-const labelStyles = css({
-  padding: 0,
-  margin: 0,
-  fontWeight: 'bold',
-});
-
 const buttonStyles = css({
+  fontWeight: 'bold',
+  fontSize: 14,
```
mcasimir, line 10: How can we centralize this?

Anemy: Added a `compass-font-sizes` file to `compass-components` with this in it. Updated the other places in the connect form to use it as well. It's definitely something only custom components should really be using, but I think at least having it in one place that things reference will keep things more manageable.

mcasimir (approved):

PR #2839, opened 2022-02-25
feat(connection-form): use loose connection string validation

addaleax:
Use loose validation in the connection form UI. This allows
editing in situations like when the URI contains no username
but a password.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/connection-form/src/utils/authentication-handler.ts
+++ b/packages/connection-form/src/utils/authentication-handler.ts
@@ -104,16 +104,18 @@ export function handleUpdateUsername({
     updatedConnectionString.toString()
   );
 
-  if (parsingError) {
+  const hasPasswordWithoutUsername =
+    updatedConnectionString.password && !action.username;
+  if (parsingError || hasPasswordWithoutUsername) {
     return {
       connectionOptions,
       errors: [
         {
           fieldName: 'username',
           fieldTab: 'authentication',
-          message: action.username
+          message: parsingError
             ? parsingError.message
-            : `Username cannot be empty: "${parsingError.message}"`,
+            : 'Username cannot be empty if password is present',
         },
       ],
     };
@@ -148,10 +150,12 @@ export function handleUpdatePassword({
     updatedConnectionString.toString()
   );
 
-  if (parsingError) {
+  const hasPasswordWithoutUsername =
+    action.password && !updatedConnectionString.username;
+  if (parsingError || hasPasswordWithoutUsername) {
     return {
       connectionOptions,
-      errors: connectionStringUrl.username
+      errors: parsingError
         ? [
             {
               fieldName: 'password',
@@ -163,7 +167,7 @@ export function handleUpdatePassword({
             {
               fieldName: 'username',
               fieldTab: 'authentication',
-              message: `Username cannot be empty: "${parsingError.message}"`,
+              message: 'Username cannot be empty if password is present',
             },
             {
               fieldName: 'password',
--- a/packages/connection-form/src/utils/authentication-handler.ts
+++ b/packages/connection-form/src/utils/authentication-handler.ts
@@ -104,18 +104,14 @@ export function handleUpdateUsername({
     updatedConnectionString.toString()
   );
 
-  const hasPasswordWithoutUsername =
-    updatedConnectionString.password && !action.username;
-  if (parsingError || hasPasswordWithoutUsername) {
+  if (parsingError) {
     return {
       connectionOptions,
       errors: [
         {
           fieldName: 'username',
           fieldTab: 'authentication',
-          message: parsingError
-            ? parsingError.message
-            : 'Username cannot be empty if password is present',
+          message: parsingError.message,
         },
       ],
     };
@@ -150,31 +146,16 @@ export function handleUpdatePassword({
     updatedConnectionString.toString()
   );
 
-  const hasPasswordWithoutUsername =
-    action.password && !updatedConnectionString.username;
-  if (parsingError || hasPasswordWithoutUsername) {
+  if (parsingError) {
     return {
       connectionOptions,
-      errors: parsingError
-        ? [
-            {
-              fieldName: 'password',
-              fieldTab: 'authentication',
-              message: parsingError.message,
-            },
-          ]
-        : [
-            {
-              fieldName: 'username',
-              fieldTab: 'authentication',
-              message: 'Username cannot be empty if password is present',
-            },
-            {
-              fieldName: 'password',
-              fieldTab: 'authentication',
-              message: 'Please enter a username first',
-            },
-          ],
+      errors: [
+        {
+          fieldName: 'password',
+          fieldTab: 'authentication',
+          message: parsingError.message,
+        },
+      ],
     };
   }
 
--- a/packages/compass-connect/package.json
+++ b/packages/compass-connect/package.json
@@ -106,7 +106,7 @@
     "mongodb-build-info": "^1.3.0",
     "mongodb-cloud-info": "^1.1.3",
     "mongodb-connection-model": "^21.13.0",
-    "mongodb-connection-string-url": "^2.4.2",
+    "mongodb-connection-string-url": "^2.5.2",
     "mongodb-data-service": "^21.17.0",
     "mongodb-runner": "^4.8.3",
     "mongodb-shell-to-url": "^0.1.0",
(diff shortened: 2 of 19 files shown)


packages/connection-form/src/utils/authentication-handler.ts
```diff
@@ -104,16 +104,18 @@ export function handleUpdateUsername({
...
       connectionOptions,
       errors: [
         {
           fieldName: 'username',
           fieldTab: 'authentication',
-          message: action.username
+          message: parsingError
             ? parsingError.message
-            : `Username cannot be empty: "${parsingError.message}"`,
+            : 'Username cannot be empty if password is present',
```
Anemy, line 118: Does this error here without username because we want it to fail earlier?

Anemy (approved):
lgtm! Nice update.

packages/connection-form/src/utils/authentication-handler.ts
```diff
@@ -148,10 +150,12 @@ export function handleUpdatePassword({
     updatedConnectionString.toString()
   );
 
-  if (parsingError) {
+  const hasPasswordWithoutUsername =
```
mcasimir, line 153: For context, this is not necessary anymore, it was a "special" error handling mechanism necessary only for those updates which were breaking the connection string, now we can move this check with all the others in `validations.ts`.

addaleax: Oh, I see – that’s nice :+1: Are there other cases like these?

mcasimir (commented):
I think i used `new ConnectionString()` somewhere in legacy-connection-model to remove the appName

addaleax:
> I think i used `new ConnectionString()` somewhere in legacy-connection-model to remove the appName

@mcasimir That’s fine though, right? We don’t actually want to be able to store invalid connection strings?

packages/connection-form/src/utils/authentication-handler.ts
mcasimir: Could even be already covered: https://github.com/mongodb-js/compass/blob/76dbf60f21e26f95dd22b714c9d3f25a9e39f57b/packages/connection-form/src/utils/validation.ts#L127

addaleax: Alright, done – I think this works well enough at this point :)

mcasimir: > > I think i used `new ConnectionString()` somewhere in legacy-connection-model to remove the appName
> 
> @mcasimir That’s fine though, right? We don’t actually want to be able to store invalid connection strings?

That's a good question. Well not `pineapple` maybe cause in that case we wouldn't be able to identify secrets, but something well-formed with a password but not a username can be ok. That aligns with all the other "validations" and with the general philosophy of new the form of trying to limit the user as least as possible. If someone wants to save a connection without the username cause they don't remember it, they have to jump to something else and just continue later, etc .. then they will be able to. As long as I'm not missing some downside I'd let users do that.

addaleax: > they have to jump to something else

They could also just leave a dummy in there…

I’d really rather not let users enter connection strings into our storage that don’t parse as valid driver connection string, tbh

mcasimir: That's doable yeah, I'm not super fond of that though. What downside you're thinking of?

The policy of trying to validate and anticipate users' mistakes only lead to issues in the past, with validations out-of-date and assumptions on driver behavior that didn't stay consistent. 

We intentionally adopted a different strategy this time by letting the user connect and save whatever they want (let the driver be the source of truth and do most of the job rejecting invalid connections), introducing warnings to inform users of potential issues, and reducing validations to the minimum.

The only argument that in my mind would support preventing users from saving any connection is the fact that we cherrypick secrets, and adding some validation to prevent new `authMechanism`s would make the save a tiny bit safer, but is not real security, for that we need to encrypt the entire connection string.

Random theories aside, if you want to save only valid connections, that's technically pretty simple, there is some code already happening on connect: https://github.com/mongodb-js/compass/blob/76dbf60f21e26f95dd22b714c9d3f25a9e39f57b/packages/connection-form/src/components/connect-form.tsx#L214-L221, should be enough to also use it in the same file in the `onSave` callback.

addaleax: > What downside you're thinking of?

I’m mostly worried that our own code won’t anticipate that we store data that is valid as a URI but not valid as a connection string. The ability to do `new ConnectionString(str)` is a great tool, and I’d try to keep deviations from that as localized as possible (i.e. here only to the connect-form, not across Compass).

> The policy of trying to validate and anticipate users' mistakes only lead to issues in the past, with validations out-of-date and assumptions on driver behavior that didn't stay consistent.

Does that really apply when we’re sharing the validation code with the driver, though? Isn’t it a *good* thing when we can (reliably) perform the same validation as the driver? Like, that’s the point here – we could let the user store this connection string, but the driver will reject it, and that’s something that we can know for sure.

mcasimir: I see so you mean that we would save string that would then break later, and we won't be able to rely on the fact that `new ConnectionString(str)` is safe.

That is possible to validate, why not. The solution would be the same (`new ConnectionString(str)` right before saving and set the error similarly to what `validate` does).

> we can (reliably) perform the same validation as the driver.
Yeah but for that I find this interface a bit restrictive, and at least at the moment not so useful. The only stuff that validates are the scheme and the password. My dream would be to have `new ConnectionString(str)` having the sole responsability to parse a string and not validating anything, including the scheme. Then a shared `uri.validate()` or something of that effect that we can actually call and returning errors that we can relate to the string params similar to what a schema validation would do (`throw new InvalidConnectionStringError([{path: 'params.tls', message: 'value should be one of ...'}])`).

I don't think having `new ConnectionString` on load doing semantic validations before the connection would be useful, for example, you may have an old connection string saved with a deprecated parameter that would break on load when updating Compass to a new version with a driver that doesn't support it anymore. I would prefer if `new ConnectionString` would have a clear intent and not validate anything. Using a connection string after the connection would imply that is valid already.

addaleax: > I see so you mean that we would save string that would then break later, and we won't be able to rely on the fact that `new ConnectionString(str)` is safe.

Not so much “later”, but rather “in other parts of the code” (than the connection form).

> That is possible to validate, why not. The solution would be the same (`new ConnectionString(str)` right before saving and set the error similarly to what `validate` does).

I’m not sure I understand – Compass currently already fails when trying to save invalid connection strings?

> My dream would be to have `new ConnectionString(str)` having the sole responsability to parse a string and not validating anything, including the scheme. Then a shared `uri.validate()` or something of that effect that we can actually call and returning errors that we can relate to the string params similar to what a schema validation would do (`throw new InvalidConnectionStringError([{path: 'params.tls', message: 'value should be one of ...'}])`).

I think that would be really nice to have, yes. But since this is not the status quo of how we use this or the Node.js driver uses it, we can maybe bring this up with the Node.js team at sync today?

addaleax: Also, note that the `ConnectionString` constructor does actually do some validation that is directly related to parsing (for example, that username and password contain no unescaped characters).

mcasimir: >  Compass currently already fails when trying to save invalid connection strings?

Not that I know: https://github.com/mongodb-js/compass/blob/71af7ab2ddb9c2f9db711991592a619e9875ce47/packages/compass-connections/src/stores/connections-store.ts#L239

We could try to `new ConnectionString(connectionInfo.connectionOptions.connectionString)` before saving, the error should be handled already nicely enough there, a toast will show reporting an error.

addaleax: Hm yeah – it reports an error saving the connection (presumably as part of the secrets extraction?), but it does add it to the favorites list. Attempting to remove it again throws an uncaught exception then – that’s the kind of scenario that I’m worried about here.

For now, I’ve pushed a change here that forbids saving invalid connection strings. I think that’s fair as a requirement for making sure that parts of Compass outside the connect-form can safely interact with the saved connection strings.

mcasimir (approved):

packages/connection-form/src/utils/authentication-handler.ts
mcasimir: > that’s the kind of scenario that I’m worried about here
yeah but it also wouldn't happen if `new ConnectionString(str)` wouldn't do validations

PR #2840, opened 2022-02-25
chore(connection-form): Add default read preference option COMPASS-5540

Anemy:
COMPASS-5540
Adds a `Default` read preference option that's selected when the readPreference is unset.

https://user-images.githubusercontent.com/1791149/155762502-7273c805-1745-40a0-bbb8-f365f2bff6ff.mp4


packages/connection-form/src/utils/read-preferences.ts
```diff
@@ -1,12 +1,18 @@
...
 interface ReadPreference {
   title: string;
-  id: ReadPreferenceMode;
+  id: ReadPreferenceMode | typeof defaultReadPreference;
 }
 
 export const readPreferences: ReadPreference[] = [
+  {
+    title: 'Default',
+    id: defaultReadPreference,
```
mcasimir, line 14: Nit: do we need to overload `id` with the value? Considering that we are also only using this "type" in the advanced tab and is quite tightly coupled with the usage there, we could just delete this file and do this instead:

What about this:

```
const readPreferences = [{
   id: 'default',
   title: 'Default',
   value: ''
}, ...]
```

the code in the other file will be much simpler as the `RadioBox` should handle everything. The only thing still needed will be `value={readPreference ?? ''}` on the `RadioBoxGroup`.

mcasimir (approved):

packages/connection-form/src/utils/read-preferences.ts
Anemy: Updated. Thanks for the suggestion, definitely reduces, sometimes with smaller changes like this its easy to just do it without thinking of a more simple way of implementing it.

PR #2853, opened 2022-03-02
feat(aggregations): toolbar setup

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass/src/index.d.ts
```diff
@@ -44,6 +44,11 @@ declare module 'process' {
          * development
          */
         COMPASS_SHOW_YOUR_QUERIES_TAB?: 'true';
+
+        /**
+         * Enable new aggregation pipeline toolbar
+         */
+        SHOW_NEW_AGGREGATION_TOOLBAR?: 'true',
```
mcasimir, line 51: What about adding the `COMPASS_` prefix like the other flag? ie: `COMPASS_SHOW_NEW_AGGREGATION_TOOLBAR`

packages/compass-aggregations/src/components/pipeline/pipeline.jsx
```diff
@@ -191,6 +193,47 @@ class Pipeline extends PureComponent {
     }
   }
 
+  renderPipelineToolbar() {
+    if (process.env.SHOW_NEW_AGGREGATION_TOOLBAR === 'true') {
+      return <PipelineToolbarLG />;
+    }
+    return (
+      <PipelineToolbar
```
mcasimir, line 201: NIT: We could rename the old one to `<PipelineToolbarLegacy />` and have the new one already as `<PipelineToolbar>`.
Also we can flip the logic and keep the new one outside the flag, so that enabling the flag is just a matter of deleting code:

``` js
if (process.env.SHOW_NEW_AGGREGATION_TOOLBAR !== 'true') {
   return <LegacyPipelineToolbar ...>
}

return <PipelineToolbarLG />;
```

packages/compass-aggregations/src/components/pipeline-toolbar-leafygreen/pipeline-actions.tsx
```diff
@@ -0,0 +1,56 @@
...
+  name,
+  onExportToLanguage,
+  onSavePipeline,
+}) => {
+  return (
+    <div className={pipelineActionsContainerStyles}>
+      <Button className={buttonStyles} variant="primary">
+        Run
+      </Button>
+      <Button className={buttonStyles} onClick={() => onSavePipeline(name)}>
```
mcasimir, line 29: What do you think about splitting this PR in 2? One that only enables the FF and renders an empty component and more following PR that would add features and interactions here?

If we are adding features here I'd expect to see also related tests for it in the same PR, in this case, the number of changes may become significant, if possible I'd consider breaking it in multiple PRs that are going to merge value completed with tests.

packages/compass-aggregations/src/components/pipeline/pipeline.jsx
```diff
@@ -186,9 +192,51 @@ class Pipeline extends PureComponent {
...
           savedPipelinesListToggle={this.props.savedPipelinesListToggle}
-          savedPipeline={this.props.savedPipeline} />
+          savedPipeline={this.props.savedPipeline}
+        />
+      );
+    }
+  }
+
+  renderPipelineToolbar() {
+    if (process.env.COMPASS_SHOW_NEW_AGGREGATION_TOOLBAR !== 'true') {
```
Anemy, line 202: On cloud `process` might be undefined 
```suggestion
    if (global?process?.env?.COMPASS_SHOW_NEW_AGGREGATION_TOOLBAR !== 'true') {
```

mcasimir (approved):

PR #2856, opened 2022-03-02
feat(compass-components): add dark mode flag for using theme in components instead of darkreader COMPASS-5520 

Anemy:
COMPASS-5520
This pr adds a `COMPASS_LG_DARKMODE` feature flag. When enabled this makes it so we theme Compass using our own components, and the `darkMode` parameter with leafygreen components, instead of using `darkreader` to globally theme the entire page. 

Most of the changes in the pr are around wrapping the leafygreen components in `compass-components` with a `withTheme` react high order components. How do y'all feel on how we're structuring these imports/exports now? Any suggestions or thoughts on improving?

It doesn't look great with it on (yet 😉 ),  but this give a way for us to test our theming overtime so we can move towards not needing `darkreader` for themes.

https://user-images.githubusercontent.com/1791149/156402326-8b1c1531-3887-4587-b6f1-75e3c803f7a8.mp4


mcasimir (approved):

packages/compass-components/src/hooks/use-theme.tsx
```diff
@@ -29,4 +29,31 @@ function useTheme(): ThemeState {
...
+  const ComponentWithTheme = (props: T) => {
+    const theme = useTheme();
+
+    const applyTheme = global?.process?.env?.COMPASS_LG_DARKMODE === 'true';
+
+    return (
+      <WrappedComponent
+        // Set the darkMode before the props so that the props can
+        // override the theme if needed.
+        darkMode={applyTheme ? theme?.theme === Theme.Dark : undefined}
```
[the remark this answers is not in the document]
Anemy: I'd prefer it to be unset in the case where down the line leafygreen has their own theme provider which supplies themes to components. In that case if we explicitly set the theme here we might be indicating to leafygreen that the color is overridden so they don't use the provider's themes.

packages/compass-components/src/components/leafygreen.tsx
```diff
@@ -0,0 +1,178 @@
...
+  Overline,
+  Label as LeafyGreenLabel,
+  Link,
+  Description as LeafyGreenDescription,
+} from '@leafygreen-ui/typography';
+
+// 2. Wrap the components that accept darkMode with Compass' theme.
+const Button = withTheme(
+  LeafyGreenButton as React.ComponentType<typeof LeafyGreenButton>
+) as typeof LeafyGreenButton;
```
Anemy, line 71: We were seeing two errors occurring when using `withTheme` to wrap the leafygreen components. It appears it would try to compare the inferred react prop types with the typescript prop definition and fail. To get around this we explicitly cast them to their leafygreen types. I'm not sure if there's a better way to do this or to avoid this typecast. Do y'all know any?

Here's examples of the two errors we were seeing:
```
Exported variable 'Toggle' has or is using name 'BaseToggleProps' from external module "/Users/rhys/Documents/mongodb/compass/node_modules/@leafygreen-ui/toggle/dist/Toggle" but cannot be named.
```

```
Argument of type 'typeof Toggle' is not assignable to parameter of type 'ComponentType<ToggleProps>'.
  Type 'typeof Toggle' is not assignable to type 'FunctionComponent<ToggleProps>'.
    Types of property 'propTypes' are incompatible.
      Type '{ size: Requireable<string>; darkMode: Requireable<boolean>; checked: Requireable<boolean>; disabled: Requireable<boolean>; className: Requireable<...>; onChange: Requireable<...>; onClick: Requireable<...>; }' is not assignable to type 'WeakValidationMap<ToggleProps> | undefined'.
        Type '{ size: Requireable<string>; darkMode: Requireable<boolean>; checked: Requireable<boolean>; disabled: Requireable<boolean>; className: Requireable<...>; onChange: Requireable<...>; onClick: Requireable<...>; }' is not assignable to type 'WeakValidationMap<Omit<BaseToggleProps & Omit<HTMLElementProps<"button", never>, keyof BaseToggleProps>, "aria-label" | "aria-labelledby"> & Required<...> & Partial<...>>'.
          The types of 'size[nominalTypeHack]' are incompatible between these types.
            Type '{ type: string | null | undefined; } | undefined' is not assignable to type '{ type: Size | null | undefined; } | undefined'.
              Type '{ type: string | null | undefined; }' is not assignable to type '{ type: Size | null | undefined; }'.
                Types of property 'type' are incompatible.
                  Type 'string | null | undefined' is not assignable to type 'Size | null | undefined'.
                    Type 'string' is not assignable to type 'Size | null | undefined'.
```

packages/connection-form/src/components/advanced-options-tabs/general-tab/scheme-input.tsx
```diff
@@ -32,7 +32,7 @@ const regularSchemaDescription =
 const srvSchemaDescription =
```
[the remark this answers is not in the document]
Anemy: Updated and updated all other occurrences in the connection-form.

packages/compass-components/src/components/leafygreen.tsx
Anemy: We'll want to use this wrapper with some of our own custom components - checking to make sure those types work well too.

mcasimir: Ok, let me know when i can review again

Anemy: Added `withTheme` to the `FileInput` as an example/test of using the theme in a custom component, works smooth. I spent a while trying to get these types for these leafygreen components to work nicely yesterday and was having trouble finding a clean way without casting explicitly here. Do you think it's a blocker for merging? I'd be down to pair on looking at the `withTheme` hoc to see if there's anything we can do to avoid casts. If we do merge I'd definitely want to solidify these types before the epic is done, but for now we would be able to unblock theming other parts of this epic.

It appears typescript tries to compare the inferred react prop types with the typescript prop definition and that results in a typing/compile error. To get around this we explicitly cast them to their leafygreen types.

Here's examples of the two errors we were seeing:
```
Exported variable 'Toggle' has or is using name 'BaseToggleProps' from external module "/Users/rhys/Documents/mongodb/compass/node_modules/@leafygreen-ui/toggle/dist/Toggle" but cannot be named.
```

```
Argument of type 'typeof Toggle' is not assignable to parameter of type 'ComponentType<ToggleProps>'.
  Type 'typeof Toggle' is not assignable to type 'FunctionComponent<ToggleProps>'.
    Types of property 'propTypes' are incompatible.
      Type '{ size: Requireable<string>; darkMode: Requireable<boolean>; checked: Requireable<boolean>; disabled: Requireable<boolean>; className: Requireable<...>; onChange: Requireable<...>; onClick: Requireable<...>; }' is not assignable to type 'WeakValidationMap<ToggleProps> | undefined'.
        Type '{ size: Requireable<string>; darkMode: Requireable<boolean>; checked: Requireable<boolean>; disabled: Requireable<boolean>; className: Requireable<...>; onChange: Requireable<...>; onClick: Requireable<...>; }' is not assignable to type 'WeakValidationMap<Omit<BaseToggleProps & Omit<HTMLElementProps<"button", never>, keyof BaseToggleProps>, "aria-label" | "aria-labelledby"> & Required<...> & Partial<...>>'.
          The types of 'size[nominalTypeHack]' are incompatible between these types.
            Type '{ type: string | null | undefined; } | undefined' is not assignable to type '{ type: Size | null | undefined; } | undefined'.
              Type '{ type: string | null | undefined; }' is not assignable to type '{ type: Size | null | undefined; }'.
                Types of property 'type' are incompatible.
                  Type 'string | null | undefined' is not assignable to type 'Size | null | undefined'.
                    Type 'string' is not assignable to type 'Size | null | undefined'.
```

Anemy: @mcasimir Pushed an update to make the components that use the `withTheme` wrapper require an optional `darkMode` prop. Also added an explicit cast to the generic type in the hoc.
Re https://stackoverflow.com/a/58473012

PR #2857, opened 2022-03-02
WIP: smoke tests for document CRUD

addaleax:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-e2e-tests/helpers/selectors.ts
+++ b/packages/compass-e2e-tests/helpers/selectors.ts
@@ -268,6 +268,7 @@ export const InsertDocumentOption =
   '[data-test-id="insert-data-dropdown-insert-document"]';
 export const ImportFileOption =
   '[data-test-id="insert-data-dropdown-import-file"]';
+export const DocumentListEntry = '[data-test-id="editable-document"]';
 
 // Insert Document modal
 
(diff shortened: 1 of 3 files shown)


packages/compass-e2e-tests/tests/collection-import.test.ts
```diff
@@ -95,6 +95,10 @@ describe('Collection import', function () {
     await browser.connectWithConnectionString('mongodb://localhost:27018/test');
   });
 
+  beforeEach(async function () {
+    await browser.shellEval('db.getSiblingDB("test").getCollection("json-array").deleteMany({});');
```
lerouxb, line 99: So far I've been deliberately making different collections for different tests, but I can't really think of a reason why I'm doing that (other than maybe that you could examine the db after tests to debug? bit of a weak reason) so I suppose this is fine.

lerouxb (approved):
Looks good!

packages/compass-e2e-tests/tests/collection-import.test.ts
lerouxb: I think I avoided using shellEval early on because it was very flaky at the time. Seems good now, so we should use it more.

PR #2858, opened 2022-03-03
chore(e2e): connection form tests COMPASS-5338 

lerouxb:
For each connection we test this will:
* Insert the connection string and make sure that the form reflects it
* then reset the form again, set all those form fields and make sure we get the same connection string back out.

This should do several things:
* tests the parsing of valid connection strings and filling of the form
* tests all the "happy paths" of inserting things in the form
* will alert us if any of the selectors we use in the tests have to be changed

**This PR doesn't perform the actual connect at the end yet.** The plan is to gradually change the connection strings (and expected resulting form state) over to real connections that we can connect to. Starting with things we can test locally and against atlas, then to all the docker machines we have available and so on. We could then also easily disable the connect parts based on which environment variables are set. ie. only test against docker images on ubuntu for now, etc.

This also doesn't test warnings and errors yet. Just the known happy paths.

I tried to hit every possible form field. This is not a complete cartesian product of every possible combination of state. We can easily expand it over time if this misses anything important.


lerouxb:
Aaand after merging in main I just noticed that one or two selectors already has to be changed. Will fix in a moment..

lerouxb:
The failures seem to be old flakes.

packages/compass-e2e-tests/helpers/commands/set-value-visible.ts
```diff
@@ -5,7 +5,18 @@ export async function setValueVisible(
...
-  await element.setValue(value);
+  await browser.waitUntil(
+    async () => {
+      const element = await browser.$(selector);
+      await element.waitForDisplayed();
+      await element.clearValue();
+      await element.setValue(value);
+      const actualValue = await element.getValue();
+      if (actualValue !== value) {
+        console.log(actualValue, '!==', value);
```
lerouxb, line 16: Trying to debug this. Some race condition where sometimes it appends the connection string on at the end rather than clearing it first. Just leaving this console.log in to keep an eye on it.

mcasimir (approved):

PR #2861, opened 2022-03-04
feat(connect-form): Allow saving invalid but well-formed strings

mcasimir:
This came up from a chat with @mmarcon and requests by customers on slack. The gist of the change is:

- We don't do semantic validations on save. Only well-formed strings are saved, ie. no `pineapple`, but users can save connections that would not work with the driver, like those missing passwords.
- If a connection has been saved in the past then we will only update the `lastUsed` field instead of saving all the changes. NOTE: this replicates the behavior of the old connection form.
- Unrelated: Changes the copy for TLS warnings to a more positive language.
Screenshot 2022-03-04 at 11 56 51
Screenshot 2022-03-04 at 11 57 00
Screenshot 2022-03-04 at 11 57 19
Screenshot 2022-03-04 at 12 57 33

### DIFF
--- a/packages/compass-connections/src/stores/connections-store.ts
+++ b/packages/compass-connections/src/stores/connections-store.ts
@@ -32,6 +32,14 @@ export function createNewConnectionInfo(): ConnectionInfo {
   };
 }
 
+function ensureWellFormedConnectionString(connectionString: string) {
+  new ConnectionString(
+    connectionString,
+
+    { looseValidation: true }
+  );
+}
+
 function setAppNameParamIfMissing(
   connectionString: string,
   appName: string
@@ -232,10 +240,17 @@ export function useConnections({
   const connectedConnectionInfo = useRef<ConnectionInfo>();
   const connectedDataService = useRef<DataService>();
 
-  async function saveConnectionInfo(connectionInfo: ConnectionInfo) {
+  async function saveConnectionInfo(
+    connectionInfo: ConnectionInfo
+  ): Promise<boolean> {
     try {
+      ensureWellFormedConnectionString(
+        connectionInfo?.connectionOptions?.connectionString
+      );
       await connectionStorage.save(connectionInfo);
       debug(`saved connection with id ${connectionInfo.id || ''}`);
+
+      return true;
     } catch (err) {
       debug(
         `error saving connection with id ${connectionInfo.id || ''}: ${
@@ -250,6 +265,8 @@ export function useConnections({
           (err as Error).message
         }`,
       });
+
+      return false;
     }
   }
 
@@ -383,7 +400,11 @@ export function useConnections({
       });
     },
     async saveConnection(connectionInfo: ConnectionInfo) {
-      await saveConnectionInfo(connectionInfo);
+      const saved = await saveConnectionInfo(connectionInfo);
+
+      if (!saved) {
+        return;
+      }
 
       const existingConnectionIndex = connections.findIndex(
         (connection) => connection.id === connectionInfo.id
--- a/packages/compass-connections/src/stores/connections-store.ts
+++ b/packages/compass-connections/src/stores/connections-store.ts
@@ -2,6 +2,7 @@ import type {
   ConnectionInfo,
   ConnectionOptions,
   DataService,
+  ConnectionStorage,
 } from 'mongodb-data-service';
 import { getConnectionTitle } from 'mongodb-data-service';
 import { useEffect, useReducer, useRef } from 'react';
@@ -29,11 +30,6 @@ export function createNewConnectionInfo(): ConnectionInfo {
       connectionString: 'mongodb://localhost:27017',
     },
   };
-}
-export interface ConnectionStore {
-  loadAll: () => Promise<ConnectionInfo[]>;
-  save: (connectionInfo: ConnectionInfo) => Promise<void>;
-  delete: (connectionInfo: ConnectionInfo) => Promise<void>;
 }
 
 function setAppNameParamIfMissing(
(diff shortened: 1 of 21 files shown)


packages/compass-connections/src/stores/connections-store.ts
```diff
@@ -263,9 +278,15 @@ export function useConnections({
     try {
       onConnected(connectionInfo, dataService);
 
-      // Update lastUsed date as now and save the connection.
-      connectionInfo.lastUsed = new Date();
-      await saveConnectionInfo(connectionInfo);
+      // if a connection has been saved already we only want to update the lastUsed
+      // attribute, otherwise we are going to save the entire connection info.
+      const connectionInfoToBeSaved =
+        (await connectionStorage.load(connectionInfo.id)) ?? connectionInfo;
```
Anemy, line 284: Do we need to reload the entire connection here? Would it make sense to check for `lastUsed` instead?

packages/data-service/src/connection-storage.ts
```diff
@@ -67,6 +67,28 @@ export class ConnectionStorage {
     const model = await convertConnectionInfoToModel(connectionOptions);
     model.destroy();
   }
+
+  /**
+   * Updates attributes of the model.
+   *
+   * @param {string} id ID of the model to update
+   * @param {object} attributes Attributes of model to update
```
Anemy, line 75: I think this comment is from another method

packages/compass-connections/src/stores/connections-store.ts
mcasimir: we need to reload for the update, but we could do a check first, something like this?

``` js
const connectionInfoToBeSaved = connectionInfo.lastUsed && (await connectionStorage.load(connectionInfo.id)) ?? connectionInfo;
```

Anemy: Ah gotcha we are reloading here to pull in the original state of the connection then so the changes that are currently there do not override. This is probably a nicer way to do it than keeping around a clone of the original state of the connection then, I think both have implications if a user is doing things in another window.

mcasimir: We are only replicating the old behavior here, i was surprised this is how the old connection model behaved.

Anemy: I think the suggestion I made for using `lastUsed` might not actually do anything then? Sorry!

Anemy (approved):
lgtm - do we want to show an indicator that there are unsaved changes similar to how the old form did? Or does the `Save` button functionality cover that now?
Screen Shot 2022-03-04 at 12 28 31 PM

packages/compass-connections/src/stores/connections-store.ts
mcasimir: ah true, actually it would cause the entire thing to be saved on connect the first time, even if it was saved before, but just never used

mcasimir: Changed it back. Let's wait for @addaleax before merging, It would be great to have some feedback, cause she wasn't fond of relaxing the validation too much. We can easily drop the `looseValidation` from `ensureWellFormedConnectionString`

packages/connection-form/src/utils/validation.ts
```diff
@@ -288,15 +285,20 @@ function validateCertificateValidationWarnings(
...
+  const tlsAllowInvalidHostnames =
+    connectionString.searchParams.get('tlsAllowInvalidHostnames') === 'true';
+  const tlsAllowInvalidCertificates =
+    connectionString.searchParams.get('tlsAllowInvalidCertificates') === 'true';
+
+  if (tlsInsecure || tlsAllowInvalidHostnames || tlsAllowInvalidCertificates) {
     warnings.push({
-      message:
-        'Disabling certificate validation is not recommended as it may create a security vulnerability',
+      message: `TLS/SSL certificate validation is disabled. For a more secure connection enable certificate validation if possible.`,
```
addaleax, line 301: I did like that this explicitly used the phrase “vulnerability”. Usually, that’s what it is, and “more secure connection” makes it sound like the connection still gives some security guarantees, which it effectively does not.

It’s nice to use positive language, but we should really push users towards never using these options.

packages/connection-form/src/utils/validation.ts
```diff
@@ -393,7 +395,7 @@ function validateTLSAndHostWarnings(
   if (nonLocalhostsCount && !isSecure(connectionString)) {
     warnings.push({
       message:
-        'Connecting without tls is not recommended as it may create a security vulnerability.',
+        'Connecting to a remote server without TLS/SSL is not recommended. For a more secure connection enable TLS/SSL if possible.',
```
addaleax, line 398: Ditto here

packages/connection-form/src/utils/validation.ts
mcasimir: @mmarcon what's your opinion here? For this PR I can just keep the old copy and maybe do another one.

mmarcon: What if we said something like:
```
TLS/SSL certificate validation is disabled. If possible, enable certificate validation to avoid security vulnerabilities.
```

mcasimir: Going to merge this one and open another PR to discuss the copy for TLS

addaleax: I think that’s fine, yeah.

(Think about how browsers [do this](https://wrong.host.badssl.com/) – in both Chrome and Firefox you’ll need to go to an  `Advanced` section and are only then presented with the option to actually go to the site. And here, this potentially affects our customer’s production databases, where the risk probably tends to be higher-impact then a person visiting a bad-HTTPS site.)

PR #2866, opened 2022-03-07
feat(aggregations): toolbar actions COMPASS-5534

mabaasit:
feat(aggregations): toolbar actions COMPASS-5534

Implement new toolbar, results workspace (where we will show the complete result of aggregation). 

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-actions.tsx
```diff
@@ -3,26 +3,90 @@ import { connect } from 'react-redux';
...
-      <Button className={buttonStyles} variant="primary">
-        Run
+      {workspace === 'builder' && (
+        <Button
+          data-testid="toolbar-run-action-button"
+          className={buttonStyles}
+          variant="primary"
+          onClick={() => {
+            onChangeWorkspace('results');
+            onRunAggregation();
```
gribnoysup, line 37: This mixes business logic and UI in a way that should be avoided, if changing workspace should trigger a run, this should happen from a single action, see [these](https://redux.js.org/style-guide/style-guide#avoid-dispatching-many-actions-sequentially) [two](https://redux.js.org/style-guide/style-guide#model-actions-as-events-not-setters) redux style guide recommendations

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-settings.tsx
```diff
@@ -0,0 +1,125 @@
...
+});
+
+const toggleLabelStyles = css({
+  marginBottom: 0,
+  textTransform: 'uppercase',
+  marginLeft: spacing[1],
+});
+
+const PipelineSettings: React.FunctionComponent<PipelineSettingsProps> = ({
+  isCollationExpanded,
```
gribnoysup, line 53: Does this need to be part of the redux state?

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-stages.tsx
```diff
@@ -1,33 +1,46 @@
...
   stages,
+  onStageAdded,
 }) => {
   if (stages.length === 0) {
     return (
       <Description className={noStagesTextStyles}>
+        Your pipeline is currently empty. To get started select the{' '}
+        <Link onClick={() => onStageAdded()} hideExternalIcon>
+          first stage.
+        </Link>
```
gribnoysup, line 36: This should be a button, right?

packages/compass-aggregations/src/components/pipeline/pipeline.jsx
```diff
@@ -239,6 +240,14 @@ class Pipeline extends PureComponent {
     return <PipelineToolbar />;
   }
 
+  renderPipelineWorkspace() {
+    return this.props.workspace === 'results' ? (
+      <PipelineResultsWorkspace />
+    ) : (
+      <PipelineBuilderWorkspace {...this.props} />
```
gribnoysup, line 247: Are all these props used by `PipelineBuilderWorkspace` component?

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-settings.tsx
mabaasit: yeah to show the toggle icon on `collation` button.

packages/compass-aggregations/src/components/pipeline/pipeline.jsx
mabaasit: definitely no. wanted to keep the existing implementation as is. but now as I'm digging into these props, I may've to   fix this.

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-stages.tsx
gribnoysup: You can do `<Link as="button" ...` and apply a few additional styles to reset the styles to be normal link (we do this in a few other places). Visually it should be a link, yes, but semantically this is a button and we should try to follow these rules

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-stages.tsx
```diff
@@ -0,0 +1,87 @@
...
+          className={addStageStyles}
+          as="button"
+          onClick={() => onStageAdded()}
+          hideExternalIcon
+        >
+          first stage.
+        </Link>
+      </Description>
+    );
+  }
```
gribnoysup, line 54: This never shows up because we always have at least one empty stage:

image

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-stages.tsx
```diff
@@ -0,0 +1,87 @@
...
+  return (
+    <div className={cx(containerStyles, stagesStyles)}>
+      <Pipeline size="small">
+        {stages.map((stage) => (
+          <Stage key={stage}>{stage}</Stage>
```
gribnoysup, line 59:
```suggestion
        {stages.map((stage, idx) => (
          <Stage key={`${idx}${stage}`}>{stage}</Stage>
```

gribnoysup, line 59: Should we show some text when this is empty or filter those out? Currently looks a bit weird:

image

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+          <Body weight="medium" className={pipelineTextStyles}>
+            Pipeline
+          </Body>
+          <div
+            role="button"
+            onClick={() => onShowSavedPipelines()}
+            className={openSavedPipelinesStyles}
+            aria-label="Open saved pipelines"
+            tabIndex={0}
+          >
```
gribnoysup, line 59: This needs to be a `button`, not a `div` with a role. Buttons come with a lot accessibility features, not only click handlers, otherwise you need to reimplement all the accessibility features that come with a `button` element yourself and it's a lot of work

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-actions.tsx
```diff
@@ -0,0 +1,87 @@
...
+} from '@mongodb-js/compass-components';
+
+import type { RootState } from '../../../modules';
+import { runAggregation } from '../../../modules/aggregation';
+import { changeWorkspace } from '../../../modules/workspace';
+import { toggleOptions } from '../../../modules/options';
+
+const containerStyles = css({
+  display: 'flex',
+});
```
gribnoysup, line 20: This needs a `flex: none` or `flex-shrink: 0` disabled (you should generally do this every time you use flex just for positioning and don't want containers to change their size) because right now it starts to collapse when you have a decent amount of pipeline stages:

image

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-actions.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+const defaultMargins = css({
+  marginLeft: spacing[1],
+  marginRight: spacing[1],
+});
```
gribnoysup, line 25: If you use `:not(:first-child)` and just required left margin instead you will not be left with weird margins that affect anything outside of this section. You can also use `gap` CSS property on the container and then your components don't need to know anything about their margins at all

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-actions.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+const optionsButtonStyles = css({
+  backgroundColor: 'transparent',
+  border: 'none',
+});
```
gribnoysup, line 30: You probably want underline to be disabled too here

packages/compass-aggregations/src/typings.d.ts
```diff
@@ -1 +1,4 @@
 declare module '@mongodb-js/mongodb-redux-common/app-registry';
+declare module 'mongodb-ace-autocompleter';
+declare module 'mongodb-query-parser';
+declare module 'semver';
```
gribnoysup, line 4: At least for semver is it possible to install [type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/semver) instead of completely ignoring the types of this package?

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-settings/pipeline-name.tsx
```diff
@@ -0,0 +1,55 @@
...
+      {name === '' ? (
+        'Untitled'
+      ) : (
+        <Tooltip
+          align="top"
+          justify="start"
+          trigger={() => <div className={nameStyles}>{name}</div>}
+        >
+          {name}
+        </Tooltip>
```
gribnoysup, line 40: This tooltip never shows because trigger fn in tooltip component has props that you have to manually pass to the childred when used the way you are using it here

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-settings/pipeline-extra-settings.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <div className={toggleStyles}>
+          <Toggle
+            id="auto-preview"
+            size="small"
+            aria-label="Toggle Auto Preview"
+            onChange={() => onToggleAutoPreview()}
+          />
+          <Label className={toggleLabelStyles} htmlFor="auto-preview">
+            Auto Preview
+          </Label>
```
gribnoysup, line 50: Seems like this needs some style adjustments due to bootstrap styles

image

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-settings/pipeline-menus.tsx
```diff
@@ -0,0 +1,169 @@
...
+          variant="primary"
+          size="xsmall"
+          leftGlyph={<Icon glyph={glyph} />}
+          onClick={(evt) => {
+            evt.stopPropagation();
+            onClick();
+          }}
+        >
+          {title}
+          {isMenuOpen && children}
```
gribnoysup, line 73: IIRC you don't need to do this yourself, it should be ok to always render children and not remove them from the component tree manually (this is probably why the closing animation is broken right now)

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+            <Icon glyph="Folder" />
+            <Icon glyph="CaretDown" />
+          </div>
+        </div>
+        <PipelineStages />
```
gribnoysup, line 64: This group of elements has its spacing definitions a bit all over the place, you have this pipeline button have most of the margins responsible for spacing here, but then the pipeline has it's own small margin (that will not stack because this is a flex container)

image
image

As a rule of thumb I really suggest you to aim for defining spacing in a way that will not mess with the layout if elements are moved around, used in another context, or removed completely (it's really conceptually similar to this other comment I left). In this case I would suggest that you remove the margin from the pipeline component and wrap it in something and have a spacing for it defined here, that way not only your pipeline component can be moved around easier, you also have one place that is responsible for defining layout of the toolbar, not three like it is right now

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-stages.tsx
mabaasit: I was thinking of filtering it out. Not sure. I'll check with Claudia what's better here.

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-actions.tsx
mabaasit: Switched to grid. And the results are much better. Thanks

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <div className={pipelineActionStyles}>
+        <PipelineActions />
+      </div>
+    </div>
+  );
+};
+
+const mapDispatch = (dispatch: Dispatch) => ({
+  onShowSavedPipelines: () => {
+    // todo: fix dispatch
```
mabaasit, line 72: Leaving this as is here - this is the old implementation and I don't want to fix this in this PR (as it has already exploded the limits). Fixing this will require changing it at many places.

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-builder-workspace.spec.jsx
```diff
@@ -113,6 +114,7 @@ function createPipelineWorkspace({
     projections={projections}
     projectionsChanged={projectionsChanged}
     newPipelineFromPaste={newPipelineFromPaste}
+    workspace={workspace}
```
gribnoysup, line 117: Why do you need to pass it here?

packages/compass-aggregations/src/components/pipeline-toolbar/index.spec.tsx
```diff
@@ -0,0 +1,281 @@
...
+describe('PipelineToolbar', function () {
+  describe('renders toolbar', function () {
+    let toolbar: HTMLElement;
+    beforeEach(function () {
+      process.env.COMPASS_SHOW_NEW_AGGREGATION_TOOLBAR = 'true';
+      const store = configureStore({});
+      render(
+        <Provider store={store}>
+          <Aggregations />
+        </Provider>
```
gribnoysup, line 34: Seems like these tests test and rely on a lot of things that are outside of the actual toolbar implementation. This means that any changes outside the toolbar will break these tests and it will be hard to debug if something really broke or it is some unrelated changes. If these are new tests that we are adding, shouldn't we focus on testing just the toolbar behavior that we expect here?

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/index.tsx
gribnoysup: You don't need to refactor whole stores, but at least adding a new thunk action export that keeps track of all the logic required to open a saved pipelines list from saved pipeline actions that combines the two instead of directly using them here would be better. That way when we eventually refactoring this code away we will at least not need to hunt down every reference to these setter actions everywhere and will just have one import here that already matches what we want it to do.

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-actions.tsx
```diff
@@ -0,0 +1,84 @@
...
+      </Button>
+      <Link
+        as="button"
+        className={optionsButtonStyles}
+        data-testid="pipeline-toolbar-options-button"
+        hideExternalIcon={true}
+        onClick={() => onToggleOptions()}
+      >
+        <div className={optionStyles}>
+          {isOptionsVisible ? 'Less' : 'More'} Options{' '}
```
gribnoysup, line 65: Because the text size here changes, the Run button above jerks a little when clicking, should we add some min-width to the whole button here so this doesn't happen?

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+import {
+  getSavedPipelines,
+  savedPipelinesListToggle,
+} from '../../../modules/saved-pipeline';
+
+const containerStyles = css({
+  display: 'grid',
+  gap: spacing[4],
+  gridTemplateAreas: '"pipelineTextAndOpen pipelineStages pipelineAction"',
+  gridTemplateColumns: 'min-content',
```
gribnoysup, line 18: This only applies size restrictions to the first column here, you need to either grid auto columns or explicitly specify size for every column here. Without this you have two columns now that try to get auto sized and this looks like this:

image

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-stages.tsx
```diff
@@ -0,0 +1,110 @@
...
+  margin: 'inherit !important',
+});
+
+const PipelineStages: React.FunctionComponent<PipelineStagesProps> = ({
+  isEditing,
+  stages,
+  onStageAdded,
+  onChangeWorkspace,
+}) => {
+  if (stages.filter(Boolean).length === 0) {
```
gribnoysup, line 57: So now that we filter out empty stages I do see this message when opening a new aggregation with a single empty stage:

image

I intuitively clicked the button to add a new stage because the message said so and nothing changed (because it's another empty stage) and this made me really confused at first, I thought something broke, clicked it a few more times, and only then noticed that I added like 10 more empty stages below the existing empty one.

I wonder if we should think through this behavior here a bit more and sync with Max (or Claudia when she's back)

packages/compass-aggregations/src/components/pipeline/collation-toolbar.jsx
```diff
@@ -73,7 +79,7 @@ class CollationToolbar extends PureComponent {
               styles['collation-toolbar-input-label'],
               { [ styles['has-error'] ]: (this.props.collation === false) }
             )}
-            data-test-id="collation-toolbar-input-label">
+            data-testId="collation-toolbar-input-label">
```
gribnoysup, line 82:
```suggestion
            data-testid="collation-toolbar-input-label">
```

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-stages.tsx
mabaasit: I had a chat about this with @mcasimir and this seems fine. However, maybe we can add an extra if to validate this use case and then show corresponding message?

packages/compass-aggregations/src/modules/options.ts
```diff
@@ -0,0 +1,22 @@
...
+export default function reducer(state = INITIAL_STATE, action: ToggleOptionAction): State {
+  if (action.type === ActionTypes.ToggleOption) {
+    return !state;
+  }
+  return state;
+}
+
+export const toggleOptions = (): ToggleOptionAction => ({
+  type: ActionTypes.ToggleOption,
+});
```
gribnoysup, line 22: So before we used existing collation slice state and actions because collation was not part of our new toolbar, I remember us talking about that and this makes sense. Now collation input is completely encapsulated  in the new toolbar so why are we using redux for a state that can be local component state and doesn't need to be global?

gribnoysup (approved):
Looks awesome! I think you might want to assign and close COMPASS-5548 and COMPASS-5547 tickets too as you also implemented these in this PR

PR #2870, opened 2022-03-08
chore: fix react prop warnings

Anemy:
Fixes a few react props warnings, mistypes, old-now-gone-away props, or places where the store isn't setup yet so it passes undefined.
Here's what they looked like:
Screen Shot 2022-03-08 at 10 29 31 AM
Screen Shot 2022-03-08 at 10 29 45 AM
Screen Shot 2022-03-08 at 10 32 32 AM


mcasimir (approved):
👍

PR #2872, opened 2022-03-08
chore(e2e): browser.scrollToVirtualItem() COMPASS-5529

lerouxb:
The virtual scrolling used on the database and collection lists mean that collections that are off screen might not exist in the page. So we have to keep scrolling down until they are loaded. This command attempts to do that.

This PR also re-enables the test that we had disabled due to a card being too far off screen before.

### DIFF
--- a/packages/compass-e2e-tests/helpers/commands/scroll-to-virtual-item.ts
+++ b/packages/compass-e2e-tests/helpers/commands/scroll-to-virtual-item.ts
@@ -1,0 +1,53 @@
+import type { CompassBrowser } from '../compass-browser';
+
+export async function scrollToVirtualItem(
+  browser: CompassBrowser,
+  itemsSelector: string,
+  targetSelector: string
+): Promise<boolean> {
+  let found = false;
+
+  let lastId: string;
+
+  // Start by scrolling down, then scroll back up. This is necessary in case we
+  // started off somewhere in the middle and the item we're looking for is above
+  // it.
+  for (const direction of [1, -1]) {
+    await browser.waitUntil(async () => {
+      const targetElement = await browser.$(targetSelector);
+      if (await targetElement.isExisting()) {
+        await targetElement.waitForDisplayed();
+        await targetElement.scrollIntoView();
+        // the item is now visible, so stop scrolling
+        found = true;
+        return true;
+      }
+
+      const elements = await browser.$$(itemsSelector);
+      const edgeElement = elements[direction === 1 ? elements.length - 1 : 0];
+      await edgeElement.waitForDisplayed();
+      await edgeElement.scrollIntoView();
+
+      const thisId = await edgeElement.getAttribute('data-id');
+
+      if (thisId === lastId) {
+        // once we reach the end, go back up the other direction
+        // NOTE: this would be the first place to look if this starts flaking.
+        // Not sure if new items are guaranteed to have appeared immediately
+        // after we scrolled to the first/last one so we might need a better way
+        // of detecting that we reached the end before going back up the other way.
+        return true;
+      }
+
+      // keep scrolling
+      lastId = thisId;
+      return false;
+    });
+
+    if (found) {
+      return true;
+    }
+  }
+
+  return found;
+}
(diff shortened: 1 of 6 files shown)


packages/compass-e2e-tests/helpers/commands/scroll-to-virtual-item.ts
```diff
@@ -0,0 +1,53 @@
...
+export async function scrollToVirtualItem(
+  browser: CompassBrowser,
+  itemsSelector: string,
+  targetSelector: string
+): Promise<boolean> {
+  let found = false;
+
+  let lastId: string;
+
+  // Start by scrolling down, then scroll back up. This is necessary in case we
```
mcasimir, line 12: Isn't it possible to just move back to `y=0` first and then just scroll down?

packages/compass-e2e-tests/helpers/commands/scroll-to-virtual-item.ts
```diff
@@ -0,0 +1,53 @@
...
+        found = true;
+        return true;
+      }
+
+      const elements = await browser.$$(itemsSelector);
+      const edgeElement = elements[direction === 1 ? elements.length - 1 : 0];
+      await edgeElement.waitForDisplayed();
+      await edgeElement.scrollIntoView();
+
+      const thisId = await edgeElement.getAttribute('data-id');
```
mcasimir, line 31: Can't we save a reference to the element directly instead of relying on `data-id`?

packages/compass-e2e-tests/helpers/commands/scroll-to-virtual-item.ts
```diff
@@ -0,0 +1,53 @@
...
+
+      if (thisId === lastId) {
```
mcasimir, line 33: probably not important and i'm not sure this can happen, but what would happen if both are undefined (like someone removed `data-id`)

mcasimir (approved):

packages/compass-e2e-tests/helpers/commands/scroll-to-virtual-item.ts
lerouxb: how?

packages/compass-e2e-tests/helpers/commands/scroll-to-virtual-item.ts
lerouxb: how? every time you look up the element it is a different reference.

packages/compass-e2e-tests/helpers/commands/scroll-to-virtual-item.ts
lerouxb: waitUntil will time out after 5 seconds, twice.

packages/compass-e2e-tests/helpers/commands/scroll-to-virtual-item.ts
mcasimir: > every time you look up the element it is a different reference

I see, is it not like a real dom tree with the same selector returning always the same node?

packages/compass-e2e-tests/helpers/commands/scroll-to-virtual-item.ts
mcasimir: that much I don't know, of course, that's why I was asking :). I would imagine you could use js of the window as a last resort:

``` js
browser.execute(function() {
    document.querySelector(containerSelector).scrollTop = 0;
});
```

Just to be clear, this is a nit, looks good to me as it is, feel free to merge.

packages/compass-e2e-tests/helpers/commands/scroll-to-virtual-item.ts
lerouxb: If we looked up the card as `[data-testid="collection-grid-item"][data-id="${dbName}.${collectionName}"]` and waited for it to exist or be displayed it would time out the exact same way if it never existed, so I think that's fine :)

packages/compass-e2e-tests/helpers/commands/scroll-to-virtual-item.ts
lerouxb: I thought about that. There's an element in the dom somewhere that seems to have the total height so that it would render scrollbars and that's probably the one that gets scrolled. Doesn't have a name or id or anything so we'd have to try and find it by some "parent of something" or "child of something" trickery, then access it with browser.execute() and passing code snippets..

Can probably work, but feels even more hacky to me and then the code has to know even more about the exact inner workings of the component and it works even less like the way a human would scroll to the card..

I'm on the fence, but this seems to work, so could go with this for now and just change it later.

PR #2878, opened 2022-03-09
chore(e2e): Don't leak secrets in E2E tests COMPASS-5569

lerouxb:


packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -581,3 +579,32 @@ export async function afterTest(
...
+  for (const field of SENSITIVE_ENV_VARS) {
+    if (process.env[field] === undefined) {
+      continue;
+    }
+
+    const quoted = `'${process.env[field] as string}'`;
+    // /regex/s would be ideal, but we'd have to escape the value to not be
+    // interpreted as a regex.
+    while (value.indexOf(quoted) !== -1) {
+      value = value.replace(quoted, "'<redacted>'");
```
addaleax, line 599: Maybe this could be 

```suggestion
      value = value.replace(quoted, "'$" + field + "'");
```

to be more meaningful if/when we add other sensitive env vars?

addaleax (approved):

packages/compass-e2e-tests/helpers/compass.ts
lerouxb: Nice.

lerouxb:
Will deal with that flake separately.

PR #2889, opened 2022-03-10
chore(e2e): Don't do the "first run experience" every time COMPASS-5262

lerouxb:
Basically: only calculate a userDataDir once. Only delete it if/when we ask for the "firstRun" experience explicitly and then during cleanup after all the tests.

### DIFF
--- a/packages/compass-e2e-tests/helpers/commands/exists-eventually.ts
+++ b/packages/compass-e2e-tests/helpers/commands/exists-eventually.ts
@@ -3,7 +3,7 @@ import type { CompassBrowser } from '../compass-browser';
 export async function existsEventually(
   browser: CompassBrowser,
   selector: string,
-  timeout = 10000
+  timeout = 1000
 ): Promise<boolean> {
   try {
     // return true if it exists before the timeout expires
--- a/packages/compass-e2e-tests/helpers/commands/exists-eventually.ts
+++ b/packages/compass-e2e-tests/helpers/commands/exists-eventually.ts
@@ -3,7 +3,7 @@ import type { CompassBrowser } from '../compass-browser';
 export async function existsEventually(
   browser: CompassBrowser,
   selector: string,
-  timeout = 1000
+  timeout = 10000
 ): Promise<boolean> {
   try {
     // return true if it exists before the timeout expires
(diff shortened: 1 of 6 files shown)


packages/compass-e2e-tests/helpers/commands/exists-eventually.ts
```diff
@@ -3,7 +3,7 @@ import type { CompassBrowser } from '../compass-browser';
 export async function existsEventually(
   browser: CompassBrowser,
   selector: string,
-  timeout = 10000
+  timeout = 1000
```
lerouxb, line 6: This seems to be faster than before, but best to keep an eye on it. It is unfortunate that we have to wait to see if a modal appears. Although technically we now know if we expect it or not, but that's still a bit hacky.

Point is that this is time added to every test because it has to reach this timeout to see if there are modals to interact with.

As things are written now, anyway. Could change things down the line.

lerouxb: Ugh. 1 second is not long enough in CI for when it does show up, so I'll have to rethink this now.

lerouxb:
The windows flake is just the same windows flake we're trying to fix elsewhere: https://github.com/mongodb-js/compass/pull/2891#issuecomment-1065200352

packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -1,6 +1,6 @@
 import { inspect } from 'util';
 import { ObjectId, EJSON } from 'bson';
-import { promises as fs } from 'fs';
+import { promises as fs, rmdirSync } from 'fs';
```
addaleax, line 3: Is there a specific reason that this method needs to be synchronous?

addaleax (approved):

packages/compass-e2e-tests/helpers/compass.ts
lerouxb: cleanup() in index.ts is synchronous and I need to call it there.

lerouxb: See https://github.com/mongodb-js/compass/blob/a58edee9ed9bc7d522a2cd6ccdccc45e2407cb67/packages/compass-e2e-tests/index.ts#L50-L74

PR #2867, opened 2022-03-07
feat(compass-saved-aggregations-queries): remove feature flag COMPASS-5565

mabaasit:
remove `COMPASS_SHOW_YOUR_QUERIES_TAB` feature flag to enable `My Queries` screen by default

### Checklist
- [x] New tests and/or benchmarks are included
- [ ] Documentation is changed or added

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*
- [ ] Patch (non-breaking change which fixes an issue)
- [x] Minor (non-breaking change which adds functionality)
- [ ] Major (fix or feature that would cause existing functionality to change)

PR #2891, opened 2022-03-11
fix: change collection name to appear first

mabaasit:
this PR fixes the [broken test](https://github.com/mongodb-js/compass/runs/5493888002?check_suite_focus=true#step:9:3023) that was introduced in #2867 

Due to less scroll area of sidebar databases, the `database.collection` is not visible in the viewport and test fails.
Here, I changed the collection name so that it appears first and the test passes. 

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-e2e-tests/tests/instance-sidebar.test.ts
+++ b/packages/compass-e2e-tests/tests/instance-sidebar.test.ts
@@ -137,7 +137,7 @@ describe('Instance sidebar', function () {
 
   it('can create a collection and drop it', async function () {
     const dbName = 'test'; // existing db
-    const collectionName = 'my-sidebar-collection';
+    const collectionName = 'a-sidebar-collection';
 
     await browser.clickVisible(Selectors.SidebarFilterInput);
     const sidebarFilterInputElement = await browser.$(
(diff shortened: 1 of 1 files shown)


lerouxb:
lol

packages/compass-e2e-tests/tests/instance-sidebar.test.ts
```diff
@@ -137,7 +137,7 @@ describe('Instance sidebar', function () {
 
   it('can create a collection and drop it', async function () {
     const dbName = 'test'; // existing db
-    const collectionName = 'my-sidebar-collection';
+    const collectionName = 'a-sidebar-collection';
```
lerouxb, line 140: why not just search in the sidebar so it is the only one being displayed?

mabaasit: ah damn. i didn't know that we can also search by collection name. that sounds much better. I always tried `db.collection` but of course it didn't work.

Anemy (approved):

PR #2894, opened 2022-03-11
chore(compass-schema): remove `mms-icon` usage in compass-schema

Anemy:
We only have a few `mms-icon` uses in Compass. This pr removes the one in compass-schema. The other occurrences are in the query bar and sidebar. Since those will be handled in seperate leafygreen projects I figured this one could go now so that when the others are removed we can remove the global `mms-icon` styles: https://github.com/mongodb-js/compass/blob/main/packages/compass/src/app/styles/mms-icons.less 

Also brings an improvement of accessibility, using a `button` and not an `i` for handling clicking in the dom.

Before:
Screen Shot 2022-03-11 at 3 42 30 PM

After:
Screen Shot 2022-03-11 at 3 43 01 PM

### DIFF
--- a/packages/compass-schema/src/components/compass-schema/compass-schema.module.less
+++ b/packages/compass-schema/src/components/compass-schema/compass-schema.module.less
@@ -328,25 +328,6 @@
 
       dt {
         color: @gray3;
-        padding-top: 12px;
-
-        a {
-          color: @gray5;
-          display: inline-block;
-
-          &:hover {
-            text-decoration: none;
-            color: @gray1;
-          }
-        }
-        i.mms-icon-continuous { // remove after wrapping this in an <a> again
-          color: @gray5;
-          cursor: pointer;
-
-          &:hover {
-            color: @gray1;
-          }
-        }
       }
       dd {
         overflow: auto;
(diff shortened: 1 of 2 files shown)


mcasimir (commented):
In the screenshot you sent the icon is not centred vertically anymore, would it be possible to keep the alignment before merging?

Anemy:
@mcasimir It's not actually vertically centered currently, the screenshot looks like it by chance. The visual differences are caused by the padding of the icon button to show its focus/hover/active state. We could re-add the `12px` padding top we're removing, but I think it looks better without it personally. How do you feel? Could do somewhere in the middle.

Keeping the padding:
Screen Shot 2022-03-14 at 11 41 15 AM

Without padding (current pr changes):
Screen Shot 2022-03-14 at 11 40 50 AM

mcasimir:
> @mcasimir It's not actually vertically centered currently, the screenshot looks like it by chance. The visual differences are caused by the padding of the icon button to show its focus/hover/active state. We could re-add the `12px` padding top we're removing, but I think it looks better without it personally. How do you feel? Could do somewhere in the middle.

I see so is a bit skewed intentionally, looks a bit odd, but I agree with you we are improving it, I guess we can't just merge. What happens when there is only 1 line? Another option would be to align it to the top?

mcasimir (approved):

PR #2896, opened 2022-03-14
feat(connect-form): limit recent connections to 10

mcasimir:
List of changes:
- derive pre-sorted favorite and recent connections lists from loaded connections
- on connect: delete the last recent if a new one was saved and 10 were already existing
- use `favoriteConncections` and `recentConnections` in the connection list
- isolate `<ResizableSidebar>` from `<ConnectionList>`

### DIFF
--- a/packages/compass-connections/src/components/resizeable-sidebar.tsx
+++ b/packages/compass-connections/src/components/resizeable-sidebar.tsx
@@ -1,17 +1,10 @@
-import React, { useState } from 'react';
+import React, { useCallback, useState } from 'react';
 import {
   ResizeHandle,
   ResizeDirection,
   uiColors,
   css,
-  spacing,
 } from '@mongodb-js/compass-components';
-import type { ConnectionInfo } from 'mongodb-data-service';
-
-import ConnectionList from './connection-list/connection-list';
-
-const initialSidebarWidth = spacing[4] * 10 + spacing[2]; // 248px
-const minSidebarWidth = spacing[4] * 7; // 168px
 
 const listContainerStyles = css({
   display: 'flex',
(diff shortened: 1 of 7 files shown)


packages/compass-connections/src/stores/connections-store.ts
```diff
@@ -283,6 +321,20 @@ export function useConnections({
...
+      // NOTE: there are edge cases that may lead to more than MAX_RECENT_CONNECTIONS_LENGTH
+      // to be saved (ie. concurrent run of Compass),
+      // however we accept it as long as the list of
+      // recents won't grow indefinitely.
+      if (
+        !connectionInfoToBeSaved.favorite &&
+        !connectionInfoToBeSaved.lastUsed &&
+        recentConnections.length >= MAX_RECENT_CONNECTIONS_LENGTH
+      ) {
+        await removeConnection(recentConnections[recentConnections.length - 1]);
```
mcasimir, line 336: NOTE: this technically abuses a little the fact that `recentConnections` are already ordered in the right way. This may change, for example, if we introduce some way to reorder or filter that list dynamically .. not really something we are planning for now, but I could easily extract a `getRecentsByLastUsed(connections)` and use it both here and in `useMemo` instead.

Let me know if you have some thoughts about it.

packages/compass-connections/src/stores/connections-store.spec.ts
```diff
@@ -75,6 +91,163 @@ describe('use-connections hook', function () {
...
+      await result.current.connect({
+        id: 'new',
+        connectionOptions: {
+          connectionString: 'mongodb://new-recent',
+        },
+      });
+
+      // this may cause a false negative, but there is no other reliable way to
+      // test this case. It would fail eventually if the functionality is broken.
+      await util.promisify(setTimeout)(100);
```
mcasimir, line 244: anything better?

packages/compass-connections/src/components/resizeable-sidebar.tsx
```diff
@@ -1,84 +1,57 @@
...
-function ResizableSidebar({
-  activeConnectionId,
-  connections,
-  createNewConnection,
-  setActiveConnectionId,
-  onConnectionDoubleClicked,
-  removeAllRecentsConnections,
-  duplicateConnection,
-  removeConnection,
+const ResizableSidebar = ({
```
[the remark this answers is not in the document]
Anemy: Makes sense, definitely shareable, will address with that sidebar work 👍

packages/compass-connections/src/stores/connections-store.spec.ts
gribnoysup: You can write the test in `onConnected` and use `done` similar to how we usually test callback behavior, right?

mcasimir: If I understand it right, `onConnected` is called before any of this happens

lerouxb:
The e2e test flake will hopefully go away if you merge in main: https://github.com/mongodb-js/compass/pull/2899

Or it will at least help us debug what's going on if it happens again.

packages/compass-connections/src/stores/connections-store.spec.ts
gribnoysup: Would it make sense to move the callback and call it only after everything else finished there successfuly then maybe? Otherwise you're communicating that connection happened and then do a bunch of stuff that can fail in the same hook, is it sorta expected?

mcasimir: Yeah we could do something Iike that. I think is intentional that one doesn't wait for the other though.

Probably better to have a `Promise.all()` to make it clear, but yeah if saving "recent" won't work users should still be able to connect and move on to the next screen.

@Anemy what's your feedback on this?

To be clear that test won't flake and it would quite likely fail all the time if a recent gets removed, but is not technically guaranteed to be red if the feature is broken.

packages/compass-connections/src/components/resizeable-sidebar.tsx
```diff
@@ -1,84 +1,57 @@
...
+  const getMaxSidebarWidth = useCallback(() => {
+    return Math.max(minWidth, window.innerWidth - 100);
+  }, [minWidth]);
 
   return (
     <div
       className={listContainerStyles}
       style={{
-        minWidth: width,
+        minWidth: minWidth,
```
Anemy, line 40: I'm not sure if it's this change, but it looks like the sidebar now defaults to a narrower width:
(beta below, this branch top) - is that intended?
Screen Shot 2022-03-16 at 12 59 57 PM

Anemy (commented):
Nice, I really like the improvements in the resizable sidebar usage and connections-store.
One comment on the sidebar width which might not be intended? Other than that looks good to go

packages/compass-connections/src/stores/connections-store.ts
```diff
@@ -266,44 +287,68 @@ export function useConnections({
...
-    ) {
-      void onConnectSuccess(
-        connectedConnectionInfo.current,
-        connectedDataService.current
-      );
-    }
-  }, [isConnected, onConnected]);
+  const onConnectSuccess = useCallback(
+    async (connectionInfo: ConnectionInfo, dataService: DataService) => {
+      // After connecting and the UI is updated we notify the rest of Compass.
```
Anemy, line 307: I think this comment can go away now that we've made this connect success not occur on a `useEffect`?

packages/compass-connections/src/stores/connections-store.spec.ts
Anemy: I see what you mean it's a bit hard to test.
I think it makes sense **not** to wait for a file system access or entry before moving the user past the connect screen? Probably super fast on most machines but could cause some delay.
Can't think of a cleaner way without bringing in extra logic or debugs in the flow, I'm alright with the setTimeout here as it's only in one test and will probably never fail, can always improve it later if somehow it does flake.

packages/compass-connections/src/components/resizeable-sidebar.tsx
mcasimir: Yeah, you're right! Will fix it

packages/compass-connections/src/stores/connections-store.spec.ts
mcasimir: > this flow has completed and then monitoring debug logs in the test work

Seems equally convoluted, If we have to change the code for this test maybe we could add an optional callback like `onConnectionProcessComplete` or something? That we could do, not really useful outside the tests but is kinda clean.

packages/compass-connections/src/stores/connections-store.ts
```diff
@@ -266,44 +287,68 @@ export function useConnections({
...
+        ) {
+          await removeConnection(
+            recentConnections[recentConnections.length - 1]
+          );
+        }
+      } catch (err) {
+        debug(
+          `error occurred connection with id ${connectionInfo.id || ''}: ${
+            (err as Error).message
+          }`
```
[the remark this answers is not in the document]
Anemy: Ah totally missed that pr when it happened! Nice - all good

Anemy (approved):
🔥

PR #2898, opened 2022-03-14
chore(e2e): generate coverage for E2E tests COMPASS-5247

lerouxb:
When running E2E tests and building compass in there (right now NOT when testing the packaged app), webpack/babel will use istanbul to instrument the code to store the global `__coverage__` variable in both the main and renderer processes.

Then in the tests we pull those out before closing compass and store it as files in `.log/coverage`

Afterwords we use `nyc merge` to merge it all together as `.nyc_output/coverage.json`. (see `npm run coverage-merge`)

This coverage.json file will also be uploaded as an artifact in evergreen and github actions if it exists.

Then you can use `nyc report` (see `npm run coverage report`) to turn it into an html report in `packages/compass-e2e-tests/coverage/lcov-report` and you can then `open packages/compass-e2e-tests/coverage/lcov-report/index.html` to view it.

---

The problem in CI is that I think we only test the packaged app and then we don't have coverage, so no coverage in CI at the moment.


packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -444,9 +482,11 @@ export async function rebuildNativeModules(
 export async function compileCompassAssets(
   compassPath = COMPASS_PATH
 ): Promise<void> {
+  process.env.COVERAGE = 'true';
```
lerouxb, line 485: So basically: When we compile compass here, we get coverage. But not when testing the packaged app. So I guess right now this is only really useful when run locally?

packages/compass-e2e-tests/package.json
```diff
@@ -22,7 +22,9 @@
     "insert-data": "ts-node ./scripts/insert-data.ts",
     "unzip-fixtures": "ts-node ./scripts/gunzip.ts fixtures/*.gz",
     "testslow": "env DEBUG=hadron*,mongo*,compass* npm run test -- --disable-start-stop --bail",
-    "testfast": "env DEBUG=hadron*,mongo*,compass* npm run test -- --no-native-modules --no-compile --disable-start-stop --bail"
+    "testfast": "env DEBUG=hadron*,mongo*,compass* npm run test -- --no-native-modules --no-compile --disable-start-stop --bail",
+    "coverage-merge": "nyc merge .log/coverage .nyc_output/coverage.json",
+    "coverage-report": "npm run coverage-merge && nyc report --reporter=lcov --cwd ../.. --temp-dir ./packages/compass-e2e-tests/.nyc_output --report-dir ./packages/compass-e2e-tests/coverage"
```
lerouxb, line 27: I had to set --cwd to make it take in all the packages, but then once you set that you have to set --temp-dir and --report-dir as well because those are interpreted relative to --cwd as well.

packages/compass/src/main/logging.ts
```diff
@@ -120,6 +120,11 @@ async function setupLogging(compassApp: typeof CompassApplication) {
       return app.getPath('logs');
     });
 
+    // TODO: there must be a better place to put this
```
lerouxb, line 123: No idea where would be the best place to put this.

mcasimir: yeah, it doesn't really belong here, what's wrong about putting it in its own file?

lerouxb: which package?

configs/webpack-config-compass/src/loaders.ts
```diff
@@ -41,6 +41,13 @@ export const javascriptLoader = (args: ConfigArgs) => ({
           require.resolve('@babel/plugin-proposal-decorators'),
           { legacy: true },
         ],
+        args.coverage && [
```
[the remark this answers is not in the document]
gribnoysup: I think having it here makes sense, what feels a bit out of place is the `cwd` option as we probably only want it applied like that for `packages/compass` not for all packages that use this configuration. Maybe have a separate argument added for that and passed to the config in `packages/compass/webpack.config.js` instead of always set to the root of the monorepo? Don't have a strong preference for that though

packages/compass-e2e-tests/package.json
gribnoysup: Tiny nit / suggestion: maybe makes sense to use a [config file](https://github.com/istanbuljs/nyc#configuration-files) instead of inline args for ease of reading the options? And then you can even have these context comments there if you use yaml or js config format

configs/webpack-config-compass/src/loaders.ts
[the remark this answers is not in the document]
lerouxb: What if I make args.coverage an optional string and if it is set we set it to the path we want to generate coverage for?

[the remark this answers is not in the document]
gribnoysup: Sounds good too, yeah

packages/compass-e2e-tests/helpers/compass.ts
lerouxb: I guess that's the only way. Kinda like going back to running e2e tests against main except with coverage? Wondering if I should do it in this PR or another.

lerouxb:
Coverage can be downloaded here now: https://evergreen.mongodb.com/task/10gen_compass_main_ubuntu_e2e_coverage_patch_ef2081e1e715949377e27599d1c86fe2188ec8b6_62307bc63e8e863e38c59dbe_22_03_15_11_43_04

The mac flake is new (and in a recently new test), but unrelated to this PR.

The windows flake is the one @mabaasit is working on.

gribnoysup (approved):

PR #2899, opened 2022-03-15
more assertions for the advanced url options

lerouxb:
Still trying to fix this flake:

screenshot-Connection-form-parses-and-formats-a-URI-with-advanced-options


lerouxb:
Notarize just failed three times for mac in one go :(

addaleax (approved):

PR #2900, opened 2022-03-16
feat(compass-workspace): Update workspace tabs to Leafy Green in compass-workspace COMPASS-5518

Anemy:
COMPASS-5518

This PR updates the workspace tabs in Compass. It adds a new package `compass-workspace` with the tabs in it. These are now required from `compass-collection` where they were previously. The plan with `compass-workspace` is to have a higher level concept for tabs so they aren't bound to a collection. Please share any thoughts y'all might have on this.

I'll be syncing with Claudia for a last pass on styles before merging so I can bring any feedback there - let me know if you try it out and think of anything. I think one thing that might change is we'll add a minimum width for the tabs so they don't resize then the title grows or shrinks a bit.

These updates to tabs should bring more accessibility - if y'all see any ways we could improve please share. Here's a resource with more information: https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html

`env COMPASS_LG_DARKMODE="true" npm start` for component darkmode.
Screen Shot 2022-03-16 at 12 40 54 AM

https://user-images.githubusercontent.com/1791149/158519730-f1cc35bb-094a-4ec6-89c7-124f7e82c29f.mp4

*Scroll now overlays, styling in video a little out of date:

https://user-images.githubusercontent.com/1791149/158519733-eea4cb08-2400-43c1-a77c-8330ad640046.mp4

Has full borders now now, styling in video a little out of date:

https://user-images.githubusercontent.com/1791149/158519735-0c6d8a1c-748b-4f01-8732-22cd4c2fac24.mp4


packages/compass-components/src/components/workspace-container.tsx
```diff
@@ -34,7 +32,10 @@ function UnthemedWorkspaceContainer({
   darkMode,
   children,
   'data-test-id': dataTestId,
-}: WorkspaceContainerProps) {
+  ...props
+}: React.PropsWithChildren<
+  WorkspaceContainerProps & React.HTMLProps<HTMLDivElement>
+>) {
```
Anemy, line 38: Unrelated drive by type improvement I accidentally scooped in this branch 😅

packages/compass-workspace/src/components/tab.tsx
```diff
@@ -0,0 +1,314 @@
...
+    content: '""',
+    pointerEvents: 'none',
+    right: -2,
+    bottom: 0,
+    left: -2,
+    border: '3px solid transparent',
+    borderTopWidth: 0,
+    borderBottomWidth: 0,
+    transition: 'border-color .16s ease-in',
+  },
```
gribnoysup, line 54: This doesn't look like it's showing up or do anything, it is a leftover?

packages/compass-workspace/src/components/tab.tsx
```diff
@@ -0,0 +1,314 @@
...
+        onClick={(e) => {
+          e.stopPropagation();
+          onClose();
+        }}
+        aria-label="Close Tab"
+      >
+        <Icon glyph="X" role="presentation" />
+      </IconButton>
+      <div
+        role="presentation"
```
gribnoysup, line 302: It's not bad to have this role here, but you don't really need it for nodes without content

packages/compass-workspace/src/components/workspace-tabs.tsx
```diff
@@ -0,0 +1,347 @@
...
+  )
+);
+
+type TabProps = {
+  namespace: string;
+  id: string;
+  activeSubTabName: string;
+  isActive: boolean;
+  isTimeSeries: boolean;
+  isReadonly: boolean;
```
gribnoysup, line 233: Does it makes sense to have this as part of the component interface? I know that it was like that before, but aren't we only interested in collection type here (like view or timeseries) similar to how it is done in the sidebar navigation

packages/compass-workspace/src/components/tab.tsx
```diff
@@ -0,0 +1,314 @@
...
+        darkMode ? tabDarkThemeStyles : tabLightThemeStyles,
+        {
+          [selectedTabStyles]: isSelected,
+          [focusedTabStyles]: isFocused,
+        }
+      )}
+      aria-selected={isSelected}
+      role="tab"
+      // The tab navigation is handled by the lab list.
+      tabIndex={-1}
```
gribnoysup, line 241: Because tabIndex never changes on these elements when using keyboard to navigate them with VO enabled you never get informed that the tab is changed

gribnoysup:
> The plan with compass-workspace is to have a higher level concept for tabs so they aren't bound to a collection. Please share any thoughts y'all might have on this.

My feeling is that it makes sense now because tabs are used by collection plugin and nothing else, but feels like when tabs are really "global" we will have a separate package with 3 components that will be used in compass-home and nowhere else

packages/compass-workspace/src/components/workspace-tabs.tsx
mcasimir: To go one step forward, is there a way to completely split the code that "knows" about the domain of what the tab content is doing from the one that renders / moves and locates the tabs?

I would imagine something like this (ideally and very pseudo-codeish):

```
<WorkspaceTabs tabs={[
    {
       id: 'my-tab-123',
       header: <CollectionTabHeader ns="my-db.my-coll" type="timeseries"> 
       content: <CollectionTabContent selectedSubTab={...}>
    }
]}>
```

packages/compass-collection/src/components/workspace/workspace.jsx
```diff
@@ -247,22 +186,13 @@ class Workspace extends PureComponent {
...
-              createNewTab={this.onCreateNewTab}
-            />
-          </div>
-          <div className={styles['workspace-tabs-right']}>
-            <div onClick={this.props.nextTab} className={styles['workspace-tabs-next']}>
-              <i className="fa fa-chevron-right" aria-hidden/>
-            </div>
-          </div>
-        </div>
+        <WorkspaceTabs
```
mcasimir, line 189: This may be a consequence of my lack of context here, but isn't there a bit of mix and match of responsibilities and knowledge split between this component and WorkspaceTabs right now, what is the idea behind both?

packages/compass-workspace/src/components/workspace-tabs.spec.tsx
```diff
@@ -0,0 +1,244 @@
...
+      );
+    });
+
+    it('should render all of the tab namespaces', function () {
+      expect(screen.getByText('mock-tab-1')).to.be.visible;
+      expect(screen.getByText('mock-tab-2')).to.be.visible;
+      expect(screen.getByText('mock-tab-3')).to.be.visible;
+    });
+
+    it('should render all of the tab subtab names', function () {
```
mcasimir, line 95: Why do we want this component to do this?

Anemy: We want the tabs to contains things like `Document` or `Explain Plan` that indicate the contents of the tab.
This test checks that the tabs rendered by `workspace-tabs` contain the correct information passed through. It's definitely a bit of a coupled test as it pairs the rendering in `tab` with `workspace-tabs`, but I think that helps build some guarantee that things are rendered and passed through. Does that answer the question?

packages/compass-workspace/src/components/workspace-tabs.tsx
gribnoysup: Yeah, something like this would be even better, I agree

packages/compass-collection/src/components/workspace/workspace.jsx
Anemy: What is the knowledge split here?
I think the idea is that `WorkspaceTabs` is a more straightforward rendering of the tabs for the workspace, and this `workspace.jsx` manages the rendering of the tabs and some of the global keyboard actions like creating a new tab with `ctrl/cmd + t`. If I'm not following your train of thought let's hop on a call?

packages/compass-workspace/src/components/tab.tsx
```diff
@@ -0,0 +1,314 @@
...
+  border: '1px solid',
+  borderTop: 'none',
+  borderBottom: 'none',
+  transition: 'border-color .16s ease-out',
+  display: 'inline-flex',
+  flexDirection: 'row',
+  alignItems: 'center',
+  margin: 0,
+  paddingTop: spacing[2],
+  paddingBottom: spacing[1] + spacing[2],
```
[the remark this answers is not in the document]
Anemy: Thanks for catching! Here's how it looks now:
Screen Shot 2022-03-16 at 8 44 55 AM

packages/compass-workspace/src/components/workspace-tabs.spec.tsx
Anemy: Would it be better to rename this `activeSubTabName` to `workspaceName` to be less collection/subtab centric in the naming?

packages/compass-collection/src/components/workspace/workspace.jsx
mcasimir: I think I'm probably just confused by the fact that there is a workspace package and then this file which seems to control `WorkspaceTabs` in every aspect through its own store. But that's just temporary or?

packages/compass-workspace/src/components/workspace-tabs.spec.tsx
mcasimir: Renaming it would definitely be better yeah. But I'll never be fond of this coupling here, in my opinion, `WorkspaceTabs` shouldn't know anything about sub-tabs, tabs content, and the fact that there is a concept of time-series and whatnot.

packages/compass-workspace/src/components/workspace-tabs.tsx
Anemy: Good thinking, I've updated to make most of the component more generic and accept an icon.
I think it might make sense to avoid moving the content rendering into this `compass-workspace` in this pr, I'll update it so that we're passing a render function from `compass-collection` as `compass-collection` has a good amount of logic tied around how the tabs are generated and even some ipc listeners for interacting with the tabs.

packages/compass-workspace/src/components/tab.tsx
Anemy: Updated - added a `useRovingTabIndex` for focusing the tabs, and now only the selected/active tab will have `tabIndex={0}`. This `useRovingTabIndex` could live in `compass-components`, but since this is the only use place of it I haven't moved it there yet. Would you prefer I move it over now? Otherwise we can just keep an eye on it and move it over if we end up needing it elsewhere.

packages/compass-collection/src/components/workspace/workspace.jsx
Anemy: Synced offline on how we can move towards a generic `compass-workspace` while addressing around the current collection centric logic in the `compass-collection` package. Down the line we'll use providers to supply the ipc handlers that are currently in this store. We can then manage listening to ipc events, like the menu actions from the main compass process as part of our react lifecycle. This'll probably clear up some bugs we currently have in Compass.

packages/compass-workspace/src/components/workspace-tabs.spec.tsx
Anemy: Updated 👍  thanks for raising.

packages/compass-workspace/src/components/tab.tsx
Anemy: Screen Shot 2022-03-16 at 10 45 40 PM

packages/compass-workspace/src/components/workspace-tabs.tsx
```diff
@@ -0,0 +1,330 @@
...
+      className={cx(
+        tabsContainerStyles,
+        darkMode ? tabsContainerDarkStyles : tabsContainerLightStyles
+      )}
+    >
+      <div className={tabsListContainerStyles}>
+        <div
+          className={tabsListStyles}
+          role="tablist"
+          aria-label="Workspace Tabs"
```
gribnoysup, line 295: This should be probably also be a prop for now and say something like "Collections" or "Collection Workspaces" for now, does this makes sense?

packages/compass-workspace/src/components/tab.tsx
gribnoysup: I think if it's something that only used in one place it would make sense not to move it. Kind of a tangent, but still related: I do feel like now with the updated, more generic tabs interface that this component has this whole thing can be part of compass-components and not a separate package, would that make sense to you?

packages/compass-collection/src/components/workspace/workspace.jsx
```diff
@@ -245,24 +225,26 @@ class Workspace extends PureComponent {
...
+          onCreateNewTab={this.onCreateNewTab}
+          onMoveTab={this.props.moveTab}
+          onSelectTab={this.props.selectTab}
+          onCloseTab={this.props.closeTab}
+          tabs={this.props.tabs.map(tab => ({
+            title: tab.activeSubTabName,
+            subtitle: tab.namespace,
+            tabContentId: tab.id,
+            renderIcon: (iconProps) => renderTabIcon(
```
mcasimir, line 241: Why can't this be just `icon: <Elem>`?

Anemy: It could, where would the styling be applied then? I'm cool with whatever here.

packages/compass-workspace/src/components/workspace-tabs.tsx
Anemy: Sounds good, created a ticket in the epic to move more of the workspace into the package and make the boundary between `Collection` view and `Workspace` view a bit more distinct: https://jira.mongodb.org/browse/COMPASS-5605

packages/compass-collection/src/components/workspace/workspace.jsx
mcasimir: ah cause we need to apply some style to the icon, makes sense

packages/compass-workspace/src/components/tab.tsx
```diff
@@ -0,0 +1,326 @@
...
+      aria-selected={isSelected}
+      role="tab"
+      // Catch navigation on the active tab when a user tabs through Compass.
+      tabIndex={isSelected ? 0 : -1}
+      aria-controls={tabContentId}
+      title={`${subtitle} - ${title}`}
+      {...tabProps}
+    >
+      <div className={tabTitleContainerStyles}>
+        {renderIcon(iconProps)}
```
gribnoysup, line 269: If we want to be so much in control of the props that are passed to the icon, wouldn't it make sense to only allow to pass valid leafygreen glyph names here and nothing else? With the current implementation I can break the layout easily by setting my own size or a class name after spreading the props and the props are not generic enough to be spreadable to anything, like if I use another icon component that doesn't support size exactly as expected here it wouldn't work

packages/compass-workspace/src/components/workspace-tabs.tsx
```diff
@@ -0,0 +1,339 @@
...
+      )}
+    >
+      <div className={tabsListContainerStyles}>
+        <div
+          className={tabsListStyles}
+          role="tablist"
+          aria-label={ariaLabel}
+          aria-orientation="horizontal"
+          ref={tabContainerRef}
+          {...tabContainerProps}
```
gribnoysup, line 307: I think this `ref={tabContainerRef}` will be overwritten by `tabContainerProps`, you sure it works as expected? Is it just not used anywhere?

gribnoysup (approved):
As I mentioned to you in person I do think that this generic component doesn't need a separate package and can be part of compass-components, maintaining separate package comes with an overhead and I don't see any benefit of having it separated. I don't have a strong enough opinion here though, so I will not block this work on this, your call if you want to keep it here or move it.

I also spotted a few small things, but nothing blocking too. Otherwise these new tabs look and work awesome!

packages/compass-components/src/compass-font-sizes.ts
```diff
@@ -1 +1,2 @@
+export const smallFontSize = '13px';
 export const defaultFontSize = '14px';
```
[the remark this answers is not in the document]
Anemy: Yes it exists there. We're not using this `defaultFontSize` in many places in Compass. I can open up a pr replacing those and removing it, but it's probably not worth doing here as its a bit unrelated. How do you feel about the addition of the `smallFontSize` ? Is there another way we might want to do this? We could add some typography components to avoid a constant here.

packages/compass-workspace/src/components/workspace-tabs.tsx
Anemy: Ah thanks! Missed removing it when cleaning up.

packages/compass-workspace/src/components/tab.tsx
Anemy: Sounds good, updated it to take a `iconGlyph` string with the `glyph` type from leafygreen.

packages/compass-components/src/compass-font-sizes.ts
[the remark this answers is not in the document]
gribnoysup: This font size is completely out of leafygreen design system, right? Is the difference that noticeable that we can't use 14px here? If it is then I guess it's fine, but 1px sounds like something that can be changed to a slightly bigger font without major issues. Anyway, totally not something that should block this PR from landing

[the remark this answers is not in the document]
Anemy: I think so yes. Claudia mentioned the 13 px size explicitly but 14px I'm thinking could work, I'll check with her. In the meantime I think I'll merge and we can update those fonts independently (and remove the 14px).

PR #2905, opened 2022-03-21
chore(e2e): e2e test favorites COMPASS-5496

lerouxb:


packages/compass-e2e-tests/tests/connection-form.test.ts
```diff
@@ -486,4 +487,84 @@ describe('Connection form', function () {
...
+    await browser.$(Selectors.FavoriteModal).waitForDisplayed();
+    await browser.$(Selectors.FavoriteNameInput).setValue(favoriteName);
+    await browser.clickVisible(
+      `${Selectors.FavoriteColorSelector} [data-testid="color-pick-color1"]`
+    );
+    await browser.$(Selectors.FavoriteSaveButton).waitForEnabled();
+    await browser.clickVisible(Selectors.FavoriteSaveButton);
+    await browser.$(Selectors.FavoriteModal).waitForExist({ reverse: true });
+
+    // HACK: The favorite doesn't always appear in the sidebar. Must be some
```
lerouxb, line 506: I suspect some of the other flakes could be a similar thing and we can probably work the sidebar database/collection flakes the same way.

But this is still a bug in the app.

packages/compass-connections/src/components/connection-list/connection-list.tsx
```diff
@@ -171,6 +172,9 @@ function ConnectionList({
           {favoriteConnections.map((connectionInfo, index) => (
             <li
               data-testid="favorite-connection"
+              data-id={`favorite-connection-${getConnectionTitle(
```
mcasimir, line 175: This function is not meant for this use case, makes sense in places where we don't know if a connection is a favorite or not, but using it here is a bit unexpected. Since this connection is guaranteed to be a favorite why not something like `connectionInfo?.favorite?.name` instead?

packages/compass-e2e-tests/tests/connection-form.test.ts
mcasimir: Can it be the fact that save is not currently awaited?

mcasimir (approved):
🙌

packages/compass-e2e-tests/tests/connection-form.test.ts
lerouxb: I don't know but it _does_ show up when you reload the app, so it did save. It just never shows up until you reload.

My gut suspects that if you save a connection at just the right time while connections are loaded then the connections get returned without that connection.

lerouxb: I'm awaiting it now. Let's see if that helps.

packages/storage-mixin/lib/backends/secure-ipc.js
```diff
@@ -117,6 +121,8 @@ if (typeof window !== 'undefined') {
 
     ipc.on('storage-mixin:find-one:result', listener);
 
+    // TODO: this returns a promise, but for some reason the result gets
```
lerouxb, line 124: This probably just pre-dates ipcRenderer.call() and ipcMain.respondTo() being able to handle promises and responses and what not for you. Or whoever wrote this initially didn't think about or know about that.

I think I'll fix this separately because there's no reason for our changes in ipcRenderer and ipcMain anymore that I can tell - ipcMain.handle() does all this stuff for us.

I don't know at which stage electron added that stuff. Maybe it was always there? No idea.

But I think we can clean it all up in a separate PR. Or just leave this until we get rid of ampersand and then redo all this stuff.

PR #2917, opened 2022-03-24
style(compass-shell): Make toggle flash on connect

Anemy:
To improve discoverability of the shell, this pr makes it so the shell header toggle flashes after you've connected. (Claudia said why not :))

https://user-images.githubusercontent.com/1791149/159954140-30ee8dde-302c-4b92-aebd-37c1c2e89250.mp4


addaleax (approved):
I love it!

Anemy:
Looks like it passed on first test run and not much changed in the main merge which I think failed due to the github githooks outage. Going to merge, let's keep in mind if this flashing feature does become bothersome we can always take it out.

PR #2919, opened 2022-03-24
deps(hadron-document): remove unused react deps

Anemy:
Looks like these were added in the mono repo work, maybe unintentional? I might be missing something cc @gribnoysup https://github.com/mongodb-js/compass/commit/e5830b8f8d688f6cbc6c7f83053fcf04fac93dcd#diff-37e62e398aa57efb932e58a64d36c435cbe728cf81c26cc9ffe5fa82b360aaf3R38
Removing to help cloud upgrading to react v17

### DIFF
--- a/packages/hadron-document/.depcheckrc
+++ b/packages/hadron-document/.depcheckrc
@@ -1,8 +1,6 @@
 ignores: [
   "debug",
   "lodash.foreach",
-  "react",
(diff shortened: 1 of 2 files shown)


gribnoysup (approved):
Yeah, does look like it was added unintentionally

PR #2922, opened 2022-03-28
refactor(compass-import-export): Refactor exporter in order to export a generic Cursor COMPASS-5542

leorossi:
## Description
This PR introduces a `CursorExporter` that will allow to export in `json`/`csv` format a Cursor returned by Node Driver.

Code has been decoupled from the react component and will allow (in another PR) to use the same exporter both for `find()` and `aggregate()` results.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-import-export/src/modules/export.js
```diff
@@ -489,62 +487,62 @@ export const startExport = () => {
...
-        all_docs: exportData.isFullCollection,
-        file_type: exportData.fileType,
-        all_fields: Object.values(exportData.fields).every(
-          (checked) => checked === 1
-        ),
-        number_of_docs: numDocsToExport,
-        success: !err,
-      });
-      if (err) {
-        log.error(mongoLogId(1001000085), 'Export', 'Export failed', {
```
mcasimir, line 513: Shouldn't we have this also in the new version?

packages/compass-import-export/src/modules/export.js
```diff
@@ -489,62 +487,62 @@ export const startExport = () => {
...
-       * "fields_selected": true|false
-       */
-      dispatch(
-        globalAppRegistryEmit(
-          'export-finished',
-          numDocsToExport,
-          exportData.fileType
-        )
-      );
+    await exporter.start();
```
mcasimir, line 509: What happens if this throws?

packages/compass-import-export/src/modules/CursorExporter.ts
```diff
@@ -0,0 +1,69 @@
...
+    );
+  }
+  private getFormatter() {
+    return this._formatter({ columns: this._columns });
+  }
+  private _getPercentageOfDocuments() {
+    if (this._totalNumberOfDocuments) {
+      return (this._exportedDocuments * 100) / this._totalNumberOfDocuments;
+    }
+    return 0;
```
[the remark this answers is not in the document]
leorossi: We don't always have the `totalNumberOfDocuments` (i.e. in aggregations) so I choose to return 0 instead of the percentage.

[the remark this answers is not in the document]
mcasimir: Since we show the num of documents processed if the count is not available, why do we calculate the percentage here? Isn't the caller a better place to own this logic since it has all the knowledge? Would it be better to only `emit('processed')` here instead of trying to give a percentage back?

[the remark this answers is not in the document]
mcasimir: > I choose to return 0 instead 

that makes sense but if there are 0 document to export that means that you are already at 100%. Just an edge case

packages/compass-import-export/src/modules/export.js
```diff
@@ -489,62 +487,72 @@ export const startExport = () => {
...
         )
     );
-    let formatter;
-    if (exportData.fileType === 'csv') {
-      formatter = createCSVFormatter({ columns });
-    } else {
-      formatter = createJSONFormatter();
-    }
-
     const dest = fs.createWriteStream(exportData.fileName);
```
mcasimir, line 490: What would happen if `createWriteStream` and  `dataService.fetch` throws? does Compass crash?

packages/compass-import-export/src/modules/cursor-exporter.ts
```diff
@@ -0,0 +1,81 @@
...
+    this._formatter =
+      opts.type === 'csv' ? createCSVFormatter : createJSONFormatter;
+    this._output = opts.output;
+    this._columns = opts.columns ? opts.columns : true;
+    this._totalNumberOfDocuments = opts.totalNumberOfDocuments || 0;
+  }
+
+  async start(): Promise<void> {
+    try {
+      const p = pipeline(
```
mcasimir, line 37: why don't you `await` it here?

packages/compass-import-export/src/modules/cursor-exporter.ts
```diff
@@ -0,0 +1,81 @@
...
+        this._cursorStream,
+        this.getProgressTransformStream(),
+        this.getFormatter(),
+        this._output
+      );
+      return p;
+    } catch (err) {
+      if (this._cancelled) {
+        // do nothing
+        console.log(err);
```
mcasimir, line 47: we use `@mongodb-js/compass-logging` to report debug messages

packages/compass-import-export/src/modules/cursor-exporter.ts
```diff
@@ -0,0 +1,81 @@
...
+      } else {
+        throw err;
+      }
+    }
+  }
+  cancel(): boolean {
+    this._cursorStream.unpipe();
+    this._cursorStream.emit('end');
+    this._cancelled = true;
```
mcasimir, line 56: if we set `this._cancelled = true;` after `unpipe` and `emit('end')` the catch above where we check `if (this._cancelled) {` is going to potentially race and find  `this._cancelled` to be still false if, for instance `this._cursorStream.unpipe();` would cause pipeline to throw. Though i dont see that we wait it above so i'm not sure if all of this is intentional.

packages/compass-import-export/src/modules/export.js
```diff
@@ -456,12 +455,12 @@ export const startExport = () => {
         projection,
       }
     );
-    const source = createReadableCollectionStream(
-      dataService,
-      ns,
-      spec,
-      projection
-    );
+    // const source = createReadableCollectionStream(
```
mcasimir, line 458: can we remove this commented code?

mcasimir (approved):

PR #2925, opened 2022-03-28
feat(data-service): use split CRUD client/Metadata client model COMPASS-5641

addaleax:
For CSFLE support, we will need Compass to be able to access both
server metadata and encrypted collection contents.

Using automatic encryption means that a number of administrative
commands are not available. As a consequence, for CSFLE-enabled
connections, Compass will maintain two separate `MongoClient`
instances, and `DataService` methods will pick a specific one
depending on what they do.

For other connections, there is only a single client.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/data-service/src/data-service.ts
+++ b/packages/data-service/src/data-service.ts
@@ -16,7 +16,6 @@ import type {
   CountDocumentsOptions,
   CreateCollectionOptions,
   CreateIndexesOptions,
-  Db,
   DeleteOptions,
   DeleteResult,
   Document,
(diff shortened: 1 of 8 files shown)


packages/data-service/src/data-service.ts
```diff
@@ -1493,27 +1529,41 @@ class DataService extends EventEmitter {
   /**
    * Create a ClientSession that can be passed to commands.
    */
-  startSession(): ClientSession {
-    return this._initializedClient.startSession();
+  startSession(clientType: ClientType): ClientSession {
```
Anemy, line 1532: Would it help the typing here if we have a `ClientTypeSession` (other names?) type that extends the `ClientSession` with the `clientType` as a key? Then in `killSessions` we might not need to do `any` casting.

Anemy (approved):
lgtm, nice comments with jira links! One typing thought/suggestion not blocker

packages/data-service/src/data-service.ts
addaleax: Sure, done! :)

PR #2928, opened 2022-03-29
feat(connection-form): add support for FLE KMS options COMPASS-5639

addaleax:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/connection-form/src/components/advanced-options-tabs/advanced-options-tabs.tsx
```diff
@@ -72,6 +73,10 @@ function AdvancedOptionsTabs({
     { name: 'Advanced', id: 'advanced', component: AdvancedTab },
   ];
 
+  if (process.env.COMPASS_CSFLE_SUPPORT === 'true') {
+    tabs.push({ name: 'CSFLE', id: 'csfle', component: CSFLETab });
+  }
```
addaleax, line 78: TBD: Figure out where exactly to place this

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/csfle-tab.tsx
```diff
@@ -0,0 +1,211 @@
...
+        // FLE1 equivalent of the FLE2 'encryptedFieldConfig[Map?]'.
+        // Once 'encryptedFieldConfig' is available, we will start
+        // using it instead.
+        encryptedFieldConfig={autoEncryptionOptions?.schemaMap}
+        errorMessage={errorMessageByFieldName(errors, 'schemaMap')}
+        onChange={(value: Document | undefined) => {
+          handleFieldChanged('schemaMap', value);
+        }}
+      />
+      {/* TODO: Add Ace editor for EncryptedFieldConfig/SchemaMap */}
```
addaleax, line 154:
```suggestion
```

(just forgot to remove this one before opening the PR)

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/csfle-tab.tsx
```diff
@@ -0,0 +1,211 @@
...
+      <Label htmlFor="TODO">KMS Providers</Label>
```
addaleax, line 155: These labels don't really look pretty right now, I should probably find something better for these

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/encrypted-field-config-input.tsx
```diff
@@ -0,0 +1,73 @@
...
+        theme="mongodb"
+        width="100%"
+        value={currentText}
+        onChange={changeCurrentText}
+        editorProps={{ $blockScrolling: Infinity }}
+        name="import-pipeline-editor"
+        setOptions={OPTIONS}
+      />
+      {
+        /* TODO: proper error message display! */ errorMessage && (
```
addaleax, line 65: @Anemy Any ideas for doing this? Do we have some easy way to display error messages for ace editor inputs?

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/kms-tls-options.tsx
```diff
@@ -0,0 +1,69 @@
...
+        useSystemCA={false /* COMPASS-5635 */}
+        disabled={false}
+        handleTlsOptionChanged={(key, value) =>
+          handleFieldChanged(
+            key as 'tlsCAFile' /* COMPASS-5635 */,
+            value ?? undefined
+          )
+        }
+      />
+      {/* TODO: Update UI messages for TLS situation (e.g. drop reference to X.509 auth) */}
```
addaleax, line 45: Reminder to self that this still needs a ticket

packages/connection-form/src/components/connect-form.tsx
```diff
@@ -31,7 +31,7 @@ const formContainerStyles = css({
   margin: 0,
   padding: 0,
   height: 'fit-content',
-  width: spacing[6] * 10,
+  width: spacing[6] * 12,
```
addaleax, line 34: Also maybe not ideal yet? :)

packages/connection-form/src/utils/csfle-handler.ts
```diff
@@ -0,0 +1,199 @@
...
+  connectionOptions = cloneDeep(connectionOptions);
+  const autoEncryptionOptions = connectionOptions.fleOptions?.autoEncryption;
+  // TODO(COMPASS-5645): schemaMap -> encryptedFieldConfig[Map?]
+  if (autoEncryptionOptions?.schemaMap?.['$compass.error'] === null) {
+    autoEncryptionOptions.schemaMap = textToEncryptedFieldConfig(
+      autoEncryptionOptions.schemaMap['$compass.rawText']
+    );
+  }
+  return connectionOptions;
+}
```
addaleax, line 199: @mabaasit Fwiw, the last three functions here are the solution that I came up with w.r.t. what we chatted about today -- this solution here has the advantages that the driver options are always in a usable state and we do keep the actual editor contents around without needing an extra top-level field for storage, if you have any thoughts, feel free to share them :)

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/encrypted-field-config-input.tsx
Anemy: Could the errors at the bottom of the form be an option - is there a reason we'd like to keep these errors shown more locally?

addaleax: @Anemy It's an option I guess, yeah -- it's a bit inconsistent with how we use the `errorMessage` prop of TextInput/TextArea fields where we display it immediately next to the fields, but it doesn't look like leafygreen exposes that specific part as a component anywhere

packages/connection-form/src/components/advanced-options-tabs/tls-ssl-tab/tls-client-certificate.tsx
```diff
@@ -41,8 +49,12 @@ function TLSClientCertificate({
             );
           }}
           showFileOnNewLine
-          optional
-          optionalMessage="Optional (required with X.509 auth)"
+          optional={optional ?? true}
+          optionalMessage={
+            optional && displayDatabaseConnectionUserHints
+              ? 'Optional (required with X.509 auth)'
```
Anemy, line 55: On the `TLS` page it looks like this message isn't being shown:
Screen Shot 2022-04-01 at 12 37 02 PM

Maybe we need to pass the `displayDatabaseConnectionUserHints` from the tls tab https://github.com/mongodb-js/compass/blob/main/packages/connection-form/src/components/advanced-options-tabs/tls-ssl-tab/tls-ssl-tab.tsx#L177

packages/connection-form/src/hooks/use-connect-form.ts
```diff
@@ -187,7 +201,9 @@ function buildStateFromConnectionInfo(
       : validateConnectionOptionsWarnings(
           initialConnectionInfo.connectionOptions
         ),
-    connectionOptions: cloneDeep(initialConnectionInfo.connectionOptions),
+    connectionOptions: adjustCSFLEParams(
+      initialConnectionInfo.connectionOptions
```
Anemy, line 205: Down the line will we remove this `adjustCSFLEParams` function? How would you feel about keeping the `cloneDeep` here as well as in that function to be sure we don't accidentally remove the cloning when we make that change. Probably a non-issue just thinking to avoid a slim chance at a possible bug way down the line.
```suggestion
      cloneDeep(initialConnectionInfo.connectionOptions)
```

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/encrypted-field-config-input.tsx
```diff
@@ -0,0 +1,98 @@
...
+          security.
+        </Description>
+      </div>
+      <div className={withMarginStyles}>
+        <AceEditor
+          mode="javascript" // will be set to mongodb as part of OPTIONS
+          theme="mongodb"
+          width="100%"
+          value={currentText}
+          onChange={changeCurrentText}
```
Anemy, line 83: I'm receiving an error when typing in this field:
Screen Shot 2022-04-01 at 12 44 31 PM

Is that expected for now?

Anemy: Since this action does some more logic with the `textToEncryptedFieldConfig` function do we want to add a test for this component that types into this editor and spies on the response to the `onChange` prop we pass to the component?
An example of a test like this: https://github.com/mongodb-js/compass/blob/main/packages/compass-find-in-page/src/components/find-in-page-input.spec.tsx#L121 - there are also probably a few in this package to reference as well.

Anemy (commented):
Nice! Couple comments/suggestions/thoughts. Looking good. Nice react work!

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/encrypted-field-config-input.tsx
addaleax: Ugh, thanks for noticing this error. No, it’s not expected, it’s a weird quirk of our `ejson-shell-parser` package that I cannot understand at all, where invalid input is mapped to an empty string as a result, rather than an exception.

(This is worrying me a little bit, because it means that we might be reading an empty string where we expect something else in other places as well. Here this would have been particularly bad, since accidentally omitting a clien-provided `EncryptedFieldConfig` can be dangerous from a security perspective.)

> Since this action does some more logic with the `textToEncryptedFieldConfig` function do we want to add a test for this component that types into this editor and spies on the response to the `onChange` prop we pass to the component?

I’ve tinkered around a bit with writing a test like this, but in the end I’m not sure how to make this work with the ace editor rather than a text input.

Fwiw, the bulk of the work here is done by `textToEncryptedFieldConfig()` and `encryptedFieldConfigToText()`, and those are tested quite well, so I’ve moved the remainder of the text conversions there as well.

---

Somewhat relatedly, I noticed that we also need to use the logic here for reviving EncryptedFieldConfig when double-clicking through the sidebar, where otherwise the stored data in the model would have led to errors. I’ve exposed a wrapper around `adjustCSFLEParams()` in the connection-form exports (see 5f8f8197a99e8def5770fdfb76ec3a181871b9ac), it doesn’t feel like an ideal solution here though and I’m happy to take suggestions on where a better place for this would be.

packages/connection-form/src/hooks/use-connect-form.ts
addaleax: Sure, makes sense. Done!

packages/connection-form/src/components/advanced-options-tabs/tls-ssl-tab/tls-client-certificate.tsx
addaleax: Ah sorry, this was actually just an editing leftover – `displayDatabaseConnectionUserHints` and `optional` should default to `true` in the TLS components. They now do :slightly_smiling_face:

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/csfle-tab.tsx
```diff
@@ -0,0 +1,187 @@
...
+  errors,
+}: {
+  errors: ConnectionFormError[];
+  updateConnectionFormField: UpdateConnectionFormField;
+  connectionOptions: ConnectionOptions;
+}): React.ReactElement {
+  const autoEncryptionOptions =
+    connectionOptions.fleOptions?.autoEncryption ?? {};
+
+  errors = errorsByFieldTab(errors, 'csfle');
```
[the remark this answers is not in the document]
addaleax: I’m not sure I understand what resetting would mean here – we are overriding a parameter, that’s true, but not affecting the value of it outside of this function.

Assigning the result to a separate variable like `fleErrors` would make me a bit worried that it’s easy to accidentally use the original parameter name (`errors`) inside this function by accident.

I’ve renamed the parameter to `errors_`, I hope that addresses both concerns here?

packages/connection-form/src/components/advanced-options-tabs/advanced-options-tabs.tsx
```diff
@@ -14,12 +14,15 @@ import GeneralTab from './general-tab/general-tab';
...
 import ProxyAndSshTunnelTab from './ssh-tunnel-tab/proxy-and-ssh-tunnel-tab';
 import TLSTab from './tls-ssl-tab/tls-ssl-tab';
+import CSFLETab from './csfle-tab/csfle-tab';
 import AdvancedTab from './advanced-tab/advanced-tab';
 import type { UpdateConnectionFormField } from '../../hooks/use-connect-form';
 import type { ConnectionFormError, TabId } from '../../utils/validation';
 import { errorsByFieldTab } from '../../utils/validation';
 import { defaultConnectionString } from '../../constants/default-connection';
 
+const csfleFeatureFlag = process.env.COMPASS_CSFLE_SUPPORT === 'true';
```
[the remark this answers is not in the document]
addaleax: Sure, done!

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/csfle-tab.tsx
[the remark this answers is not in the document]
Anemy: Sounds good, sure does! I should have left a longer comment, I was hoping to avoid any chances of modifying the errors although that wouldn't probably happen, the recent changes definitely protect against that 👍  . Makes sense on avoiding the `fleErrors` naming with errors present.

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/csfle-tab.tsx
```diff
@@ -0,0 +1,188 @@
...
+          value={autoEncryptionOptions.keyVaultNamespace || ''}
+          errorMessage={errorMessageByFieldName(errors, 'keyVaultNamespace')}
+          state={
+            fieldNameHasError(errors, 'keyVaultNamespace') ? 'error' : 'none'
+          }
+          spellCheck={false}
+          description="Specify a collection in which data encryption keys are stored in the format <db>.<collection>."
+        />
+      </FormFieldContainer>
+      <EncryptedFieldConfigInput
```
Anemy, line 135: How would you feel about wrapping this in the `FormFieldContainer` so that it gets a little bit of top margin?

We might be able to remove the first `withMarginStyles` from that component then 
https://github.com/mongodb-js/compass/pull/2928/files#diff-02ecb09ae0d4678618b443da67f6a9cdf5c3043d424727b39d494a2dc53fb9a9R55

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/csfle-tab.tsx
```diff
@@ -0,0 +1,188 @@
...
+        // FLE1 equivalent of the FLE2 'encryptedFieldConfig[Map?]'.
+        // Once 'encryptedFieldConfig' is available, we will start
+        // using it instead.
+        encryptedFieldConfig={autoEncryptionOptions?.schemaMap}
+        errorMessage={errorMessageByFieldName(errors, 'schemaMap')}
+        onChange={(value: Document | undefined) => {
+          handleFieldChanged('schemaMap', value);
+        }}
+      />
+      <div className={withMarginStyles}>
```
Anemy, line 146: Same as above, how would you feel about wrapping or replacing this with the `FormFieldContainer` so that it gets a little bit of top margin? We might be able to remove the `withMarginStyles` then.

packages/connection-form/src/utils/csfle-handler.ts
```diff
@@ -0,0 +1,222 @@
...
+  try {
+    parsed = queryParser(text);
+    if (!parsed || typeof parsed !== 'object') {
+      // We've hit the condition in
+      // https://github.com/mongodb-js/ejson-shell-parser/blob/c9c0145ababae52536ccd2244ac2ad01a4bbdef3/src/index.ts#L36
+      // in which instead of returning a parsed value or throwing an error,
+      // the query parser just returns an empty string when encountering
+      // input that can be parsed as JS but not as a valid query.
+      // Unfortunately, this also means that all context around what
+      // caused this error is unavailable here.
```
Anemy, line 178: Should we ticket to improve this? I guess it's maybe tech debt, but could be a feature for this project if we want. No worries if now is not the time.

packages/connection-form/src/utils/csfle-kms-fields.ts
```diff
@@ -0,0 +1,172 @@
...
+  },
+  {
+    name: 'endpoint',
+    label: 'Endpoint',
+    type: 'text',
+    optional: true,
+    value: (autoEncryption) =>
+      autoEncryption?.kmsProviders?.gcp?.endpoint ?? '',
+    state: 'none',
+    description: 'A host with optional port.',
```
Anemy, line 54: Maybe
```suggestion
    description: 'A host with an optional port.',
```
reads better?

packages/connection-form/src/utils/csfle-kms-fields.ts
```diff
@@ -0,0 +1,172 @@
...
+  },
+  {
+    name: 'identityPlatformEndpoint',
+    label: 'Identity Platform Endpoint',
+    type: 'text',
+    optional: true,
+    value: (autoEncryption) =>
+      autoEncryption?.kmsProviders?.azure?.identityPlatformEndpoint ?? '',
+    state: 'none',
+    description: 'A host with optional port.',
```
Anemy, line 131: Same as above, maybe:
```suggestion
    description: 'A host with an optional port.',
```
reads better? or maybe:
```suggestion
    description: 'A host with port optional.',
```

Anemy (approved):
lgtm! Couple styling thoughts/suggestions, not blockers. Nice work – I like how you've made each dropdown accordion show the state and error state if there is one.

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/encrypted-field-config-input.tsx
Anemy: Gotcha - nice find. 

For the location of `adjustCSFLEParams` - would it make sense to be in the data-service instead?
Something on load/save? I'm thinking then the connect layer would only interact with the expected connectionOptions.
https://github.com/mongodb-js/compass/blob/main/packages/data-service/src/legacy/legacy-connection-model.ts#L206
Used here: https://github.com/mongodb-js/compass/blob/main/packages/data-service/src/connection-storage.ts#L46
Not sure if that's a better spot, just a thought. I think ideally we want this connection form to be usable in places like VSCode or other projects down the line - it would make sense to keep the connection form logic/usage to be minimal where we can then.

addaleax: Mhh … I thought about that, and actually wrote a version that placed it in data-service at some point, *but*:

- This is solving a problem with how options are stored (i.e. not in a way that properly serializes/deserializes BSON values), not with how we connect
- It makes sense to me to keep the `$compass.xxx` special field handling as localized as possible, not spread across packages

The ideal way to handle this would probably be adjusting the connection storage to handle BSON data well, e.g. by using EJSON instead of JSON, but that might not be trivial (since some of the data might *not* be handled well by EJSON, and we’re aiming to move the storage layer away from Ampersand entirely at some point).

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/csfle-tab.tsx
addaleax: Sure, done – I’ve removed these everywhere and used `FormFieldContainer` instead. I’m not sure if this is what you had in mind exactly, but in any case these feel like things that are fairly easy to adjust :)

packages/connection-form/src/utils/csfle-handler.ts
addaleax: Yes :+1: Opened COMPASS-5689 for this.

PR #2930, opened 2022-03-30
feat(aggregations): run aggregation COMPASS-5549

mabaasit:
feat(aggregations): run aggregation COMPASS-5549

Runs and paginates an aggregation with support for cancelling it as well.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/components/pipeline-results-workspace/index.tsx
```diff
@@ -1,24 +1,134 @@
...
+    const showingFrom = (page - 1) * perPage;
+    const showingTo = showingFrom + documents.length;
     return (
       <div data-testid="pipeline-results-workspace">
-        <pre>
-          <code>{JSON.stringify(documents, null, 2)}</code>
-        </pre>
+        <div className={topStyles}>
+          <Body>
+            Showing {showingFrom + 1} - {showingTo}
```
addaleax, line 70:
```suggestion
            Showing {showingFrom + 1} – {showingTo}
```

This is a really minor thing, but en dashes look a bit nicer for ranges than hyphen-minus and are generally the right type of dash here :slightly_smiling_face:

packages/compass-aggregations/src/components/pipeline-results-workspace/index.tsx
```diff
@@ -1,24 +1,134 @@
...
+              variant="primaryOutline"
+              size="xsmall"
+            >
+              Cancel
+            </Button>
+          </div>
+        )}
+        {documents.length > 0 && (
+          <pre>
+            <code>{JSON.stringify(documents, null, 2)}</code>
```
addaleax, line 109: Without being entirely sure about the context here, should this be EJSON rather than JSON? (I assume it’s not the final state of things because we’d probably use one of our existing components with code hightlighting etc. here anyway, so feel free to just resolve this if I’m being unhelpful)

packages/compass-aggregations/src/modules/aggregation.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+  limit(count: number) {
+    this.options.limit = count;
+    return this;
+  }
+  async hasNext() {
+    await waitFor(this.options.delayInResponse ?? 0);
+    return Promise.resolve(this.options.hasNext);
+  }
+  async toArray() {
+    await waitFor(this.options.delayInResponse ?? 0);
```
addaleax, line 42: I think these timers keep the test running – would it make sense to use `.unref()`ed timers for these? Can we avoid using timers altogether (it seems like what we really need here is just a Promise that does not resolve in time, which can also be a Promise that never fulfills or rejects)?

packages/compass-aggregations/src/modules/aggregation.ts
```diff
@@ -39,32 +101,120 @@ export const runAggregation = (): ThunkAction<
...
-      collation: collation || undefined,
-    };
-    if (dataService) {
+    if (!dataService) {
+      return;
+    }
+
+    const abortController = new AbortController();
+    const signal = abortController.signal;
+    abortController.signal.addEventListener('abort', cancelAggregation, { once: true });
```
addaleax, line 178: I feel like I’m missing something here – the `abortController` here is the same one that we get form the state inside of `cancelAggregation` itself (right?), so this line is saying that we should call `cancelAggregation()`, which calls `.abort()` on the AbortController, right after that has already happened? Are these different controllers/signals?

packages/compass-aggregations/src/utils/cancellablePromise.ts
```diff
@@ -0,0 +1,50 @@
...
+  };
+
+  const succeed = () => {
+    // if this task succeeds it will never abort, so clean up that event listener
+    // (successSignal's event handler is already removed due to { once: true })
+    abortSignal.removeEventListener('abort', abort);
+  };
+
+  abortSignal.addEventListener('abort', abort, { once: true });
+  successSignal.addEventListener('abort', succeed, { once: true });
```
addaleax, line 28: We should probably also check whether `signal.aborted` is already true before adding these listeners.

packages/compass-aggregations/src/modules/aggregation.ts
```diff
@@ -39,32 +101,120 @@ export const runAggregation = (): ThunkAction<
...
+        page,
+        isLast: documents.length < limit,
+      });
+    } catch (e) {
+      dispatch({
+        type: ActionTypes.AggregationFailed,
+        error: (e as Error).message,
+      });
+    } finally {
+      abortController.signal.removeEventListener('abort', cancelAggregation);
```
addaleax, line 215: Should we also implement the session-based logic for stopping the aggregations server-side that we use in compass-crud? Otherwise a user cancellation will stop the operation in the UI but not on the server afaict.

packages/compass-aggregations/src/modules/aggregation.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+        dataService: new class {
+          aggregate() {
+            return aggregateMock;
+          }
+        }
+      });
+
+      await store.dispatch(fetchNextPage() as any);
+
+      await waitFor(500);
```
lerouxb, line 183: why 500?

mabaasit: wait for redux to update the state. little high just to be on the safer side.

addaleax (approved):

lerouxb (approved):

packages/compass-aggregations/src/modules/aggregation.ts
```diff
@@ -39,32 +113,131 @@ export const runAggregation = (): ThunkAction<
...
+      const abortController = new AbortController();
+      const signal = abortController.signal;
+      const session = dataService.startSession('CRUD');
+      dispatch({
+        type: ActionTypes.AggregationStarted,
+        abortController,
+        session,
+      });
+
+      const stages = pipeline.map(generateStage);
```
gribnoysup, line 206: `generateStage` returns disabled stages as `{}` which breaks pipeline when you are trying to run it:

image

packages/compass-aggregations/src/modules/aggregation.ts
```diff
@@ -39,32 +113,131 @@ export const runAggregation = (): ThunkAction<
...
+      aggregation: { abortController, session },
+    } = getState();
+    abortController?.abort();
+    if (session && dataService) {
+      try {
+        await dataService.killSessions(session);
+      } catch (e) {
+        log.warn(mongoLogId(1001000105), 'Aggregations', 'Attempting to kill the session failed');
+      }
+    }
```
gribnoysup, line 172: If this code cancels aggregation, should it also clean it up from the store?

packages/compass-aggregations/src/modules/aggregation.ts
```diff
@@ -39,32 +113,131 @@ export const runAggregation = (): ThunkAction<
...
-    if (dataService) {
-      const cursor = dataService.aggregate(
+    if (!dataService) {
+      return;
+    }
+
+    try {
+      const abortController = new AbortController();
+      const signal = abortController.signal;
+      const session = dataService.startSession('CRUD');
```
gribnoysup, line 199: I might be missing something, do we need both session and signal stored separately in the state for some reason? Session should be attached to the aggregate and canceled with it, so all this logic can be part of your `aggregatePipeline` method I think, I don't see why you would need to spread it across multiple actions

packages/compass-aggregations/src/utils/cancellableAggregation.ts
```diff
@@ -0,0 +1,32 @@
...
+  let result;
+  try {
+    result = await raceWithAbort(cursor.toArray(), signal);
+  } finally {
+    signal.removeEventListener('abort', abort);
+    result = result || [];
+  }
+
+  return result;
+}
```
gribnoysup, line 32: Doesn't look like the session you created is getting passed to the aggregate at all so why do we have it?

packages/compass-aggregations/src/components/pipeline-results-workspace/index.tsx
```diff
@@ -1,24 +1,149 @@
...
+            >
+              <Icon glyph="ChevronLeft" />
+            </IconButton>
+            <IconButton
+              aria-label="Next page"
+              disabled={isNextDisabled}
+              onClick={() => onNext()}
+            >
+              <Icon glyph="ChevronRight" />
+            </IconButton>
```
gribnoysup, line 99: Seems like I'm able to paginate even when my pipeline is broken:

image

gribnoysup, line 99: This section should probably take the error state in the account somehow, otherwise I'm getting both "showing documents" and an error message on the same screen:

image

gribnoysup (commented):
Left a few comments in the code. A few other things that caught my eye.

Seems like you currently can run a broken aggregation, is this intentional? If not, it would probably make sense to disable the Run button if stages have issues

image

On smaller screen sizes button labels collapse:

image

Have we considered maybe hiding button labels for a certain viewport size? This seems to be a very common pattern for avoiding content breaking like that on smaller screens, would look something like this:

image

Or maybe at least allowing text to overflow and hide it with ellipsis

packages/compass-aggregations/src/components/pipeline-results-workspace/index.tsx
mcasimir: A little out of topic, is it intentional that the pagination and the view selection breaks in 2 lines? seems like a lot of space thrown away

packages/compass-aggregations/src/modules/aggregation.ts
mabaasit: moved session handling to aggregatePipeline :)

packages/compass-aggregations/src/components/pipeline-results-workspace/index.tsx
mabaasit: the UX comments, i'll be fixing in my next PR.

PR #2933, opened 2022-03-30
chore: Add .gitignore for new plugins, add .gitignore to create-workspace script for plugins

Anemy:
When publishing the packages recently I noticed we had a few untracked files. This adds the `lib` in those two folders to the gitignore:
Screen Shot 2022-03-30 at 1 14 35 PM

This is because the `outputPath` in the webpack config for plugins is always `lib`: https://github.com/mongodb-js/compass/blob/main/configs/webpack-config-compass/src/index.ts#L248
Our `create-workspace` script currently sets up new packages to use the `dist` folder which works for normal react packages, but conflicts when creating new plugin packages as they're always using `lib`. Because of that I updated the `create-workspace` script to use the `lib` folder for new plugins. 
We could alternatively allow for overriding the `output` path for plugins so new ones can use `dist`, I'm not sure if there's an underlying reason for the `dist` vs `lib` usage here. cc @gribnoysup 

Removed a few extra items in a package.json's `files` since they're always included by default - https://docs.npmjs.com/cli/v7/configuring-npm/package-json#files


scripts/create-workspace.js
```diff
@@ -184,6 +184,8 @@ async function main(argv) {
 
   console.log();
 
+  const outDirName = isPlugin ? 'lib' : 'dist';
```
mcasimir, line 187: ❤️

mcasimir (approved):

gribnoysup (commented):
> We could alternatively allow for overriding the output path for plugins so new ones can use dist, I'm not sure if there's an underlying reason for the dist vs lib usage here

All existing plugins were using `lib` and so to make this new config a drop-in replacement for older plugins we made the plugin webpack config default to `lib`

If you don't think it would be too much work, I would maybe suggest to go another way about this and change the webpack config and the plugins that use it to point to `dist` instead of `lib`, then we don't need additional config file in every package as dist is globally ignored and it will be a smaller diff in general, wdyt?

PR #2934, opened 2022-03-31
chore(e2e): un-skip timeseries collection test

lerouxb:
I could have re-enabled this in the PR where we started testing multiple mongodb versions, but I didn't think of it.

### DIFF
--- a/packages/compass-e2e-tests/helpers/compass.ts
+++ b/packages/compass-e2e-tests/helpers/compass.ts
@@ -36,6 +36,15 @@ export const LOG_PATH = path.resolve(__dirname, '..', '.log');
 const OUTPUT_PATH = path.join(LOG_PATH, 'output');
 const COVERAGE_PATH = path.join(LOG_PATH, 'coverage');
 
+// mongodb-runner defaults to stable if the env var isn't there
+export const MONGODB_VERSION = (process.env.MONGODB_VERSION || '5.0.6')
+  // semver interprets these suffixes like a prerelease (ie. alpha or rc) and it
+  // is irrelevant for our version comparisons anyway
+  .replace('-community', '')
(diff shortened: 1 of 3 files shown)


packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -36,6 +36,15 @@ export const LOG_PATH = path.resolve(__dirname, '..', '.log');
...
 const COVERAGE_PATH = path.join(LOG_PATH, 'coverage');
 
+// mongodb-runner defaults to stable if the env var isn't there
+export const MONGODB_VERSION = (process.env.MONGODB_VERSION || '5.0.6')
+  // semver interprets these suffixes like a prerelease (ie. alpha or rc) and it
+  // is irrelevant for our version comparisons anyway
+  .replace('-community', '')
+  // HACK: comparisons don't allow X-Ranges and 5.x or 5.x.x installs 5.2.1 so
+  // we can't just map it to 5.0.0
+  .replace(/x/g, '999');
```
addaleax, line 46: By the way, this is what we do in mongosh: https://github.com/mongodb-js/mongosh/blob/f1e6417374a8aba22f299961b992241d7376d2af/testing/integration-testing-hooks.ts#L562-L564

addaleax (approved):

packages/compass-e2e-tests/helpers/compass.ts
lerouxb: So the reason I went with 999 rather than 0 is because it would turn 5.x.x into 5.0.0 and then that won't match >= 5.1 whereas it actually installs 5.2.1 which is > 5.1.

lerouxb: Luckily we're in control of both ends because 5 or 5.x are perfectly valid version numbers for mongodb runner but would require padding to three parts to work with semver.

PR #2937, opened 2022-04-01
chore(e2e): Rudimentary tests for connections with reduced permissions COMPASS-5335

lerouxb:
I added new secrets to both evergreen and github actions to make this work.


packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -110,6 +115,127 @@ function generateIamSessionToken(): {
...
+  browser: CompassBrowser,
+  dbName: string,
+  collectionName: string
+): Promise<void> {
+  await browser.navigateToCollectionTab(dbName, collectionName, 'Documents');
+  await browser.waitUntil(async () => {
+    const text = await browser
+      .$(Selectors.DocumentListActionBarMessage)
+      .getText();
+    return !!/Displaying documents \d+ - \d+ of \d+/.exec(text);
```
addaleax, line 128: tiny nit, feel free to ignore, but since there is a regexp method that is specifically for returning booleans… :)

```suggestion
    return /Displaying documents \d+ - \d+ of \d+/.test(text);
```

addaleax (approved):

packages/compass-e2e-tests/tests/connection.test.ts
lerouxb: I can never remember regex method names and regex related methods on strings for some reason :)

PR #2939, opened 2022-04-01
chore: remove compass-connect

mcasimir:


addaleax (approved):
:partying_face:

Anemy (approved):
lgtm! One more place we can remove it from is the main README: https://github.com/mongodb-js/compass/blob/main/README.md?plain=1#L18

PR #2945, opened 2022-04-05
chore: remove metrics and compass-metrics

mcasimir:
WIP PR to remove `metrics` and `compass-metrics`, is missing tests for `intercom.ts` and the bugsnag setup.


packages/compass/src/app/intercom.ts
```diff
@@ -0,0 +1,200 @@
...
+    user_id: user.id,
+    app_id: intercomAppId,
+    created_at: Math.floor(user.createdAt.getTime() / 1000),
+    app_name: process.env.HADRON_PRODUCT_NAME,
+    app_version: process.env.HADRON_APP_VERSION,
+    app_stage: process.env.NODE_ENV,
+  };
+
+  ('MongoDB Compass Dev Local');
+
```
addaleax, line 152:
```suggestion
```

packages/compass/src/app/intercom.ts
```diff
@@ -0,0 +1,200 @@
...
+  const intercomWidgetUrl =
+    'https://widget.intercom.io/widget/' + intercomAppId;
+
+  // in some environment the network can be firewalled, this is a safeguard to avoid
+  // uncaught errors when injecting the script.
+
+  debug('testing intercom availability');
+  const response = await fetch(intercomWidgetUrl);
+  if (response.status >= 400) {
+    debug('intercom unreachable, skipping setup');
```
addaleax, line 162: Should some of these events be logged to the Compass log?

packages/compass/src/app/intercom.ts
```diff
@@ -0,0 +1,200 @@
...
+    return;
+  }
+
+  debug('intercom is reachable, proceeding with the setup');
+
+  // we use a queue to ensure that the state of intercom will eventually reflect
+  // what is selected in the preferences.
+  const queue = new PQueue({ concurrency: 1 });
```
addaleax, line 170: Feel free to ignore, but fwiw, this seems like something that’s similarly easy to do without an added external dependency:

```js
const queue = Promise.resolve();

...

queue = queue
  .then(() => loadIntercomScript(intercomWidgetUrl, metadata))
  .catch((e) => debug('queue error', e));
```

addaleax (approved):

packages/compass/src/app/intercom.ts
mcasimir: I'm so wondering where that came from 🤣, thanks!

packages/compass/src/app/intercom.ts
mcasimir: makes a lot of sense, much better without the dep

packages/compass/src/app/intercom.ts
mcasimir: I think so, is it ready at the point where i've hooked this loading procedure?

mcasimir: Should we just log when is activated / deactivated? Users won't really know that an NPS is not working since they just won't really receive one, could be useful for us, especially if we disable the devtools in prod.

addaleax: Logging should always be available in the renderer, and yeah, I think it’s fine to log activate/deactivate only, but I feel like adding _something_ might be helpful for us at some point :)

packages/compass/src/app/intercom.ts
```diff
@@ -0,0 +1,217 @@
...
+          s.id = INTERCOM_SCRIPT_ELEM_ID;
+          s.async = true;
+          s.src = intercomWidgetUrl;
+          const x = d.getElementsByTagName('script')[0];
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          x.parentNode!.insertBefore(s, x);
+          debug('intercom script injected');
+          w.Intercom('boot', metadata);
+          log.info(
+            mongoLogId(1_001_000_106),
```
addaleax, line 67: You’ll probably need to bump these numbers

addaleax (approved):

packages/compass/src/app/intercom/setup-intercom.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const { debug } = createLoggerAndTelemetry('COMPASS-INTERCOM');
+
+type User = {
+  id: string;
+  createdAt: Date;
+};
+
+type Preferences = Pick<EventEmitter, 'on'> & {
+  isFeatureEnabled: (feature: string & 'enableFeedbackPanel') => boolean;
```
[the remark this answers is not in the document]
addaleax: This is just a type cleanup, because `string & 'enableFeedbackPanel'` is already the same type as `'enableFeedbackPanel'`.

But that’s okay, because `isFeatureEnabled: (feature: string) => boolean` is assignable to `(feature: 'enableFeedbackPanel') => boolean` – function arguments are contravariant, i.e. making an argument type more specific is actually making the requirements on the function itself *less* specific.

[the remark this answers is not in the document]
mcasimir: Nice TIL, i thought that it would have only accepted `'enableFeedbackPanel'` as the type for that argument then, thanks for the explanation!

PR #2950, opened 2022-04-05
chore(compass-serverstats): Update Performance Tab toolbar to leafy green COMPASS-5588

Anemy:
Updates the Performance Tab's toolbar to use leafy green components from `compass-components`. This doesn't update the error banners, those will be updated in [COMPASS-5671](https://jira.mongodb.org/browse/COMPASS-5671).

I was going to create a new `Toolbar` component in `compass-components` for this and the new toolbars to use, but I'm going to hold off on that until we know if there are any things that a shared toolbar would help with styling (currently it might be only background) (thread in slack). We have a separate ticket for that [COMPASS-5589](https://jira.mongodb.org/browse/COMPASS-5589) so I think it's fine if we hold off on that and move a shared toolbar in later once we've got a bit more idea on the styles it might need.

https://user-images.githubusercontent.com/1791149/161823880-09d3f052-0f27-4300-856a-60d26717bd44.mp4


Anemy:
Ah I'll re-open this with a shared toolbar component, probably better to add one now than later even if we do change up the styles a bit.

PR #2951, opened 2022-04-05
feat: visually identify CSFLE-decrypted fields COMPASS-5628

addaleax:
Note: This only works with the changes in https://github.com/mongodb-js/devtools-connect/pull/9 and https://github.com/mongodb/libmongocrypt/commit/adfdce0e29e4091a5006bce90665198e11076c0f.

Screenshot from 2022-04-05 22-21-59

(For now not placing this on the right-hand side of the screen as we had discussed last week, because we only actually show the types in edit mode currently)

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-components/src/components/document-list/element.tsx
```diff
@@ -466,6 +466,17 @@ export const HadronElement: React.FunctionComponent<{
         </div>
         <div className={elementDivider} role="presentation">
           :&nbsp;
+          {
+            /* TODO: figure out exact placement */
```
Anemy, line 470: How do we figure out exact placement? Is that pending w/ Claudia? If there is a ticket maybe we can link it.

Anemy (commented):
Screen Shot 2022-04-07 at 9 13 42 PM

One or two questions, looking good - also handles that long string case nicely ^^
Looking at the `maybeDecorateWithDecryptedKeys` it looks like we have to repeat that in a lot of places, is there a way we could keep it more in one place? Not sure if that's possible without some larger refactoring so I'm thinking its fine as is?

addaleax:
> Looking at the `maybeDecorateWithDecryptedKeys` it looks like we have to repeat that in a lot of places

I don’t think we’ll need to repeat it in more places than the ones that are present here. It *is* a bit weird that we duplicate the object generation code four times essentially, but I don’t think this PR is making things much worse? :woman_shrugging: It’s certainly possible to merge those (I certainly feel like I’m leaning towards DRY a bit more than others) but it might also be okay to keep it simple here.

packages/hadron-document/src/element.js
```diff
@@ -486,13 +498,42 @@ class Element extends EventEmitter {
...
+   */
+  isValueDecrypted() {
+    return this.decrypted;
+  }
+
+  /**
+   * Detemine if this value or any of its children were marked
+   * as having been decrypted with CSFLE.
+   *
+   * Warning: This does *not* apply to the children of these elements!
```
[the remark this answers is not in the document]
addaleax: > Maybe I'm misunderstanding this function, but aren't we evaluating here if the children of this element have decrypted children? 

I’ve made the comments a bit more explicit – it returns whether a child of this element has been decrypted. That would not apply to a child of a decrypted field, because decrypted fields generally don’t have decrypted children.

I will admit that I’ve never actually checked what happens when nested encryption happens on nested fields. I don’t think anybody would ever really do this on purpose, but I’ve opened COMPASS-5705 to remind myself to look into that, and how we would handle it in Compass. :slightly_smiling_face: 

> Should this warning be for `isValueDecrypted`?

It also applies there, that’s true :slightly_smiling_face: Added it.

packages/compass-components/src/components/document-list/element.tsx
addaleax: Yeah, I’ve opened a ticket and assigned it to her, added the reference here ;)

Anemy (approved):
lgtm! nice

PR #2955, opened 2022-04-06
feat(connection-form): implement optional CSFLE secrets storage COMPASS-5643

addaleax:
Screenshot from 2022-04-06 13-41-09

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/connection-form/src/components/advanced-options-tabs/csfle-tab/csfle-tab.tsx
+++ b/packages/connection-form/src/components/advanced-options-tabs/csfle-tab/csfle-tab.tsx
@@ -4,6 +4,7 @@ import type { ConnectionOptions } from 'mongodb-data-service';
 import {
   Accordion,
   Banner,
+  Checkbox,
   Label,
   Link,
   Description,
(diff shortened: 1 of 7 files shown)


packages/connection-form/src/components/advanced-options-tabs/csfle-tab/csfle-tab.tsx
```diff
@@ -146,6 +157,29 @@ function CSFLETab({
...
+          }}
+          data-testid="csfle-storeCredentials-input"
+          id="csfle-storeCredentials-input"
+          label={
+            <>
+              <Label htmlFor="csfle-storeCredentials-input">
+                Store KMS provider secrets
+              </Label>
+              <Description>
+                Control whether Compass stores KMS secrets, protected by the OS
```
Anemy, line 174: I'm wondering if there's a way, if we find it useful, for us to make this text more explicit that we're either storing or not storing here. Maybe a stretch, but I'm thinking some people might think it's a checkbox for storing in the OS keychain or not in the keychain.
Maybe something like
```
Control whether KMS secrets are stored on disk (protected by the OS keychain) or discarded after disconnecting.
```

Anemy (approved):
lgtm! One thought on possible copy, not too sure on it so feel free to go any direction (or make no changes).

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/csfle-tab.tsx
addaleax: Sure, done!

PR #2957, opened 2022-04-06
chore(compass-serverstats): Update Performance Tab toolbar to leafy green COMPASS-5588

Anemy:
COMPASS-5588
Updates the Performance Tab's toolbar to use leafy green components from `compass-components`. This doesn't update the error banners, those will be updated in [COMPASS-5671](https://jira.mongodb.org/browse/COMPASS-5671).

COMPASS-5589
This PR also adds a new `Toolbar` component in `compass-components` for the performance tab toolbar and the new toolbars to use, currently it's fairly minimal, but if we add something like drop shadow it should have some more use (thread in devtools design slack). We update the new aggregations toolbar to use this shared toolbar as well cc @mabaasit 

https://user-images.githubusercontent.com/1791149/162021550-e44067c6-5100-4a48-9b3d-6f770174a2a5.mp4


packages/compass-serverstats/src/components/index.jsx
```diff
@@ -1,21 +1,35 @@
...
-const { StatusRow } = require('hadron-react-components');
+const { ServerStatsToolbar } = require('./server-stats-toolbar');
 
 /**
  * The default interval.
  */
 const INTERVAL = 1000;
 
+const workspaceContainerStyles = css({
+  background: '#3D4247'
```
alenakhineika, line 21: Do we have this value somewhere in constants? And if not maybe we could introduce one?

alenakhineika (approved):

packages/compass-serverstats/src/components/index.jsx
Anemy: I couldn't find this value used anywhere else in Compass. Chatting with Claudia and looking through mocks the styles (and background color) here are soon to change so I'm thinking having this one is probably fine. I'll make it a variable in this file though 👍

Anemy:
Merging, looks like it was just data lake flaking.

PR #2958, opened 2022-04-06
feat(collection): move collection package to TypeScript COMPASS-5695

alenakhineika:
## Description

- Move collection package to TypeScript and CSS-in-JS
- Use color themes in compass stats package

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-collection-stats/src/components/document-stats-item/document-stats-item.tsx
```diff
@@ -31,8 +28,6 @@ type DocumentStatsItemProps = {
 const DocumentStatsItem: React.FunctionComponent<DocumentStatsItemProps> = ({
   documentCount,
   isTimeSeries,
```
addaleax, line 30: This component is essentially empty if `isTimeSeries` is true. For the index stats, we skip them here: https://github.com/mongodb-js/compass/pull/2958/files#diff-f6340fbdf9bd7dca1d628c1ea60ace9041137ca0d43f9e58d2221cc83aaa4212R84

Should the two be skipped the same way, i.e. outside of the component itself?

(We currently still do display the storage size for time series.)

packages/compass-collection/src/components/collection-header/collection-header.tsx
```diff
@@ -1,33 +1,77 @@
...
+  textOverflow: 'ellipsis',
+  whiteSpace: 'nowrap',
+});
+
+type CollectionHeaderProps = {
+  globalAppRegistry: AppRegistry;
+  namespace: string;
+  isReadonly: boolean;
+  isTimeSeries: boolean;
+  statsPlugin: React.FunctionComponent<any>;
```
addaleax, line 61: I think it’s fine to be a tiny bit more specific here, feel free to ignore:

```suggestion
  statsPlugin: React.FunctionComponent<{ store: any }>;
```

packages/compass-collection/src/components/collection-header/collection-header.tsx
```diff
@@ -1,33 +1,77 @@
...
+  selectOrCreateTab: React.FunctionComponent<any>;
```
addaleax, line 62: This doesn’t look as if it’s used as a component, but as an action (I think that’s the right term – never quite sure about these things … :slightly_smiling_face:)

Either way, I think typing it as `React.FunctionComponent` doesn’t quite communicate the right thing, even if the type might technically fit. I get that it’s a lot of work to actually go and replace all the `any` with concrete types, but I think this should be fine?

```suggestion
  selectOrCreateTab: (options: any) => any;
```

packages/compass-collection/src/components/collection/collection.tsx
```diff
@@ -0,0 +1,157 @@
...
+type CollectionProps = {
+  namespace: string;
+  isReadonly: boolean;
+  isTimeSeries: boolean;
+  statsPlugin: React.FunctionComponent<any>;
+  statsStore: any;
+  editViewName?: string;
+  sourceReadonly: boolean;
+  sourceViewOn?: string;
+  selectOrCreateTab: React.FunctionComponent<any>;
```
addaleax, line 43: Same as above:

```suggestion
  selectOrCreateTab: (options: any) => any;
```

packages/compass-collection/src/components/collection/collection.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+const collectionModalContainerStyles = css({
+  zIndex: 100,
+});
+
+type CollectionProps = {
+  namespace: string;
+  isReadonly: boolean;
+  isTimeSeries: boolean;
+  statsPlugin: React.FunctionComponent<any>;
```
addaleax, line 38: Same as above:

```suggestion
  statsPlugin: React.FunctionComponent<{ store: any }>;
```

packages/compass-collection/src/components/collection/collection.tsx
```diff
@@ -0,0 +1,157 @@
...
+  editViewName?: string;
+  sourceReadonly: boolean;
+  sourceViewOn?: string;
+  selectOrCreateTab: React.FunctionComponent<any>;
+  pipeline: Document[];
+  sourceName: string;
+  activeSubTab: number;
+  id: string;
+  queryHistoryIndexes: number[];
+  tabs: any;
```
addaleax, line 49:
```suggestion
  tabs: string[];
```

packages/compass-collection/src/components/workspace/workspace.tsx
```diff
@@ -69,65 +85,87 @@ function getIconGlyphForCollectionType(type) {
...
+  };
+};
+
 /**
  * The collection workspace contains tabs of multiple collections.
  */
-class Workspace extends PureComponent {
-  static displayName = 'Workspace';
+class Workspace extends PureComponent<WorkspaceProps> {
+  boundHandleKeypress: (evt: any) => void;
```
addaleax, line 125:
```suggestion
  boundHandleKeypress: (evt: React.KeyboardEvent) => void;
```

packages/compass-collection/src/components/workspace/workspace.tsx
```diff
@@ -136,7 +174,7 @@ class Workspace extends PureComponent {
    *
    * @param {Event} evt - The event.
    */
-  handleKeypress = (evt) => {
+  handleKeypress = (evt: any): void => {
```
addaleax, line 177:
```suggestion
  handleKeypress = (evt: React.KeyboardEvent): void => {
```

packages/compass-collection/src/components/workspace/workspace.tsx
```diff
@@ -69,65 +85,87 @@ function getIconGlyphForCollectionType(type) {
   }
 }
 
+type WorkspaceProps = {
+  tabs: any;
```
addaleax, line 89:
```suggestion
  tabs: any[];
```

packages/compass-collection/src/index.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { expect } from 'chai';
```
addaleax, line 1: Can’t comment on `packages/compass-collection/src/index.d.ts`, so I’ll do it here – it’s an empty file that should not be necessary, should we just drop it?

packages/compass-collection/src/stores/context.tsx
```diff
@@ -16,12 +17,40 @@ const { debug } = createLoggerAndTelemetry(
...
+  editViewName?: string;
+  sourcePipeline?: Document[];
+  query?: any;
+  aggregation?: any;
+  isEditing?: boolean;
+  key?: number;
+  state?: any;
+  isDataLake?: boolean;
+  queryHistoryIndexes?: number[];
+  statsPlugin?: React.FunctionComponent<any>;
```
addaleax, line 49:
```suggestion
  statsPlugin?: React.FunctionComponent<{ store: any }>;
```

packages/compass-collection/src/stores/is-data-lake.ts
```diff
@@ -21,7 +21,7 @@ export const INITIAL_STATE = false;
  *
  * @returns {String} The new state.
  */
-export default function reducer(state = INITIAL_STATE, action) {
+export default function reducer(state = INITIAL_STATE, action: any): any {
```
addaleax, line 24:
```suggestion
export default function reducer(state = INITIAL_STATE, action: any): boolean {
```

packages/compass-collection/src/stores/server-version.ts
```diff
@@ -21,7 +21,7 @@ export const INITIAL_STATE = '4.0.0';
  *
  * @returns {String} The new state.
  */
-export default function reducer(state = INITIAL_STATE, action) {
+export default function reducer(state = INITIAL_STATE, action: any): any {
```
addaleax, line 24:
```suggestion
export default function reducer(state = INITIAL_STATE, action: any): string {
```

packages/compass-collection/src/stores/tabs.ts
```diff
@@ -206,25 +210,25 @@ const doCloseTab = (state, action) => {
...
-const doDatabaseDropped = (state, action) => {
-  const tabs = state.filter((tab) => {
+const doDatabaseDropped = (state: any, action: any) => {
+  const tabs = state.filter((tab: any) => {
     const tabDbName = toNS(tab.namespace).database;
     return tabDbName !== action.name;
   });
   if (tabs.length > 0) {
-    if (tabs.findIndex(tab => tab.isActive) < 0) {
+    if (tabs.findIndex((tab: any) => tab.isActive) < 0) {
```
addaleax, line 231: Totally up to you, but `tab: any` or `tabs: any[]` pops up a lot here, maybe it’s worth actually typing them?

addaleax (commented):
Great work! It’s a large PR, nothing much to add here, mostly left a few comments for a few cases where I think we can avoid `any` without too much effort :+1:

packages/compass-collection/src/components/collection-header/collection-header.tsx
alenakhineika: For some `any` i really put an effort to figure out what should be used instead. In some cases, I didn't fully understand from where a prop is coming, therefore instead of confusing people later, who will take a look at the types I introduce here :) I left them as `any` with the hope that later someone who will work with this package and knows for sure what it is replaces those any.

With this particular case yeah I agree! Will push this change with the next commit.

packages/compass-collection/src/components/workspace/workspace.tsx
alenakhineika:
```
Type '(evt: KeyboardEvent<Element>) => void' is not assignable to type 'EventListener'.
  Types of parameters 'evt' and 'evt' are incompatible.
    Type 'Event' is missing the following properties from type 'KeyboardEvent<Element>': altKey, charCode, ctrlKey, code, and 13 more.
```

packages/compass-collection/src/index.spec.ts
alenakhineika: Thanks for noticing it! Added it temporarily and forgot to delete ✨

packages/compass-collection/src/components/workspace/workspace.tsx
addaleax: :woman_shrugging: Then let’s leave it as is :)

addaleax (approved):
Looks good! And thank you for doing `WorkspaceTabObject`, really glad to see that we’re a bit more specific about those :slightly_smiling_face:

Issue #2953, opened 2022-04-06
Compass Ui doesn´t show DB after ATLAS upgrade to 5.0.6


Sebs030:
When connecting to an Atlas hosted MongoDB the compass UI no longer shows the DB and the collections. 
![image](https://user-images.githubusercontent.com/29453191/161926908-e2b574d6-9e3f-4a9d-84b6-ab94b07f8e1d.png)

This happened after MongoDB was upgraded to 5.0.6
In the mongoshell I can run `show collections` and the Collections of the DB will be listed: 
<img width="1503" alt="image" src="https://user-images.githubusercontent.com/29453191/161927502-a52283e0-5598-454b-a020-84a4cf32d724.png">

skasyn:
I have the same issue.
<img width="572" alt="image" src="https://user-images.githubusercontent.com/38563623/162018382-039bac72-726e-4521-ba80-821d92786cf6.png">

Operating System: macOS Monterey 12.1
Compass Version: 1.30.1

Henrixounez:
Same problem here. Tried to run `show dbs` or `show databases` but it gives me an error but I don't know if that's regular behavior or not.

<img width="733" alt="image" src="https://user-images.githubusercontent.com/30901439/162019368-e2419974-4b4d-4748-b073-e2bc2bdf9d8a.png">

Operating System: macOS Monterey 12.1
MongoDB Compass Version: 1.31.1

Anemy:
Hi @Sebs030 , thanks for reporting. A couple questions if you have a moment to help us understand the issue:
What permissions does the user you're connecting with have?
Is the user able to `listDatabases` in the shell?

I think this is the same issue we have reported on JIRA: https://jira.mongodb.org/browse/COMPASS-5691
We'll use that to track the issue and we'll comment back here when we've got something to update on. Hopefully if this is an issue we can get a fix out to y'all quick.

Sebs030:
@Anemy , that´s interesting, so `show dbs` does not work, but `db` works. I actually did not change the connection string, it stopped working from one day to the other: 
![image](https://user-images.githubusercontent.com/29453191/162031366-a208a3ec-ad7c-44d3-aa5f-16f7bfc8d590.png)

The user for the connection String has readWrite access and even after adding dbAdmin rights I still get the above error

Sebs030:
[...]

PR #2959, opened 2022-04-06
fix(data-service): use databases where user has built in roles in `listDatabases` COMPASS-5691

Anemy:
COMPASS-5691 https://github.com/mongodb-js/compass/issues/2953

## Description
Atlas shared tier clusters (serverless, M0, M2) with MongoDB server version 5+ do not currently return privileges with the `connectionStatus` command response (fix might be in soon/already w/ [CLOUDP-115805](https://jira.mongodb.org/browse/CLOUDP-115805)). It also fails to `listDatabases`. As a result, currently a user with permissions to only one database would not be able to see their database. 

While there may be a release of Atlas soon making this issue go away, the changes in this PR introduce a workaround so users with exclusive database permissions can use their databases. These changes expand our extra `listDatabases` affordances by referencing the [authenticatedUserRoles](https://www.mongodb.com/docs/manual/reference/command/connectionStatus/#mongodb-data-connectionStatus.authinfo.authenticatedUserRoles) in the `connectionStatus` response. We then add all of the databases where a user has any of the `read`, `readWrite`, `dbAdmin`, `dbOwner` [built-in roles](https://www.mongodb.com/docs/manual/reference/built-in-roles/).

## Motivation and Context
- [x] Bugfix

## Open Questions
Is there another way to get this information? Is there anything we can do for cases where users are using custom roles with custom privileges for their access management?

## Types of changes

### DIFF
--- a/packages/data-service/src/data-service.ts
+++ b/packages/data-service/src/data-service.ts
@@ -342,7 +342,6 @@ class DataService extends EventEmitter {
     const {
       authInfo: { authenticatedUserPrivileges },
     } = await this.connectionStatus();
-
     return authenticatedUserPrivileges;
   }
 
--- a/packages/data-service/src/data-service.ts
+++ b/packages/data-service/src/data-service.ts
@@ -58,6 +58,7 @@ import {
   getPrivilegesByDatabaseAndCollection,
   checkIsCSFLEConnection,
   getInstance,
+  getDatabasesByRoles,
 } from './instance-detail-helper';
 import { redactConnectionString } from './redact';
 import connectMongoClient from './connect-mongo-client';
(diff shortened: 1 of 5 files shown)


packages/data-service/src/data-service.ts
```diff
@@ -500,16 +520,33 @@ class DataService extends EventEmitter {
...
+          listDatabases(),
+          getDatabasesFromPrivileges(),
+          getDatabasesFromRoles(),
+        ]);
 
       const databases = uniqueBy(
         // NB: Order is important, we want listed collections to take precedence
         // if they were fetched successfully
-        [...databasesFromPrivileges, ...listedDatabases],
+        [...databasesFromRoles, ...databasesFromPrivileges, ...listedDatabases],
```
lerouxb, line 549: I'm trying to make sense of the comment that was here before. Where's listed collections coming from?

lerouxb (approved):
It will be interesting to see if we can get a release out with this fix before cloud just fixes it.

lerouxb:
The ubuntu failure is a flake (something about the timing of that leafygreen component I don't understand well enough to reliably work around) and the others are failing due to 5.3. Should be fixed by https://github.com/mongodb-js/compass/pull/2937

lerouxb:
I updated the branch so long to pull in the e2e fix for 5.3

packages/data-service/src/data-service.ts
mcasimir: I'm assuming this is a merge of databases taken from the privileges and the result of `adminDb.command({listDatabases: ...})` + `adminDb.command({listCollections: dbName})`, probably the db objects returned by the commands are more accurate than the privileges one

PR #2964, opened 2022-04-07
Add DeclarationStore to transpiler (POC)

prestonvasquez:
## Description

[COMPASS-5685](https://jira.mongodb.org/browse/COMPASS-5685)

See the the updated [README.md ](https://github.com/prestonvasquez/compass/tree/COMPASS-5685.addVariableDeclarationsToBSONTranspiler/packages/bson-transpilers#declarationstore) for the bson-transpilers package.

### Checklist

## Motivation and Context

The go driver export has a need to prepend the driver syntax with variable declarations to avoid using closures within the CRUD operations.  For example, the following would be what is generated under the current architecture:

```go
// This would be embedded in some CRUD Operation and is what would be transpiled from
// { _id: ObjectID("5ab901c29ee65f5c8550c5b9") }
bson.D{{"_id", func(hex string) primitive.ObjectID {
objectID, err := primitive.ObjectIDFromHex(hex)
if err != nil {
log.Fatal(err)
}
return objectID
}("5ab901c29ee65f5c8550c5b9")}}
```

The proposed changes would add an optional functionality to prepend the document body with variable declarations
```go
objectID err := primitive.ObjectIDFromHex("5ab901c29ee65f5c8550c5b9")
if err != nil {
log.Fatal(err)
}
 
// this would be embedded in some CRUD Operation
bson.D{{"_id", objectID}}
```

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions

1. Does the method for passing the declarationStore into the `argsTemplate` hold up?  That is, is it best practices? 
2. Does `DeclarationStore` follow good JS class-naming standards?

## Dependents
[GODRIVER-2268](https://jira.mongodb.org/browse/GODRIVER-2268)

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/bson-transpilers/README.md
+++ b/packages/bson-transpilers/README.md
@@ -59,6 +59,38 @@ Any transpiler errors that occur will be thrown. To catch them, wrap the
 - __error.column:__ If it is a syntax error, will have the column.
 - __error.symbol:__ If it is a syntax error, will have the symbol associated with the error.
 
+### DeclarationStore
+The motivation for using `DeclarationStore` is to prepend the driver syntax with variable declarations rather than using closures in the CRUD logic.
+
+More specifically, the `DeclarationStore` class maintains state concerning variable declarations in the driver syntax.  Within the context of the symbols template, the use case is to pass the declaration store as a parameter for the `argsTemplate` anonymous function.  For example,
+
+```javascript
+// within the args template
+(arg, declarationStore) => {
+  return declarationStore.add("objectID", (varName) => {
+    return [
+      `${varName}, err := primitive.ObjectIDFromHex(${arg})`,
+      'if err != nil {',
+      '   log.Fatal(err)',
+      '}'
+    ].join('\n')
+  })
+}
+```
+
+Note that each use of the same variable name will result in an incrament being added to the declaration statement. For example, if the variable name `objectID` is used two times the resulting declaration statements will use `objectID` for the first declaration and `objectID2` for the second declaration.  The `add` method returns the incremented variable name, and is therefore what would be expected as the right-hand side of the statement defined by the `template` function.
+
+The instance of the `DeclarationStore` constructed by the transpiler class is passed into the driver syntax for use:
+
+```javascript
+(spec, declarationStore) => {
+  const comment = '// some comment'
+  const client = 'client, err := mongo.Connect(context.Background(), options.Client().ApplyURI(cs.String()))'
+  const declarations = declarationStore.toString()
+  return "#{comment}\n\n#{client}\n\n${declarations}"
+}
+```
+
 ### Errors
 There are a few different error classes thrown by `bson-transpilers`, each with
 their own error code:
--- a/packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js
+++ b/packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js
@@ -10,6 +10,7 @@ const {
 } = require('../helper/error');
 
 const { removeQuotes } = require('../helper/format');
+const DeclarationStore = require('./DeclarationStore');
 
 /**
  * Class for code generation. Goes in between ANTLR generated visitor and
@@ -24,6 +25,7 @@ module.exports = (ANTLRVisitor) => class CodeGenerationVisitor extends ANTLRVisi
     super();
     this.idiomatic = true; // PUBLIC
     this.clearImports();
+    this.declarationStore = new DeclarationStore();
   }
 
   clearImports() {
@@ -62,6 +64,10 @@ module.exports = (ANTLRVisitor) => class CodeGenerationVisitor extends ANTLRVisi
    */
   start(ctx) {
     return this.returnResult(ctx).trim();
+  }
+
+  getDeclarationStore() {
+    return this.declarationStore;
   }
 
   /**
@@ -413,6 +419,7 @@ module.exports = (ANTLRVisitor) => class CodeGenerationVisitor extends ANTLRVisi
     this.requiredImports[ctx.type.code] = true;
 
     if (ctx.type.template) {
+      // ! of note
       return ctx.type.template();
     }
     return this.returnFunctionCallLhs(ctx.type.code, name);
@@ -490,6 +497,12 @@ module.exports = (ANTLRVisitor) => class CodeGenerationVisitor extends ANTLRVisi
     if (`emit${lhsType.id}` in this) {
       return this[`emit${lhsType.id}`](ctx);
     }
+    // if there are no args, but a buffer in for argTemplate methods that are
+    // expecting to key off of the fact that there are no args.
+    if (args.length === 0) {
+      args = [undefined];
+    }
+    args = [...args, this.declarationStore];
     const lhsArg = lhsType.template
       ? lhsType.template()
       : defaultT;
--- a/packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js
+++ b/packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js
@@ -496,8 +496,8 @@ module.exports = (ANTLRVisitor) => class CodeGenerationVisitor extends ANTLRVisi
     if (`emit${lhsType.id}` in this) {
       return this[`emit${lhsType.id}`](ctx);
     }
-    // if there are no args, but a buffer in for argTemplate methods that are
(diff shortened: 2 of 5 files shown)


packages/bson-transpilers/README.md
```diff
@@ -59,6 +59,38 @@ Any transpiler errors that occur will be thrown. To catch them, wrap the
...
+      `${varName}, err := primitive.ObjectIDFromHex(${arg})`,
+      'if err != nil {',
+      '   log.Fatal(err)',
+      '}'
+    ].join('\n')
+  })
+}
+```
+
+Note that each use of the same variable name will result in an incrament being added to the declaration statement. For example, if the variable name `objectID` is used two times the resulting declaration statements will use `objectID` for the first declaration and `objectID1` for the second declaration.  The `add` method returns the incremented variable name, and is therefore what would be expected as the right-hand side of the statement defined by the `template` function.
```
addaleax, line 81:
```suggestion
Note that each use of the same variable name will result in an increment being added to the declaration statement. For example, if the variable name `objectID` is used two times the resulting declaration statements will use `objectID` for the first declaration and `objectID1` for the second declaration.  The `add` method returns the incremented variable name, and is therefore what would be expected as the right-hand side of the statement defined by the `template` function.
```

packages/bson-transpilers/codegeneration/DeclarationStore.js
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Add declaration statements by a pre-incremented variable root-name
+   *
+   * @param {string} varRoot - The root of the variable name to be appended by the occurance count
+   * @param {function} declaration - The code block to be prepended to the driver syntax
+   * @returns {string} the variable name with root and appended count
+   */
+  add(varRoot, declaration) {
+    const varName = this.next(varRoot);
+    const data = {varName: varName, declaration: declaration(varName) };
```
addaleax, line 20:
```suggestion
    const data = { varName: varName, declaration: declaration(varName) };
```

(surprised the linter didn’t say anything)

alenakhineika:
Looks good! Since you are updating readme, could you please also add `ruby` to the output languages at the beginning of the readme file? We forgot to do it when added its support.

packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js
```diff
@@ -24,6 +25,7 @@ module.exports = (ANTLRVisitor) => class CodeGenerationVisitor extends ANTLRVisi
     super();
     this.idiomatic = true; // PUBLIC
     this.clearImports();
+    this.state = { declarationStore: new DeclarationStore() };
```
prestonvasquez, line 28: @alenakhineika @addaleax @mcasimir I ran into an issue concerning the way we handle the `argsTemplate` functions.  Since they are variadic, passing in the declaration store as the last value will cause a lot of problems with existing language templates.  [This one](https://github.com/mongodb-js/compass/blob/main/packages/bson-transpilers/symbols/csharp/templates.yaml#L770), in particular, comes to mind.  To fix this, I  propose

1. Adding this state hash in case we ever need to add state for something else in the future
2. Updating all uses of `argsTemplate` so that the first argument will be be `this.state`

packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js
```diff
@@ -53,15 +55,32 @@ module.exports = (ANTLRVisitor) => class CodeGenerationVisitor extends ANTLRVisi
...
    * an attribute called "startNode".
    *
    * @param {ParserRuleContext} ctx
+   * @param {Boolean} useDeclarations - prepend the result string with declarations
    * @return {String}
    */
-  start(ctx) {
-    return this.returnResult(ctx).trim();
+  start(ctx, useDeclarations = false) {
+    return (useDeclarations ? this.returnResultWithDeclarations(ctx) : this.returnResult(ctx)).trim();
```
prestonvasquez, line 75: This change is definitely of note, it's what allows us to prepend the raw document output with functions/variables:

Screen Shot 2022-04-13 at 11 41 22 AM

There might be a much better way of going about this, though.

prestonvasquez:
@matthewdale @benjirewis Here is the final draft for the code to prepend declarations to statements.

packages/bson-transpilers/README.md
```diff
@@ -3,7 +3,7 @@
 [![downloads][5]][6]
 
 Transpilers for building BSON documents in any language. Current support
-provided for `shell` `javascript` and `python` as inputs. `java`, `c#`, `node`, `shell` and `python` as
+provided for `shell` `javascript` and `python` as inputs. `java`, `c#`, `node`, `shell`, `python` and `ruby` as
```
benjirewis, line 6: Should this say `go`?

packages/bson-transpilers/README.md
```diff
@@ -59,6 +59,57 @@ Any transpiler errors that occur will be thrown. To catch them, wrap the
...
+        return '';
+    }
+
+DriverTemplate: &DriverTemplate !!js/function >
+    (_spec) => {
+      return this.oneLineStatement;
+    }
+```
+
+The output of the driver syntax for this language will be the one-line statement `Hello World`.
```
benjirewis, line 80: Cool! 🧑‍🔧

packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js
```diff
@@ -545,7 +564,7 @@ module.exports = (ANTLRVisitor) => class CodeGenerationVisitor extends ANTLRVisi
     if (ctx.type.argsTemplate) { // NOTE: not currently being used anywhere.
       args = visitedElements.map((arg, index) => {
         const last = !visitedElements[index + 1];
-        return ctx.type.argsTemplate(arg, ctx.indentDepth, last);
+        return ctx.type.argsTemplate.bind(this.getState())(arg, ctx.indentDepth, last);
```
benjirewis, line 567: Does `bind` just merge the variables in `state` with the ones already in `this`?

packages/bson-transpilers/codegeneration/DeclarationStore.js
```diff
@@ -0,0 +1,90 @@
...
+  constructor() {
+    this.clear();
+  }
+
+  /**
+   * Add declarations by templateID + varRoot + declaration combo.  Duplications will not be collected, rather the add
+   * method will return the existing declaration's variable name.
+   *
+   * @param {string} templateID - Name/alias of the template the declaration is being made for
+   * @param {string} varRoot - The root of the variable name to be appended by the occurance count
```
benjirewis, line 16:
```suggestion
   * @param {string} varRoot - The root of the variable name to be appended by the occurrence count
```

packages/bson-transpilers/test/declaration-store.test.js
```diff
@@ -0,0 +1,231 @@
...
+    const declaration3 = 'var y := func() {}';
+
+    ds.addFunc(declaration1);
+    ds.addFunc(declaration2);
+    ds.addFunc(declaration3);
+    ds.addVar('x', 'y', () => 'z');
+
+    assert.strictEqual(ds.length(), 3);
+  });
+});
```
benjirewis, line 231: Great testing 🧑‍🔧

packages/bson-transpilers/README.md
prestonvasquez: Ah yeah, we should definitely include `go`.  Let me add that to [the ](https://github.com/mongodb-js/compass/pull/2991) so that this PR can remain language agnostic.  I added ruby here based for [this request](https://github.com/mongodb-js/compass/pull/2964#issuecomment-1092782188).

packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js
prestonvasquez: It actually replaces the existing `this` of the prototypes with `state`.  The solutions _seems_ safe based on my conversations with @mcasimir , but there may definitely be underlying problems I'm not aware of.  An alternative would be to merge the two objects before the bind.

packages/bson-transpilers/codegeneration/DeclarationStore.js
prestonvasquez: Good catch!

PR #2965, opened 2022-04-11
feat(header): update collection header styles COMPASS-5519

alenakhineika:
## Description

Update styles and components used in the collection header. This is the area with the namespace and collection stats. To simplify the codebase and get rid of the reflux store, combine the collection stats and collection packages into one new package.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-components/src/index.ts
```diff
@@ -34,6 +34,9 @@ export {
 export { Checkbox } from './components/checkbox';
 export { default as LeafyGreenProvider } from '@leafygreen-ui/leafygreen-provider';
 
+import Box from '@leafygreen-ui/box';
+export { Box };
```
gribnoysup, line 38: This one is really low level, so I wanted to ask if this is something we are planning to do more now, using Box instead of just html elements?

alenakhineika: Not sure. They kind of use it as a link here in example: https://www.mongodb.design/component/box/example/
When a link from typography has a very specific style with an icon: https://www.mongodb.design/component/typography/example/
I am ok with having just `<a></a>` here, but I thought the idea was to use leafy green as much as possible.

gribnoysup: Actionable elements on the screen should be buttons or links otherwise they can't be accessed by keyboard. If you need it to look like a link, you can check out how we do this in other places:

https://github.com/mongodb-js/compass/blob/b06c5ed05524d60d4705c7bea0df33b018a7342a/packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-actions.tsx#L55-L66

packages/compass-collection/src/modules/stats.ts
```diff
@@ -0,0 +1,126 @@
...
+ * Collection details reset action name.
+ */
+export const RESET_COLLECTION_DETAILS = `${PREFIX}/stats/RESET_COLLECTION_DETAILS`;
+
+/**
+ * Invalid stats.
+ */
+const INVALID = 'N/A';
+
+export interface StatsObject {
```
Anemy, line 24: Small idea or suggestion, since this `StatsObject` looks like it always pertains to the stats of a collection, how do you feel about the name `CollectionStats`?
```suggestion
export interface CollectionStats {
```

packages/compass-collection/src/components/collection/collection.tsx
```diff
@@ -59,37 +73,41 @@ type CollectionProps = {
...
+    stats,
+  } = props;
   useEffect(() => {
     if (tabs && tabs.length > 0) {
       track('Screen', {
         name: trackingIdForTabName(tabs[activeSubTab] || 'Unknown'),
       });
     }
-  }, []);
+  });
```
Anemy, line 110: Giving `useEffect` no second argument acts as both componentDidMount and componentDidUpdate, as in it runs first on mount and then on every re-render. Do we want to keep this using the `[]`? I'm not sure I follow the change. If we want to track the sub tab changes maybe we make it use `activeSubTab` and `tabs` as dependencies?

Anemy (commented):
Nice work, I like that we're doing some simplification here. I left a couple comments, mostly on some ways we might improve types now that we have types here

packages/compass-collection/src/components/collection/collection.tsx
alenakhineika: Lint complaints with []: `React Hook useEffect has missing dependencies: 'activeSubTab' and 'tabs'. Either include them or remove the dependency array.`

What would be the proper approach?

Anemy: Lint is showing that error because function is using the `activeSubTab` and `tabs` props inside of it. Because of the function's scope those variables could be stale in the useEffect, or not the current, correct value if we pass only `, []` because the function is only set on first mount. This might be alright if we only want it to run once, but in this case do we want the tracking to occur every time the `activeSubTab` changes? 
If we want the tracking event every time the `activeSubTab` or `tabs` change we can add them to the dependency list like this:
```suggestion
  }, ['activeSubTab', 'tabs']);
```

More information on how `useEffect` and other hooks use these dependencies and why the `exhaustive-deps` lint shows up can be found here: https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect

Anemy: If we want the event to only occur when the `activeSubTab` changes and not the `tabs` we could use another hook, `useMemo` to track the active subtab name and then add that new variable to this dependency list.
Something like:

```ts
const activeSubtabName = useMemo(() => trackingIdForTabName((tabs && tabs[activeSubTab]) || 'Unknown'), ['activeSubTab', 'tabs']);
```

```suggestion
  }, [activeSubtabName]);
```

packages/compass-components/src/index.ts
alenakhineika: Thx for sharing the example! Will do this change!

packages/compass-collection/src/modules/stats.ts
```diff
@@ -0,0 +1,126 @@
...
+  };
+};
+
+/**
+ * Action creator for clearing tabs.
+ *
+ * @returns {Object} The action.
+ */
+export const updateCollectionDetails = (
+  collectionModel: any
```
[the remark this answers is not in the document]
gribnoysup: We do have collection-model typedefs

packages/compass-collection/src/modules/stats.ts
alenakhineika: It will overlap with the component name then, but I will make it `CollectionStatsObject` to be more explicit.

packages/compass-collection/src/components/collection-header-actions/view-information.tsx
```diff
@@ -1,10 +1,10 @@
-import { css } from '@mongodb-js/compass-components';
+import { css, spacing } from '@mongodb-js/compass-components';
 import React from 'react';
 
 const collectionHeaderActionsReadonlyStyles = css({
-  margin: '0px 8px',
+  margin: `0px ${spacing[2]}px`,
   marginLeft: 'auto',
-  fontSize: '16px',
+  fontSize: spacing[3],
```
gribnoysup, line 7: Is it allowed to use spacing for font sizes and line heights? I can't see this font size in leafygreen anywhere

packages/compass-collection/src/components/collection-header/collection-header.tsx
```diff
@@ -47,26 +55,49 @@ const collectionHeaderTitleDBStyles = css({
...
+  flexBasis: 'auto',
+  overflow: 'hidden',
+  textOverflow: 'ellipsis',
+  whiteSpace: 'nowrap',
+  cursor: 'pointer',
+  textDecoration: 'none',
+  '&:hover,&:focus': {
+    textDecoration: 'underline',
+  },
+});
```
gribnoysup, line 70: Seems like we are duplicating quite a lot of styles between light and dark here, would it make sense to have only properties that change defined separately?

packages/compass-collection/src/components/collection-stats-item/collection-stats-item.tsx
```diff
@@ -18,40 +18,41 @@ const collectionStatsItemStyles = css({
...
 
 const lightThemeLabelStyles = css({
   textTransform: 'uppercase',
-  fontSize: '10px',
-  fontWeight: 'bold',
+  color: uiColors.gray.dark1,
   display: 'inline-block',
-  color: uiColors.gray.dark2,
+  fontSize: spacing[2] * 1.5,
+  lineHeight: `${spacing[3]}px`,
```
gribnoysup, line 33: Is it okay that we use these very custom font sizes here? My understanding was that we stick to what typography packages allows us to use

packages/compass-collection/src/components/collection/collection.tsx
```diff
@@ -2,21 +2,36 @@ import type AppRegistry from 'hadron-app-registry';
...
-const collectionStyles = css({
+const collectionLightStyles = css({
   display: 'flex',
   alignItems: 'stretch',
   height: '100%',
   width: '100%',
+  background: uiColors.white,
+});
+
+const collectionDarkStyles = css({
```
gribnoysup, line 29: Same as above, wouldn't it make sense to have only styles that change extracted? This should make maintaining this a bit easier over time

packages/compass-collection/src/stores/index.ts
```diff
@@ -123,6 +171,7 @@ store.onActivated = (appRegistry: AppRegistry) => {
             sourcePipeline: metadata.sourcePipeline,
             query: metadata.query,
             aggregation: metadata.aggregation,
+            stats: metadata.stats,
```
gribnoysup, line 174: This `metadata` object doesn't have `stats` available

packages/compass-collection/src/stores/index.ts
```diff
@@ -76,6 +88,8 @@ const appReducer = combineReducers({
   serverVersion,
   tabs,
   isDataLake,
+  stats,
```
gribnoysup, line 91: I see that you have `stats` as a slice now and it inherits most of the previous plugin logic (like getting updated when collection status updates), but it doesn't look like it's ever used in the collections as you are also store stats info with every tab and this is what gets passed to the header. I think we should decide which approach to take here: either have the singleton logic copied over, or keep stats data copied over to all the tabs, but then keeping them in sync when collection status changes (this is not happening at the moment, you can reproduce by opening a collection from the sidebar without opening collections list first)

packages/compass-collection/src/components/collection-header-actions/view-information.tsx
alenakhineika: I don't know, I have seen it in other places in compass implemented like this, so I thought you have some kind of agreement. What should be used instead?

packages/compass-collection/src/components/collection-header/collection-header.tsx
alenakhineika: I was also considering this but was wondering if it reduces code readability with lots of styles combined with each other.

packages/compass-collection/src/components/collection-stats-item/collection-stats-item.tsx
alenakhineika: With sizes out of the box, it doesn't match the design 😞

packages/compass-collection/src/components/collection-header/collection-header.tsx
Anemy: While it does introduce more complexity at the rendering level I think I would agree that separating the theme styles from the main component styles that both share is probably worth doing, or it's at least my preference. I think having general css rules like `flexShrink` or `overflow` in both of the styles can fairly easily lead to one style being out of sync with the other if one is changed without the other.

Here's how I was doing it in `find-in-page`:
https://github.com/mongodb-js/compass/blob/3bfdea024eff1067e282681d729039fb1c6301b4/packages/compass-find-in-page/src/components/find-in-page-input.tsx#L138
It's definitely more things going on in the rendering, but I think `cx` might help improve readability concerns.

packages/compass-collection/src/components/collection/collection.tsx
alenakhineika: Returned []!

packages/compass-collection/src/components/collection-stats-item/collection-stats-item.tsx
gribnoysup: Hmmm, shouldn't designs be using spacings from leafygreen though? @Anemy is it some special case here that we decided doesn't need to be aligned with leafygreen?

packages/compass-collection/src/components/collection-header-actions/view-information.tsx
alenakhineika: I replaced all places in `compass-collection` where I used spacing for font sizes with some elements from typography. Pls let me know if this is ok I need to update it in some other way. I also removed line heights.

packages/compass-collection/src/stores/index.ts
alenakhineika: Fixes opening new tabs with updated stats info. Hope I understood the data flow correctly. Pls, let me know if misused something.

gribnoysup:
Something in this PR is causing a very weird e2e test failure, I'm trying to figure out what it is

packages/compass-collection/src/components/collection-stats-item/collection-stats-item.tsx
Anemy: Sorry for missing this, looks like it's gone now so I'll resolve. I think we typically should avoid using `lineHeight` if possible. We should aim to use `spacings` from leafygreen as much as possible. There are rare cases where I think it makes sense to deviate and use custom styles though.

packages/compass-collection/src/components/collection-header/collection-header.tsx
```diff
@@ -9,65 +16,90 @@ import CollectionHeaderActions from '../collection-header-actions';
...
-  paddingTop: '15px',
-  paddingBottom: '5px',
-  minHeight: '64px',
+  paddingTop: spacing[3],
+  paddingBottom: spacing[1],
+  minHeight: spacing[6] + spacing[1],
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'space-between',
   background: uiColors.white,
```
Anemy, line 29: Not totally related, looks like this background might not be needed anymore? If we remove this background setting it looks like the background adjusts based on the background color. (with the env variable `COMPASS_LG_DARKMODE`)
```suggestion
```

Screen Shot 2022-04-20 at 5 59 13 PM
Screen Shot 2022-04-20 at 6 03 11 PM

packages/compass-collection/src/components/collection-header-actions/collection-header-actions.tsx
```diff
@@ -1,22 +1,29 @@
...
+  ButtonSize,
+  css,
+  spacing,
+} from '@mongodb-js/compass-components';
 import React from 'react';
 
 import ViewInformation from './view-information';
 
+const editViewButtonStyles = css({
+  marginLeft: spacing[2],
```
Anemy, line 12: Maybe
```suggestion
    flexShrink: 0,
    marginLeft: spacing[2],
```

Screen Shot 2022-04-20 at 6 07 05 PM
Screen Shot 2022-04-20 at 6 08 32 PM

packages/compass-collection/src/modules/stats.ts
```diff
@@ -0,0 +1,130 @@
...
+  const {
+    document_count,
+    index_count,
+    index_size,
+    status,
+    avg_document_size,
+    storage_size,
+    free_storage_size,
+  } = collectionModel;
+  let stats = INITIAL_STATE;
```
Anemy, line 91: Should we 
```suggestion
  let stats = { ...INITIAL_STATE };
```
here to ensure we don't mutate the `INITIAL_STATE` object?

packages/compass-collection/src/modules/stats.ts
```diff
@@ -0,0 +1,130 @@
...
+ * @returns {any} The new state.
+ */
+const reducer = (
+  state = INITIAL_STATE,
+  action: AnyAction
+): CollectionStatsObject => {
+  if (action.type === UPDATE_COLLECTION_DETAILS) {
+    return action.stats;
+  } else if (action.type === RESET_COLLECTION_DETAILS) {
+    return INITIAL_STATE;
```
Anemy, line 125: Same as above, would object spreading this make the INTIAL_STATE and `action.stats;` usage a little safer?

packages/compass-collection/src/components/collection-header-actions/collection-header-actions.tsx
gribnoysup: Sometimes I feel like doing a single PR that just puts `flex: none` on literally all the buttons in the app, I don't think we ever want them to collapse like that 😅

gribnoysup:
@Anemy following up on our discussion [here's a fix](https://github.com/mongodb-js/compass/pull/2965/commits/1663bf7972633f98c8a0c7483c0c096674707e76) I came up with: we are still cleaning up the tabs on disconnect, and at the same time the `all-tabs-closed` is now only emitted when all tabs are closed by the user and not programmatically, check it out and tell me if that makes sense to you

packages/compass-collection/src/components/collection-header/collection-header.tsx
```diff
@@ -18,11 +18,12 @@ import TimeSeriesBadge from './time-series-badge';
 import ViewBadge from './view-badge';
 import CollectionStats from '../collection-stats';
 import type { CollectionStatsObject } from '../../modules/stats';
+import ClusteredBadge from './clustered-badge';
 
 const collectionHeaderStyles = css({
   paddingTop: spacing[3],
   paddingBottom: spacing[1],
-  minHeight: spacing[6] + spacing[1],
+  minHeight: +spacing[6] + +spacing[1],
```
Anemy, line 26: Do we need a min height here? I'm thinking letting the elements define the height and adding some bottom padding here would be a solution the might make the height not required to change if the elements shrink in new leafygreen iterations for instance?

Anemy (approved):
lgtm! A couple open questions on a few places we might be able to do some improvements probably worth looking into before merging.

I like that the e2e failure fix you added removes the very general `store.subscribe` which definitely could have led to similar issues in future, nice.

packages/compass-collection/src/components/collection-header/collection-header.tsx
Anemy: We need it for keeping consistent height when swapping between views and non-views as the elements on the right are a different height.

Anemy:
Added a commit updating some styles and fixing some overflow with long collection names.

gribnoysup:
Looks good to me. Do we have tickets for fixing all the any types that we are introducing here? Might be worth adding so we don't forget to clean them up

packages/compass-collection/src/components/collection-header/collection-header.tsx
```diff
@@ -23,11 +23,11 @@ import ClusteredBadge from './clustered-badge';
 const collectionHeaderStyles = css({
   paddingTop: spacing[3],
   paddingBottom: spacing[1],
-  minHeight: +spacing[6] + +spacing[1],
+  height: +spacing[6] + +spacing[2],
```
alenakhineika, line 26: To be honest this looks very weird :) Don't we know for sure that spacing value is a number? We could add an extra step to `compass-components` to ensure that it is a number or 0 if someone misused an index.

gribnoysup: Ah, that's on me, for some reason TS doesn't compile for me locally if I don't explicitly convert these to numbers, if it doesn't cause errors for y'all totally feel free to remove this

alenakhineika: Ah got ya! You probably have a newer version than we do. From my experience, it is better to keep it then, because later we will all encounter it anyway :)

PR #2966, opened 2022-04-11
feat(collections): Clustered collections COMPASS-5666

lerouxb:
* new feature flag COMPASS_CLUSTERED_COLLECTIONS. If that's set everything below will show up when connected to 5.3+.
* new checkbox when adding a collection, name and expireAfterSeconds fields
* show the badge in the documents header
* show the badge in the collections list

For now not showing the badge in the sidebar because we only have type and name. For these special kinds of collections the type is "collection" and you have to identify the specialness via options which is not included when you do nameOnly which is what we're currently using to fetch collections in the sidebar.

I'm unclear about whether Capped Collection and Clustered Collection are really mutually exclusive in the same way they are with Time-Series. I'm treating it the same way for now. Checkboxes that mostly act like radio buttons are becoming an increasingly weird UI paradigm in the Create Collection modal IMHO. We'll probably have to revisit that.

**Edit:** Actually I just tested it:

```
Enterprise test> db.createCollection('cappedAndClustered', {
...   clusteredIndex: {
.....     key: { _id: 1 },
.....     unique: true,
.....     name: "my clustered key"
.....   },
...   capped: true,
...   size: 1024,
...   expireAfterSeconds: 5
... })
```

```
MongoServerError: Clustered capped collection only available with 'enableTestCommands' server parameter
```

This is in 5.3 and also latest 6.0 alpha. Not sure if that will change by the time 6.0 gets released, but for now all the checkboxes except Use Custom Collation seem to be mutually exclusive.

Right now the form passes `unique: true` and `key: { _id: 1 }` because those properties are required and each only has one valid value for now. We can change this in future if unique: false or different keys become possible. I confirmed this with Michael Gargiulo.

He also pointed out that `expireAfterSeconds` is a bit weird here compared to with time-series. `To create a TTL index, use the createIndex() method on a field whose value is either a date or an array that contains date values`. While Object IDs are indeed date based, if you use one as a doc's `_id` then that document will not expire. I tested this manually via mongosh to confirm. So I updated the help text for that accordingly.

Screenshot 2022-04-11 at 13 59 19

Screenshot 2022-04-19 at 17 02 29

Screenshot 2022-04-19 at 17 02 34


lerouxb:
Actually thinking about this again expireAfterSeconds has nothing to do with clustered collections and I think I should remove it again.

packages/data-service/src/data-service.ts
```diff
@@ -370,7 +370,9 @@ class DataService extends EventEmitter {
 
     const listCollections = async () => {
       try {
-        return await db.listCollections(filter, { nameOnly }).toArray();
+        const collections = await db.listCollections(filter, { nameOnly }).toArray();
+        console.log('all collecitons', collections);
+        return collections;
```
addaleax, line 375: Debugging leftover? :)

packages/databases-collections/src/components/collection-fields/clustered-collection-fields.jsx
```diff
@@ -0,0 +1,79 @@
+import React, { useCallback } from 'react';
+import PropTypes from 'prop-types';
+import { TextInput } from '@mongodb-js/compass-components';
+
+import CollapsibleFieldSet from '../collapsible-field-set/collapsible-field-set';
+import FieldSet from '../field-set/field-set';
+
+const HELP_URL_CLUSTERED = 'https://www.mongodb.com/docs/v5.3/core/clustered-collections/';
```
addaleax, line 8: Maybe leave a TODO to update this after 6.0?

packages/data-service/src/data-service.ts
lerouxb: Just a draft that I pushed to be able to point at stuff while I'm working on it ;)

packages/databases-collections/src/components/collection-fields/clustered-collection-fields.jsx
lerouxb: You mean change v5.3 to current?

addaleax: I think it would be just `manual` instead of `v5.3` :)

packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -367,6 +367,7 @@ async function startCompass(opts: StartCompassOptions = {}): Promise<Compass> {
     waitforInterval: 100, // default is 500ms
   };
 
+  process.env.COMPASS_CLUSTERED_COLLECTIONS = 'true';
```
lerouxb, line 370: Always enabling the feature flag in e2e tests so we can test this. Wondering if we shouldn't make that policy, actually.

packages/compass-e2e-tests/tests/database-collections-tab.test.ts
```diff
@@ -179,7 +179,9 @@ describe('Database collections tab', function () {
     const collectionCard = await browser.$(selector);
     await collectionCard.waitForDisplayed();
 
-    // TODO: how do we make sure this is really a collection with a custom collation?
+    await collectionCard
```
lerouxb, line 182: As a driveby I just made sure the collation collection has the collation badge as a cheap way to try and ensure it really did what we asked for.

packages/compass-e2e-tests/tests/database-collections-tab.test.ts
```diff
@@ -210,6 +212,40 @@ describe('Database collections tab', function () {
     const collectionCard = await browser.$(selector);
     await collectionCard.waitForDisplayed();
 
-    // TODO: how do we make sure this is really a timeseries collection?
+    await collectionCard
+      .$('[data-testid="collection-badge-timeseries"]')
```
lerouxb, line 216: Same for timeseries.

packages/compass/src/index.d.ts
```diff
@@ -45,6 +45,8 @@ declare module 'process' {
 
         COMPASS_CSFLE_SUPPORT?: 'true' | 'false';
 
+        COMPASS_CLUSTERED_COLLECTIONS?: 'true' | 'false';
```
lerouxb, line 48: Feature-flag we will have to remember to remove at/after mongodb world!!!

Anemy: Do we have a ticket to remove it? Might help keep track of it so we can't easily forget.

Anemy (approved):
Left a couple comments/questions on small things we might want to update, looks good! Nice work. 👍 on the e2e test adds for the other collections

Screen Shot 2022-04-20 at 4 08 16 PM

packages/databases-collections/src/components/collection-fields/collection-fields.jsx
```diff
@@ -149,17 +166,35 @@ export default class CollectionFields extends PureComponent {
...
       )}
+      {hasClusteredCollectionSupport(serverVersion) && (
+        <ClusteredCollectionFields
+          isCapped={isCapped}
+          isTimeSeries={isTimeSeries}
+          isClustered={isClustered}
+          clusteredIndex={clusteredIndex}
+          expireAfterSeconds={expireAfterSeconds}
+          onChangeIsClustered={(newIsClustered) => this.setState(
+            { isClustered: newIsClustered },
```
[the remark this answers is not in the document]
lerouxb: Hmm good point. I just copied this blindly from timeseries code. Should probably clear it there as well.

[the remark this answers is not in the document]
lerouxb: That expireAfterSeconds property is weird because it is kinda a collection option, not necessarily timeseries or clustered collections specific. But it is only becoming an option in conjunction with those. I think eventually you'll be able to put it on any collection.

packages/compass-e2e-tests/helpers/compass.ts
Anemy: I'm thinking it would be better to not have feature flags enabled in end to end tests. That way we know what we're releasing is what we've tested. I wouldn't be opposed to running tests again with flags enabled if we want to double up some tests, but I think we run a risk by enabling feature flags in end to end tests.

lerouxb: I added these questions to the Technical Alignment Sync Notes doc for future discussion.

PR #2973, opened 2022-04-11
chore(compass-serverstats): Move `ErrorSummary` to compass-components, use in compass-serverstats COMPASS-5671 COMPASS-4148

Anemy:
Part of COMPASS-5671 - breaking into multiple prs to keep things easier to review.
Also fixes COMPASS-4148

This pr moves the `ErrorSummary` and `WarningSummary` that was used by the `connection-form` to `compass-components`. We then use the `ErrorSummary` from `compass-serverstats` to show the errors that can occur when showing the performance page. 

Two other changes:
- We now reset the stores of `compass-serverstats` on unmount. Previously when a user would encounter an error on the performance page, it would make the performance page unusable. When a user would disconnect and connect to an instance where they have all permissions for the performance page it would still be in an errored state and display no errors as well. This should fix that issue. I haven't found any side effects to this change yet, but it is one we should read through. I'll leave a comment where this is. Fixes COMPASS-4148
- Removed `rtss-databases.less` as it appeared to be unused.

Here's how it looks with three errors that users might run into:
Screen Shot 2022-04-11 at 3 01 23 PM


packages/compass-serverstats/src/components/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+ * Renders the entire performance tab, including charts and lists.
+ */
+function PerformanceComponent() {
+  const eventDispatcher = useRef(realTimeDispatcher());
+
+  useEffect(() => {
+    return () => {
+      // Reset the store on unmount so that when this is remounted
+      // the errors and previous state are reset.
+      Actions.restart();
```
Anemy, line 53: This is a change in behavior from before. This makes it so we reset the stores of the `compass-serverstats` plugin when we unmount. This should fix both the issue where the performance page would not display on disconnect and connect, and the issue where the errors would not display if a user leaves the performance tab and goes back to it.  COMPASS-4148 

Places that listen to this action:
https://github.com/mongodb-js/compass/search?q=Actions.restart

packages/compass-serverstats/src/components/dberror-component.tsx
```diff
@@ -0,0 +1,57 @@
...
+import type { Store } from 'reflux';
+import { ErrorSummary, css, spacing } from '@mongodb-js/compass-components';
+
+const errorContainerStyles = css({
+  padding: spacing[2],
+  position: 'relative'
+});
+
+// const debug = require('debug')('mongodb-compass:server-stats:dberror-component');
+
```
mcasimir, line 11:
```suggestion
```

packages/compass-serverstats/src/components/index.tsx
mcasimir: Do we need some manual cleanup for whatever state was there before restarting too, or that's enough?

packages/connection-form/src/components/connect-form-actions.tsx
```diff
@@ -41,8 +42,10 @@ function ConnectFormActions({
 }): React.ReactElement {
   return (
     <div className={formActionStyles}>
-      {warnings.length ? <WarningSummary warnings={warnings} /> : ''}
-      {errors.length ? <ErrorSummary errors={errors} /> : ''}
+      {warnings && <WarningSummary warnings={warnings} />}
```
mcasimir, line 45: I'm not sure checking for `warnings` here makes sense, looking at the code before it seems that `warnings` is always set (`warnings.length` would have thrown otherwise), should we keep `warnings.length` as a guard? otherwise we can remove the guard entirely since you have one inside `WarningSummary`

mcasimir (approved):

packages/compass-serverstats/src/components/index.tsx
Anemy: As part of the startup, the entire store runs this restart command. The state there is already manually cleaned up.
The places that are listening to this `restart` event currently already perform the cleanup. Those places can be found here: https://github.com/mongodb-js/compass/search?q=Actions.restart

PR #2976, opened 2022-04-12
chore(compass-e2e-tests): make tests more idempotent

addaleax:
Improve resiliency and DX by making tests more self-contained
by inserting test data in a mocha hook fresh for each test,
rather than using a separate script.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-e2e-tests/helpers/commands/navigate-to-collection-tab.ts
```diff
@@ -29,7 +29,33 @@ async function navigateToCollection(
...
   const collectionElement = await browser.$(collectionSelector);
-  await collectionElement.waitForDisplayed();
+
+  // If we're not finding the collection in the sidebar immediately,
+  // click the refresh button on the sidebar. This mostly handles the case
+  // in which we've connected Compass in a before() hook, but the
+  // insert-test-data hooks in the beforeEach() hook have not run yet,
+  // but also makes the tests more resilient in general.
+  let foundCollection = false;
+  const refreshSidebarPromise = async () => {
```
lerouxb, line 39: I agree that the insert script that ran before all the tests wasn't going to scale (it just grew out of a quick script that inserted a few numbers after all) and it makes sense to have test files bring their own data.

But I don't really like this kind of workaround in a command. I'd rather fix the test file or the hooks if hooks run after we cal this command. This can hide flaky behaviour, but it is also totally normal for a collection to not appear in the sidebar in 1 second (think datalake or anything on atlas).

ie. the difference between working around our own tests vs working around a browser or webdriver issue.

packages/compass-e2e-tests/helpers/insert-data.ts
```diff
@@ -0,0 +1,71 @@
...
+  // (named loosely after the relevant test)
+  promises.push(createBlankCollection(db, 'json-array'));
+  promises.push(createBlankCollection(db, 'json-file'));
+  promises.push(createBlankCollection(db, 'extended-json-file'));
+  promises.push(createBlankCollection(db, 'csv-file'));
+  promises.push(createBlankCollection(db, 'bom-csv-file'));
+  promises.push(createBlankCollection(db, 'numbers'));
+
+  // lots of collections to test virtual scrolling
+  for (let i = 0; i < 26; ++i) {
```
lerouxb, line 57: If tests are going to bring their own fixtures then they should probably only install what they want? Creating 26 empty collections is surprisingly slow and only some sidebar and collection screen tests care about that so that things reliably have to paginate.

lerouxb (approved):
Looks good!

PR #2980, opened 2022-04-12
feat(compass-sidebar): allow toggling CSFLE mode COMPASS-5717

addaleax:
This is a small UX improvement for:
- Cases in which some features are not available when using
  CSFLE, e.g. views are currently unsupported
- Verifying that CSFLE is working as it should, i.e.
  encrypted data is fully inaccessible when disabled

out

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-sidebar/src/components/csfle-connection-modal/csfle-connection-modal.jsx
```diff
@@ -0,0 +1,64 @@
...
+          connection must be fully closed.
+      </Description>
+    </div>
+    <Banner>
+        Client-side Field-Level Encryption is an Enterprise/Atlas-only feature
+        of MongoDB.&nbsp;
+      <Link href="https://www.mongodb.com/docs/drivers/security/client-side-field-level-encryption-guide/">
+          Learn More
+      </Link>
+    </Banner>
```
addaleax, line 52: This text is copied from the connection form. Is this the kind of thing that should go into compass-components, or not so much?

packages/compass-sidebar/src/stores/store.js
```diff
@@ -63,6 +63,11 @@ store.onActivated = (appRegistry) => {
 
   appRegistry.on('data-service-connected', (_, dataService, connectionInfo) => {
     store.dispatch(changeConnectionInfo(connectionInfo));
+
+    appRegistry.on('toggle-csfle-enabled', (enabled) => {
+      dataService.setCSFLEEnabled(enabled);
+      appRegistry.emit('refresh-data');
+    });
```
addaleax, line 70: Probably commit a state management sin here, but it’s simple and works?

packages/compass-sidebar/src/components/csfle-marker/csfle-marker.jsx
```diff
@@ -8,23 +8,38 @@ const badgeContainerStyles = css({
...
+  if (!csfleMode || csfleMode === 'unavailable') {
     return null;
   }
   return (
     <div className={badgeContainerStyles}>
-      <Badge variant={BadgeVariant.DarkGray}>
-        <Icon glyph="Key" />
-        CSFLE
-      </Badge>
+      <button className={badgeButtonStyles} onClick={() => toggleCSFLEModalVisible()}>
```
Anemy, line 29: I think for this button we probably want to `useFocusRing` and include an `aria-label` to improve accessibility and make the interactions consistent with some of the other newer buttons in Compass.
Here's an example of a place we're doing this already:
https://github.com/mongodb-js/compass/blob/47e5bfae7ee7b3424901c01bf66da21a0a9096e4/packages/compass-sidebar/src/components/sidebar/navigation-items.jsx#L63

packages/compass-sidebar/src/components/csfle-marker/csfle-marker.jsx
```diff
@@ -8,23 +8,38 @@ const badgeContainerStyles = css({
   padding: '0px 0px 6px 36px'
 });
 
-function CSFLEMarker({ isCSFLEConnection }) {
-  if (!isCSFLEConnection) {
+const badgeButtonStyles = css({
+  '&': {
```
Anemy, line 12: What is the `&` here used for? Does it work without it?

packages/compass-sidebar/src/components/csfle-connection-modal/csfle-connection-modal.jsx
Anemy: I think it's probably better outside of Compass components since it's not that generic. If we were using it in more places I think we could maybe add it or pull it from a separate package. I think it's fine as is.

packages/compass-sidebar/src/components/csfle-connection-modal/csfle-connection-modal.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+function CSFLEConnectionModal({ csfleMode, open, setOpen, setConnectionIsCSFLEEnabled }) {
+  return (<Modal
+    open={open}
+    title="Client-Side Field-Level Encryption"
+    trackingId="csfle_connection_modal"
+    setOpen={setOpen}
+  >
+    <div>
+      <p>
```
Anemy, line 20: Would the `Body` or `Subtitle` from leafygreen or one of the other `Typography` elements be better for this?
https://www.mongodb.design/component/typography/example/
```suggestion
      <Body>
```

packages/data-service/src/data-service.spec.ts
```diff
@@ -1646,5 +1646,21 @@ describe('DataService', function () {
...
+      dataService._crudClient = a;
+      dataService._metadataClient = b;
+
+      expect(dataService._initializedClient('CRUD')).to.equal(a);
+      expect(dataService._initializedClient('META')).to.equal(b);
+
+      dataService.setCSFLEEnabled(false);
+
+      expect(dataService._initializedClient('CRUD')).to.equal(b);
+      expect(dataService._initializedClient('META')).to.equal(b);
```
Anemy, line 1663: Maybe a bit unnecessary, but if it's easy to add maybe we can add another check to ensure re-enabling it works as expected:

```suggestion
      expect(dataService._initializedClient('META')).to.equal(b);
      
      dataService.setCSFLEEnabled(true);
      
      expect(dataService._initializedClient('CRUD')).to.equal(a);
      expect(dataService._initializedClient('META')).to.equal(b);
```

packages/compass-sidebar/src/components/sidebar/sidebar.jsx
```diff
@@ -232,7 +237,7 @@ const mapStateToProps = (state) => ({
   isDataLake: state.isDataLake,
   isGenuineMongoDB: state.isGenuineMongoDB,
   isGenuineMongoDBVisible: state.isGenuineMongoDBVisible,
-  isCSFLEConnection: state.isCSFLEConnection,
+  csfleMode: state.csfleMode,
```
Anemy, line 240: Unused prop? It looks like we pull `csfleMode` from `instance` in `sidebar-instance.jsx`

packages/compass-sidebar/src/components/csfle-connection-modal/csfle-connection-modal.jsx
```diff
@@ -0,0 +1,64 @@
...
+      >
+          Enable CSFLE for this connection
+      </Label>
+      <Description>
+          Disabling CSFLE only affects how Compass accesses data.
+          In order to make Compass forget KMS credentials, the
+          connection must be fully closed.
+      </Description>
+    </div>
+    <Banner>
```
Anemy, line 46: How do you feel about `marginTop: spacing[3]` on this banner? I think it will give a bit more space to the toggle then.

packages/compass-sidebar/src/components/csfle-connection-modal/csfle-connection-modal.jsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  css, spacing, Banner, Description, Label, Link, Toggle, Modal
+} from '@mongodb-js/compass-components';
+
+const toggleStyles = css({
+  marginBottom: spacing[3],
```
Anemy, line 8: How do you feel about a bit more spacing between the top title and this toggle?
```suggestion
  marginTop: spacing[3],
  marginBottom: spacing[3],
```

packages/compass-sidebar/src/components/csfle-connection-modal/csfle-connection-modal.jsx
```diff
@@ -0,0 +1,64 @@
...
+    open={open}
+    title="Client-Side Field-Level Encryption"
+    trackingId="csfle_connection_modal"
+    setOpen={setOpen}
+  >
+    <div>
+      <p>
+          This connection is configured with Client-Side Field-Level Encryption enabled.
+      </p>
+      <Toggle
```
Anemy, line 23: To get the toggle label and the toggle vertically aligned I'm thinking it might be help to wrap the `Toggle` and `Label` in a `div` with some styles to align them. How do you feel about:
```ts
const toggleContainerStyles = css({
  display: 'flex',
  alignItems: 'center'
});
```

```suggestion
      <div className={toggleContainerStyles}>
        <Toggle
```

It would look something like
Screen Shot 2022-04-12 at 1 49 35 PM

Anemy (commented):
Nice ux feature. Could comments, mostly on styling suggestions. Is this something that will also live in the settings (which we don't have yet) down the line?

packages/compass-sidebar/src/components/csfle-marker/csfle-marker.jsx
addaleax: Sure, done!

packages/compass-sidebar/src/components/csfle-connection-modal/csfle-connection-modal.jsx
addaleax: Sure – `Body` seems good here!

packages/data-service/src/data-service.spec.ts
addaleax: Yeah, makes absolute sense – done!

packages/compass-sidebar/src/components/sidebar/sidebar.jsx
addaleax: Yup, thanks for catching 👍

packages/compass-sidebar/src/components/csfle-connection-modal/csfle-connection-modal.jsx
addaleax: Done 👍

packages/compass-sidebar/src/components/csfle-connection-modal/csfle-connection-modal.jsx
addaleax: Done 👍

packages/compass-sidebar/src/components/csfle-connection-modal/csfle-connection-modal.jsx
addaleax: Perfect, thank you!

addaleax:
> Is this something that will also live in the settings (which we don't have yet) down the line?

@Anemy Sorry, forgot to reply to this part – I would assume that settings are for options that are persisted across Compass sessions, right? If so, no, this wouldn’t make sense… if somebody doesn’t want CSFLE enabled at all, they are probably not going to configure it in the connection options to begin with :slightly_smiling_face:

packages/instance-model/lib/model.js
```diff
@@ -254,6 +254,7 @@ const InstanceModel = AmpersandModel.extend(
 
         this.set({ refreshingStatus: 'ready', refreshingStatusError: null });
       } catch (err) {
+        console.log("refreshingStatusError!", err.stack)
```
Anemy, line 257: Do we want to `log` these errors?

Anemy (approved):
lgtm - left one comment on a `console.log` that made it in that I'm not sure is intended?

packages/instance-model/lib/model.js
addaleax: Yeah, this is definitely a debugging leftover :) I think we don’t quite need to do that here because we already log failed driver commands in general, and errors here will typically fall into that category, but we can look into adding it if this does turn out to be a problem

PR #2991, opened 2022-04-13
feat(export-to-language): Add Go Support

prestonvasquez:
## Description

[GODRIVER-2268](https://jira.mongodb.org/browse/GODRIVER-2268)

This adds the Go language as an output for the language transpiler, enabling exports for find and aggregation pipelines.

For the dbx-go team, to test on compass locally you can read the instructions for installing [here](https://github.com/mongodb-js/compass/blob/main/CONTRIBUTING.md#working-with-the-monorepo):

1. Install Node ^14.17.5 `brew install node@^14.17.5`
2. Install npm 7 `npm install -g npm7`
3. Clone/Fork the compas repository,
4. check out this branch `git checkout prestonvasquez:GODRIVER-2268.compassGoExport`
5. `npm run bootstrap`
6. `npm run start`

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions

There are a few open questions from the dbx-go team:

**Q1. Is it possible to do the string parsing within the transpiler code?**. This is in reference to how data is transpiled when trying to convert a string to a long type.  In its current state the code will generate a function to handle this:

Screen Shot 2022-04-13 at 11 41 22 AM

But perhaps it would be better to just cast the literal directly:

```
bson.D{{ "x": int64("1") }}
```

Two considerations for casting directly:
1. This code will fail at compile time 
7. Including the function adds consistency across other drivers as well as other symbols transpilers within the go driver.

**Q2. Are there any cases where the query might contain NumberLong(variable)?** This is a follow-up to Q1.

**Q3. Also, is it possible to just print an error message, like "unsupported conversion"?** This is also a follow-up to Q1.  If we are not wanting to include the function in the raw output, should we just not support any raw output that require declarations?

**Q4. If we decide to use declarations, do we want them to be functions or variables?**

## Dependents

This ticket depends on https://github.com/mongodb-js/compass/pull/2964 , which adds the capability of prepending statements with declarations.

## Types of changes

- [ ] *Backport Needed*


prestonvasquez:
Can we add @matthewdale and @benjirewis as reviewers alongside anyone reviewing from the compass team?

packages/bson-transpilers/symbols/go/templates.yaml
```diff
@@ -0,0 +1,924 @@
...
+        () => ''
+    Int32SymbolArgsTemplate: &Int32SymbolArgsTemplate !!js/function >
+        (lhs, arg, type) => {
+            if (!arg)
+                arg = 0
+            switch(type) {
+                case '_string':
+                    this.declarations.addFunc([]
+                        .concat('var parseInt32 = func(str string) int32 {')
+                        .concat('i64, err := sstrconv.ParseInt(str, 10, 32)')
```
benjirewis, line 369:
```suggestion
                        .concat('i64, err := strconv.ParseInt(str, 10, 32)')
```

packages/bson-transpilers/symbols/go/templates.yaml
```diff
@@ -0,0 +1,924 @@
...
+        () => ''
+    TimestampGreaterThanOrEqualArgsTemplate: &TimestampGreaterThanOrEqualArgsTemplate !!js/function >
+        (arg1, arg2) => {
+            this.declarations.addFunc([]
+                .concat('var timestampGEQ = func(tm1, tm2) bool {')
+                .concat('   c := primitive.CompareTimestamp(tm1, tm2)')
+                .concat('   return c == 1 || c == 0')
+                .concat('}')
+                .join('\n'));
+            return `timestampGEQ(${arg1}, ${arg2})`;
```
benjirewis, line 705: [opt] You could probably one-line this:
```
(lhs, rhs) => `primitive.CompareTimestamp(${lhs}, ${rhs}) >= 0`
```

packages/bson-transpilers/symbols/go/templates.yaml
```diff
@@ -0,0 +1,924 @@
...
+        () => ''
+    TimestampLessThanOrEqualArgsTemplate: &TimestampLessThanOrEqualArgsTemplate !!js/function >
+        (arg1, arg2) => {
+            this.declarations.addFunc([]
+                .concat('var timestampLEQ = func(tm1, tm2) bool {')
+                .concat('   c := primitive.CompareTimestamp(tm1, tm2)')
+                .concat('   return c == -1 || c == 0')
+                .concat('}')
+                .join('\n'));
+            return `timestampLEQ(${arg1}, ${arg2})`;
```
benjirewis, line 721: [opt] Same comment as above:
```
(lhs, rhs) => `primitive.CompareTimestamp(${lhs}, ${rhs}) <= 0`
```

packages/bson-transpilers/symbols/go/templates.yaml
```diff
@@ -0,0 +1,924 @@
...
+    200ImportTemplate: &200ImportTemplate !!js/function >
+        (args) => ['time.Time']
+    201ImportTemplate: &201ImportTemplate null
+    300ImportTemplate: &300ImportTemplate null
+    301ImportTemplate: &301ImportTemplate null
+    302ImportTemplate: &302ImportTemplate null
+    303ImportTemplate: &303ImportTemplate null
+    304ImportTemplate: &304ImportTemplate null
+    305ImportTemplate: &305ImportTemplate null
+    306ImportTemplate: &306ImportTemplate null
```
benjirewis, line 924:
```suggestion
    306ImportTemplate: &306ImportTemplate null
    
```
Unsure if needed.

packages/bson-transpilers/test/yaml/bson-methods.yaml
```diff
@@ -51,10 +53,8 @@ tests:
             shell: new Code('test code').code
             object: "'test code'"
             ruby: BSON::Code.new('test code').javascript
-            # Rust doesn't have a code attribute
```
benjirewis, line 54: Probably don't want to lose this comment.

packages/bson-transpilers/test/yaml/bson-methods.yaml
```diff
@@ -557,6 +667,13 @@ tests:
             shell: new Timestamp(1, 100) >= new Timestamp(2, 99)
             object: 'true'
             ruby: BSON::Timestamp.new(1, 100) >= BSON::Timestamp.new(2, 99)
+            go: |-
+                var timestampGEQ = func(tm1, tm2) bool {
+                   c := primitive.CompareTimestamp(tm1, tm2)
+                   return c == 1 || c == 0
+                }
+
+                timestampGEQ(primitive.Timestamp{T: 1, I: 100}, primitive.Timestamp{T: 2, I: 99})
```
benjirewis, line 676: _Could_ one-line this and L704 as I pointed out above.

packages/bson-transpilers/test/yaml/bson.yaml
```diff
@@ -254,6 +284,16 @@ tests:
           csharp: Convert.ToInt32("3")
           object: new Int32('3')
           ruby: "'3'.to_i"
+          go: |-
+            var parseInt32 = func(str string) int32 {
+            i64, err := sstrconv.ParseInt(str, 10, 32)
```
benjirewis, line 289:
```suggestion
            i64, err := strconv.ParseInt(str, 10, 32)
```
I think also missing an indent.

packages/bson-transpilers/test/yaml/driver-syntax.yaml
```diff
@@ -81,6 +81,30 @@ tests:
...
+
+                    // Set client options
+                    clientOptions := options.Client().ApplyURI("mongodb://localhost:27017")
+
+                    // Connect to MongoDB
+                    client, err := mongo.Connect(ctx, clientOptions)
+                    if err != nil {
+                      log.Fatal(err)
+                    }
+                    defer client.Disconnect(ctx)
```
benjirewis, line 97: Same comment about capturing error from `client.Disconnect`.

packages/bson-transpilers/test/yaml/driver-syntax.yaml
```diff
@@ -81,6 +81,30 @@ tests:
                       'x' => 0,
                       'y' => BSON::ObjectId.new
                     })
+                go: |-
+                    // Requires the MongodDB Go Driver
```
benjirewis, line 85:
```suggestion
                    // Requires the MongoDB Go Driver
```
Here and elsewhere.

benjirewis (commented):
Looking great! Some initial comments about the output Go.

packages/bson-transpilers/symbols/go/templates.yaml
prestonvasquez: Ah I love that!

packages/bson-transpilers/symbols/go/templates.yaml
```diff
@@ -0,0 +1,912 @@
...
+    NumericTypeTemplate: &NumericTypeTemplate null # args: literal, argType
+    ArrayTypeTemplate: &ArrayTypeTemplate !!js/function >
+        (literal, depth) => {
+            depth++;
+            const indent = '\n' + '    '.repeat(depth);
+            const closingIndent = '\n' + '    '.repeat(depth - 1);
+            if (literal === '')
+              return 'bson.A{}';
+            return `bson.A{${literal}${closingIndent}}`;
+        }
```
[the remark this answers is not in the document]
prestonvasquez: Absolutely, I think this makes sense.

packages/bson-transpilers/symbols/go/templates.yaml
```diff
@@ -0,0 +1,912 @@
...
+            const options = spec.options;
+            const filter = spec.filter || {};
+            delete spec.options;
+            delete spec.filter;
+
+            const indent = (depth) => '  '.repeat(depth)
+
+            const comment = []
+                .concat('// Requires the MongodDB Go Driver')
+                .concat('// https://github.com/mongodb/mongo-go-driver')
```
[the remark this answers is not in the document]
prestonvasquez: Ah I think the docs for sure!

packages/bson-transpilers/symbols/go/templates.yaml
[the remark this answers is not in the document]
prestonvasquez: It's almost certainly possible, but at the moment there is no wrapper specifically for aggregations, it just treats `[]` as whatever our transpiler defines for the `ArrayTypeTemplate`.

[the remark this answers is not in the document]
matthewdale: Sounds reasonable to defer to COMPASS-5742.

packages/bson-transpilers/symbols/go/templates.yaml
benjirewis: Bumping.

benjirewis (approved):
The Go code looks good to me.

In some examples you ignore the result of a `Find` or `Aggregate` operation (I'm assuming because it introduces an unused variable), but it should be pretty obvious to users that they should replace `_` with their own variable and continue logic with that variable.

packages/bson-transpilers/symbols/go/templates.yaml
```diff
@@ -0,0 +1,912 @@
...
+                        .concat('}')
+                        .join('\n'));
+                    return `parseInt64(${arg})`;
+                default:
+                    return `int64(${arg})`;
+            }
+        }
+    RegExpSymbolTemplate: &RegExpSymbolTemplate !!js/function >
+        () => 'regex'
+    RegExpSymbolArgsTemplate: &RegExpSymbolArgsTemplate null # Args: lhs, pattern, flags
```
matthewdale, line 409: I think there may be a bug handling the `RegExp` function. If you write a query like:
```
{x: RegExp('a', '')}
```

The generated Go code looks like:
```go
bson.D{{"x", RegExp}}
```
which doesn't compile and seems to be missing the arguments.

For comparison, the Python code looks like:
```python
{
    'x': re.compile(r"a")
}
```

packages/bson-transpilers/symbols/go/templates.yaml
```diff
@@ -0,0 +1,912 @@
...
+                return `${lhs}.Now()`;
+
+            const dateStr = [
+                date.getUTCFullYear(),
+                date.getUTCMonth() + 1,
+                date.getUTCDate(),
+                date.getUTCHours(),
+                date.getUTCMinutes(),
+                date.getUTCSeconds(),
+                'nil',
```
matthewdale, line 492: The `nsec` value must be an int, so I believe this should be `'0'` instead of `'nil'`.

E.g. filter:
```
{x: ISODate('2020-01-01')}
```
and generated Go code:
```go
bson.D{{"x", time.Date(2020, 1, 1, 0, 0, 0, nil, time.UTC)}}
```
Compilation error:
```
cannot use nil as int value in argument to time.Date
```

matthewdale (commented):
Found a few bugs, but everything else looks good!

packages/bson-transpilers/symbols/go/templates.yaml
```diff
@@ -0,0 +1,991 @@
...
+                if (singleQuoted || doubleQuoted)
+                    str = str.substr(1, str.length - 2);
+                return `"${str.replace(/\\([\s\S])|(")/g, '\\$1$2')}"`;
+            }
+
+            const structParts = [];
+            structParts.push(`Pattern: ${doubleStringify(pattern)}`);
+            if (flags)
+                structParts.push(`Options: ${doubleStringify(flags)}`);
+            return `primitive.Regex(${structParts.join(", ")})`;
```
matthewdale, line 217: Should use `{}` instead of `()` for declaring the `primitive.Regex`.

E.g.
```go
primitive.Regex{Pattern: "...", Options: "..."}
```

matthewdale (approved):
Looks awesome! 👍

packages/bson-transpilers/symbols/go/templates.yaml
```diff
@@ -0,0 +1,991 @@
...
+
+    #############################################
+    #                  Syntax                   #
+    #                                           #
+    # Templates for language syntax expressions #
+    #                                           #
+    #############################################
+
+    DriverTemplate: &DriverTemplate !!js/function >
+        (spec) => {
```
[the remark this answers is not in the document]
mcasimir: other languages don't seem to do that, cc: @mmarcon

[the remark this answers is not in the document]
mmarcon: @mcasimir and I chatted about this. The other languages don't do this but they probably should. A good UX compromise could be to wrap the code inside a function `RunAggregation` or something like that which then a user can call from their `main()`. Maybe let's do that as a follow-up PR.

PR #2996, opened 2022-04-18
chore(compass-indexes): Update compass indexes to shared webpack config COMPASS-5748

Anemy:
COMPASS-5748
This PR updates the `compass-indexes` package to use the [shared Compass configs](https://github.com/mongodb-js/compass/tree/main/configs) (webpack, eslint, etc). This is so that the upcoming 6.0 work and leafy green updates will be able to make use of typescript where changes are made.

The diff is large because of lint and prettier fixes. 
We are removing the electron running environment with these changes.

No visual changes in this pr. There are three places I had to change the logic a bit to fix lint issues so I'll leave a comment by each. In these places I tried to do the minimum changes as we're going to be addressing and improving all of them with the upcoming leafy green indexes work. I'm totally up for improving them more here if folks would prefer that.


packages/compass-indexes/src/components/create-index-modal/create-index-modal.jsx
```diff
@@ -329,54 +348,71 @@ class CreateIndexModal extends PureComponent {
...
+              >
                 Add another field
               </button>
             </div>
 
-            <div className={styles['create-index-modal-toggle-bar']}
-              onClick={this.handleToggleBarClick.bind(this)}>
+            <button
+              className={styles['create-index-modal-toggle-bar']}
+              onClick={this.handleToggleBarClick.bind(this)}
```
Anemy, line 377: We're changing this to a button as it was a `div` with an `onClick` before. Happy that lint picks this up now.
> Visible, non-interactive elements with click handlers must have at least one keyboard listener.eslint [jsx-a11y/click-events-have-key-events](https://github.com/evcohen/eslint-plugin-jsx-a11y/tree/master/docs/rules/click-events-have-key-events.md)

> Static HTML elements with event handlers require a role.eslint [jsx-a11y/no-static-element-interactions](https://github.com/evcohen/eslint-plugin-jsx-a11y/tree/master/docs/rules/no-static-element-interactions.md)

packages/compass-indexes/src/components/index-header-column/index-header-column.jsx
```diff
@@ -53,11 +52,15 @@ class IndexHeaderColumn extends PureComponent {
       <th
         data-test-id={this.props.dataTestId}
         className={this._renderClassName()}
-        onClick={this.handleIndexSort.bind(this)}>
+        onClick={this.handleIndexSort.bind(this)}
+      >
         {this.props.name}
         <i
-          className={classnames(styles['index-header-column-sort'], `fa fa-fw ${this.props.sortOrder}`)}
-          onClick={this.handleIndexSort.bind(this)}/>
```
Anemy, line 60: Removed this `onClick` as it was on a non-interactive element and the parent already has an `onClick`.

packages/compass-indexes/src/components/type-column/type-column.jsx
```diff
@@ -37,18 +42,27 @@ class TypeColumn extends PureComponent {
    */
   renderType() {
     let tooltipOptions = {};
-    if (this.props.index.type === 'text' || 'wildcard') {
+    if (
+      this.props.index.type === 'text' ||
+      this.props.index.type === 'wildcard'
```
Anemy, line 47: I think this was a bug before. Doesn't look like it impacted anything though.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.jsx
addaleax: :+1:

Do we want to make this an `import(compass-components).Accordion` at some point?

addaleax (approved):
Nice work, LGTM!

packages/compass-indexes/src/components/create-index-modal/create-index-modal.jsx
Anemy: Good suggestion, I think so yes, we have an upcoming epic for updating the indexes page to leafy green, so I was thinking to do it then:
https://jira.mongodb.org/browse/COMPASS-5485
I started to update to the Accordion it in this pr, but because a lot of these ui states are tied into the store it ended up changing a bit too much logic and I wanted to avoid too many changes for this pr.

PR #3011, opened 2022-04-21
feat(aggregation-count): count docs COMPASS-5756

mabaasit:
feat(aggregation-count): count docs COMPASS-5756

count aggregation documents

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/modules/count-documents.ts
```diff
@@ -0,0 +1,106 @@
...
+};
+
+export const INITIAL_STATE: State = {
+  loading: false,
+};
+
+const reducer: Reducer<State, Actions | WorkspaceActions> = (state = INITIAL_STATE, action) => {
+  switch (action.type) {
+    case WorkspaceActionTypes.WorkspaceChanged:
+      return INITIAL_STATE;
```
gribnoysup, line 47: As we are not cleaning up inflight counts when workspace type is changed (both for count and for results), you can run into a pretty nasty race condition where slower, older pipeline results and counts replace content of the newer one. Here's how it looks in UI:

Kapture 2022-04-21 at 18 16 46

We should definitely fix that both for counts and for results

packages/compass-aggregations/src/modules/workspace.ts
```diff
@@ -30,9 +33,17 @@ const reducer: Reducer<State, Actions | AggregationActions> = (state = INITIAL_S
...
-  view,
-});
-
+export const changeWorkspace = (view: Workspace): ThunkAction<void, RootState, void, Actions> => {
+  return (dispatch) => {
+    // As user switches to builder view, we cancel running ops
+    if (view === 'builder') {
+      dispatch(cancelAggregation());
+      dispatch(cancelCount());
+    }
```
gribnoysup, line 42: We're definitely getting there. I think you would also want to do this when Run button is pressed if there is an in-flight request running, otherwise I can overload server pretty easily (but more importantly if I'm running a log query and maybe change collation or maxtimems there is a chance of running into race condition issue):

Kapture 2022-04-25 at 17 50 21

packages/compass-import-export/src/modules/export.js
```diff
@@ -383,10 +383,10 @@ export const openExport = ({
...
     } = getState();
     try {
-      const count = isAggregation
-        ? null
-        : maybeCount ??
-          (await fetchDocumentCount(dataService, namespace, query));
+      const count =
+        !isAggregation && !maybeCount
+          ? await fetchDocumentCount(dataService, namespace, query)
+          : maybeCount;
```
gribnoysup, line 389: `!maybeCount` will be `true` in case when count is zero, nullish coalescing will handle this case. Maybe something like this to be a bit more explicit about our intended order (if count passed, use it, otherwise if not aggregation fetch results, otherwise `null`)?

```suggestion
      const count =
        maybeCount ??
        (!isAggregation
          ? await fetchDocumentCount(dataService, namespace, query)
          : null);
```

packages/compass-aggregations/src/modules/aggregation.ts
```diff
@@ -204,19 +236,33 @@ const fetchAggregationData = (page: number): ThunkAction<
...
       const abortController = new AbortController();
       const signal = abortController.signal;
       dispatch({
         type: ActionTypes.AggregationStarted,
         abortController,
+        previousPageData: {
+          page: _page,
+          isLast: _isLast,
+          documents: _documents,
+        },
```
gribnoysup, line 265: Actions should not be the ones deriving new state from current one, that's the responsibility of the reducer

gribnoysup (approved):
I think you should improve the aggregation state slice a bit by not deriving state in the action and doing it in a reducer instead. Otherwise looks good

PR #3024, opened 2022-04-27
feat(compass-indexes): Columnstore indexes support COMPASS-5665

Anemy:
COMPASS-5665

This PR adds support for creating and viewing `columnstore`/columnar indexes. When creating a `columnstore` index a user also has the ability to configure the `columnstoreProjection` options. We present this in the same way as we present the `wildcardProjection` option.

There's definitely a lot of room for code improvement (and ux improvement) in this package, but in this pr we are focusing on adding the 6.0.0 (or 6.1.0 ;) feature support. In the indexes leafygreen epic I'm hoping we can streamline a lot of this create index modal and the indexes table view.

Note: This feature is gated behind server version `6.1.0`. A build is not yet publicly available so we can't totally test this feature out yet. It's also not explicitly supported in the driver yet. 

There are restrictions and nuances around when a columnstore index can be used (for instance they won't be able to be used with clustered or time-series collections to start). In this PR we take the approach of sending it to the server and exposing the response for failures. We could try to do more in terms of restricting and hiding this feature in these cases but it might be more work that it's worth, and down the line these things will change with different server versions.

Server version less than 6.1.0 create index view (same as before):
Screen Shot 2022-04-26 at 8 32 03 PM

Server version greater than 6.1.0 create index view:
Screen Shot 2022-04-26 at 11 09 22 PM
Screen Shot 2022-04-26 at 11 09 32 PM

A `columnstore` index in the indexes table:
Screen Shot 2022-04-26 at 10 57 06 PM

We can hold for a bit more information on the landing times of this feature before looking to merge this pr. Things may be subject to change. That being said I think what's here can be looked at as it's likely in about the shape of how we'll support this index. In our technical designs for this feature we also use the wordage `columnar` instead of `columnstore`. I'm double checking on the wording we should expose in Compass now.


packages/compass-indexes/src/modules/create-index/index.js
```diff
@@ -199,6 +221,7 @@ export const createIndex = () => {
           background: state.isBackground,
           unique: state.isUnique,
           ttl: state.isTtl,
+          columnstore: state.isColumnstore,
```
Anemy, line 224: This fulfills the telemetry ticket https://jira.mongodb.org/browse/COMPASS-5766

packages/compass-e2e-tests/tests/collection-indexes-tab.test.ts
```diff
@@ -88,4 +95,73 @@ describe('Collection indexes tab', function () {
...
+      const textOption = await typeList.$('div=columnstore'); // div element with the text "columnstore"
+      textOption.waitForDisplayed();
+      textOption.click();
+
+      await browser.clickVisible(Selectors.CreateIndexConfirmButton);
+
+      await createModal.waitForDisplayed({ reverse: true });
+
+      const indexComponent = await browser.$(
+        Selectors.indexComponent('columnstore')
```
Anemy, line 147: @lerouxb Added a feature flag, `COMPASS_COLUMNSTORE_INDEXES`, to enable the columnstore index creation feature. Also added this e2e test. Looks like the default name for indexes created will be `columnstore` from https://github.com/mongodb/mongo/blob/master/src/mongo/db/index_names.cpp
When we remove the feature flag we'll know if the test is failing as well.

Once tests pass I'll merge tomorrow if that's cool.

packages/compass-e2e-tests/tests/collection-indexes-tab.test.ts
```diff
@@ -88,4 +95,73 @@ describe('Collection indexes tab', function () {
...
+
+      process.env.COMPASS_COLUMNSTORE_INDEXES = 'true';
+    });
+
+    after(function () {
+      process.env = initialEnvVars;
+    });
+
+    it('supports creating a columnstore index', async function () {
+      if (semver.lt(MONGODB_VERSION, '6.1.0')) {
```
lerouxb, line 113: we can maybe try to include alpha or rcs as well so we'll find out sooner. I think it should be possible to just add -alpha or something like here https://github.com/mongodb-js/compass/blob/da327e148b3c3d763af88eac7c4e385c67958779/packages/databases-collections/src/components/collection-fields/has-fle2-support.js#L3

packages/compass-e2e-tests/tests/collection-indexes-tab.test.ts
lerouxb: Sounds good!

lerouxb (approved):

packages/compass-e2e-tests/tests/collection-indexes-tab.test.ts
lerouxb: That will only help if we ever actually run these tests against alphas or rcs, but might also make it easier to run your test manually. by specifying MONGODB_VERSION so mongodb runner installs something that supports it.

Anemy, line 113:
```suggestion
      if (semver.lt(MONGODB_VERSION, '6.1.0-alpha0')) {
```

PR #3028, opened 2022-04-27
feat(telemetry): Track Database Created and Collection Created COMPASS-5765

lerouxb:
Since create database is basically create collection and we were missing that as well I just did both.


addaleax (approved):

packages/databases-collections/src/modules/create-collection/index.js
```diff
@@ -111,6 +113,16 @@ export const createCollection = (data) => {
           return stopWithError(dispatch, err);
         }
 
+        const trackEvent = {
+          is_capped: !!data.options.capped,
```
mcasimir, line 117: should we add some guard to check if `data.options` is there, or by this time will always be there?

lerouxb: I've checked all the usage and it seems to always be at least `{}`.

PR #3029, opened 2022-04-27
chore: group AceEditor usage in compass-components COMPASS-5687

addaleax:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-crud/src/components/insert-json-document.jsx
```diff
@@ -63,15 +42,15 @@ class InsertJsonDocument extends Component {
...
-          defaultValue={EDITOR_COMMENT}
-          value={value}
-          onChange={this.onChange.bind(this)}
-          theme="mongodb"
-          width="100%"
-          editorProps={{$blockScrolling: Infinity}}
-          setOptions={OPTIONS}
+        <Editor
+          variant={EditorVariant.EJSON}
+          placeholder={EDITOR_COMMENT}
```
Anemy, line 47: Do we want the `placeholder` to stay? Might be a nice feature?

```suggestion
          defaultValue={EDITOR_COMMENT}
```
or

```suggestion
          defaultValue={EDITOR_COMMENT}
          placeholder={EDITOR_COMMENT}
```

Screen Shot 2022-04-27 at 4 28 25 PM
hmm getting some lag and no highlighting here. Restarting my instance, but may be a different mode or something, I'll remove this when it finishes building if it was a dep install issue.

packages/compass-components/src/components/editor.tsx
```diff
@@ -0,0 +1,104 @@
...
+          ? 'json'
+          : 'javascript' // set to 'mongodb' as part of setOptions
+      }
+      theme="mongodb"
+      width="100%"
+      value={text}
+      onChange={onChangeText}
+      editorProps={{ $blockScrolling: Infinity }}
+      setOptions={setOptions}
+      defaultValue={placeholder}
```
Anemy, line 91: Screen Shot 2022-04-27 at 3 54 18 PM
Do we want to be setting the default value to the placeholder? Is it a change of behavior?

```suggestion
      placeholder={placeholder}
      defaultValue={defaultValue}
```

Anemy (commented):
Nice - this is real good for how we can customize and improve the editor.

Left 2 comments on param changes, still trying some out to double check behavior.

Should we add a test for the editor? `editor.spec.tsx`
Not sure if it's easy to test, if it's another ticket no worries we can add later. I can add a commit adding it tmrw or Monday (I'm out friday).

packages/compass-export-to-language/src/components/editor/editor.jsx
```diff
@@ -34,34 +24,24 @@ class Editor extends PureComponent {
...
+        <Editor
+          variant={EditorVariant.Generic}
           mode={this.props.language}
-          defaultValue=""
-          value={this.props.value}
-          theme="mongodb"
-          width="100%"
-          editorProps={{$blockScrolling: Infinity}}
-          setOptions={OPTIONS}
+          text={this.props.value}
```
Anemy, line 34: Since we're removing the `defaultValue` here, should we default to passing `defaultValue=""}` in the Editor in `editor.tsx`?

packages/compass-components/src/components/editor.tsx
```diff
@@ -0,0 +1,104 @@
...
+      mode={
+        variant === 'Generic'
+          ? undefined
+          : variant === 'EJSON'
+          ? 'json'
+          : 'javascript' // set to 'mongodb' as part of setOptions
+      }
+      theme="mongodb"
+      width="100%"
+      value={text}
```
Anemy, line 87: Definitely a small thing, no change needed, Is there a benefit to using the wording `text` here instead of value? Is it for readonly cases? I'm wondering if using `value` everywhere might be a bit more intuitive.

packages/compass-components/src/components/editor.tsx
```diff
@@ -0,0 +1,104 @@
...
+  enableLiveAutocompletion: false,
+  tabSize: 2,
+  fontSize: 11,
+  minLines: 10,
+  maxLines: Infinity,
+  showGutter: true,
+  useWorker: false,
+};
+
+// Currently, only EJSON
```
Anemy, line 40: Maybe a half started or left over comment?

packages/compass-crud/src/components/json-editor.jsx
```diff
@@ -244,36 +238,26 @@ class EditableJson extends React.Component {
...
       showLineNumbers: this.state.editing,
-      vScrollBarAlwaysVisible: false,
-      hScrollBarAlwaysVisible: false,
       fixedWidthGutter: false,
       showPrintMargin: false,
       displayIndentGuides: false,
       wrapBehavioursEnabled: true,
-      foldStyle: 'markbegin',
-      useWorker: false
+      foldStyle: 'markbegin'
```
Anemy, line 250: I'm not seeing this fold style show up, maybe we don't pass it down? Can look more tmrw/Monday also.
(top is this branch)
Screen Shot 2022-04-28 at 3 24 25 AM

packages/compass-components/src/components/editor.tsx
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+
+if (typeof window === 'undefined' && typeof globalThis !== 'undefined') {
+  // ace-builds wants to install itself on `window`, which
+  // is not available when this package is loaded through
+  // (non-Electron) Node.js. That's an atypical case, but it's
+  // easier to account for it here than to handle all cases
+  // in which this package is loaded from Node.js.
```
gribnoysup, line 8: What are the cases where we load this in Node.js environment with no window available? In mocha test envs it should be provided by jsdom, in electron and browser it should exist 🤔

packages/compass-crud/src/components/insert-json-document.jsx
addaleax: Ah no, I actually got mixed up a bit here. I thought `defaultValue` was the ace version of `placeholder`, because I wasn’t sure what `defaultValue` would mean otherwise (since we’re always passing `value` as well). I’m still not sure but it’s probably better if I don’t touch this for now. :slightly_smiling_face: 

The missing JSON highlighting (and the `foldStyle` issue below) was just a missing import, my bad :)

packages/compass-components/src/components/editor.tsx
addaleax: Yeah – I’d like it if we could _eventually_ merge the parsing parts as well, and then have e.g. an `onChangeValue(value: any, error: Error | unknown, text: string) => void` callback as well.

Just because it’s not great that we use at least three different parsing backends (acorn, mongodb-query-parser, EJSON, …?) that all accept different input without the user being made aware of which backend they are writing their code for right now, and encapsulating this logic inside the editor seems like it could be a good way to solve this. Felt a bit out of scope, though :slightly_smiling_face:

packages/compass-components/src/components/editor.tsx
addaleax: @gribnoysup Was running into this with gen-esm-wrapper in databases-collections and was not a fan of trying to remove it there for this, but let me know if you prefer that :sweat_smile:

gribnoysup: Gotcha! I think it's fine to keep then, wasn't obvious where it can happen and so I thought that maybe this was actually highlighting a real issue (like running tests in wrong env or something)

packages/compass-aggregations/src/components/stage-editor/stage-editor.jsx
```diff
@@ -1,32 +1,13 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import 'ace-builds';
-import AceEditor from 'react-ace';
+import { Editor, EditorVariant,EditorTextCompleter } from '@mongodb-js/compass-components';
```
mcasimir, line 3:
```suggestion
import { Editor, EditorVariant, EditorTextCompleter } from '@mongodb-js/compass-components';
```

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/encrypted-field-config-input.tsx
```diff
@@ -55,15 +36,12 @@ function EncryptedFieldConfigInput({
...
-          mode="javascript" // will be set to mongodb as part of OPTIONS
-          theme="mongodb"
-          width="100%"
-          value={encryptedFieldConfigToText(encryptedFieldConfig)}
-          onChange={(newText) => onChange(textToEncryptedFieldConfig(newText))}
-          editorProps={{ $blockScrolling: Infinity }}
-          name="import-pipeline-editor"
-          setOptions={OPTIONS}
+        <Editor
+          variant="Shell"
```
mcasimir, line 40: don't we use the enum here?

mcasimir: Line 40 (sorry :))

mcasimir (commented):
🔥

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/encrypted-field-config-input.tsx
addaleax: the enum is just so we don’t do typos in non-TS files, here TS already takes care of that for us

packages/compass-crud/src/components/insert-json-document.jsx
gribnoysup: If we are always passing value, we probably can remove the defaultValue prop, yeah. Although I'm wondering if there might be cases where having only a defaultValue passed can help with ace editor performace, maybe opting it out of the controlled component behavior can help with the lag we are noticing and in some cases we probably don't need it to be synced back and forth all the time

packages/compass-components/src/components/editor.tsx
Anemy: Sounds good, definitely out of scope.

Anemy (approved):
lgtm . nice work

PR #3030, opened 2022-04-28
chore: make CI bash scripts fail on error

mcasimir:
- Add `set -e` to bash scripts on EVG so they exit on failures
- Add a `retry-with-backoff.sh` script used to invoke curl to mitigate CI flakes due to failures during downloads of `node.js`

### DIFF
--- a/.evergreen/node-gyp-bug-workaround.sh
+++ b/.evergreen/node-gyp-bug-workaround.sh
@@ -32,17 +32,20 @@ if [ x"$LOCALAPPDATA" = x"" ]; then
   exit
 fi
 
-set -x
+set -ex
+
+SCRIPTDIR="$(cd $(dirname "$0"); pwd)"
 CACHEDIR="$LOCALAPPDATA/node-gyp/Cache"
 rm -rvf "$CACHEDIR"
 mkdir -p "$CACHEDIR/$NODE_JS_VERSION"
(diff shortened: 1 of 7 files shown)


.evergreen/node-gyp-bug-workaround.sh
```diff
@@ -32,17 +32,20 @@ if [ x"$LOCALAPPDATA" = x"" ]; then
...
+set -ex
+
+SCRIPTDIR="$(cd $(dirname "$0"); pwd)"
 CACHEDIR="$LOCALAPPDATA/node-gyp/Cache"
 rm -rvf "$CACHEDIR"
 mkdir -p "$CACHEDIR/$NODE_JS_VERSION"
 cd "$CACHEDIR/$NODE_JS_VERSION"
-curl -sSfLO "https://nodejs.org/download/release/v$NODE_JS_VERSION/node-v$NODE_JS_VERSION-headers.tar.gz"
+
+bash "${SCRIPTDIR}/retry-with-backoff.sh" curl -sSfLO "https://nodejs.org/download/release/v$NODE_JS_VERSION/node-v$NODE_JS_VERSION-headers.tar.gz"
```
mcasimir, line 43: `curl` actually already has a `retry` feature but it seems reliable only in more recent versions that I'm not sure are available everywhere in EVG, just opted for a separate script

addaleax (approved):

PR #3034, opened 2022-04-28
feat(databases-collections): Add advanced collection options accordion for showing collection options

Anemy:
This pr adds an `Advanced Collection Options` dropdown to the create database and create collection modals.

Now that we're adding more collection configuration options, we should ensure the collection creation experience is streamlined for all users. If most users aren't using these options, I think it makes sense to hide them behind a dropdown so that we don't confuse or mislead users.

I'll check with design if this is something we can add (might have been a bit too early to open this pr oops).

https://user-images.githubusercontent.com/1791149/165797132-2b347ce5-821b-4839-b964-3ffd4a7a6ac5.mp4


addaleax (approved):
Looks good! :slightly_smiling_face:

PR #3044, opened 2022-05-03
feat(explain-aggregations): aggregations in mongodb-explain-compat COMPASS-5785

mabaasit:
feat(explain-aggregations): add aggregation explain support in mongodb-explain-compat

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/mongodb-explain-compat/lib/index.js
```diff
@@ -72,18 +72,56 @@ function mapStages(queryPlan, sbeExecutionStages) {
...
+}
+
+/**
+ * https://wiki.corp.mongodb.com/display/QUERY/Explain+Notes
+ */
+function mapAggregationExplain(explain) {
+  // In aggregations for SBE, the cursor based stage is always first.
+  const firstStage = explain.stages[0]; // {$cursor: {queryPlanner: {}, executionStats: {} }}
+  const firstStageKey = Object.keys(firstStage)[0]; // $cursor
+  if (!firstStageKey.match(/cursor/)) {
```
addaleax, line 86: Just curious, is there a specific reason for the [0] and the regular expression here? I don’t think the server would typically guarantee ordering of keys, so if they added a key before `cursor`, this would not match anymore.

packages/mongodb-explain-compat/lib/index.js
```diff
@@ -72,18 +72,56 @@ function mapStages(queryPlan, sbeExecutionStages) {
...
+  const winningPlan =
+    firstStage[firstStageKey].queryPlanner.winningPlan.queryPlan;
+  firstStage[firstStageKey].queryPlanner.winningPlan = winningPlan;
+
+  firstStage[firstStageKey].executionStats.executionStages = mapStages(
+    winningPlan,
+    firstStage[firstStageKey].executionStats.executionStages
+  );
+  firstStage[firstStageKey].queryPlanner.plannerVersion = 1;
+  explain.stages[0] = firstStage;
```
addaleax, line 98: I think this line is a no-op, since `firstStage` is initialized to be `explain.stages[0]`.

packages/mongodb-explain-compat/lib/index.js
```diff
@@ -72,18 +72,56 @@ function mapStages(queryPlan, sbeExecutionStages) {
...
+function isAggregationExplain(explain) {
+  return !!explain.stages;
+}
+
+/**
+ * https://wiki.corp.mongodb.com/display/QUERY/Explain+Notes
+ */
+function mapAggregationExplain(explain) {
+  // In aggregations for SBE, the cursor based stage is always first.
+  const firstStage = explain.stages[0]; // {$cursor: {queryPlanner: {}, executionStats: {} }}
```
addaleax, line 84: What happens to the other stages?

mabaasit: so while executing aggregation, the server will get the necessary data from the collection first and then do the required transformations that are in the pipeline. so, it creates a cursor stage that absorbs all the stages responsible for fetching data where we have all the necessary information for explain. And for rest of the stages, we don't have any `queryPlanner` or `executionStats` information.

And if the cursor absorbs all the stages, then the output of explain is same as find operation (also pipeline is run as find)

packages/mongodb-explain-compat/lib/index.js
mabaasit: awesome! fixed.

addaleax: I’m still not quite sure about the regular expression part – is there any other key than `$cursor` or maybe `cursor` without the `$` that we need to take care of? https://github.com/mongodb/mongo/blob/e9e4e99d0ff87ec9b6ed2a689a4c42cdba7f5527/src/mongo/shell/explainable.js#L607 has a literal check as well

packages/mongodb-explain-compat/lib/index.js
```diff
@@ -72,18 +72,67 @@ function mapStages(queryPlan, sbeExecutionStages) {
...
+  explain = JSON.parse(JSON.stringify(explain));
+  if ((explain.explainVersion || 0) < 2) {
+    return explain;
+  }
+  delete explain.explainVersion;
+  if (isAggregationExplain(explain)) {
+    explain = mapAggregationExplain(explain);
+  } else {
+    explain = mapQueryExplain(explain);
+  }
```
addaleax, line 136: In https://github.com/mongodb/mongo/blob/e9e4e99d0ff87ec9b6ed2a689a4c42cdba7f5527/src/mongo/shell/explainable.js the outermost check is a sharding check, not an agg vs query check – which makes sense to me, because shard handling is always the last thing that happens.

Does this handle sharding properly? Can we add a test for that case as well?

packages/mongodb-explain-compat/lib/index.js
mabaasit: Yeah! The first stage is a general cursor stage and it be $cursor, $geoNearCursor (i know so far these two only, but i assume there are more). 

I had a chat with the team in #aggregation-framework regarding this and accordingly updated the documentation (linked in comments of code).

addaleax: Yeah, I see – feels like the regexp could be tightened up a bit to `/^\$.*cursor/i` if the distinguishing feature of the key is that it starts with a `$`, but this makes sense :+1: Might be good to leave a comment about this?

packages/mongodb-explain-compat/lib/index.js
```diff
@@ -72,18 +72,143 @@ function mapStages(queryPlan, sbeExecutionStages) {
...
+      );
+
+      queryPlanner.winningPlan.shards[index].winningPlan = winningPlan;
+      executionStats.executionStages.shards[index].executionStages =
+        executionStages;
+    }
+  });
+  explain.queryPlanner = queryPlanner;
+  explain.executionStats = executionStats;
+  return JSON.parse(JSON.stringify(explain));
```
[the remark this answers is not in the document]
mabaasit: We already have that in place at main export. This, I think has been there to clear out all the information which is not needed like the undefined props, symbols which are added while mapping.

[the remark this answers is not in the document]
addaleax: Then maybe I’d suggest keeping it there for consistency/to avoid duplicating this bit every time, since it’s part of the external interface of this package and not part of each function’s individual functionality?

addaleax (approved):

PR #3045, opened 2022-05-03
chore: hide `__safeContent__` field COMPASS-5780

addaleax:
Grey out the field in the CRUD view, mark it non-editable, and
deselect it by default when exporting from a collection.

Screenshot from 2022-05-03 16-15-37
Screenshot from 2022-05-03 16-15-49
Screenshot from 2022-05-03 16-27-39

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-import-export/src/modules/load-fields.js
```diff
@@ -54,5 +54,8 @@ export function getSelectableFields(fields, { maxDepth } = {}) {
   const selectableFields = Object.keys(fields).map((field) =>
     truncateFieldToDepth(field, maxDepth)
   );
-  return Object.fromEntries(selectableFields.map((field) => [field, ENABLED]));
+  return Object.fromEntries(selectableFields.map((field) => {
+    const enabled = !/^__safeContent__\b/.test(field);
```
Anemy, line 58: I'm thinking this is worth a comment since the `__safeContent__` field might be a bit unexpected. Would it help to break this into a helper so we can test it independently?
The regex looks good, I think it makes me sleep good when they're tested though.

I think you mentioned that this internal field is useful in mongo dump/restore. Is that relevant to importing/exporting?

```suggestion
    // The `__safeContent__` field is used internally in MongoDB and can cause
    // issues when it is duplicated or changed by users.
    // We default to not export it unless the user explicitly enables it.
    const enabled = !/^__safeContent__\b/.test(field);
```

Anemy (commented):
Should we hide this field on the `schema` page as well?

packages/compass-import-export/src/modules/load-fields.js
mcasimir: I'm actually thinking if the idea of having that as a projection in data service is not better here, we can figure ways to share the logic with DE too, maybe we just need to keep that in mind?

addaleax: @Anemy Just fyi, I restructured this a bit so that the logic for determining this field handling is more bundled in hadron-document – I think that makes sense here, I also really hope that we don’t ever have a field like this introduced by the server, but who knows :woman_shrugging: 

@mcasimir I mentioned this on Slack already, but yeah, I think it’s too confusing to have fields that are displayed only under some circumstances, and filtering in the data service is also complicated because there are situations, like the export one, in which a user might have a good reason to want to include the field depending on the situation

Anemy (approved):

PR #3048, opened 2022-05-03
fix(compass-aggregations): remove ace editor snippet completion, complete operator snippets on our end COMPASS-5683

Anemy:
This PR removes our use of [ace editor's](https://cloud9-sdk.readme.io/docs/snippets) `insertSnippet` command. This command was causing any items which matched the token, which was by default anything starting with `$` to be removed when the stage operator was changed. This unintentional replacing was made more evident when we added the ability for the stage content not to change when the operator is changed.

To fix this we are now replacing the ace snippet completion by replacing the tokens explicitly (see the `replaceOperatorSnippetTokens` function). As a result of this approach we are able to remove the `fromStageOperators` flag, which, from what it looks like, was used to prevent an additional preview reload when the stage operator changed. We're also removing the `snippet` from the stage model as it is no longer needed.
We could alternatively try to update how ace does its snippet completion, but I think that approach would make us more tied to ace editor internals, and might require us to keep these extra things in the state which do not need to be.

before:

https://user-images.githubusercontent.com/1791149/166514717-f1a263a5-fcb3-4192-b6a6-a54ecad7d6e9.mp4

after:

https://user-images.githubusercontent.com/1791149/166514088-c633e580-eccf-4c52-9c5f-697e4da56fb8.mp4

### DIFF
--- a/packages/compass-aggregations/src/components/stage-editor/stage-editor.jsx
+++ b/packages/compass-aggregations/src/components/stage-editor/stage-editor.jsx
@@ -15,7 +15,6 @@ class StageEditor extends Component {
   static propTypes = {
     stage: PropTypes.string,
     stageOperator: PropTypes.string,
-    snippet: PropTypes.string,
     error: PropTypes.string,
     syntaxError: PropTypes.string,
     runStage: PropTypes.func.isRequired,
@@ -73,8 +72,6 @@ class StageEditor extends Component {
   }
 
   /**
-   * On update if the stage operator is changed insert the snippet and focus on the editor.
-   *
    * @param {Object} prevProps - The previous properties.
    */
   componentDidUpdate(prevProps) {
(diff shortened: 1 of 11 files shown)


addaleax (approved):

packages/compass-aggregations/src/components/stage-editor/stage-editor.jsx
```diff
@@ -87,9 +81,13 @@ class StageEditor extends Component {
     );
     this.completer.version = this.props.serverVersion;
     if (this.props.stageOperator !== prevProps.stageOperator && this.editor) {
-      this.editor.setValue('');
-      this.editor.insertSnippet(this.props.snippet || '');
```
Anemy, line 91: Here's where we used to perform the snippet completion. This would also call the `onStageChange` twice I believe, and by doing so cause the stage preview to be recalculated. These changes make it so we explicitly try to run now when the operator changes. This might be something we could flag in the store instead.

PR #3049, opened 2022-05-03
chore: run linting in compass package

addaleax:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass/package.json
+++ b/packages/compass/package.json
@@ -141,7 +141,7 @@
     "evergreen-expansions": "hadron-build info --format=yaml --flatten > expansions.yml",
     "build-info": "hadron-build info",
     "jsx": "babel ./src --out-dir ./lib",
-    "lint": "eslint",
+    "lint": "eslint .",
     "depcheck": "depcheck",
     "test-ci": "npm run test-release-tasks",
     "test-ci-electron": "npm run test-electron",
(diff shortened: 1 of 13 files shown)


packages/compass/package.json
```diff
@@ -141,7 +141,7 @@
     "evergreen-expansions": "hadron-build info --format=yaml --flatten > expansions.yml",
     "build-info": "hadron-build info",
     "jsx": "babel ./src --out-dir ./lib",
-    "lint": "eslint",
+    "lint": "eslint .",
```
addaleax, line 144: This was missing previously … :slightly_smiling_face:

packages/compass/src/main/logging.ts
```diff
@@ -194,7 +194,7 @@ async function showLogFileDialog(logFilePath: string) {
     case 3: {
       try {
         const tempFilePath = await extractPartialLogFile(logFilePath);
-        shell.openPath(tempFilePath);
+        await shell.openPath(tempFilePath);
```
addaleax, line 197: … and this is the only code change of substance :slightly_smiling_face:

Anemy (approved):
nice catch

addaleax:
> nice catch - makes me think there could be other places we're missing the `.` thinking

I think we usually do something like

```js
    "eslint": "eslint",
    "prettier": "prettier",
    "lint": "npm run eslint . && npm run prettier -- --check .",
```

where omitting the `.` is fine, I guess the compass package was just special this way? With a quick grep I couldn’t find other problems of this type, at least :slightly_smiling_face:

PR #3065, opened 2022-05-09
fix(connect-form): allow empty hosts

mcasimir:
Fixes:

> "Selecting the whole host and trying to delete it keeps the value in the form but also shows a bunch of errors"

image

With the loose validation in `ConnectionString` is possible to leave the host empty. Also edge cases like `mongodb://,,,/` are handled just fine.


Anemy (approved):
lgtm! tried it out, I think this check was preventing an edge case that went away with improvements to how we were storing the connection string being edited.

PR #3079, opened 2022-05-11
chore: one should be enough?

lerouxb:


packages/compass-e2e-tests/helpers/commands/shell-eval.ts
```diff
@@ -57,6 +57,7 @@ export async function shellEval(
     } catch (err) {
       // just leave it unparsed for now if there's a parse error because
       // that's really helpful when debugging
+      console.error('Could not parse result:', result);
```
lerouxb, line 60: Trying to figure out `AssertionError: expected 'TypeError: Cannot read property \'res…' to have property 'ok'`

packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -628,18 +628,32 @@ export async function afterTests(
...
+      } catch (_) {
+        debug('browser already closed');
+      }
     }
-  }
+    clearTimeout(timeoutId);
+    return;
+  })();
+
+  return Promise.race([timeoutPromise, closePromise]);
```
lerouxb, line 656: For some reason afterAll for some tests (where it just calls this function that closes compass) often times out after 120 seconds. Even though I can see this reach the end..

Figured I can just time it out because we're already suppressing all errors that can happen while we close compass anyway.

The reason we suppress errors is because you can't close compass if it already crashed which is the most likely reason for this stuff to fail. And we don't want to get into a situation where the entire e2e test run has to time out before it fails. Rather fail fast and with the real error, not some compounding error's eventual fallout.

.evergreen/functions.yml
```diff
@@ -81,17 +81,9 @@ post:
...
       local_files_include_filter:
         - src/packages/compass-e2e-tests/.log/**/*
-  - <<: *save-diagnostic-file
-    params:
-      local_files_include_filter:
         - src/packages/compass-e2e-tests/.nyc_output/coverage.json
-  - <<: *save-diagnostic-file
-    params:
-      local_files_include_filter:
+        - src/packages/compass-e2e-tests/.log/report.json
```
lerouxb, line 85: This should help with it complaining in red that report.json doesn't exist

PR #3088, opened 2022-05-11
feat(explain-aggregations): aggregations in explain-helper COMPASS-5786

mabaasit:
feat(explain-aggregations): aggregations in explain-plan-helper COMPASS-5786

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/explain-plan-helper/src/getExecutionStats.ts
```diff
@@ -0,0 +1,86 @@
...
+  const totalDocsExamined = sumArrayProp(shardStats, 'totalDocsExamined');
+  const response = {
+    nReturned,
+    executionTimeMillis,
+    totalKeysExamined,
+    totalDocsExamined,
+    allPlansExecution: [],
+    executionSuccess: true,
+    executionStages: {
+      stage: shardStats.length === 1 ? "SINGLE_SHARD" : "SHARD_MERGE",
```
addaleax, line 69: Does the linter not complain about this?
```suggestion
      stage: shardStats.length === 1 ? 'SINGLE_SHARD' : 'SHARD_MERGE',
```

Where is this stage name consumed/why these specific values? I can’t see anything outside of tests that references them

packages/explain-plan-helper/src/getExecutionStats.ts
```diff
@@ -0,0 +1,86 @@
+import {
```
addaleax, line 1: not important, but don’t we usually stick with kebab-case for filenames?

packages/explain-plan-helper/src/index.ts
```diff
@@ -1,8 +1,11 @@
 import convertExplainCompat from 'mongodb-explain-compat';
+import { getPlannerInfo } from './getPlannerInfo';
+import { getExecutionStats } from './getExecutionStats';
+import type { ExecutionStats } from './getExecutionStats';
 
 const kParent = Symbol('ExplainPlan.kParent');
 
-type Stage = Record<string, any> & { stage: string; [kParent]: Stage | null };
+export type Stage = Record<string, any> & { stage: string;[kParent]: Stage | null };
```
addaleax, line 8: also something the linter should catch, I think:
```suggestion
export type Stage = Record<string, any> & { stage: string; [kParent]: Stage | null };
```

packages/explain-plan-helper/src/getExecutionStats.ts
mabaasit: Great catch! Here i am creating a response similar to shardedFind by merging execution stats from various shards (something similar to [sharded_geo_query_3.2.json](https://github.com/mongodb-js/compass/blob/COMPASS-5786-explain-plan-helper/packages/explain-plan-helper/test/fixtures/sharded_geo_query_3.2.json)). In case we want to recreate a UI similar to Query Explain, we will have all the data already processed. 

Also we need these aggregated stats in Query Performance Summary of Aggregation Explain.

addaleax (approved):

packages/explain-plan-helper/src/index.ts
mabaasit: i got to fix my linter!

PR #3096, opened 2022-05-16
fix(ssh-tunnel): reconnect the ssh tunnel when it gets disconnected COMPASS-5454

lerouxb:


packages/ssh-tunnel/src/index.ts
```diff
@@ -70,57 +73,16 @@ export class SshTunnel extends EventEmitter {
 
     this.sshClient = new SshClient();
 
+    this.sshClient.on('close', () => {
```
lerouxb, line 76: https://github.com/mscdex/ssh2/issues/491#issuecomment-258041615

> Yes, the 'close' event will always be called when a socket is closed, whether a connection was ended gracefully or closed due to error. Speaking of which, you'll want to add an 'error' event handler for the connection in case of errors, to prevent the process from dying by default.

packages/ssh-tunnel/src/index.ts
```diff
@@ -195,7 +147,55 @@ export class SshTunnel extends EventEmitter {
...
+      return this.connectingPromise;
+    }
+
+    if (!stayConnected) {
+      // A socks5 request could come in after we deliberately closed the connection. Don't reconnect in that case.
+      throw new Error('Disconnected.');
+    }
+
+    debug('creating SSH connection');
+    const ac = new AbortController();
```
lerouxb, line 169: The AbortController allows us to cancel listening to the 'error' and 'ready' events when the Promise.race resolves or rejects. A but minor in the bigger scheme of things, but would technically be a leak?

mcasimir: nice! didn't know `once` accepted a signal

addaleax: It’s only supported in Node.js 15+, I think this might not be working yet? But I guess it’s fine to leave it like this and see it start working when we upgrade Electron :)

packages/ssh-tunnel/src/index.ts
```diff
@@ -195,7 +145,55 @@ export class SshTunnel extends EventEmitter {
...
+      debug('already connected');
+      return;
+    }
+
+    if (this.connectingPromise) {
+      debug('reusing connectingPromise');
+      return this.connectingPromise;
+    }
+
+    if (!stayConnected) {
```
mcasimir, line 161: super NIT, I would have probably this variable easier if this variable was called something like `closed`:

```
if (closed) {
   throw new Error('Tunnel already closed.');
}
```

packages/ssh-tunnel/src/index.ts
```diff
@@ -187,6 +136,7 @@ export class SshTunnel extends EventEmitter {
       // close error
       this.serverClose().catch<Error>((e) => e),
       this.closeSshClient(),
+      // TODO: shouldn't we close the open connections first since they depend on the others?
       this.closeOpenConnections(),
```
mcasimir, line 140: good question, there is the `.catch<Error>((e) => e)` with the comment "we want to await on closing the connections before re-throwing server", that would probably make that part simpler? I don't know if is important to do these 3 operations in parallel

packages/ssh-tunnel/src/index.ts
```diff
@@ -159,24 +125,7 @@ export class SshTunnel extends EventEmitter {
...
-          return waitForReady;
-        })(),
-      ]);
-      debug('created SSH connection');
-    } catch (err) {
-      debug('failed to establish SSH connection', err);
-      await this.serverClose();
-      throw err;
-    }
+    await this.connectSsh(true);
```
addaleax, line 128: Unused argument? :)

packages/ssh-tunnel/src/index.ts
```diff
@@ -195,7 +144,53 @@ export class SshTunnel extends EventEmitter {
...
+
+    debug('creating SSH connection');
+    const ac = new AbortController();
+
+    this.connectingPromise = Promise.race([
+      once(this.sshClient, 'error', { signal: ac.signal }).then(([err]) => {
+        throw err;
+      }),
+      (() => {
+        const waitForReady = once(this.sshClient, 'ready', { signal: ac.signal }).then(() => { return; });
```
addaleax, line 171:
```suggestion
        const waitForReady = once(this.sshClient, 'ready', { signal: ac.signal });
```

packages/ssh-tunnel/src/index.ts
```diff
@@ -70,57 +73,16 @@ export class SshTunnel extends EventEmitter {
 
     this.sshClient = new SshClient();
 
+    this.sshClient.on('close', () => {
+      debug('sshClient closed');
+      this.connected = false;
+    });
```
addaleax, line 79: Is there a race condition here? When the ssh conneciton is in the process of timing out, `.connected` is still `true`?

packages/ssh-tunnel/src/index.ts
```diff
@@ -133,6 +95,10 @@ export class SshTunnel extends EventEmitter {
         )
       );
     }
+    else {
+      debug('skipping auth setup for this server');
+      this.server.useAuth(socks5AuthNone());
+    }
```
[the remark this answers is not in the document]
lerouxb: No I just cleaned up `if (!false) {} else {}`.

packages/ssh-tunnel/src/index.ts
lerouxb: waitForReady isn't `Promise<void>`. There was some `Promise<[void]>` going on that I'm trying to figure out how to cast.

addaleax: I think in that case I’d just do

```suggestion
        const waitForReady = once(this.sshClient, 'ready', { signal: ac.signal }).then(() => {});
```

since otherwise the `return;` looks a bit out-of-place, as if the intention was to actually return something from the `.then()` callback

addaleax: Ugh. Yeah, this is fine, but still icky that we need to do this (what’s wrong with empty arrow functions?) :confused:

packages/ssh-tunnel/src/index.ts
lerouxb: I think it is fine this way - the leaks were already there and they are super minor. You're only going to make so many ssh tunnels.

lerouxb: Actually I just removed the AbortController stuff for now because it was spiralling out and out. It requires jsdom so I'd have to systematically move more and more packages onto our react mocha config so that the polyfill is there when it uses ssh-tunnel.

packages/ssh-tunnel/src/index.ts
lerouxb: I slapped a try/catch around where it forwards to retry now and wrote more tests.

packages/ssh-tunnel/src/index.spec.ts
```diff
@@ -106,6 +112,14 @@ async function stopTestSshTunnel() {
   }
 }
 
+async function breakSshTunnelConnection() {
+  try {
+    return once(sshTunnel.sshClient, 'close');
```
addaleax, line 117: is this supposed to be 
```suggestion
    return await once(sshTunnel.sshClient, 'close');
```

? Otherwise I’m not sure why there’s a try/finally :slightly_smiling_face:

addaleax (approved):
Looks good! Only one comment on the tests :)

packages/ssh-tunnel/src/index.spec.ts
lerouxb: Aah good point.

lerouxb: Should be

```javascript
function breakSshTunnelConnection() {
  const promise = once(sshTunnel.sshClient, 'close');
  sshTunnel.sshClient.end();
  return promise;
}
```

No idea where I copied that from :)

lerouxb: Aah. Just found it. Will change _that_ as well..

PR #3100, opened 2022-05-16
chore(compass-query-bar): update compass query bar to shared config COMPASS-5823

Anemy:
This updates the `compass-query-bar` package to the shared Compass config.

There are two places where changes happen brought up from eslint that I'll highlight. I'll open a quick follow up pr after this one with `prettier` updates to keep the diff low.


packages/compass-query-bar/src/stores/query-changed-store.js
```diff
@@ -65,7 +65,7 @@ const configureStore = (options = {}) => {
         const copyable = state.lastExecutedQuery || this.getInitialState();
 
         for (const key in copyable) {
-          if (copyable.hasOwnProperty(key)) {
+          if (Object.prototype.hasOwnProperty.call(copyable, key)) {
```
Anemy, line 68: Fixes `error  Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins`

packages/compass-query-bar/src/components/options-toggle/options-toggle.jsx
```diff
@@ -36,13 +36,14 @@ class OptionsToggle extends PureComponent {
     );
 
     return (
-      <div
+      <button
```
Anemy, line 39: Fixes 
```
  error  Visible, non-interactive elements with click handlers must have at least one keyboard listener  jsx-a11y/click-events-have-key-events
  error  Static HTML elements with event handlers require a role                                         jsx-a11y/no-static-element-interactions
  ```

addaleax (approved):
:shipit:

PR #3101, opened 2022-05-17
feat: allow import into fle2 collections COMPASS-5810

alenakhineika:
## Description

Currently, the server doesn't support batched inserts for FLE2. We can handle this by catching [this specific error](https://github.com/10gen/mongo/blob/a3efbb9a26eba7f40392b3921ed0162c80b6b6d1/src/mongo/db/fle_crud.cpp#L195) and re-trying to insert documents one by one.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-import-export/src/utils/collection-stream.js
+++ b/packages/compass-import-export/src/utils/collection-stream.js
@@ -1,6 +1,8 @@
 /* eslint-disable no-console */
 import { Writable } from 'stream';
 import { createDebug } from './logger';
+import { promisify } from 'util';
+
 const debug = createDebug('collection-stream');
 
 function mongodbServerErrorToJSError({ index, code, errmsg, op, errInfo }) {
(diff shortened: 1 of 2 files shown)


packages/compass-import-export/src/utils/collection-stream.js
```diff
@@ -69,57 +71,85 @@ class WritableCollectionStream extends Writable {
     this._executeBatch(callback);
   }
 
-  _executeBatch(callback) {
+  async _executeBatch(callback) {
```
addaleax, line 74: :blue_heart:

packages/compass-import-export/src/utils/collection-stream.js
```diff
@@ -69,57 +71,85 @@ class WritableCollectionStream extends Writable {
...
-    this._collection().bulkWrite(
-      documents.map((doc) => ({ insertOne: doc })),
-      {
-        ordered: this.stopOnErrors,
-        retryWrites: false,
-        checkKeys: false,
-      },
-      (err, res) => {
+    const bulkWrite = promisify(
+      this._collection().bulkWrite.bind(this._collection())
```
addaleax, line 80: No need to promisify, right? `collection.bulkWrite()` returns a Promise if it doesn’t receive a callback argument

packages/compass-import-export/src/utils/collection-stream.js
```diff
@@ -69,57 +71,85 @@ class WritableCollectionStream extends Writable {
...
+        }
+      );
+    } catch (bulkWriteError) {
+      // Currently, the server does not support batched inserts for FLE2:
+      // https://jira.mongodb.org/browse/SERVER-66315
+      // We check for this specific error and re-try inserting documents one by one.
+      if (bulkWriteError.code === 6371202) {
+        const insertOne = promisify(
+          this._collection().insertOne.bind(this._collection())
+        );
```
addaleax, line 101: Ditto here :)

packages/compass-import-export/src/utils/collection-stream.js
```diff
@@ -69,57 +71,85 @@ class WritableCollectionStream extends Writable {
...
+          await Promise.allSettled(
+            documents.map(async (doc) => {
+              try {
+                await insertOne(doc, {});
+                nInserted += 1;
+              } catch (error) {
+                this._errors.push(error);
+              }
+            })
+          );
```
addaleax, line 114: This would currently run in parallel, which is nice for performance, but I’m not sure how bad it is to potentially insert in an order which doesn’t match the import source file’s order, or to not stop the insert if `stopOnErrors` is set and an error occurs on one of the inserts.

Maybe this could just be a plain `for` loop? I think the performance impact here would be acceptable either way, it’s fine for this to be slow, especially if the server will add support for batched writes sometime in the future anyway.

packages/compass-import-export/src/utils/collection-stream.js
```diff
@@ -69,57 +71,85 @@ class WritableCollectionStream extends Writable {
...
+        this._errors.push(bulkWriteError);
+      }
+    }
 
-        this._batchCounter++;
+    // Driver seems to return null instead of undefined in some rare cases
+    // when the operation ends in error, instead of relying on
+    // `_mergeBulkOpResult` default argument substitution, we need to keep
+    // this OR expression here
+    this._mergeBulkOpResult(result);
```
addaleax, line 132: The `||` expression is gone, so this comment might need some updating? :)

packages/compass-import-export/src/utils/collection-stream.js
```diff
@@ -69,57 +71,85 @@ class WritableCollectionStream extends Writable {
...
+          ordered: this.stopOnErrors,
+          retryWrites: false,
+          checkKeys: false,
+        }
+      );
+    } catch (bulkWriteError) {
+      // Currently, the server does not support batched inserts for FLE2:
+      // https://jira.mongodb.org/browse/SERVER-66315
+      // We check for this specific error and re-try inserting documents one by one.
+      if (bulkWriteError.code === 6371202) {
```
addaleax, line 98: Do we want to also do `this.BATCH_SIZE = 1;` in this case, to reduce the number of times that we run into this condition? (The `BATCH_SIZE` check above would need to be turned into `>=` instead of `===` in that case, but as you mentioned yesterday, that might be more intuitive anyway)

packages/compass-import-export/src/utils/collection-stream.js
alenakhineika: Ths for pointing at this! Post this here for reference: https://www.mongodb.com/docs/manual/core/bulk-write-operations/#ordered-vs-unordered-operations

addaleax (approved):

PR #3102, opened 2022-05-17
feat(explain-aggregation): explain aggregation COMPASS-5788

mabaasit:
feat(explain-aggregation): explain aggregation COMPASS-5788

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/components/pipeline-explain/explain-error.tsx
```diff
@@ -0,0 +1,58 @@
...
+export const ExplainError: React.FunctionComponent<ExplainErrorProps> = ({
+  isNetworkError,
+  message,
+  onRetry,
+}) => {
+  if (!isNetworkError) {
+    return (
+      <ErrorSummary data-testid="pipeline-explain-error" errors={[message]} />
+    );
+  }
```
addaleax, line 41: I’m curious about this. We’re not giving network errors special treatment in other parts of Compass – where did we make the decision to add this? If we do want to handle network errors, it might make sense to think about how we can address them more generally, not just for explain execution (especially since users are unlikely to reach the point of being able to run explain on aggregations if there are connectivity issues).

packages/compass-aggregations/src/components/pipeline-explain/explain-indexes.tsx
```diff
@@ -0,0 +1,25 @@
...
+}) => {
+  if (indexes.filter(({ index }) => index).length === 0) {
+    return <Body weight="medium">No index available for this query.</Body>;
+  }
+
+  return (
+    <div>
+      {indexes.map((info, idx) => (
+        <Badge key={idx} variant={BadgeVariant.LightGray}>
+          {info.index} {info.shard && <>({info.shard})</>}
```
addaleax, line 20: React noob question – what’s difference between `{info.shard && <>({info.shard})</>}` and just `{info.shard}`?

packages/compass-aggregations/src/utils/cancellable-aggregation.ts
```diff
@@ -63,3 +63,54 @@ export async function aggregatePipeline({
...
+    pipeline,
+    options
+  );
+  const abort = () => {
+    Promise.all([
+      cursor.close(),
+      dataService.killSessions(session)
+    ]).catch((err) => {
+      log.warn(
+        mongoLogId(1001000139),
```
addaleax, line 99: tiny nit, should be okay here since this is .ts:

```suggestion
        mongoLogId(1_001_000_139),
```

packages/compass-aggregations/src/components/pipeline-explain/explain-error.tsx
mabaasit: This came from the designs that Claudia shared with me. I'll ping her, see if can align the workflow and avoid this use case.

packages/compass-aggregations/src/components/pipeline-explain/explain-indexes.tsx
mabaasit: Goal here is to show the index along with its shard name, if defined.
In case `info.shard` is not defined, the former does not render empty parenthesis, while the latter does.

Just to avoid something like this: `host_id_idx ()`

packages/compass-aggregations/src/components/pipeline-explain/explain-error.tsx
addaleax: Yeah, I mean, it's not like anything's bad about this, it just feels like we're doing a lot of work to handle a very unlikely error condition well

packages/compass-aggregations/src/components/pipeline-explain/explain-indexes.tsx
addaleax: Well I did miss those parentheses :) All good then!

addaleax (approved):

packages/compass-aggregations/src/components/pipeline-explain/explain-indexes.tsx
gribnoysup: FWIW making it a template literal will make it a bit more readable in this case otherwise it's really easy to miss these parentheses

```
`(${info.shard})`
```

packages/compass-aggregations/src/components/pipeline-explain/explain-results.tsx
```diff
@@ -0,0 +1,50 @@
...
+      )}
+      <Card className={cardStyles} data-testid="pipeline-explain-results-json">
+        <DocumentListView
+          docs={[new HadronDocument(plan)]}
+          isEditable={false}
+          copyToClipboard={(doc) => {
+            const str = doc.toEJSON();
+            void navigator.clipboard.writeText(str);
+          }}
+        />
```
gribnoysup, line 46: Why are we using a list component if we have just a Document component for that?

packages/compass-aggregations/src/components/pipeline-explain/index.tsx
```diff
@@ -1,36 +1,112 @@
...
+  const modalFooter = (
+    <ModalFooter className={footerStyles}>
+      <Button
+        onClick={onCloseModal}
+        data-testid="pipeline-explain-footer-close-button"
+      >
+        Close
+      </Button>
+    </ModalFooter>
+  );
```
gribnoysup, line 83: I would suggest you to do these types of things only for the parts of the rendering tree that are complex to grasp at a glance or have a non-trivial branching logic. Otherwise this has the same problem that having these `renderThing` methods that we have all over the place: you can't get a view of a component structure at a glance and have to jump back and forth to understand it

packages/compass-aggregations/src/components/pipeline-explain/explain-results.tsx
mabaasit: Didn't know we have that. Implemented.

packages/compass-aggregations/src/components/pipeline-explain/explain-results.tsx
```diff
@@ -0,0 +1,50 @@
...
+      {stats && (
+        <ExplainQueryPerformance
+          nReturned={stats.nReturned}
+          executionTimeMillis={stats.executionTimeMillis}
+          usedIndexes={stats.usedIndexes}
+        />
+      )}
+      <Card className={cardStyles} data-testid="pipeline-explain-results-json">
+        <Document
+          doc={doc}
```
gribnoysup, line 41: As we recently discovered re-creating `HadronDocument` in render [is really not great](https://github.com/mongodb-js/compass/pull/3073) so instead of passing a document instance created on every render, you can pass the object that the document is based on instead

```suggestion
          doc={plan}
```

The memoization of the `HadronDocument` instance will be handled by the `Document` component itself

packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
```diff
@@ -19,6 +19,10 @@ async function waitForAnyText(
     return text !== '';
   });
 }
+const initialAggregationToolbarValue =
```
gribnoysup, line 22: Do you think you can cherry pick e2e test changes to a separate PR / postpone them a bit? I'm cleaning up feature flags and updating the e2e and this already had conflicts with the explain plan work that I had to resolve and I would really prefer to avoid another conflict resolution or being blocked on this PR landing. It's okay if it's not possible, I'll just postpone feature flag removal and work on something else, just tell me what your preference would be here

packages/compass-aggregations/src/modules/explain.ts
```diff
@@ -1,51 +1,217 @@
...
+    try {
+      const abortController = new AbortController();
+      const signal = abortController.signal;
+      dispatch({
+        type: ActionTypes.ExplainStarted,
+        abortController,
+      });
+
+      const options: AggregateOptions = {
+        maxTimeMS: maxTimeMS || DEFAULT_MAX_TIME_MS,
```
gribnoysup, line 157: Zero is a valid `maxTimeMS` value

```suggestion
        maxTimeMS: maxTimeMS ?? DEFAULT_MAX_TIME_MS,
```

packages/compass-aggregations/src/components/pipeline-explain/explain-query-performance.tsx
```diff
@@ -0,0 +1,60 @@
...
+    return (
+      <div
+        className={containerStyles}
+        data-testid="pipeline-explain-results-summary"
+      >
+        <Subtitle>Query Performance Summary</Subtitle>
+        <div className={statsStyles}>
+          <div className={statItemStyles}>
+            <Body>Documents returned:</Body>
+            <Body weight="medium">{nReturned}</Body>
```
gribnoysup, line 45: Seems like nReturned is not always returned here, maybe we should show a placeholder or something? This is how an explain looks for me for a random aggregation I tried:

image

In other places we use em dash (`—`) for that

packages/compass-aggregations/src/components/pipeline-explain/index.tsx
```diff
@@ -1,36 +1,103 @@
...
+  } else if (explain) {
+    content = <ExplainResults plan={explain.plan} stats={explain.stats} />;
+  }
+
+  if (!content) {
+    return null;
+  }
+
   return (
     <Modal
```
gribnoysup, line 69: On a default compass window size this modal height makes the modal scroll even when there is not that much content on the screen, I think it would make sense to fix that

Kapture 2022-05-18 at 10 59 45

packages/compass-aggregations/src/components/pipeline-explain/explain-query-performance.tsx
```diff
@@ -0,0 +1,60 @@
...
+          </div>
+          <div className={statItemStyles}>
+            <Body>Actual query execution time(ms):</Body>
+            <Body weight="medium">{executionTimeMillis}</Body>
```
gribnoysup, line 49: Even on a very long running aggregation I'm getting a zero execution time, not sure if that's expected (also no `nReturned`):

image

This is from running following aggregation on `Corporate-Registrations.corporations`

```json
[
    {
        "$sort": {
            "name": -1
        }
    },
    {
        "$limit": 100
    }
]
```

mabaasit: hmm. great catch! i didn't find this scenario when i was working on processing raw server explain in my previous tickets. also since the raw explain is not documented anywhere, its hard to handle all the cases.

in this case, i'll hide the section if we don't have any info, while I work on fixing it in another pr.

packages/compass-aggregations/src/components/pipeline-explain/explain-query-performance.tsx
mabaasit: In explain for queries, we are using colon. Not sure if we should change it here.
But if we are in the process of transitioning over to dash throughout compass, I'd be happy to do that.

gribnoysup: I'm okay with any placeholder, this was just a suggestion that I saw that we use often in other places, whatever you prefer that is not just empty space would be great

gribnoysup:
Should we disable "Explain" button in the tab header when last stage is merge or out similar to what we do for "Export"?

image

packages/compass-aggregations/src/utils/cancellable-aggregation.ts
```diff
@@ -63,3 +63,54 @@ export async function aggregatePipeline({
...
+  };
+  signal.addEventListener('abort', abort, { once: true });
+  let result = {};
+  try {
+    const { dataLake: { isDataLake } } = await dataService.instance();
+    const verbosity = isDataLake ? 'queryPlannerExtended' : 'allPlansExecution';
+    result = await raceWithAbort(cursor.explain(verbosity), signal);
+  } finally {
+    signal.removeEventListener('abort', abort);
+  }
```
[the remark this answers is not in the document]
mcasimir: I vaguely recall discussing this, but another option more self-contained, would be to just ignore this, do the call, and if the error is `allPlansExecution` is not supported we can repeat with  `queryPlannerExtended`.

I'm saying this having in perspective to move this to the `DataService` and considering that the `explain` for ADL is not all that interesting, so here we could just "privilege" the more straightforward usage with regular clusters and have a completely stateless implementation.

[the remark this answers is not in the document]
mcasimir: although maybe even better if the `allPlansExecution` and `queryPlannerExtended` are passed as arguments maybe to the data service at least for DE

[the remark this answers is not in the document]
gribnoysup: Trying one and then falling back to another might also cover [this case](https://github.com/mongodb-js/compass/pull/3102#issuecomment-1129817823) I guess. Although I'm still not sure if we should allow to explain an out / merge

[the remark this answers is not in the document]
mabaasit: I am not sure of `$out` either, but @mcasimir with [this design](https://docs.google.com/document/d/1ysoL5tYfaUB0w0gFk8-8XDoA_YoX5RaqYPiT0xTNUSY/edit#), we should be able to handle that.

packages/compass-aggregations/src/components/pipeline-explain/explain-query-performance.tsx
```diff
@@ -40,14 +40,18 @@ export const ExplainQueryPerformance: React.FunctionComponent<ExplainQueryPerfor
...
-          <div className={statItemStyles}>
-            <Body>Actual query execution time(ms):</Body>
-            <Body weight="medium">{executionTimeMillis}</Body>
-          </div>
+          {nReturned > 0 && (
+            <div className={statItemStyles}>
+              <Body>Documents returned:</Body>
+              <Body weight="medium">{nReturned}</Body>
+            </div>
+          )}
```
gribnoysup, line 48: It is probably possible for `nReturned` to be zero if query did not match any documents, right? 🤔

packages/compass-aggregations/src/components/pipeline-explain/explain-results.tsx
```diff
@@ -18,7 +18,7 @@ const containerStyles = css({
 });
 
 const cardStyles = css({
-  height: '350px',
+  height: `calc(100vh - 450px)`,
```
gribnoysup, line 21: Where's the 450px number comes from? Is it a calculated value based on some other spacings? Might be worth adding a comment with explanation in case leafygreen changes some spacing of their modal component and we would need to recalculate it or something.

FWIW it feels like maybe something like this belongs more on a modal itself (if that's even possible) not on the card? Feels like what we're trying to do here is for modal container to always as much height as possible (but not more than some `x` value probably so it doesn't stretches too long when people are working with Compass on vertical screens), so something like this for the modal:

```css
.modal {
  height: calc(100vh - var(--leafygreen-modal-vertical-margin));
  max-height: var(--whatever-we-feel-is-appropriate-here-like-600-px-or-something);
}
```

And then the modal content layout should be structured  in a way where card takes as much space as possible?

This feels like something a bit more resilient to changes to the modal content while covering the use case of giving card as much of allowed space.

But also maybe it's a bit of an overkill, I would at least add a comment where the number comes from though 😄

packages/compass-collection/src/stores/index.ts
```diff
@@ -248,13 +250,16 @@ store.onActivated = (appRegistry: AppRegistry) => {
...
-  });
+  appRegistry.on(
+    'data-service-connected',
+    (error, dataService: DataService) => {
+      store.dispatch(dataServiceConnected(error, dataService));
+      void dataService.instance().then(({ dataLake: { isDataLake } }) => {
+        store.dispatch(dataLakeChanged(isDataLake));
+      });
+    }
+  );
```
gribnoysup, line 262: Hmmm, that's weird that you have to do this here at all. But also I think there is an issue that seems to be in Compass for awhile where `isDataLake` is not passed to compass-collection plugin correctly. I do think thought that we should resolve this without making yet another call to `dataService.instance` outside the instance model that already has this information.

As this is something that can be addressed independently and the feature is behind a feature flag for now how do you feel about reverting the changes in this file (and keeping everything else you changed in compass-collection) and I can take a deeper look at this in parallel while you are wrapping up work on this PR?

mabaasit: thanks for helping me with its resolution.
```
instance.dataLake.on(
  'change:isDataLake',
  (_model: unknown, value: boolean) => {
    store.dispatch(dataLakeChanged(value));
   }
);
```

packages/compass-aggregations/src/components/pipeline-explain/explain-query-performance.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const ExplainQueryPerformance: React.FunctionComponent<ExplainQueryPerformanceProps> =
+  ({ nReturned, executionTimeMillis, usedIndexes }) => {
+    return (
+      <div
+        className={containerStyles}
+        data-testid="pipeline-explain-results-summary"
+      >
+        <Subtitle>Query Performance Summary</Subtitle>
+        <div className={statsStyles}>
+          {nReturned && (
```
gribnoysup, line 43: Zero is falsy, so if we want zeroes to show up, you probably want something like this instead

```suggestion
          {typeof nReturned === 'number' && (
```

packages/compass-aggregations/src/components/pipeline-explain/explain-results.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const containerStyles = css({
+  display: 'flex',
+  flexDirection: 'column',
+  gap: spacing[4],
+});
+
+const cardStyles = css({
+  // 170px works with minimum-height of compass
+  // todo: handle height for bigger sized compass
```
gribnoysup, line 22: As an addition to this TODO, currently because "copy to clipboard" button is part of the Document component and not the card, the whole thing scrolls including the button:

image

I think we might want to set the height here on the document fields container and not the wrapper card as a way to prevent it, but there is no way to do it now and it's a very small thing so totally can be addressed later

mabaasit: sure :)

gribnoysup (approved):

packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
```diff
@@ -549,6 +549,8 @@ describe('Collection aggregations tab', function () {
       await browser.waitForAnimations(Selectors.AggregationExplainModal);
 
       const modal = await browser.$(Selectors.AggregationExplainModal);
+      await browser.waitForAnimations(Selectors.AggregationExplainModal);
+      await modal.waitForDisplayed();
```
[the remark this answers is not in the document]
mabaasit: Checked it @lerouxb and flipping them makes sense.

packages/compass-aggregations/src/components/pipeline-explain/explain-query-performance.tsx
mabaasit: Added a fix in #3170

PR #3103, opened 2022-05-17
fix(export-to-language): Go Export Improvements COMPASS-5837

prestonvasquez:
## Description

Apply improvements for
- the URL in ApplyURI is always "localhost", in the other languages we pass that value as a variable somehow so it reflects the current connection
- the import: "go.mongodb.org/mongo-driver/mongo/options" is missing with the generated code
- Missing struct initialization: primitive.Null -> primitive.Null{} and primitive.Undefined -> primitive.Undefined{}

### Checklist

## Motivation and Context

Minor issues in language export

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


prestonvasquez:
@mcasimir Here is the patch for COMPASS-5837 .

mcasimir (approved):

PR #3114, opened 2022-05-19
fix(aggregations): Hide the stage operator because it only works on db level COMPASS-5843

lerouxb:


packages/compass-aggregations/src/utils/stage.js
```diff
@@ -103,6 +103,17 @@ const isSearchOnView = ({ oName, isTimeSeries, isReadonly, sourceName }) =>
...
+
+/**
+ * Some stages only work with db.aggregate and not coll.aggregate
+ *
+ * @param {Object} options
+ * @property {String} oName - The stage name.
+ *
+ * @returns {boolean} If the stage is only works on db aggregations.
+ */
+const isDbOnly = ({ oName }) => [DOCUMENTS].includes(oName);
```
addaleax, line 115: This information should probably at least _eventually_ be made part of the mongodb-ace-autocompleter package. Should we do this as part of this PR?

$currentOp is also a db-only stage, but it’s not even listed in that package right now (possibly for this reason).

lerouxb: I don't think the autocompleter knows anything about the required context, though. Sounds like a bigger change? To be honest I don't really know that package.  I did ask about removing it from the list there entirely and apparently that's a no, but I can't think of where else we'd autocomplete these operators https://mongodb.slack.com/archives/G2L10JAV7/p1652954616265999

lerouxb: Oh in vscode extension https://github.com/mongodb-js/vscode/blob/1e98a4e9634fe1f298821a1b894b9a5e24902522/scripts/update-snippets.ts#L52

and mongosh https://github.com/mongodb-js/mongosh/blob/9e5064654780d2a61bad6904afaca250c7030b31/packages/autocomplete/src/index.ts#L8

addaleax: > I don't think the autocompleter knows anything about the required context, though.

What’s the “required context” in this context? We already provide context for autocompletion, like server version etc., so we could probably use this here as well? Or even if the code structurally remains the same, it’s probably still better for the actual “this stage can only be applied to db.aggregate()” information to be part of the `STAGE_OPERATORS` list entries.

lerouxb: I can't spot any logic around autocompleting STAGE_OPERATORS inside the ace autocompleter repo. It is just reexporting a const array.

Do you mean something like export two or three separate lists? ie. the unfiltered one, the one that works for dbs, the one that works for collections?

lerouxb: And would I move the rest of filterStageOperators() there too? I don't mind, I just don't know what makes most sense to you.

addaleax: I’m thinking about something like

```diff
diff --git a/lib/constants/stage-operators.js b/lib/constants/stage-operators.js
index b05a605c14c6..5b3918a9c686 100644
--- a/lib/constants/stage-operators.js
+++ b/lib/constants/stage-operators.js
@@ -90,6 +90,7 @@ const STAGE_OPERATORS = [
     name: '$documents',
     value: '$documents',
     label: '$documents',
+    dbOnly: true,
     score: 1,
     env: [ ATLAS ],
     meta: 'stage',
```

(or a `contexts: ['db']` array which is then set to `contexts: ['coll']` for the other stage operators)

I don’t care so much about where the filtering happens as I care about the stage operator metadata being stored in one place, not spread across projects :)

lerouxb: Sounds good. Will do that.

lerouxb: https://github.com/mongodb-js/ace-autocompleter/pull/126

mabaasit (approved):

PR #3117, opened 2022-05-19
chore(compass-query-bar): add new toolbars env flag, use in `compass-query-bar` COMPASS-5670

Anemy:
COMPASS-5670

## Description

This pr adds a `COMPASS_SHOW_NEW_TOOLBARS` feature flag to Compass. We then use use it in `compass-query-bar` to show the new query bar while it's being developed. Once this feature is ready we'll remove the feature flag and expose it.  Also a few drive by file updates to typescript where easily done (more will happen when we move deeper into [COMPASS-5673](https://jira.mongodb.org/browse/COMPASS-5673)

Screen Shot 2022-05-19 at 12 51 55 PM

## Notes for Reviewers

I exposed the `focusRingStyles` and `focusRingVisibleStyles` from [use-focus-ring.ts](https://github.com/mongodb-js/compass/compare/COMPASS-5670-add-toolbars-env-flag?expand=1#diff-57f6739adcc3ac73c61e6882a02f5198a3bd269d5f5fe6ffb7f487cfa42f1bc3) in `compass-components` and used it for one of the more custom buttons. How do y'all feel about this? We could use the hook instead, but it didn't seem to be needed for this implementation.


packages/compass-query-bar/src/components/query-bar/query-bar.tsx
```diff
@@ -0,0 +1,122 @@
...
+  uiColors,
+} from '@mongodb-js/compass-components';
+
+const queryBarStyles = css({
+  display: 'flex',
+  alignItems: 'center',
+  gap: spacing[2],
+  padding: spacing[2],
+  border: `1px solid ${uiColors.gray.light2}`,
+  borderRadius: '6px',
```
mcasimir, line 21: Is this radius shared with other components, for example something in the Aggregation Tab?

mcasimir (approved):

packages/compass-query-bar/src/components/query-bar/query-bar.tsx
Anemy: Yes, how about an exported `toolbarBorderRadius` variable in the Toolbar component in compass-components so we can share it? I'll merge this and can include in next pr if that sounds like a good move, open to other ways to share too. Probably not worth it's own component I think, but the `border` is also duplicated.

PR #3118, opened 2022-05-20
chore(compass-crud): Update compass crud to shared config COMPASS-5824

Anemy:
COMPASS-5824
This pr updates the `compass-crud` package to the shared Compass config. This sets us up to do the `compass-crud` toolbar updating work in COMPASS-5672 in typescript and with the config improvements. 
This removes the electron runtime.

I'll open up a reformatting/prettier pr after this is merged. 
Also after this is merged we'll do COMPASS-5847 which is around removing the `eslint-disable`s added in this pr (Also added TODOs linking the COMPASS-5847 ticket in those places). Most of the things that the new eslint config brought up are non-interactive html tags being used with `onClick` handlers, mostly in the table view.

Since the diff is a bit big I will add comments in the places where there are actual changes.


packages/compass-crud/src/components/table-view/cell-editor.jsx
```diff
@@ -236,15 +239,6 @@ class CellEditor extends React.Component {
     this.props.api.refreshCells({rowNodes: [this.props.node], force: true});
   }
 
-  focus() {
-    setTimeout(() => {
-      const container = ReactDOM.findDOMNode(this.props.reactContainer);
```
Anemy, line 241: Looks like this was unused, also it was using the deprecated `findDOMNode` so I removed.

```
Do not use findDOMNode. It doesn’t work with function components and is deprecated in StrictMode. See https://reactjs.org/docs/react-dom.html#finddomnode  react/no-find-dom-node
```

packages/compass-crud/src/components/table-view/cell-editor.jsx
```diff
@@ -54,7 +58,7 @@ class CellEditor extends React.Component {
    * in this cell, get the type of the column from this.props.column and add a
    * field to the HadronDocument that is empty.
    */
-  componentWillMount() {
+  UNSAFE_componentWillMount() {
```
Anemy, line 61: I think fixing this is probably not worth doing. Might be worth investigating using a table library before investing much more time into the table view. Open to other ideas, could lump it into COMPASS-5847 but it would blow up the scope and might introduce some unintentional bugs.

packages/compass-crud/src/components/breadcrumb.jsx
```diff
@@ -32,23 +32,23 @@ class BreadcrumbComponent extends React.PureComponent {
...
+        </button>
         {this.props.path.map((name, i) => {
           let displayName = '';
           if (typeof name === 'number' && i > 0) {
             displayName = this.props.path[i - 1] + '.';
           }
           displayName = displayName.concat(name);
           return (
-            <span
+            <button
```
Anemy, line 46: Fixes eslint `jsx-a11y/no-noninteractive-element-interactions`
Screen Shot 2022-05-19 at 11 46 42 PM
Same with the `button` update above.

packages/compass-crud/src/components/table-view/cell-renderer.jsx
```diff
@@ -238,9 +237,9 @@ class CellRenderer extends React.Component {
       return null;
     }
     return (
-      <div className={'table-view-cell-circle-button'} onClick={this.handleDrillDown.bind(this)}>
+      <button className={BUTTON_CLASS} onClick={this.handleDrillDown.bind(this)}>
```
Anemy, line 240: Updated the `div` here to be an interactive html element (and the button above as well.

Here's how they look (same as before):
Screen Shot 2022-05-19 at 10 12 21 PM
Screen Shot 2022-05-19 at 10 13 18 PM

packages/compass-crud/src/stores/crud-store.js
```diff
@@ -378,17 +353,9 @@ const configureStore = (options = {}) => {
...
-      input.value = documentJSON;
-      document.body.appendChild(input);
-      input.select();
-      const success = document.execCommand(COPY);
-      document.body.removeChild(input);
-      input = null;
-      return success;
+      const documentEJSON = doc.toEJSON();
+      // eslint-disable-next-line no-undef
+      void navigator.clipboard.writeText(documentEJSON);
```
Anemy, line 358: Updated the copy pasting, also added 2 tests https://github.com/mongodb-js/compass/pull/3118/files#diff-4ac7fe029a57f95f04bb2b93dbcfc6a83fe0fcbf2b46b89103577efbe3329199R213

packages/compass-crud/src/components/table-view/cell-editor.jsx
addaleax: You mean, the `.focus()` call didn’t work?

addaleax (approved):

packages/compass-crud/src/components/table-view/cell-editor.jsx
Anemy: Ah sorry, totally misworded that, yes the `.focus()` call didn't work. This function was called/used, but it wasn't doing anything.
I'll spend a bit and see if we can get this behavior as intended.

Anemy: Probably better we're not autofocusing here for now. I think we'll revisit the table separately and we can see if it's something we want then. As we're not losing functionality here compared to what was happening I think it's okay to remove. I can spend more time re-implementing this part if we're thinking it's useful.

configs/webpack-config-compass/src/externals.ts
```diff
@@ -12,6 +12,8 @@ export const sharedExternals: string[] = [
   'win-export-certificate-and-key',
   'macos-export-certificate-and-key',
   'mongodb-client-encryption',
+  // We externalize react as we want to avoid multiple instances of it.
+  'react',
```
Anemy, line 16: Currently `compass-explain-plan` and `compass-aggregations` use `compass-crud`. This works fine in the hot reloading app, however in the testing mocha build we're seeing two react instances which crashes it. We need to externalize react so only the top consumer brings it in. This isn't really the place to do that, we probably want it somewhere in the `index.ts` of this folder. There might be another way to do it too, looking into.

Anemy:
Posting update here - a few places use `compass-crud` directly as a dependency (for the document component). As a result of the config updates here we'll need to update those dependents to resolve this package differently. `compass-schema-validation`, and possibly another package. We can also look into moving the shared dependency out of `compass-crud` so it's more of a leaf. Investigation in progress

packages/compass-crud/src/utils/cancellable-queries.spec.js
```diff
@@ -73,6 +74,12 @@ describe('cancellable-queries', function() {
     await insertOne('config.collections', { _id: 'cancel.sharded', key: { a: 1 } }, {});
   });
 
+  after(async function() {
+    if (dataService) {
+      await dataService.disconnect();
```
Anemy, line 79: With the `mocha` version update we're now not immediately quitting after tests, so this cleanup ensures we can exit gracefully.

packages/compass-crud/package.json
```diff
@@ -3,39 +3,67 @@
...
-    "pretest": "mongodb-runner start --port 27018",
-    "test": "cross-env NODE_ENV=test mocha-webpack \"./src/**/*.spec.js\"",
-    "test:watch": "cross-env NODE_ENV=test mocha-webpack \"./src/**/*.spec.js\" --watch",
-    "posttest": "mongodb-runner stop --port 27018",
-    "cover": "nyc npm run test",
-    "test-check-ci": "npm run check && npm run cover",
-    "depcheck": "depcheck",
-    "lint": "eslint \"./src/**/*.{js,jsx}\" \"./test/**/*.js\" \"./electron/**/*.js\" \"./config/**/*.{js,jsx}\"",
-    "check": "npm run depcheck && npm run lint",
+    "bootstrap": "npm run compile",
```
Anemy, line 36: Not needed anymore
```suggestion
```

PR #3121, opened 2022-05-23
feat(data-service): cancellable explain COMPASS-5808

mabaasit:
feat(data-service): cancellable explain in data-service

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/modules/explain.ts
```diff
@@ -216,4 +220,21 @@ export const explainAggregation = (): ThunkAction<
   }
 };
 
+const _getExplainVerbosity = (
```
mcasimir, line 223: NIT: Prefixing module functions that are not exported with `_` shouldn't be necessary. Are other functions in this file / package following this convention?

Normally we prefix private class methods with `_` to signal that a method is private without TS so it is clear that we don't need to call that when we use the class outside.

packages/data-service/src/data-service.spec.ts
```diff
@@ -1477,6 +1478,65 @@ describe('DataService', function () {
...
+          abortSignal,
+        };
+
+        dataService.explainAggregate(
+          testNamespace,
+          pipeline,
+          {},
+          executionOptions as any
+        )
+          .catch((error) => {
```
mcasimir, line 1533: What about we use the `async/await` for of this and remove these callback style tests?

```
const error = dataService.explainAggregate(...).catch(err => err);
expect(error).to.be.instance.of(Error);
expect(dataService.isOperationCancelledError(error)).to.true;
```

packages/data-service/src/cancellable-promise.ts
```diff
@@ -0,0 +1,73 @@
...
+  };
+
+  abortSignal.addEventListener('abort', abort, { once: true });
+  successSignal.addEventListener('abort', succeed, { once: true });
+
+  return promise;
+}
+
+/*
+ * We need a promise that will reject as soon as the operation is aborted since
```
mcasimir, line 51: This comment is a bit out of context here, talks about "closing the cursor" .. not sure if it helps who reads too much, is it a leftover?

packages/data-service/src/cancellable-promise.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const OPERATION_CANCELLED_MESSAGE = 'The operation was cancelled.';
+
+class PromiseCancelledError extends Error {
+  name = OPERATION_CANCELLED_ERROR;
+}
+
+export const createCancelPromiseError = (): Error => {
+  return new PromiseCancelledError(OPERATION_CANCELLED_MESSAGE);
+};
```
mcasimir, line 11: This is a bit nicer:

```suggestion
class PromiseCancelledError extends Error {
  name = OPERATION_CANCELLED_ERROR;
  
  constructor() {
     super(OPERATION_CANCELLED_MESSAGE);
  }
}
```

Like that we don't need `createCancelPromiseError` anymore and just call `new PromiseCancelledError()` everywhere.

packages/data-service/src/cancellable-promise.ts
```diff
@@ -0,0 +1,73 @@
...
+ * We need a promise that will reject as soon as the operation is aborted since
+ * closing the cursor isn't enough to immediately make the cursor method's
+ * promise reject.
+ */
+export async function raceWithAbort<T>(
+  promise: Promise<T>,
+  signal: AbortSignal
+): Promise<T> {
+  if (signal.aborted) {
+    return Promise.reject(createCancelPromiseError());
```
mcasimir, line 60: Do we need to return a Promise directly here? or we could just throw?

packages/data-service/src/data-service.ts
```diff
@@ -1996,6 +2033,44 @@ export class DataServiceImpl extends EventEmitter implements DataService {
...
+          'CancelOp',
+          'Attempting to kill the session failed',
+          { error: err.message }
+        );
+      });
+    };
+    abortSignal.addEventListener('abort', abort, { once: true });
+    let result: T;
+    try {
+      result = await raceWithAbort(start(), abortSignal);
```
mcasimir, line 2067: This is more a personal opinion .. I don't love all that much that we are adding this external `raceWithAbort` and "cancellable promise", is super neat to put that in a different file, but unless I'm missing something is not necessary and creates a new "Concept" that I'd have a hard time to wrap my head around. 

Unless I misunderstood what that does, the whole "raceWithAbort can be replaced here by this code directly:

```
let abortListener;

const aborted = new Promise(,reject=> {
  abortListener = reject;
  abortController.addEventListener('abort', () => reject(new OperationCancelledError()), {once: true});
});

try {
  await Promise.race([aborted, promise]);
} finally {
  abortController.removeEventListener('abort', abortListener);
}
```

packages/data-service/tsconfig.json
```diff
@@ -2,7 +2,12 @@
   "extends": "@mongodb-js/tsconfig-compass/tsconfig.common.json",
   "compilerOptions": {
     "allowJs": true,
-    "lib": ["ES2020"],
+    "lib": [
+      "ES2020",
+      // With the implementation of cancellable actions using AbortController, we need to include DOM for types.
+      // In nodejs, AbortController support is from 15.x and we are currently on 14.x.
+      "DOM"
```
mcasimir, line 9: This should not be necessary if we don't create `AbortController`

 instances in `data-service`, you should be able to do `type AbortController = EventEmitter` and the interface should match. That way ts would still complain if we try to use browser/dom stuff in data-service which is nice.

packages/data-service/src/data-service.ts
mabaasit: @mcasimir thanks for pairing up on this one and helping me clean it!

packages/data-service/tsconfig.json
mabaasit: Added a custom type for AbortSignal and avoided using `DOM` completely. 
Unfortunately we can not use `EventEmitter` because of different interfaces.

mcasimir (approved):

mcasimir:
Would be nice to add 1 or 2 tests for `cancellableOperation` but looks great!

johnjackweir (approved):
LGTM but just for my own understanding, would you mind confirming the motivations for this ticket? From my reading, we're adding a new "cancellableExplain" operation here. Will this be used in the "Explain Plan" section of Compass? FWIW Data Explorer doesn't even support that tab and there isn't yet a mechanism for users to get query details in DE

mabaasit:
> LGTM but just for my own understanding, would you mind confirming the motivations for this ticket? From my reading, we're adding a new "cancellableExplain" operation here. Will this be used in the "Explain Plan" section of Compass? FWIW Data Explorer doesn't even support that tab and there isn't yet a mechanism for users to get query details in DE

We are adding this here in order to support explain for aggregations. With the [redesign of aggregation toolbar](https://www.figma.com/file/7t6AJJjaZG0mX47X8MWubY/Export-Pipeline?node-id=431%3A15238) in compass, we added support for Explain ([within the same aggregation tab](https://www.figma.com/file/vbjoD4dcisPYRYJ5RRu72k/Compass-LG?node-id=1657%3A44222)). So, this will not be used in Explain Plan tab (which supports only queries).

And for Data Explorer, just like we support Run and Export of aggregations, we also added support for Explain.

packages/data-service/src/data-service.ts
```diff
@@ -110,6 +110,16 @@ function isEmptyObject(obj: Record<string, unknown>) {
   return Object.keys(obj).length === 0;
 }
 
+class OperationCancelledError extends Error {
+  name = 'OperationCancelledError';
```
addaleax, line 114: Do we want to namespace this, e.g. `DataServiceOperationCancelledError`?

packages/data-service/src/data-service.ts
```diff
@@ -118,6 +128,27 @@ interface CompassClientSession extends ClientSession {
...
+type AbortSignal = {
+  aborted: boolean;
+  onabort: ((this: AbortSignal, event: Event) => void) | null;
+  addEventListener: (
+    type: string,
+    listener: (event: Event) => void,
+    options: Record<string, unknown>
+  ) => void;
+  removeEventListener: (type: string, listener: (event: Event) => void) => void;
+  dispatchEvent: (event: Event) => boolean;
```
addaleax, line 143: Where is `Event` included from? If it’s the DOM `Event` class, then we probably already require the DOM type definitions here.

packages/data-service/src/data-service.ts
```diff
@@ -1996,6 +2056,63 @@ export class DataServiceImpl extends EventEmitter implements DataService {
     return !!this._metadataClient;
   }
 
+  isOperationCancelledError(error: Error): boolean {
+    return OperationCancelledError.isOperationCancelledError(error);
+  }
+
+  private async cancellableOperation<T>(
```
addaleax, line 2063: I can’t really find anything on *why* we are doing this in the tech design/ticket/PR description. What makes this better than having the cancellation feature implemented outside of data-service? Are there follow-up tickets to migrate the other cancelable-query features to this API as well?

My understanding is that it can be implemented on top of the data-service API, and so even if we have the desire for a shared implementation, it wouldn’t be something that needs to be part of the `DataService` interface itself and could be turned into a separate helper (which would also have the advantage that other providers of the `DataService` interface wouldn’t need to copy this code here).

PR #3125, opened 2022-05-23
test(connection-storage): test csfle secrets storage COMPASS-5839

mcasimir:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/data-service/src/connection-storage.spec.ts
+++ b/packages/data-service/src/connection-storage.spec.ts
@@ -284,6 +284,96 @@ describe('ConnectionStorage', function () {
 
       expect(error.message).to.be.equal('id must be a uuid');
     });
+
+    it('does not save fle secrets if fleOptions.storeCredentials is false', async function () {
+      const connectionStorage = new ConnectionStorage();
+      const id = uuid();
+      await connectionStorage.save({
+        id,
+        connectionOptions: {
+          connectionString: 'mongodb://localhost:27017',
+          fleOptions: {
+            storeCredentials: false,
+            autoEncryption: {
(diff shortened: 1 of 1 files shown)


addaleax (approved):
Fwiw, we already have unit tests for the extract-secrets part of this.

Are we skipping a full e2e test for this? I’m okay with that, just checking

mcasimir:
> Fwiw, we already have unit tests for the extract-secrets part of this.
> 
> Are we skipping a full e2e test for this? I’m okay with that, just checking

Yeah, that's a good point... I don't know, that's what I thought since we still need to do this: https://jira.mongodb.org/browse/COMPASS-5393 for the e2e tests, I could spend some time trying to add that test suite.

I ended up adding those tests there cause we actually didn't have anything that really tests that secrets are extracted when a connection is saved, basically now we have:
- a test that ensures that checking the "Store Credentials" input sets `storeCredentials` accordingly in `ConnectionInfo
- a test that ensures that credentials are stored according to `storeCredentials` when `ConnectionInfo` is saved

which I thought is a good enough safeguard for now, but it would be nice to also have https://jira.mongodb.org/browse/COMPASS-5393 at some point and make sure that all the secrets are moved in the keytar

addaleax:
@mcasimir Fwiw, what I imagined here would be some kind of test that connects Compass with KMS configured but with “Store Credentials” off, then disconnects and checks that the connection form contains the non-credential KMS configuration but not the credentials.

PR #3128, opened 2022-05-23
chore(compass-query-bar): Add query bar option inputs, wire to store COMPASS-5673

Anemy:
Part of COMPASS-5673 - there will be a following pr updating the query option editors.

To use the new toolbar with the query bar updates:
```bash
```

This PR adds the query options to the new query bar. We'll be updating the query options themselves to typescript in an upcoming pr, this one uses the old ones. It uses the previous interface. No api changes on the outside, so this won't impact cloud, except on the styling side of things.

Because we're still using the old interface there are a few `any` I added when passing to the old interface. In the upcoming pr we'll have these types fully defined and remove them. I'll leave a comment in the code where this happens.

https://user-images.githubusercontent.com/1791149/169877489-2cb296f5-6d6d-47eb-920e-3dae9770310e.mp4

### DIFF
--- a/packages/compass-query-bar/src/components/legacy-query-bar/query-bar.jsx
+++ b/packages/compass-query-bar/src/components/legacy-query-bar/query-bar.jsx
@@ -62,7 +62,6 @@ class QueryBar extends Component {
   static displayName = 'QueryBar';
 
   static propTypes = {
-    store: PropTypes.object.isRequired,
     filter: PropTypes.object,
     project: PropTypes.object,
     sort: PropTypes.object,
@@ -190,7 +189,7 @@ class QueryBar extends Component {
         hasError={hasError}
         key={`query-option-${id}`}
         value={value}
-        actions={this.props.actions}
+        refreshEditorAction={this.props.actions.refreshEditor}
         placeholder={placeholder}
         link={OPTION_DEFINITION[option].link}
         inputType={OPTION_DEFINITION[option].type}
--- a/packages/compass-query-bar/src/components/legacy-query-bar/query-bar.jsx
+++ b/packages/compass-query-bar/src/components/legacy-query-bar/query-bar.jsx
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { Dropdown, MenuItem } from 'react-bootstrap';
-import { isFunction, pick, isEqual, isString, isArray, map } from 'lodash';
+import { pick, isEqual, isString, isArray, map } from 'lodash';
 import FontAwesome from 'react-fontawesome';
 
 import QueryOption from '../query-option';
@@ -79,7 +79,6 @@ class QueryBar extends Component {
     skipValid: PropTypes.bool,
     limitValid: PropTypes.bool,
 
-    featureFlag: PropTypes.bool,
     autoPopulated: PropTypes.bool,
     filterString: PropTypes.string,
     projectString: PropTypes.string,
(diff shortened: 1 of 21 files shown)


packages/compass-query-bar/src/components/legacy-query-bar/query-bar.jsx
```diff
@@ -243,13 +228,11 @@ class QueryBar extends Component {
    * @return {Component}          the option component
    */
   renderOption(option, id, hasToggle) {
-    const { filterValid, featureFlag, autoPopulated } = this.props;
+    const { filterValid, autoPopulated } = this.props;
```
Anemy, line 231: `featureFlag` doesn't appear to be used anywhere, so I removed it.

packages/compass-query-bar/src/components/query-bar/query-bar.tsx
```diff
@@ -49,74 +60,218 @@ const openQueryHistoryStyles = cx(
...
+  serverVersion,
   showQueryHistoryButton = true,
   toggleExpandQueryOptions,
-  toggleQueryHistory,
+  toggleQueryHistory: _toggleQueryHistory,
+  valid: isQueryValid,
+
+  // This form uses extra props from the store for placeholders and
+  // input values for the query options. `filterValue`, `sortValue`, etc.
+  ...props
```
Anemy, line 112: In the next pr we'll type out all of the various query option props:
`filterValue`, `filterPlaceholder`, `sortValue`, etc. For this reason there are 3 `any` in this file. I could do it in this pr  if preferred.

packages/compass-query-bar/src/components/query-bar/query-bar.tsx
```diff
@@ -49,74 +60,218 @@ const openQueryHistoryStyles = cx(
...
+    [_toggleQueryHistory]
+  );
+
+  const onApply = useCallback(() => {
+    if (isQueryValid) {
+      _onApply();
+    }
+  }, [_onApply, isQueryValid]);
+
+  const renderQueryOption = useCallback(
```
Anemy, line 140: In the follow up pr we'll make this all part of the `QueryOption` component interface so we won't need this function declaration happening here.

packages/compass-query-bar/src/constants/query-option-definition.ts
```diff
@@ -0,0 +1,56 @@
...
+    type: 'numeric',
+    placeholder: '0',
+    link: 'https://docs.mongodb.com/manual/reference/method/cursor.limit/',
+  },
+  maxTimeMS: {
+    label: 'Max Time MS',
+    type: 'numeric',
+    placeholder: '60000',
+    link: 'https://docs.mongodb.com/manual/reference/method/cursor.maxTimeMS/',
+  },
```
Anemy, line 53: Removed `sample` as it doesn't appear to be used anywhere (including cloud).

packages/compass-query-bar/src/plugin.jsx
```diff
@@ -1,38 +1,66 @@
...
+    }
+  }, [_onApply, actions]);
+
+  const onReset = useCallback(() => {
+    actions.reset();
+
+    if (isFunction(_onReset)) {
+      _onReset();
+    }
+  }, [_onReset, actions]);
```
Anemy, line 32: Moved this function and `onApply` to here from https://github.com/mongodb-js/compass/pull/3128/files#diff-c590b2e9bf139a99b4be59d7e761c1ad61bbe4080dfff45af0dec1511195725eL159 so that the components don't need the higher level prop logic.

Anemy:
Thinking on the more and starting on the `QueryOption` updates I think it will probably make sense to have it all in one pr. Going to wip this one and maybe close and reopen a new one tmrw.

PR #3134, opened 2022-05-25
test: create FLE2 collection and check badge COMPASS-5850 COMPASS-5851

alenakhineika:
## Description

- [COMPASS-5850](https://jira.mongodb.org/browse/COMPASS-5850) Verify that creating a collection with FLE2 options works
- [COMPASS-5851](https://jira.mongodb.org/browse/COMPASS-5851) Verify that the collection is shown with an FLE2 badge

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-e2e-tests/helpers/commands/set-connect-form-state.ts
```diff
@@ -137,37 +137,28 @@ export async function setConnectFormState(
...
+    if (state.fleKeyVaultNamespace) {
+      await browser.setValueVisible(
+        Selectors.ConnectionFormInputFLEKeyVaultNamespace,
+        state.fleKeyVaultNamespace
+      );
+    }
+    if (state.fleKey) {
+      await browser.expandAccordion(Selectors.ConnectionFormInputFLELocalKMS);
+      await browser.setValueVisible(
+        Selectors.ConnectionFormInputFLEKey,
```
addaleax, line 151: Just a side note, but I feel like we should namespace selectors for KMS options based on the KMS name; e.g. this should probably be 

```suggestion
        Selectors.ConnectionFormInputFLELocalKey,
```

and the `data-testid` should be `csfle-kms-local-key`, not `csfle-kms-key`

addaleax (approved):

packages/compass-e2e-tests/tests/in-use-encryption.test.ts
```diff
@@ -0,0 +1,164 @@
...
+  before(async function () {
+    if (
+      semver.lt(MONGODB_VERSION, '6.0.0-rc0') ||
+      process.env.MONGODB_USE_ENTERPRISE !== 'yes'
+    ) {
+      return this.skip();
+    }
+
+    initialEnvVars = Object.assign({}, process.env);
+    process.env.COMPASS_CSFLE_SUPPORT = 'true';
```
mcasimir, line 22: I'm not sure about this .. I have another comment where we set `COMPASS_CSFLE_SUPPORT` in the connection form tests. We decided not to enable FF tests in CI, maybe we need a way to run them in some instances (we should be able to do that with `evergreen patch`). 

The idea is that we need the CI to test the real state of the code that we would be releasing, so to give us the confidence nothing is broken.

For this PR if you enable the FF right after, I think we can keep it, but bear in mind for next time that normally we don't want to enable FF on CI by default.

packages/connection-form/package.json
```diff
@@ -38,9 +38,9 @@
     "depcheck": "depcheck",
     "check": "npm run lint && npm run depcheck",
     "check-ci": "npm run check",
-    "test": "mocha",
-    "test-electron": "xvfb-maybe electron-mocha --no-sandbox",
-    "test-cov": "nyc -x \"**/*.spec.*\" --reporter=lcov --reporter=text --reporter=html npm run test",
+    "test": "cross-env COMPASS_CSFLE_SUPPORT='true' mocha",
```
mcasimir, line 41: The way we normally work with FF locally is this:

```
export COMPASS_CSFLE_SUPPORT='true'

npm run any-kind-of-test
```

This is intentional, we want the CI to always test what's released not what will be released when the FF is enabled.

mcasimir (approved):
Nice!

alenakhineika:
evergreen retry

PR #3150, opened 2022-05-31
fix(saved-pipelines): open pipeline in correct namespace COMPASS-5871

mabaasit:
fixes opening of saved pipeline in a different namespace.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-saved-aggregations-queries/src/stores/open-item.ts
+++ b/packages/compass-saved-aggregations-queries/src/stores/open-item.ts
@@ -3,6 +3,8 @@ import type { ThunkAction } from 'redux-thunk';
 import type { RootState } from '.';
 import type { Item } from './aggregations-queries-items';
 import { createLoggerAndTelemetry } from '@mongodb-js/compass-logging';
+import type { Query } from '@mongodb-js/compass-query-history';
+import type { Aggregation } from '@mongodb-js/compass-aggregations';
 
 const { track } = createLoggerAndTelemetry('COMPASS-MY-QUERIES-UI');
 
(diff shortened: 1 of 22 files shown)


packages/compass-saved-aggregations-queries/src/stores/open-item.ts
```diff
@@ -237,11 +237,19 @@ const openItem =
...
-      aggregation: item.type === 'aggregation' ? item.aggregation : null,
-      query: item.type === 'query' ? item.query : null,
-    });
+      [item.type]: (item.type === 'aggregation')
+        ? {
+          ...item.aggregation,
+          namespace: metadata.namespace,
+        } : {
+          ...item.query,
+          _ns: metadata.namespace,
```
Anemy, line 248: Where is this `_ns` used? It looks like the main consumer of this `open-namespace-in-new-tab` looks explicitly for `namespace` and disregards `_ns`, looking at https://github.com/mongodb-js/compass/blob/29964950b387b3b68d605a798040bba90a219c24/packages/compass-collection/src/stores/index.ts#L170
Do the `aggregation` and `query` items open their contexts differently? I'm wondering if we should also be putting `namespace: metadata.namespace` here.

Anemy (approved):
lgtm!

packages/compass-saved-aggregations-queries/src/stores/open-item.ts
mabaasit: Yeah totally!

PR #3151, opened 2022-05-31
feat(compass-query-bar): update query options and layout COMPASS-5673

Anemy:
COMPASS-5673

To use the new toolbar with the query bar updates:
```bash
```

This PR adds the query options to the new query bar. No api changes on the outside, so this won't impact cloud, except on the styling side of things. This is all still behind a feature flag, the current query bar looks and works the same.

There are a few `any` usages around javascript modules. I'd like these to go away in future work, we'll have to find a way to expose some Ace editor typings or recreate them in this package.

\- **Outdated** - Here's how it looks:

https://user-images.githubusercontent.com/1791149/171466723-590ee1e3-f22e-4c91-8df5-632e49c89713.mp4


packages/compass-components/src/components/more-options-toggle.tsx
```diff
@@ -57,6 +57,13 @@ export const MoreOptionsToggle: React.FunctionComponent<MoreOptionsToggleProps>
       // We cast here so that the `as` prop of link can be properly typed.
     ) as Partial<React.ComponentType<React.ComponentProps<typeof Link>>>;
 
+    const onClick = useCallback((evt: React.MouseEvent) => {
+      // Don't submit forms.
+      evt.preventDefault();
+
+      onToggleOptions();
+    }, []);
```
gribnoysup, line 65: This button should have `type=button` instead

packages/compass-components/src/hooks/use-focus-ring.ts
```diff
@@ -16,7 +16,7 @@ export const focusRingStyles = css({
     bottom: 3,
     left: 3,
     borderRadius: spacing[1],
-    boxShadow: `0 0 0 0 ${uiColors.focus}`,
+    boxShadow: '0 0 0 0 transparent',
```
gribnoysup, line 19: Why are we removing the focus outline color?

packages/compass-components/src/components/editor.tsx
```diff
@@ -75,10 +77,21 @@ function Editor({
...
 
+  const editorRef = useRef<AceEditor | null>(null);
+
+  useEffect(() => {
+    if (id && editorRef.current) {
+      // After initial load, assign the id to the text area used by ace.
+      // This is so labels can `htmlFor` the input.
+      editorRef.current.editor.textInput.getElement().id = id;
+    }
+  }, [id]);
```
gribnoysup, line 88: What is the `editorId` that we have below that for then? Do we need two? Should it be one or the other?

packages/compass-components/src/components/editor.tsx
```diff
@@ -75,10 +77,21 @@ function Editor({
     ...(!!completer && { enableLiveAutocompletion: true }),
   };
 
+  const editorRef = useRef<AceEditor | null>(null);
+
+  useEffect(() => {
```
gribnoysup, line 82: This probably can be a layout effect so that the id is assigned before anything is even flushed to the dom

packages/compass-query-bar/src/components/option-editor/option-editor.tsx
```diff
@@ -0,0 +1,183 @@
...
+    });
+  }
+
+  /**
+   * @param {Object} nextProps - The next properties.
+   *
+   * @returns {Boolean} If the component should update.
+   */
+  shouldComponentUpdate(nextProps: OptionEditorProps): boolean {
+    this.boundOnFieldsChanged(nextProps.schemaFields);
```
gribnoysup, line 115: `shouldComponentUpdate` is really not a great place for side-effects. What's exactly is happening here? Why are we always updating completer but not allowing component to update?

packages/compass-query-bar/src/components/query-bar/query-bar.tsx
```diff
@@ -49,74 +61,260 @@ const openQueryHistoryStyles = cx(
...
+
+  const additionalQueryOptions = useMemo(
+    () =>
+      layout.map((queryOption: string | string[], index: number) =>
+        isQueryOptionsExpanded && index > 0
+          ? renderQueryOptionRow(queryOption, index)
+          : null
+      ),
+    [isQueryOptionsExpanded, layout, renderQueryOptionRow]
+  );
```
gribnoysup, line 256: To make it more readable, would it make sense to just split the layout in two first and then just render it without using memo excessively? Something like:

```jsx
const [firstRow, ....rows] = layout;

return <>{renderRow(firstRow)}{isQueryOptionsExpanded && rows.map(row => renderRow(row))}</>
```

packages/compass-query-bar/src/components/query-bar/query-bar.tsx
```diff
@@ -49,74 +61,260 @@ const openQueryHistoryStyles = cx(
...
           >
-            Query
-          </Label>
+            Learn more
+          </Link>
+        )}
+      </div>
+    ),
+    [renderQueryOption]
+  );
```
gribnoysup, line 238: Maybe would make sense to actually make them into components instead of memoised functions inside render method? This is getting pretty hard to decypher when reading through the code

packages/compass-query-bar/src/components/query-option/query-option.tsx
```diff
@@ -0,0 +1,156 @@
...
+      className={cx(
+        queryOptionStyles,
+        isAutoCompleteInput ? autocompleteOptionStyles : numericOptionStyles
+      )}
+      data-test-id="query-bar-option"
+    >
+      {queryOption !== 'filter' && (
+        <div
+          className={queryOptionLabelContainerStyles}
+          data-test-id="query-bar-option-label"
```
gribnoysup, line 107: Are we keeping it test-id and not testid for legacy compat reasons?

gribnoysup (commented):
For some reason when I run it locally all the input placeholders are not aligned with labels (this is after doing a git clean and fresh bootstrap)

image

packages/compass-query-bar/src/components/query-bar/query-bar.tsx
```diff
@@ -49,74 +61,260 @@ const openQueryHistoryStyles = cx(
...
+
+        <MoreOptionsToggle
+          aria-controls="additional-query-options-container"
+          data-testid="query-bar-options-toggle"
+          isExpanded={isQueryOptionsExpanded}
+          onToggleOptions={toggleExpandQueryOptions}
+        />
+      </div>
+      <div id="additional-query-options-container">
+        {additionalQueryOptions}
```
gribnoysup, line 316: If this was a grid and not flex you'd be able to simplify the layout and styles a bunch and this would allow so avoid this annoying misalignment between the row cells

image

packages/compass-query-bar/src/components/query-option/query-option.tsx
```diff
@@ -0,0 +1,156 @@
...
+              hasError && optionInputWithErrorStyles
+            )}
+            type="text"
+            state={hasError ? 'error' : 'none'}
+            value={`${value}`}
+            onChange={(evt: React.ChangeEvent<HTMLInputElement>) =>
+              onChange(evt.target.value)
+            }
+            placeholder={placeholder}
+          />
```
gribnoysup, line 149: Are these so big by design?

image

packages/compass-query-bar/src/components/query-bar/query-bar.tsx
```diff
@@ -49,74 +61,260 @@ const openQueryHistoryStyles = cx(
...
 export const QueryBar: React.FunctionComponent<QueryBarProps> = ({
+  autoPopulated,
   buttonLabel = 'Apply',
   expanded: isQueryOptionsExpanded = false,
-  valid: isQueryValid,
+  layout = [
+    'filter',
+    ['project', 'sort'],
+    ['collation', 'skip', 'limit', 'maxTimeMS'],
+  ],
```
gribnoysup, line 140: Is there ever a case where `filter` is not the first and only option in the first row? Seems like we are complicating both interface and implementation (both component logic and styles) to support that, but it doesn't look like this is really needed?

packages/compass-query-bar/src/components/option-editor/option-editor.tsx
Anemy: Thanks for bringing it up, I had copied this over without going too deep into why we were doing it this way.
From my understanding it's a bit of a performance boost mixed with a workaround for non-displayed ace editor behavior. I think we can do better here. What previously would happen is the editor would be initialized and then its value would only be controlled when the `refreshEditor` action occurred. This action happened when the `subtab-changed` global app registry event occurs here: https://github.com/mongodb-js/compass/blob/d78f63c28ed5ff516ce6453c44245f5e020c7b70/packages/compass-query-bar/src/stores/query-bar-store.js#L760
That's emitted when a user is changing from the Schema tab to the Documents tab for instance. When that happened we would then set the editor's value to the new value: https://github.com/mongodb-js/compass/blob/d78f63c28ed5ff516ce6453c44245f5e020c7b70/packages/compass-query-bar/src/components/option-editor/option-editor.jsx#L54
We were doing this because when the ace editor is not visible (not the active subtab in Compass), [ace does not update the view](https://github.com/ajaxorg/ace/issues/2497). So we would perform this `value` setting after the subtab was changed to ensure we showed the new value after the editor's container has become visible (I'm thinking this could lead to race conditions).

I'm investigating how we can ensure the view is updated when the view is visible without unrelated global event chain listening. It may be a bit out of the scope of this pr (in which case would it be alright to keep the existing implementation? With added comments)

packages/compass-components/src/components/editor.tsx
Anemy: Ah thanks for pointing out, I'll update the `editorId` variable to be `editorName` as it's not used for the textarea id, and is passed to ReactAce as the `name`.

packages/compass-query-bar/src/components/option-editor/option-editor.tsx
Anemy: Looks like we won't need the `autoPopulated` flag if we do this differently also.

packages/compass-components/src/hooks/use-focus-ring.ts
Anemy: Moved to the `option-editor` so we aren't impacting everything. We want the hover color to not transition from the focus color, setting it to transparent makes the hover not flash the focus color.

packages/compass-components/src/components/editor.tsx
Anemy: Good call, updated

packages/compass-query-bar/src/components/option-editor/option-editor.tsx
Anemy: Pushed an update with a comment on the behavior and why we do it. Cleaned up this component a bit as well. I think other solutions which would allow us to remove the global event listening would be out of the scope of this work, interested to see how you feel on it though.

packages/compass-query-bar/src/components/query-bar/query-bar.tsx
Anemy: Updated - added a `QueryBarRow` component let me know how it looks now. I think it will be a good bit more readable. I was definitely going a bit crazy with useMemo here

packages/compass-query-bar/src/components/query-option/query-option.tsx
Anemy: Updated to testid 👍 . There are a few more test-ids in the legacy query bar used in e2e tests still. I'm thinking we will update those when we remove the feature flag and update tests, can do it in this pr if preferred.

Anemy:
@gribnoysup merged in main which had the offsets for the input areas from https://github.com/mongodb-js/compass/pull/3168 (separated from this pr since it had more unrelated changes). Sorry for the confusion!

packages/compass-query-bar/src/components/query-option/query-option.tsx
Anemy: Pinged Claudia on this, currently they're the default size variant from leafygreen.
I'm thinking `small` looks a bit better (before/after):
Screen Shot 2022-06-08 at 5 40 33 PM
Screen Shot 2022-06-08 at 5 47 40 PM
Pushed up the smaller one for now, might change based on feedback.

packages/compass-query-bar/src/components/query-bar/query-bar.tsx
Anemy: Just pushed an update to make the styles on grid. I couldn't find a way to do grid without introducing some extra complexity, let me know if you can think of some ways to simplify the styles. 
The following constraints were making it a bit tricky:
- Having the learn more link under the less/more options button when possible (larger screen sizes)
- Having the react ace items take up 50% of width
- Having a variable amount of layout items take up the rest of the width while still having their inputs usable (used breakpoints for this one).

I tried out how it looks with large/small screen sizes and the various `layout` configurations cloud uses. Here are the screenshots:
Screen Shot 2022-06-08 at 7 38 14 PM
Screen Shot 2022-06-08 at 7 38 33 PM
Screen Shot 2022-06-08 at 7 40 16 PM
Screen Shot 2022-06-08 at 7 40 41 PM
Screen Shot 2022-06-08 at 7 40 48 PM

packages/compass-query-bar/src/components/query-bar/query-bar.tsx
gribnoysup: Yeah, I checked mms layouts and there `filter` is always first and single row, do we really have the use-case for having it so flexible?

If you don't want to do this in this scope I won't try to convince you otherwise, FWIW if the goal is to "simplify", moving existing code to typescript without re-evaluating the purpose of it is just covering up possible existing issues

packages/compass-query-bar/src/components/query-bar/query-bar.tsx
gribnoysup: I think you should be able to dynamically build a template area (using `grid-template-areas` css property) based on provided layout to cover most of these cases instead of using template columns, also I think the way you have it implemented right now doesn't cover the requirements you specify (i.e. if I add `skip` as the first item in second row it will take 50% of the space and `project`, `sort` will be 25% because you always have 50% as the first column width) and the first row when has more than a few items is just not rendering correctly due to the space constraints, some items just fall to the next row, not sure if we even want the first row to be a grid here.

Worth nothing that some of the requirements that you mention will just not work with the interface that we allow, nothing stopping me from doing a row like `['filter', 'project', 'sort']` that would make it impossible to have 50% width allocated for ace editor inputs.

I'll once again will say that this `layout` prop interface doesn't make sense to me and if we keep it I guarantee you will have troubles building a robust layout here (there is already cases that don't work correctly). But as you consider changing this out of scope and you have a feeling that grid brings more complexity here feel free to revert it back to flex

packages/compass-query-bar/src/components/query-option/query-option.tsx
Anemy: Claudia likes the smaller too! Let's keep it - do you like it?

gribnoysup: Yeah, does look better to me too 👍

packages/compass-query-bar/src/components/query-bar/query-bar.tsx
Anemy: I spent a few hours today making the query bar options build a dynamic `grid-template-areas` property and I feel it adds too much complexity to cover all of the possible cases (missing one query option for instance). I think we should keep the current interface for now, that way cloud doesn’t have to change their interface, and when we want to do larger changes here we should simplify.

For the basic case we create a `grid-template-areas` like so:
'project project project sort sort sort'
'collation collation skip limit maxTimeMS docs'
When something is missing, or we want to only display one line, everything has to shift or be accounted for.

The current implementation with the overly generic `layout` prop has limitations around what layouts can be passed to it, like you mentioned, passing a row like `[‘filter’, ‘project’, ‘sort’]` will definitely break it and look weird, but that isn’t a current or intended use case. I think that would be apparent in how it breaks when using it so if that’s a use down the line we can accommodate it, and use the information to make a better generalized interface.

This isn’t the best solution, but I also don’t think we should spend the time here to do it correctly to cover use cases we don’t use.

Anemy: Updated to use `grid-template-areas`, built with the query options passed to the QueryBar (replaced the `layout` prop with `queryOptions`). How does the implementation look? Before merging this I'll ping cloud folks on the interface change.

packages/compass-query-bar/src/components/query-options-grid.tsx
```diff
@@ -0,0 +1,150 @@
...
+  paddingRight: spacing[2],
+  minWidth: 'max-content', // Don't let the link collapse to two lines.
+  textAlign: 'right',
+});
+
+const queryDocsLinkStyles = css({
+  flexShrink: 0,
+});
+
+export function getGridTemplateForQueryOptions(
```
gribnoysup, line 34: This is super awesome and absolutely step in the right direction in my opinion, great work!

I totally appreciate that building this layout dynamically with the requirements we have is no easy task, and so as there were a few of these corner cases I highlighted in my review I spent some time today expanding on this work you did and trying to see if there is a universal approach here that we can take that will cover the rest of the stuff that is falling through the cracks a little at the moment.

I first sketched all possible options that we have taking into account requirement you mentioned before and after juggling elements around came to some conclusions including following in no particular order:

- The colums size of 6 that you chose is an almost* perfect match for the content we have here
  - \* The only thing that is hard to fit in it while giving them enough space is numeric inputs, especially because maxTimeMs label is so much longer than other ones, they feel like a hard fit for grid and seem more like a flex material
- Whatever ace input is on one line with numeric ones is very hard to fit so it might need some special logic to cover it
- If we want inputs to have more or less predictable and usable sizes, grid should have strictly sized columns
- It might be that the only way to make "Learn more" link not take too much space and be aligned with the end of the content is to have it as part of some other grid cell

Based on that I thought that maybe we should try a combined approach and use grid for the main layout of the inputs, but make numerical inputs a single cell with a flex layout inside. This allows ace editors to have strict minimal size while allowing some auto flex sizing to numerical inputs inside their cell.

The final sketch with all the options that would need to be covered by the algorithm looks like this:

Untitled-2021-04-27-1028(1)

After sketching there were some very noticeable patterns here that helped to formulate the algorithm that would cover all the cases we can encounter

- Link can be part of numeric inputs group, as numeric input group is always last cell in the grid and it's a perfect place to align link with the end of the grid.
- If there are no ace editors, give a full row to numeric inputs group
- If there are no numeric inputs, stick the group (that will contain just the link) at the very end of the very last row
- If both are present distribute the space evenly if the row is just ace editors, otherwise give numeric inputs priority if there are a lot of them or ace editor if there are not that many numeric ones

I pulled your branch locally and sketched some code to try out how it works and this is how the code for this might look like:

```ts
const COLUMNS = 6;

export function getGridTemplateForQueryOptions(
  queryOptions: QueryOption[]
): string {
  const documentEditors = Object.values(OPTION_DEFINITION).filter((opt) => {
    return (
      opt.name !== 'filter' &&
      queryOptions.includes(opt.name) &&
      opt.type === 'document'
    );
  });

  const numericEditors = Object.values(OPTION_DEFINITION).filter((opt) => {
    return (
      opt.name !== 'filter' &&
      queryOptions.includes(opt.name) &&
      opt.type === 'numeric'
    );
  });

  const cells = (area: string, n: number) => {
    return Array(n).fill(area).join(' ');
  };

  const toTemplateArea = (rows: string[]) => {
    return rows
      .filter(Boolean)
      .map((row) => `'${row}'`)
      .join('\n');
  };

  // If there are no document editors, just make numeric options and link take
  // all the space of one row
  if (documentEditors.length === 0) {
    return toTemplateArea([cells('rest', COLUMNS)]);
  }

  // When there are no numeric options, we just want link to be last cell in
  // last row
  if (numericEditors.length === 0) {
    const lastDocumentEditor = documentEditors.pop();
    const linkRow = `${cells(lastDocumentEditor!.name, COLUMNS - 1)} rest`;
    const rows = [
      documentEditors
        // First row will have either 1 or 2 cells evenly distributed in 6
        // available columns
        .map((option) => {
          return cells(option.name, COLUMNS / documentEditors.length);
        })
        .join(' '),
      linkRow,
    ];
    return toTemplateArea(rows);
  }

  // For every other case we will do some special handling where numeric inputs
  // will take 3 to 4 cells based on the amount of inputs provided and will
  // always be on the end of the second row and document inputs will adjust
  // around that
  // const lastDocumentEditor = documentEditors.pop();
  const numericCellsLength = numericEditors.length > 2 ? 4 : 3;
  const freeCellsLength = COLUMNS - numericCellsLength;
  const numericCells = cells('rest', numericCellsLength);

  // If there is more than two editors in total, move last one to the next row
  if (documentEditors.length > 2) {
    const lastDocumentEditor = documentEditors.pop();
    const rows = [
      documentEditors
        .map((option) => {
          return cells(option.name, COLUMNS / documentEditors.length);
        })
        .join(' '),
      `${cells(lastDocumentEditor!.name, freeCellsLength)} ${numericCells}`,
    ];
    return toTemplateArea(rows);
  }

  // Otherwise keep numeric cells on the last row separately from document
  // editors
  const rows = [
    documentEditors
      .map((option) => {
        return cells(option.name, COLUMNS / documentEditors.length);
      })
      .join(' '),
    `${cells('.', freeCellsLength)} ${numericCells}`,
  ];
  return toTemplateArea(rows);
}
```

In some cases it's pretty similar to yours with the main difference being that numeric inputs case is handled more thoroughly though this single area that they are handled through.

This of course requires some slight adjustment to the grid rendering, but it's really not more than rendering numeric inputs separately and wrapping them in a single cell with link inside.

And this is how this implementation compares in some tricky cases:

|Before|After|
|---|---|
|localhost_4242_index html|localhost_4242_index html|
|image|image|
|image|image|

packages/compass-query-bar/src/components/query-options-grid.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const gridTemplateAreas = useMemo(
+      () => getGridTemplateForQueryOptions(queryOptions),
+      [queryOptions]
+    );
+
+    return (
+      <div
+        className={gridStyles}
+        style={{
+          gridTemplateAreas,
```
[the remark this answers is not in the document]
Anemy: It's intended, without collation shrinking the `maxTimeMS` field does not have enough input room. We've seen collation not getting much usage, so I think it's a good compromise.

packages/compass-query-bar/src/constants/query-option-definition.ts
```diff
@@ -0,0 +1,87 @@
...
+  skipString: string;
+
+  limitPlaceholder?: string;
+  limitValid: boolean;
+  limitString: string;
+
+  maxTimeMSPlaceholder?: string;
+  maxTimeMSValid: boolean;
+  maxTimeMSString: string;
+};
```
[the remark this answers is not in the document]
Anemy: YESSSS

packages/compass-query-bar/src/components/query-options-grid.tsx
```diff
@@ -0,0 +1,227 @@
...
+            return (
+              opt.name !== 'filter' &&
+              queryOptions.includes(opt.name) &&
+              opt.type === 'document'
+            );
+          })
+          .map((optionDefinition) => optionDefinition.name),
+      [queryOptions]
+    );
+    const numericEditors = useMemo(
```
Anemy, line 161: We compute the numeric & document editors twice, here and in `getGridTemplateForQueryOptions`. I'm thinking I prefer keeping the argument interface of `getGridTemplateForQueryOptions` with only the query options array instead of passing these in for the slight reduction in duplication. I'm up to update it to reduce this duplication if folks prefer it.

packages/compass-query-bar/src/components/query-option.tsx
```diff
@@ -0,0 +1,146 @@
...
+    display: 'flex',
+    flexGrow: 1,
+  },
+  focusRingStyles
+);
+
+const numericTextInputStyles = css({
+  input: {
+    borderColor: 'transparent',
+  },
```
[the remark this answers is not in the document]
Anemy: I'll bring it up with Claudia, I also feel we might be losing something by not including borders in this component. Thanks for raising.

packages/compass-query-bar/src/components/query-option.tsx
```diff
@@ -0,0 +1,146 @@
...
+            onChange={onChange}
+            placeholder={placeholder}
+            refreshEditorAction={refreshEditorAction}
+            schemaFields={schemaFields}
+            serverVersion={serverVersion}
+            value={value}
+          />
+        ) : (
+          <TextInput
+            aria-labelledby={`query-bar-option-input-${queryOption}-label`}
```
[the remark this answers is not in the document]
Anemy: We'll need to keep `aria-labelledby` for now as it's part of the leafygreen type definition: https://github.com/mongodb/leafygreen-ui/tree/main/packages/text-input#special-case-aria-labels
Updated the `Filter` label to be visually hidden but in the dom for the screen readers.

packages/compass-query-bar/src/components/option-editor.tsx
```diff
@@ -0,0 +1,157 @@
...
+        bindKey: {
+          win: 'Enter',
+          mac: 'Enter',
+        },
+        exec: () => {
+          onApply();
+        },
+      });
+    },
+    [onApply]
```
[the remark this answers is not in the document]
Anemy: Updated, haven't really done this pattern before, is what I implemented what you had in mind? Thanks for noticing!

packages/compass-query-bar/src/components/query-options-grid.tsx
```diff
@@ -0,0 +1,236 @@
...
+const gridStyles = css({
+  alignItems: 'center',
+  display: 'grid',
+  flexGrow: 1,
+  position: 'relative',
+  margin: `0 ${spacing[1]}px`,
+  marginTop: spacing[1],
+  padding: `0 ${spacing[2]}px`,
+  gap: `${spacing[1]}px ${spacing[2]}px`,
+});
```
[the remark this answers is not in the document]
Anemy: Added `gridTemplateColumns` to the grid styles 👍

Anemy:
Syncing with Claudia on input borders, and on positioning of the export to language (following discussions that export to language isn't very intuitive if it's not part of the query bar).

packages/compass-query-bar/src/components/query-option.tsx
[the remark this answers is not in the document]
gribnoysup: Oh wow, well then, I guess we don't really have a choice here 👍

packages/compass-query-bar/src/components/option-editor.tsx
```diff
@@ -0,0 +1,157 @@
...
+      unsubscribeRefreshEditorAction();
+    };
+  }, [refreshEditorAction]);
+
+  useLayoutEffect(() => {
+    completer.current.update(schemaFields);
+  }, [schemaFields]);
+
+  const onApplyClicked = useCallback(() => onApply(), [onApply]);
+  const onApplyRef = useRef(onApplyClicked);
```
gribnoysup, line 123: You are on the right track here, but this inherits the same issue as before. Even though the `onApplyClicked` will be updated, you are only passing it to `useRef` as initial argument and so it will not update after that. The solution here is even simpler in a way, but totally not the most intuitive thing ever:

```suggestion
  const onApplyRef = useRef(onApplyClicked);
  onApplyRef.current = onApplyClicked;
```

Basically what you want is for `current` to always hold the freshest possible reference to the callback function. This is actually kinda similar to what the new (not yet added IIRC) hook `useEvent` will be for in React, you can read up on a very detailed explanation of what sort of issues it solves [here](https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md) (and if you check the "internal implementation" you'll see that it will be similar to this, just moved to the special hook, we probably don't really have enough cases to abstract it somewhere, at least for now)

packages/compass-query-bar/src/components/option-editor.tsx
```diff
@@ -0,0 +1,157 @@
...
+      name: 'executeQuery',
+      bindKey: {
+        win: 'Enter',
+        mac: 'Enter',
+      },
+      exec: () => {
+        onApplyRef.current();
+      },
+    });
+  };
```
gribnoysup, line 138: Even though it won't make much difference due to how ace-editor component works, I think it does make sense to have it wrapped in `useCallback` as you had it before (with `[]` as deps) to express in the code that this method will only be created (and called) once

packages/compass-query-bar/src/components/query-bar.tsx
```diff
@@ -0,0 +1,214 @@
...
+          </IconButton>
+        )}
+
+        {queryOptions && queryOptions.length > 0 && (
+          <MoreOptionsToggle
+            aria-controls="additional-query-options-container"
+            data-testid="query-bar-options-toggle"
+            isExpanded={isQueryOptionsExpanded}
+            onToggleOptions={toggleExpandQueryOptions}
+          />
```
gribnoysup, line 194: This toggle is a button, but doesn't have a `type` and so causes a document re-fetch when you toggle it. We probably want it to always be of type button as part of internal component implementation

packages/compass-query-bar/src/components/query-option.tsx
```diff
@@ -0,0 +1,146 @@
...
+            id={`query-bar-option-input-${queryOption}-label`}
+            className={queryOptionLabelStyles}
+            // We hide the `Filter` label, but keep it in the dom for
+            // screen reader label support.
+            hidden={queryOption === 'filter'}
+          >
+            {queryOption}
+          </Label>
+        </div>
+      }
```
gribnoysup, line 107: Not really an issue, but these curlies here are not really needed (I'm guessing this is a leftover from a condition?)

gribnoysup (approved):
Except for [this button issue](https://github.com/mongodb-js/compass/pull/3151#discussion_r921195234) everything else LGTM, great work!

packages/compass-query-bar/src/components/query-bar.tsx
```diff
@@ -0,0 +1,214 @@
...
+            <QueryOptionsGrid
+              queryOptions={queryOptions}
+              queryOptionProps={queryOptionProps}
+              onChangeQueryOption={onChangeQueryOption}
+              onApply={onApply}
+              refreshEditorAction={refreshEditorAction}
+              schemaFields={schemaFields}
+              serverVersion={serverVersion}
+            />
+          )}
```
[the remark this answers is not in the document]
Anemy: I think the render lag on mount was significant enough that we didn't want to add them to the view before because it caused too much lag on showing the collection tabs initially. We could get fancier here and maybe render them incrementally or something, but I think we'd probably find better performance improvements in changing our base ace editor.
We aren't addressing that issue with this work, but definitely a good thing to bring up and keep in mind for future.

PR #3152, opened 2022-05-31
[do not merge] styles(ace-theme-query): update theme for new query option input styles COMPASS-5673

Anemy:
COMPASS-5673

This PR updates the query bar ace theme for the new styles. Because the theme is used in the current query bar we've separated this work into it's own pr so it can be merged later when we're ready for the old query bar to go away.


Anemy:
Replaced by https://github.com/mongodb-js/compass/pull/3168 which is probably a cleaner way to do this.

PR #3153, opened 2022-05-31
fix(document-view): isRoot fix COMPASS-5872

mabaasit:


packages/compass-crud/src/components/document.jsx
```diff
@@ -14,7 +14,10 @@ const Document = (props) => {
   } = props;
 
   const doc = useMemo(() => {
-    if (_doc?.isRoot?.()) {
+    // COMPASS-5872 If _doc is a plain js object, it can have isRoot prop,
+    // which would cause the isRoot() to throw an error. So an extended check.
+    const { isRoot } = _doc;
```
mcasimir, line 19: is it safe to destructure and call without bind?

packages/compass-crud/src/components/document.jsx
```diff
@@ -14,7 +14,10 @@ const Document = (props) => {
   } = props;
 
   const doc = useMemo(() => {
-    if (_doc?.isRoot?.()) {
+    // COMPASS-5872 If _doc is a plain js object, it can have isRoot prop,
+    // which would cause the isRoot() to throw an error. So an extended check.
```
mcasimir, line 18:
```suggestion
    // COMPASS-5872 If _doc is a plain js object rather than an instance of hadron-document Document
    // it will not have an isRoot prop, which would cause the isRoot() to throw an error.
```

mcasimir:
Do we know why we get a plain object? Is it ok to tolerate that case, or do we need to look into it more?

mabaasit:
> Do we know why we get a plain object? Is it ok to tolerate that case, or do we need to look into it more?

We did that change recently in #3073 to improve the performance of the Document View by using memoization.

addaleax (approved):

mcasimir (approved):

PR #3159, opened 2022-06-01
feat(explain-plan, query-history): use leafygreen code in explain and query history, remove global code styles COMPASS-5873

Anemy:
COMPASS-5873

This pr removes the global `pre` and `code` styles we had in Compass. In the places depending on those styles we're now using leafygreen's `Code` component https://www.mongodb.design/component/code/documentation/

Explain plan (before -> after) (we now get better scrolling and syntax highlighting).
Screen Shot 2022-06-01 at 7 42 38 PM

Query history (before -> after) (we get to drop `highlight.js` and (imo) get better syntax highlighting).
Screen Shot 2022-06-01 at 7 22 42 PM


mabaasit:
Looks cool! Just one note though: previously for explain we used smaller font size (11) and now its 13. not sure if it fits with other information there.

PR #3160, opened 2022-06-02
chore(deps): bump mongosh, devtools-connect, mongodb-client-encryption

addaleax:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [x] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/data-service/src/csfle-collection-tracker.spec.ts
```diff
@@ -33,8 +26,22 @@ describe('CSFLECollectionTracker', function () {
...
+      // This autoEncryption config depends on the keys, so we first need
+      // to set up the keys with one DataService instance,
+      // then create the 'real' DataService instance.
+      const [, tempDataService, someKey1, someKey2] = await createTracker({});
+      await tempDataService.disconnect();
+      const [tracker, dataService] = await createTracker(
+        autoEncryption([someKey1, someKey2])
+      );
+      return [tracker, dataService, someKey1, someKey2];
+    }
```
addaleax, line 43: The updates for this file were quite a bit larger than I anticipated, unfortunately.

Most of this is because libmongocrypt has become a bit smarter about collection management, including `createCollection` calls, and that messes with the expectations in this test.

packages/data-service/src/csfle-collection-tracker.spec.ts
```diff
@@ -360,7 +381,7 @@ describe('CSFLECollectionTracker', function () {
           dataService.findOneAndUpdate(
             `${dbName}.test2`,
             {},
-            { $set: { a: 2 } },
+            { $set: { a: '2' } },
```
addaleax, line 384: csfle library was, very understandably, complaining about the fact that the schema above says this should be a `string` :)

lerouxb (approved):

PR #3161, opened 2022-06-02
test: e2e copy clone encrypted fields COMPASS-5874, COMPASS-5885, COMPASS-5886

alenakhineika:
## Description

In the CRUD view:
- In a non-CSFLE background MongoClient, insert a copy of the document with the encrypted field, that contains a second field with the same encrypted value
- Verify that the second document can not be edited in the UI
- Verify that cloning the document with the second field displays a warning banner
- Verify that when disabling CSFLE through the sidebar, encrypted fields will be shown as redacted again

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


alenakhineika:
I wanted to include more test cases in this PR, but since we are in different time zones and I am ooo already, decided to let you review what is ready)

addaleax:
@alenakhineika Yeah, that's fine :) Feel free to leave this PR as it is and add new subtasks if you prefer though (or keep adding to this PR, I'm good either way 🙂)

packages/compass-e2e-tests/tests/in-use-encryption.test.ts
```diff
@@ -278,12 +291,12 @@ describe('FLE2', function () {
...
       const document = await browser.$(Selectors.DocumentListEntry);
       const value = await document.$(
-        `${Selectors.HadronDocumentElement}:nth-child(2) ${Selectors.HadronDocumentClickableValue}`
+        `${Selectors.HadronDocumentElement}[data-field="phoneNumber"] ${Selectors.HadronDocumentClickableValue}`
       );
       await value.doubleClick();
 
       const input = await document.$(
-        `${Selectors.HadronDocumentElement}:nth-child(2) ${Selectors.HadronDocumentValueEditor}`
+        `${Selectors.HadronDocumentElement}[data-field="phoneNumber"] ${Selectors.HadronDocumentValueEditor}`
```
addaleax, line 299: Yeah, this seems like a very good solution :)

packages/compass-e2e-tests/tests/in-use-encryption.test.ts
```diff
@@ -126,9 +135,11 @@ describe('FLE2', function () {
           '})'
       );
       await browser.clickVisible(Selectors.SidebarInstanceRefreshButton);
+      plainMongo = await MongoClient.connect(CONNECTION_STRING);
     });
 
     after(async function () {
+      await plainMongo.close();
```
addaleax, line 142: If the connection is established in `beforeEach`, this should probably run in an `afterEach` hook

packages/compass-e2e-tests/tests/in-use-encryption.test.ts
```diff
@@ -305,5 +318,230 @@ describe('FLE2', function () {
...
+
+      // wait for the modal to appear
+      const insertDialog = await browser.$(Selectors.InsertDialog);
+      await insertDialog.waitForDisplayed();
+
+      // set the text in the editor
+      await browser.setAceValue(
+        Selectors.InsertJSONEditor,
+        '{ "phoneNumber": "30303030", "name": "Copy" }'
+      );
```
addaleax, line 444: Just curious, why are we setting contents here? Is it to more easily identify the cloned object or is there more to it? I’m asking because it makes the `isClonedDocumentFaxNumberDecryptedIconExisting` check below a bit odd – the field isn’t there to begin with, so it won’t be displayed at all, right? Neither as encrypted nor plaintext?

addaleax (approved):

packages/compass-e2e-tests/tests/in-use-encryption.test.ts
alenakhineika: Yes, I did it to be able to find the cloned doc. How would you approach this? I was considering sorting the list, but don't know, went with this implementation. but agree that it look a bit odd.

addaleax: I don’ think we have a helper for getting the text in an ace editor, but maybe this could get the text, perform a replacement in it, and then set it? So that you only change the `name` property, not fully reset the input object?

Alternatively, it might still be better than the current state to also include the `faxNumber` property in the JSON we set here. It’s not quite the same, but it will still give more meaning to the `isClonedDocumentFaxNumberDecryptedIconExisting` check below. :)

PR #3162, opened 2022-06-02
chore(aggregations): explain tracking COMPASS-5875

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/modules/aggregation.ts
```diff
@@ -151,8 +151,14 @@ export const runAggregation = (): ThunkAction<
   void,
   Actions
 > => {
-  return (dispatch) => {
-    track('Aggregation Executed');
+  return (dispatch, getState) => {
+    const { pipeline } = getState();
+    track('Aggregation Executed', {
+      num_stages: pipeline
```
mcasimir, line 157: Question: .. what do we use `num_stages` for and why is getting the number of stages so involved here, while below is just `pipeline.length`?

Looks like this code may throw, would be unfortunate if customers can't run an aggregation cause of some telemetry .. we could use the function form of track for that right?

```
track('Aggregation Executed', () => ({ num_stages ... }))
```

mabaasit: Below we track after the explain is fetched (because we need number of indexes used in the event). Plus the `pipeline` is already filtered list.

Here, we have raw list of stages, so we map and filter it.

mabaasit: That way we will be tracking users pagination as well, which I think is not intended.

mcasimir: mmm I see, what other options do we have? I am assuming at some point we do this parsing already, wouldn't we know there if is paginated or not ie. `page === 1`? after all to paginate we need to know that info to alter the pipeline.

If we really need to do this here, what about we at least avoid the duplicated code and extract this:

```
      const nonEmptyStages = pipeline
        .map(generateStage)
        .filter((stage) => Object.keys(stage).length > 0);
```

to a function right in this file and use it everywhere we are doing the same?

Sorry for being a bit picky here .. is a bit of detail, but chaos spreads each time we don't pay attention to it

mabaasit: We can do it when running aggregation, but I refactor this piece of code into a function and cleaned it up. I'd still prefer to have it here to avoid conditional tracking for just `page === 1`.

mcasimir: Makes sense! Thanks!

mcasimir (approved):

PR #3163, opened 2022-06-03
feat(connect): Save & Connect button COMPASS-5776

lerouxb:
In the New Connection dialog (ie. when there's no existing favorite connection involved), display a `Save & Connect` button in addition to the existing `Save` and `Connect` buttons. This will pop up the `Save connection to favorites` dialog and replace the dialog's `Save` button with a `Save & Connect` button. Clicking that button saves the connection as a favorite and connects.


packages/connection-form/src/components/connect-form.tsx
```diff
@@ -296,6 +310,10 @@ function ConnectForm({
                 ...favoriteInfo,
               },
             });
+
+            if (saveConnectionModal === 'saveAndConnect') {
+              onSubmitForm();
```
Anemy, line 315: I think the connection options in the form don't have the recently set favorite settings on the first connection.
After save and connect with a new connection:
(Lack of connection name)
Screen Shot 2022-06-14 at 1 45 02 PM
After reloading the favorites list and connecting to the favorite:
(has connection name)
Screen Shot 2022-06-14 at 1 45 07 PM

Not sure this is a blocker, but I could see it tripping up some folks. 

Due to the nature of how the `useConnectForm` is setup it may take some extra work to have it update the current connection options and then connect from there. Maybe we could add an optional argument overload to the `onSubmitForm` function to explicitly connect to the passed options? Totally up to you implementation wise, just suggesting a thought.

lerouxb: Same as this? https://mongodb.slack.com/archives/GDTJXPHD0/p1654259449438809

lerouxb: I think this is an existing problem. ie if you hit save via the favorite modal and then connect on main now it also happens.

Due to that I'm not sure what's best practise - make this PR larger and fix it now or fix it in a followup?

Anemy: Looks like it's the same, yes. When saving happens we don't update the current connection settings. As for best practice, it's up to you, since it won't change this code much I think it would be nice to have it in another pr. It could be a separate `fix` then.

Anemy (approved):

packages/connection-form/src/components/connect-form-actions.tsx
```diff
@@ -75,6 +85,19 @@ function ConnectFormActions({
           </Button>
         )}
 
+        {saveAndConnectButton !== 'hidden' && (
+          <div className={cx(saveAndConnectStyles)}>
```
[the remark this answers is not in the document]
lerouxb: Cool I didn't know that. Thanks!

packages/connection-form/src/components/connect-form.tsx
lerouxb: https://github.com/mongodb-js/compass/pull/3189

PR #3166, opened 2022-06-03
fix: always display $merge and $out COMPASS-5876

mcasimir:
To the best of my understanding `$merge` and `$out` should be available on:
- any server and any namespace
- on all distributions except Compass Readonly (`isReadonly` won't help here cause it would be `true` for views)

And the template should change based on the server type and should be different on ADL.

In this PR I've just removed any check regarding `$merge` and `$out` which to the best of my knowledge should not have been correct, hopefully even if I'm wrong, we can use this as a quick fix for MDBW.

Also:
- add an `outputStage` as operator property in "ACE Autocompleter" and filter out any operator where `outputStage === true` on Compass Readonly
- add `namespaces` as operator property in "ACE Autocompleter" and filter out any operator where current namespace doesn't match
- let Atlas only stages be filtered by version

Screenshot 2022-06-06 at 18 42 57

Screenshot 2022-06-06 at 18 43 33

Screenshot 2022-06-06 at 18 42 22

Screenshot 2022-06-06 at 18 42 50


packages/compass-aggregations/src/utils/stage.js
```diff
@@ -42,92 +40,98 @@ export const parseNamespace = (currentDb, stage) => {
...
+    if (operator.dbOnly) return false;
 
-    return isSupportedVersion({ oVersion: o.version, version: cleanVersion }) &&
-      isSupportedEnv({ oEnv: o.env, env }) ||
-      isAtlasOnly(o.env);
+    return isSupportedVersion({
+      operatorVersion: operator.version,
+      serverVersion: cleanServerVersion
+    }) &&
+      isSupportedEnv({
```
mcasimir, line 132: This is a pass-by change, I moved `isAtlasOnly` inside `isSupportedEnv` so even `isAtlasOnly` stages will be filtered by version. 

We often use the server version to hide stages that will be released in the future, I thought we don't want to ignore the version here. Happy to revert if this was intentional.

Anemy (approved):
nice fix. Happy this variable name is gone since we also have `isReadonly`

Anemy (approved):
lgtm! If it doesn't require much extra work this might be a good time to make `stage.js` in typescript? The `compass-aggregations` package does allow typescript, since we're making a good amount of changes to [stage.js](https://github.com/mongodb-js/compass/pull/3166/commits/0f390239c42f2a5b95e88cac526f6a600230c556#diff-741dac17f9b49a7c771314d753802c961cb8a0179ee12f3d58715f76bd81aabf) it may make sense. Just a suggestion, changes look good let's ship it!

*Might be an e2e test worth looking at:
```
[2022/06/06 17:31:00.166]   128 passing (10m)
[2022/06/06 17:31:00.166]   19 pending
[2022/06/06 17:31:00.166]   1 failing
[2022/06/06 17:31:00.166]   1) Collection aggregations tab
[2022/06/06 17:31:00.166]        shows atlas only stage preview:
[2022/06/06 17:31:00.166]      Error: waitUntil condition timed out after 10000ms
[2022/06/06 17:31:00.166]       at Z:\data\mci\c2fef572be1f494ad0752fb3a0f63228\src\packages\compass-e2e-tests\node_modules\webdriverio\build\commands\browser\waitUntil.js:68:19
```

mcasimir:
> this might be a good time to make `stage.js` in typescript

I tried, but it would require faking the interface of `mongodb-ace-autocompleter` entirely, I'd wait to convert that package first, should take 5 minutes after that.

PR #3168, opened 2022-06-06
chore(compass-components): Remove ace-theme packages, move to directly in compass-components

Anemy:
This PR removes the `mongodb-ace-theme` and `mongodb-ace-theme-query` packages and moves them to in `compass-components`, which was their only consumer.

We also make the query bar theme different based on the `COMPASS_SHOW_NEW_TOOLBARS` feature flag, which is part of COMPASS-5673 (closes https://github.com/mongodb-js/compass/pull/3152)


packages/compass-components/src/constants/mongodb-ace-theme-query.ts
```diff
@@ -0,0 +1,242 @@
+const mongodbAceThemeQueryCssTextLegacy =
+  '\
```
mcasimir, line 2: should be possible to use a template literal here or?

mcasimir (approved):
👍👍👍

PR #3171, opened 2022-06-07
fix(queries): Ignore duplicate recent queries COMPASS-2237

lerouxb:
This won't save a query that already exists in the recents again. It also moves a bunch of tests that were in test/renderer and therefore ignored to the right places so they run again.


Anemy (commented):
This does solve the duplicate entries pain point, however I think as a user I'd expect/like if the recent query I use goes to the top of the recents list when it's used. 
Is it possible to update the `_lastExecuted` entry for the matching recent when a query is performed?

packages/compass-query-history/src/stores/recent-list-store.js
```diff
@@ -67,6 +78,12 @@ const configureStore = (options = {}) => {
         return r._ns === ns;
       });
 
+      /* Ignore duplicate queries */
+      const existingQuery = this.state.items.find((item) => _.isEqual(comparableQuery(item), recent));
```
Anemy, line 82: Looks like we do a similar equality check in this file in `runQuery`: https://github.com/mongodb-js/compass/blob/0ae8aa6f7a2429e7a2081445a490aa43af98f2c1/packages/compass-query-history/src/stores/recent-list-store.js#L102
Would it make sense to share this comparison somehow?

lerouxb: I considered this, but then I'm not 100% sure I understand what this line is trying to do. Is it definitely the same as what I wrote above? Then I suppose we should reuse it here. If it is trying to do a looser comparison, then I don't know if that's safe.

lerouxb: Actually I suppose that's just tracking, so should be safe enough.

packages/compass-query-history/src/stores/recent-list-store.js
```diff
@@ -25,7 +24,9 @@ const configureStore = (options = {}) => {
      * Filter attributes that aren't query fields or have default/empty values.
      * @param {object} attributes
      */
-    _filterDefaults(attributes) {
+    _filterDefaults(_attributes) {
+      // don't mutate the passed in parameter
```
lerouxb, line 28: Lost some time here. My test that added the same query twice actually never executed the code that would check if it is a duplicate because this stripped ns from the query the first time, causing it to early return for a different reason.

I think it is just better style to not implicitly modify things that came in from the outside.

packages/compass-query-history/src/stores/favorite-list-store.js
```diff
@@ -60,18 +61,12 @@ const configureStore = (options = {}) => {
...
-      // (because there are more options than just .filter), but that
-      // is probably fine as a temporary measure.
-      // https://jira.mongodb.org/browse/COMPASS-5243
-      const item = this.state.items
-        .map((_item) => _item.serialize())
-        .find((_item) => {
-          return isDeepStrictEqual(_item.filter, query.filter);
+      const existingQuery = this.state.items
+        .find((item) => {
+          return _.isEqual(comparableQuery(item), query);
```
lerouxb, line 66: Found a second copy of this logic / a third place that could benefit from the new function.

PR #3172, opened 2022-06-07
fix(aggregation-explain): show indexes COMPASS-5879

mabaasit:
fix(aggregation-explain): show indexes COMPASS-5879

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-aggregations/src/modules/explain.ts
+++ b/packages/compass-aggregations/src/modules/explain.ts
@@ -256,16 +256,13 @@ const mapIndexesInformation = function (
   return explainIndexes
     .filter(x => x.index)
     .map((explainIndex) => {
-      const index = collectionIndexes.find(
+      const collectionIndex = collectionIndexes.find(
         (collectionIndex) => collectionIndex.name === explainIndex.index
       );
-      if (!index) {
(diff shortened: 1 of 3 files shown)


packages/compass-aggregations/src/modules/explain.ts
```diff
@@ -256,16 +256,13 @@ const mapIndexesInformation = function (
   return explainIndexes
     .filter(x => x.index)
     .map((explainIndex) => {
-      const index = collectionIndexes.find(
+      const collectionIndex = collectionIndexes.find(
```
Anemy, line 259: Do we / should we test this function/module? I'm thinking it could help prevent any regressions.

Anemy (approved):
lgtm, maybe an opportunity to add some more tests for the `explain.ts` module? Fine as is too.

packages/compass-aggregations/src/modules/explain.ts
mabaasit: done :)

PR #3184, opened 2022-06-13
fix(import): pre-create an empty object before creating its properties COMPASS-5076

lerouxb:
As explained in the ticket, _.set() will create an array for `foo.1` if foo doesn't already exist and 1 is a numeric key:

```javascript
JSON.stringify(_.set({}, 'foo.1', 'a'))
'{"foo":[null,"a"]}'
```

but not if it doesn't look like a number:

```javascript
JSON.stringify(_.set({}, 'foo.a', 'b'))
'{"foo":{"a":"b"}}'
```

In order to make this safe and still use `_.set()` this PR post-processes the flattened ("dotnotation") object, making sure that every object is created ie. `_.set({}, 'path.to.object', {})` before the keys for its properties. ie. `_.set({}, 'path.to.object.foo', 'bar'')`.

This does mean that the key order in the flattened object matters, though, otherwise _.set() will override. It also does it as a second pass. Maybe that's dodgy?

There are lots of other ways to solve this including replacing `flatten` and just drilling down obj and doing the flattening ourselves. Or replacing the use of _.set. Or just reimagining this code entirely. None of it would be hard to do if we choose.

But maybe this is good enough? I don't have particularly strong feelings about it either way.


Anemy (approved):
🔥  typescript 🙌 nice fix

PR #3187, opened 2022-06-13
styles(compass-components): Add shadow to toolbar

Anemy:
This shadow can be best seen in the aggregations view. This will also impact the new toolbars we're building behind feature flags.

before:
Screen Shot 2022-06-13 at 3 46 45 PM

after
Screen Shot 2022-06-13 at 3 31 28 PM


Anemy:
Going to close, we can revisit if we feel this is needed in the new aggregations/toolbars work.

PR #3192, opened 2022-06-16
feat(compass-indexes): update toolbar to leafygreen components COMPASS-5676

Anemy:
COMPASS-5676

Note: This work is **not** behind the `COMPASS_SHOW_NEW_TOOLBARS` feature flag. This is because it's not depending on other parts, like the query bar, to update first. This will be in the live product similar to the new aggregations toolbar.

This PR updates the indexes plugin's toolbar to use leafygreen components for the button, tooltip, and banners. 

Previously we were using a `TextWriteButton` from `compass-deployment-awareness` for rendering the button and syncing it with the `writable` state in Compass. Since these items were already synced with the compass-indexes store we're now using that state directly. Down the line I think we'll want the writable + readonly state in context providers with hooks. 

## Before:
Screen Shot 2022-06-15 at 1 58 18 PM

Screen Shot 2022-06-15 at 8 53 25 PM

Screen Shot 2022-06-15 at 2 36 24 PM

## After:
Screen Shot 2022-06-15 at 1 58 05 PM

Screen Shot 2022-06-15 at 2 19 40 PM

Screen Shot 2022-06-15 at 7 04 21 PM


packages/compass-indexes/src/stores/store.js
```diff
@@ -48,12 +48,20 @@ const configureStore = (options = {}) => {
...
+    deploymentAwarenessStore.listen((state) => {
+      store.dispatch(writeStateChanged(state.isWritable));
+      store.dispatch(getDescription(state.description));
+    });
+
+    // Load the initial deployment awareness state.
+    store.dispatch(
+      writeStateChanged(deploymentAwarenessStore.state.isWritable)
+    );
+    store.dispatch(getDescription(deploymentAwarenessStore.state.description));
```
Anemy, line 64: Adding this initial loading. This was a bug before, the `deploymentAwarenessStore`'s state would be out of line in this package as this compass-indexes store is initialized at a time when the deploymentAwareness state is already set.
Will be a nice thing to tackle in state management improvements, not including in this pr as we probably want those to be more easily discussed.

PR #3222, opened 2022-06-29
feat(compass-settings): compass settings COMPASS-5927

mabaasit:
feat(compass-settings): setup compass settings (LG)

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-settings/src/stores/settings.ts
```diff
@@ -0,0 +1,77 @@
...
+export const fetchSettings = (): ThunkAction<void, RootState, void, Actions> => {
+  return async (dispatch: Dispatch<Actions>): Promise<void> => {
+    try {
+      const settings = await fetchPreferences();
+      dispatch({
+        type: ActionTypes.SettingsFetched,
+        settings
+      });
+    } catch (e) {
+      console.error(`Failed to fetch settings: ${(e as Error).message}`);
```
gribnoysup, line 57: Use logger instead?

packages/compass-settings/src/utils/user-preferences.ts
```diff
@@ -0,0 +1,59 @@
...
+  showedNetworkOptIn: boolean;
+  singleDocumentCrud: boolean;
+  sslAll: boolean;
+  sslServer: boolean;
+  sslUnvalidated: boolean;
+  telemetryAnonymousId: string;
+  theme: string; // todo use THEMES
+  trackErrors: boolean;
+  trackUsageStatistics: boolean;
+};
```
gribnoysup, line 33: Are all of these used? Do we allow to change them?

mabaasit: skimmed it down.

packages/compass-settings/src/components/modal.tsx
```diff
@@ -0,0 +1,100 @@
...
+    if (isOpen) {
+      onInit();
+    }
+  }, [onInit, isOpen]);
+
+  useEffect(() => {
+    (ipc as any).on('window:show-network-optin', () => {
+      setIsOpen(!isOpen);
+    });
+  }, [setIsOpen, isOpen]);
```
gribnoysup, line 64: This will recreate the subscription every time `isOpen` is flipped. If you need this to happen, you need to either use a return value to unsubscribe, or use a callback inside `setState` instead of directly refererring to value, which is also [a recommeded approach](https://reactjs.org/docs/hooks-reference.html#functional-updates) when new state depends on the previous one

I would also maybe consider not doing this, FWIW this is really not a common pattern in the applications to use menu as a toggle, unless the label in the menu explicitly says "Toggle <smth>"

packages/compass-settings/src/components/modal.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const [selectedSetting, setSelectedSettings] = useState(settings[0].name);
+
+  const SettingComponent =
+    settings.find((x) => x.name === selectedSetting)?.component ?? null;
+
+  useEffect(() => {
+    if (isOpen) {
+      onInit();
+    }
+  }, [onInit, isOpen]);
```
gribnoysup, line 58: What would be the reason to re-fetch them every time on modal open? Also the callback name sorta doesn't make sense in this case, so I'd suggest either updating the callback name or going back to doing init only once 🙂

packages/compass-settings/src/components/modal.tsx
mabaasit: I removed the toggle completely.

packages/compass-settings/src/components/sidebar.tsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+const SettingsSideNav: React.FunctionComponent<SidebarProps> = ({
+  activeItem,
+  items,
+  onSelectItem,
+}) => {
+  return (
+    <nav data-testid="settings-sidebar">
+      <ul role="menu">
```
gribnoysup, line 35: Menus are widgets that usually provide access to the list of possible actions, sometimes they include navigation, but in most cases this is a full page one, not a localized one. This side navigation is a tab group, you select a tab and a controlled content is displayed in the main area. Here are some references to related roles: [`tablist`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role), [`tabpanel`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tabpanel_role), [`tab`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role)

packages/compass-settings/src/components/sidebar.tsx
```diff
@@ -0,0 +1,62 @@
...
+        {items.map((item) => (
+          <li
+            data-testid={`sidebar-${item}-item`}
+            role="option"
+            tabIndex={0}
```
gribnoysup, line 40: This really should be an interactive element, a button, you should not need to provide these attributes yourself

packages/compass-settings/src/components/modal.spec.tsx
```diff
@@ -0,0 +1,6 @@
+// todo: mock hadron-ipc
+describe('SettingsModal', function () {
+  it.skip('renders settings modal');
+  it.skip('calls onInit on load');
+  it.skip('closes the modal');
+});
```
gribnoysup, line 6: You're planning to add tests in the same PR?

packages/compass-settings/src/components/sidebar.tsx
```diff
@@ -32,26 +36,19 @@ const SettingsSideNav: React.FunctionComponent<SidebarProps> = ({
...
+              className={cx(itemStyles, {
+                [itemHoverStyles]: item !== activeItem,
+                [itemActiveStyles]: item === activeItem,
+              })}
+              data-testid={`sidebar-${item}-item`}
+              onClick={() => onSelectItem(item)}
+            >
+              {item}
+            </button>
           </li>
```
gribnoysup, line 52: A few things still need addressing here: role should be applied to the interactive element, not to the wrapper, and element with the tab role should have aria-selected and aria-controls attributes set. See [this example](https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html)

gribnoysup:
I see that you requested a review, but I don't see any changes that are addressing https://github.com/mongodb-js/compass/pull/3222#discussion_r913618505. Am I missing something?

gribnoysup:
You would also probably need to update e2e tests to handle this new settings screen, seems like a lot of tests are failing because of that at the moment

packages/compass/src/app/index.js
```diff
@@ -251,21 +249,14 @@ var Application = View.extend({
...
-  },
   showSecurity: function() {
     app.appRegistry.getAction('Security.Actions').show();
   },
   tourClosed: function() {
     app.preferences.unset('showFeatureTour');
     app.preferences.save();
     if (!app.preferences.showedNetworkOptIn) {
-      this.showOptIn();
+      ipc.ipcMain.broadcastFocused('window:show-network-optin');
```
gribnoysup, line 259: This code doesn't run in the main thread, and so currently when running the app locally it breaks instead of showing the settings

image

alenakhineika:
FYI this work is planned to be finished as part of the [Improvements for Enterprise Customers in Compass](https://docs.google.com/document/d/18FMePMkgpm9SOEmDs0-i1EZWujdQF7tLBbrg4mV9RHM/edit#heading=h.wtq2vbtosuas) project. Let's ensure that the new sensings package supports the dark theme when enabled via the COMPASS_LG_DARKMODE feature flag and follow [Compass Components Conventions](https://docs.google.com/document/d/1ZLz_DJYpwTiS_DiPcQMKoLoz2iNxSsMKRTuXXxHb6ak/edit#heading=h.2s9zdv6a57gn).

packages/compass-settings/src/utils/user-preferences.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { THEMES } from 'compass-preferences-model';
+
+export type UserPreferences = {
+  autoUpdates: boolean;
+  enableMaps: boolean;
+  trackErrors: boolean;
+  trackUsageStatistics: boolean;
+  enableFeedbackPanel: boolean;
+  theme: THEMES.DARK | THEMES.LIGHT | THEMES.OS_THEME;
+};
```
alenakhineika, line 12: Let's include also networkTraffic here.

packages/compass-preferences-model/index.d.ts
```diff
@@ -1,8 +1,12 @@
...
-
-export = CompassPreferencesModel;
+declare class CompassPreferencesModel {
+  fetch: () => void;
+  save: () => void;
+  set: (
+    key: string,
+    value: unknown
+  ) => void;
+};
```
alenakhineika, line 8: How detailed should this declaration be? I think we also use unset, once, on, and probably other methods from Ampersand. If we don't get rid of Ampersand, maybe it makes sense to still export the whole AmpersandModel type? And I also can't find in the code, where is this CompassPreferencesModel being used?

mabaasit: exporting the whole ampersand model type definition is very cumbersome. here i am trying to export on the parts we are using and as a result trying to keep it minimal and simple. and as for using this, its being used in `compass-settings/src/utils/user-preferences.ts`

packages/compass-settings/src/utils/user-preferences.ts
addaleax: We can defer this to the actual ticket for making networkTraffic work as expected, and we’ll be moving this list to compass-preferences-model anyway.

packages/compass-settings/src/utils/user-preferences.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const fetchPreferences = async (): Promise<UserPreferences> => {
+  const model = new Preferences();
+  const fetch = promisifyAmpersandMethod(model.fetch.bind(model));
+  const settings = await fetch();
+  return (settings as any).getAttributes({ props: true }, true);
```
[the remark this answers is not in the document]
addaleax: Fwiw, I talked about this with Basit earlier, I think it’s fine to leave this as-is since it will be changing fairly soon anyway

packages/compass/src/app/index.js
```diff
@@ -224,22 +222,16 @@ var Application = View.extend({
       this.tourClosed();
     }
   },
-  showOptIn: function() {
-    if (process.env.HADRON_ISOLATED !== 'true') {
```
addaleax, line 228: How would this currently work in Isolated edition? Would we display settings that don’t have any effect on users?

mabaasit: we hide the menu item in ([menu.ts](https://github.com/mongodb-js/compass/pull/3222/files#diff-94ecbcd9e84609b4725e01f12f7f8c01168f17639ffff65dc8f6f27ef11a4cddR336)). and as a follow up, we also need to add `@mongodb-js/compass-settings` to `compass/src/app/plugins/default.js`

addaleax: The `tourClosed()` function belowed used this one’s logic though – right now it always broadcasts `window:show-network-optin`

mabaasit: nice, i'll add this check there.

packages/compass-settings/src/utils/user-preferences.ts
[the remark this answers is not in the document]
gribnoysup: If we are planning to remove these any assertions later that's fine by me

addaleax (approved):

PR #3224, opened 2022-06-30
chore(sidebar): merge deploymentAwareness, serverVersion, sshTunnelStatus plugins into the sidebar COMPASS-5900

lerouxb:
Still trying to get this green, still some tests to write and I accidentally ran prettier on all of compass-sidebar that I was specifically trying to do hold off on until a separate PR and I'm trying to find a way to undo that without losing all my work, but I think it is close enough to start to have a look at.

### Notes to reviewers:

I would recommend collapsing all the folders and then going through it folder by folder, ignoring the packages that were just removed. Then it should be simpler to follow.

* **compass-aggregations/src/stores**: it follows what I outlined below and I annotated some interesting bits with review comments

* **compass-app-stores**: instance-store got changed a little bit so that the new toplogyDescription state can live in there

* **compass-collection/src/stores**: it now just gets version changes from the instance which it already had access to

* **compass-crud/src/components**: just prop-drilling to get isWritable and instanceDescription to the toolbar

* **compass-crud/src/stores**: fixes to instance-created, gets isWritable and description from the instance model. It already tried to get isDataLake (and this isEditable) and version from there but as explained that wasn't working.

* **compass-deployment-awareness**: removed. see info below

* **compass-indexes/src/stores**: Usage of DeploymentAwareness.WriteStateStore replaced with instance model found on App.InstanceStore

* **compass-schema-validation/src/stores**: server version and isWritable now comes from instance.build.version. Cleaned up usage of changeZeroState() and editModeChanged() a bit

* **compass-server-version**: removed

* **compass-sidebar**: This all started in packages/compass-sidebar/src/components/sidebar-instance-details/sidebar-instance-details.jsx. That's where usage of the 'InstanceDetails.Item' role was removed and the three components that make up the instance details in the sidebar just got used directly.

* **compass-ssh-tunnel-status**: removed

* **compass**: just removed the three plugins from default, isolated and readonly

* **databases-collections/src/stores**: version and isWritable now comes from instance model rather than the removed 'server-version-changed' event and WriteStateStore.

* **instance-model**: the default export got replaced so it can export more things now. Things that got moved in from deployment awareness. Most notably TopologyDescription and some constants for topology and server types. Added topologyDescription and some derived properties so we could get rid of compass-deployment-awareness

---

### compass-server-version and compass-ssh-tunnel-status

These were relatively straight forward to move. Their stores were just merged into the sidebar store and their components are moved into the sidebar as-is for now. They used reflux and compass-sidebar uses redux, so some conversion happened there. A future PR can leafygreenify those ie. probably just remove/replace them.

compass-server-version also owned the `'server-version-changed'` event. I removed it and the server version can now be read from the instance model.

### compass-deployment-awareness

The `'compass:deployment-awareness:topology-changed'` event on globalAppRegistry is now removed. This was emitted in two places in compass-deployment-awareness and only ever listened to in compass-aggregations once. The properties that were sent through it now live in instance-model. However it had four properties: `topologyType`, `setName`, `servers` and `env` and only the `env` one was used.

compass-deployment-awareness also had two more stores: `WriteStateStore` and `ReadStateStore`. `ReadStateStore` was never used, so this is now removed. `WriteStateStore` was used directly by compass-indexes, compass-schema-validation, compass-sidebar and databases-collections (for collections-store and databases-store) to propagate the `isWritable` and `description` properties. **These properties now live in instance-model.**

To get the `instance` model, plugins that aren't scoped use the instance-created event because the instance is only created after their stores are instantiated. Scoped plugins now get the instance from the globally registered `App.InstanceStore` when they get created. From there they can pull `env`, `isWritable`, `description`, `build.version` and so on and then also listen to changes on those properties.

**Drive-by fix:** There was a bug where compass-crud is a scoped plugin and it used instance-created to get the instance. But that code never executed because the event already fired before the crud-store was instantiated. This is fixed now.

CompassDeploymentAwareness also registered the components `TextReadButton`, `TextWriteButton` and `OptionWriteSelector`. Out of the three only `OptionWriteSelector` still gets used and it is only used in one place - the `Add Data` button in the crud toolbar. These are kinda weird components in that they know of the `isWritable` and instance `description` state directly (from WriteStateStore), bypassing the common react patterns of props or providers. 

`TextReadButton` and `TextWriteButton` got removed as all remaining usage of it already got removed recently and (for now) `OptionWriteSelector` was changed to take `isWritable` and `description` as props. I also moved it to compass-crud as that's the only place we are using it. In a future PR we can just leafygreenify that component and it can go away as well, probably with other changes in crud's toolbars.

Just FYI this is a rarely used pattern in the codebase. The only remaining uses of `appRegistry.getComponent()` (or indirectly through `useAppRegistryComponent()`) are `Home.Home`, `Collation.Select`, `Collection.Workspace`, `Database.Workspace`, `Instance.Workspace`, `Sidebar.Component` and `Global.Shell`.


packages/compass-aggregations/src/stores/store.js
```diff
@@ -249,6 +251,8 @@ const configureStore = (options = {}) => {
     );
   }
 
+  // options.isAtlasDeployed is only used by mms to change some behaviour in the
```
lerouxb, line 254: As part of the above I decided to typescript this file so we can start to capture which options these stores can even take, but yeah.. this file is only the interface for mms whereas for us the plugin stuff in `packages/compass-collection/src/stores/context.tsx` is more relevant but even that gets called in `packages/compass-collection/src/modules/tabs.ts` which is called in `packages/compass-collection/src/stores/index.ts` when the 'select-namespace' event arrives but really all the options are actually contained in that `metadata` parameter and that metadata param gets constructed in at least two places..

Figured that kind of work should just be in a separate PR.

packages/compass-crud/src/components/option-write-selector.jsx
```diff
@@ -27,40 +26,9 @@ class OptionWriteSelector extends React.Component {
...
-
-  /**
-   * Handle write state changes.
-   *
-   * @param {Object} state - The write state.
-   */
-  writeStateChanged(state) {
-    this.setState(state);
+    tooltipId: PropTypes.string.isRequired,
+    isWritable: PropTypes.bool.isRequired,
```
lerouxb, line 30: See for now this component (which is only used for the Add Data button in crud that lets you pick insert/import) just gets isWritable as a prop rather than listening to a dedicated store.

packages/compass-crud/src/stores/crud-store.js
```diff
@@ -1252,8 +1230,29 @@ const configureStore = (options = {}) => {
   if (options.globalAppRegistry) {
     const globalAppRegistry = options.globalAppRegistry;
 
-    globalAppRegistry.on('instance-created', ({ instance }) => {
```
lerouxb, line 1255: This event would never fire because this is a scoped plugin and the instance already exists by the time this gets executed.

packages/compass-crud/src/stores/crud-store.js
```diff
@@ -698,7 +677,7 @@ const configureStore = (options = {}) => {
       }
 
       const csfleState = { state: 'none' };
-      const dataServiceCSFLEMode = this.dataService.getCSFLEMode && this.dataService.getCSFLEMode();
+      const dataServiceCSFLEMode = this.dataService && this.dataService.getCSFLEMode && this.dataService.getCSFLEMode();
```
lerouxb, line 680: For some reason a test started failing because that one test did some extensive mocking and the store doesn't have a dataService. Now that the onInstanceCreated code actually executes it got here and blew up because dataService doesn't exist in that case.

Not crazy about this kind of thing where we aren't strict about what should exist and then use proper test helpers to make sure everything exists, but for now I think I can live with this.

packages/compass-crud/src/stores/crud-store.js
lerouxb, line 1255: This event would never fire because this is a scoped plugin and the instance already exists by the time this gets executed.

packages/instance-model/lib/model.js
```diff
@@ -135,13 +139,75 @@ const InstanceModel = AmpersandModel.extend(
...
+            if (topologyType === TopologyType.SINGLE) {
+              const message = this.isServerWritable ? 'is writable' : 'is not writable';
+              return `Single connection to server type: ${ServerType.humanize(this.singleServerType)} ${message}`;
+            }
+            return `Topology type: ${TopologyType.humanize(topologyType)} is writable`;
+          }
+          return `Topology type: ${TopologyType.humanize(topologyType)} is not writable`;
+        }
+      },
+      env: {
```
lerouxb, line 192: We have env and then various isAtlas, isDataLake, etc things all over the codebase, unfortunately. But at least for now we can know this as soon as we get the instance details. We should probably pass it to the collection tab plugins when creating their stores.

packages/instance-model/lib/model.js
```diff
@@ -135,13 +139,75 @@ const InstanceModel = AmpersandModel.extend(
...
+              return this.isServerWritable;
+            } else {
+              return true;
+            }
+          } else {
+            return false;
+          }
+        }
+      },
+      description: {
```
lerouxb, line 177: Kept this as a derived thing because a number of places care about it changing.

packages/compass-aggregations/src/stores/store.js
gribnoysup: Amazing effort, thank you for documenting this in the code!

packages/compass-crud/electron/renderer/components/option-write-selector.jsx
```diff
@@ -14,7 +14,6 @@ const WRAPPER = 'tooltip-button-wrapper';
  * OptionSelector is an extension of React-Bootstrap DropdownButton.
  */
 class OptionWriteSelector extends React.Component {
```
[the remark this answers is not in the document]
lerouxb: Aah but it isn't just this one :)

There are also the ones in compass-app-stores, compass and databases-collections. Probably more once you factor in indirect effects.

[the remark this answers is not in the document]
gribnoysup: Oh yeah, that's true! I think we all agreed that we don't want them, and removing them one by one is fine I think (I was tempted to nuke them all at once, but with all the changes across multiple packages going on right now I don't want to cause unnecessary conflicts for other PRs). But just to clarify totally do what you feel is more comfortable for you here, maybe when we have a quieter period in the repo I'll do a big clean up for whatever is left

Anemy (approved):
lgtm, left a couple comments on a few things I noticed. Looks really good. Solid cleanup and nice to have updated tests and fewer plugins!

packages/compass-app-stores/src/stores/instance-store.spec.js
```diff
@@ -92,6 +104,8 @@ describe('InstanceStore [Store]', () => {
       });
     });
 
+    // TODO: when the topologyDescription changed
```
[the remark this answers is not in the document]
lerouxb: Aah no that became redundant and I meant to remove it.

packages/compass-indexes/src/stores/store.js
```diff
@@ -48,20 +48,20 @@ const configureStore = (options = {}) => {
...
-      writeStateChanged(deploymentAwarenessStore.state.isWritable)
-    );
-    store.dispatch(getDescription(deploymentAwarenessStore.state.description));
+    // these can change later
+    instance.on('change:isWritable', () => {
+      store.dispatch(writeStateChanged(instance.isWritable));
+    });
+    instance.on('change:description', () => {
+      store.dispatch(getDescription(instance.description));
+    });
```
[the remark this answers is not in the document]
lerouxb: I kinda went back and forth on this. It shouldn't be possible for these scoped plugins for the version to change afterwards because they only get instantiated after we received the fully loaded instance with details. So I guess the answer is to remove it elsewhere.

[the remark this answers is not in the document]
lerouxb: Actually no I think they get an instance but the details can arrive later 🤔 Now I'm confused again.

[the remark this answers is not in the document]
lerouxb: (which I've done now)

PR #3228, opened 2022-07-05
fix: automatically refresh after CSFLE insert COMPASS-5806

alenakhineika:
## Description

Similary to the insert many documents functionality, refresh the documents list after inserting a single document. Counting documents in the `refreshDocuments` function happens asynchronously, therefore the expected count in the related test is 0 now, the same as for `insertMany` tests.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-crud/src/stores/crud-store.js
+++ b/packages/compass-crud/src/stores/crud-store.js
@@ -926,40 +926,18 @@ const configureStore = (options = {}) => {
           });
         }
 
-        // check if the newly inserted document matches the current filter, by
-        // running the same filter but targeted only to the doc's _id.
-        const filter = Object.assign({}, this.state.query.filter, { _id: doc._id });
-        this.dataService.count(this.state.ns, filter, {}, (err, count) => {
-          if (err) {
-            return this.setState({
-              insert: this.getInitialInsertState()
-            });
-          }
-          // track mode for analytics events
-          const payload = {
-            ns: this.state.ns,
-            view: this.state.view,
-            mode: this.state.insert.jsonView ? 'json' : 'default',
-            multiple: false,
-            docs: [doc],
-          };
-          this.localAppRegistry.emit('document-inserted', payload);
-          this.globalAppRegistry.emit('document-inserted', payload);
+        const payload = {
+          ns: this.state.ns,
+          view: this.state.view,
+          mode: this.state.insert.jsonView ? 'json' : 'default',
+          multiple: false,
+          docs: [doc],
+        };
+        this.localAppRegistry.emit('document-inserted', payload);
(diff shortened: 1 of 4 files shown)


packages/compass-e2e-tests/tests/in-use-encryption.test.ts
```diff
@@ -126,12 +103,8 @@ describe('FLE2', function () {
       }
 
       await delay(10000);
-
-      console.log('Read favorites after disconnecting and delay');
-      await readSavedConnectionsFolder(compass);
```
addaleax, line 131: Is removing these things intentional?

packages/compass-crud/src/stores/crud-store.js
```diff
@@ -926,40 +926,18 @@ const configureStore = (options = {}) => {
...
-              count: this.state.count + 1,
-              end: this.state.end + 1,
-              insert: this.getInitialInsertState()
-            });
-          }
-          this.setState({
-            count: this.state.count + 1,
-            insert: this.getInitialInsertState()
-          });
-        });
```
addaleax, line 962: Are we good with removing this “smart” behavior even for non-CSFLE/QE inserts? I don’t mind, I think it adds consistency and has no major downsides, but I think it’s worth calling out if we do it this way.

packages/compass-e2e-tests/tests/in-use-encryption.test.ts
alenakhineika: Yeah I used this previously to check if recents are actually exist in the folder, and haven't removed it after.

packages/compass-crud/src/stores/crud-store.js
alenakhineika: But how this is different from counting asynchronously while refreshing a list of documents? I mean visually it still shows the proper number, or I miss something?

addaleax: Yeah, visually the end result should be the same, it’s just that refreshing is a more expensive operation than a count (presumably, otherwise this would already have always been refreshing to begin with).

addaleax (approved):

packages/compass-crud/src/stores/crud-store.js
alenakhineika: True. What do you have in mind by `calling out`? Like our conversation here in PR? :) or some other place like comment/ticket?

addaleax: Yeah, I’m not sure, maybe just merge this and mention it during standup or sth? :)

PR #3240, opened 2022-07-11
feat(compass): Update icons and installers to new brand COMPASS-5120

Anemy:
COMPASS-5120

Reopening from https://github.com/mongodb-js/compass/pull/3214 - evergreen was having some clone issues which might be persisting from commit hashes. Testing a quick separate pr to see if it helps.

Before merging we'll look at all 3 platforms and ensure they all look good. A while back we once shipped Compass with some of the image sizes missing in the windows `.ico` (let's manually check to make sure they look good before merging).

The windows `.gif` is smaller (800 ×499 -> 640 ×480)px - that should be fine: https://github.com/electron/windows-installer/blob/master/src/index.ts#L65
The default loading gif used is 268 × 167 when none is supplied. We could definitely make this gif a bit more engaging, however as design is currently working on some loading animation and since users only see it once it's probably not worth investing much time now. We can update it when we have the newer assets. (If y'all think of anything cool you'd like to see here though I'm up to update it).

Before merging:
- [x] Mac icon looks good (with various icon sizes in the file/application explorer)
- [x] Mac installer looks good
- [x] Linux icon looks good
- [x] Windows icon looks good (with various icon sizes in the start menu and file explorer)
- [x] Windows banner looks good
- [x] Windows installer background looks good
- [x] Windows installing gif looks good


Anemy:
evergreen retry

PR #3242, opened 2022-07-11
feat: use Leafy Green Select and Combobox components COMPASS-5946

alenakhineika:
## Description

- Use [Leafy Green Select](https://www.mongodb.design/component/select/documentation/) and [Leafy Green Combobox](https://www.mongodb.design/component/combobox/example/) components instead of React Select Plus component.
- Update Combobox so we not only list existing fields but also cant create new ones from the dropdown.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-components/package.json
```diff
@@ -37,11 +37,12 @@
     "@leafygreen-ui/card": "^5.1.4",
     "@leafygreen-ui/checkbox": "^6.0.6",
     "@leafygreen-ui/code": "^9.4.0",
+    "@leafygreen-ui/combobox": "^2.0.1",
```
gribnoysup, line 40: This combobox component is implemented in the new brand style, right? Just wanted to double-check whether we are okay with mixing components with new brand styles with the old leafygreen style that Compass is currently using (like this dropdown select that is right after the combobox)

image

packages/compass-crud/electron/renderer/components/query-bar.jsx
```diff
@@ -0,0 +1,63 @@
+const React = require('react');
+const PropTypes = require('prop-types');
+
+class QueryBar extends React.Component {
```
Anemy, line 4: This file was removed on main, we can remove it here as well, maybe it was checked in by accident?

packages/compass-indexes/src/modules/create-index/new-index-field.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * Create a new index field in memory.
+ */
+export const CREATE_NEW_INDEX_FIELD =
+  'indexes/create-index/name/CREATE_NEW_INDEX_FIELD';
```
Anemy, line 5: Can we do these new files in typescript? (Also new-index-field.spec.js)

packages/compass-crud/electron/renderer/components/query-bar.jsx
alenakhineika: Probably some merge conflict that wasn't properly handled. Will remove it!

packages/compass-components/package.json
alenakhineika: Switched to the older version! Thx for pointing at this 🔥

packages/compass-indexes/src/modules/create-index/new-index-field.ts
```diff
@@ -23,7 +25,7 @@ export const INITIAL_STATE = null;
  *
  * @returns {String} The new state.
  */
-export default function reducer(state = INITIAL_STATE, action) {
+export default function reducer(state = INITIAL_STATE, action: AnyAction) {
```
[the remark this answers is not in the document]
alenakhineika: Written like this, 'CREATE_NEW_INDEX_FIELD' and 'CLEAR_NEW_INDEX_FIELD' would refer to a value but be being used as a type here. It would work with `type: string`, but in this case, there would be not so much advantage compared with AnyAction usage.

Anemy (approved):
lgtm, one comment/question on e2e test performance. Tried it out works nice 👍

PR #3246, opened 2022-07-12
feat(compass-crud): Update toolbar to leafygreen components COMPASS-5672

Anemy:
COMPASS-5672

To use the new toolbars in Compass:
```bash
```

https://user-images.githubusercontent.com/1791149/178534297-54fc3925-11f9-46c9-b9c1-368306282daa.mp4


packages/compass-crud/src/components/crud-toolbar.tsx
```diff
@@ -0,0 +1,238 @@
...
+              title="Previous Page"
+              onClick={() => getPage(page - 1)}
+              disabled={prevButtonDisabled}
+            >
+              <Icon glyph="ChevronLeft" />
+            </IconButton>
+            <IconButton
+              data-testid="docs-toolbar-next-page-btn"
+              aria-label="Previous Page"
+              title="Previous Page"
```
lerouxb, line 193: Both these buttons say Previous Page, btw. This one should probably be Next Page.

packages/compass-crud/src/components/crud-toolbar.tsx
```diff
@@ -0,0 +1,238 @@
...
+              onClick={() => getPage(page + 1)}
+              disabled={nextButtonDisabled}
```
lerouxb, line 195: I'm on the zero state and the next button is enabled. Is that intended? Now not sure what the behaviour was before 🤔 

Screenshot 2022-07-27 at 10 27 25

packages/compass-crud/src/components/crud-toolbar.tsx
```diff
@@ -0,0 +1,238 @@
...
+            </SegmentedControlOption>
+            <SegmentedControlOption
+              data-testid="toolbar-view-json"
+              aria-label="E-JSON View"
+              value="JSON"
+            >
+              <Icon glyph="CurlyBraces" />
+            </SegmentedControlOption>
+            <SegmentedControlOption
+              data-testid="toolbar-view-table"
```
lerouxb, line 225: Probably not the best place to report this and probably an existing bug, but I'm on the zero state and I clicked through to the table view and it shows below the zero state. Which looks a bit broken:
Screenshot 2022-07-27 at 10 31 57

lerouxb: Hmm actually it starts a bit low even when not on the zero state:

Screenshot 2022-07-27 at 10 33 51

packages/compass-crud/package.json
```diff
@@ -42,10 +42,9 @@
     "typecheck": "tsc --noEmit",
     "eslint": "eslint",
     "prettier": "prettier",
-    "lint": "npm run eslint .",
-    "prettier-skipped-for-now": "&& npm run prettier -- --check .",
+    "lint": "npm run eslint . && npm run prettier -- --check .",
     "depcheck": "compass-scripts check-peer-deps && depcheck",
-    "check": "npm run typecheck && npm run lint && npm run depcheck",
+    "check": "npm run lint && npm run depcheck",
```
gribnoysup, line 47: Why is the typecheck command removed from the check script? This PR seems to have typescript files added to the package source

packages/compass-crud/src/components/crud-toolbar.tsx
Anemy: It should be disabled there, fixed and added a test, thanks for catching.

packages/compass-crud/src/components/crud-toolbar.tsx
Anemy: That is an existing bug, I'd like to fix that in a separate pr if that's cool.

packages/compass-crud/package.json
Anemy: Nice catch, this was a merge conflict as we just reformatted this file and updated this check to use prettier. Re-added it 👍

lerouxb (approved):

PR #3256, opened 2022-07-13
feat(compass-aggregations): Add right glyph caret down to buttons with menus in compass-aggregations

Anemy:
As these two buttons open menus, the right glyph caret down gives a hint to that behavior.

Before:
Screen Shot 2022-07-13 at 2 55 30 PM

After:
Screen Shot 2022-07-13 at 2 55 34 PM


alenakhineika (approved):
👍

PR #3260, opened 2022-07-15
chore: Use the new node-runtime-worker-thread from mongosh COMPASS-5767

lerouxb:
This brings this fix in to compass: https://github.com/mongodb-js/mongosh/pull/1308

### DIFF
--- a/packages/compass/package.json
+++ b/packages/compass/package.json
@@ -247,7 +247,7 @@
     "make-fetch-happen": "^8.0.14",
     "marky": "^1.2.1",
     "mocha": "^8.4.0",
-    "mongodb": "^4.6.0",
+    "mongodb": "^4.8.0",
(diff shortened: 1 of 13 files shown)


packages/compass/package.json
```diff
@@ -160,7 +160,7 @@
     "email": "compass@mongodb.com"
   },
   "dependencies": {
-    "@mongosh/node-runtime-worker-thread": "^1.5.0",
+    "@mongosh/node-runtime-worker-thread": "^1.5.1",
```
addaleax, line 163: Can you bump driver and the other shell versions as well, similar to a8ff75c22ef7570375e565866f7378888457e11a? Otherwise shell versions are out of sync and we’d be running different driver versions in Compass and mongosh.

addaleax (changes requested):

addaleax (approved):

PR #3262, opened 2022-07-15
fix: allow updating unindexed encrypted fields COMPASS-5809

addaleax:
##### chore(data-service): track equality-searchable encrypted fields

In addition to tracking all encrypted fields for a given collection’s
schema sources, also track equality-searchable encrypted fields.
This is helpful for knowing whether a field can be included in
the query part of a Compass CRUD update operation that includes
original values of fields in the query portion of that update.

As part of this, the list of encrypted fields reported by
`knownSchemaForCollection()` and used internally in the encrypted
field tracker is changed from a simple array of strings to a
pair of field path lists, one for all encrypted and one for
equality-searchable encrypted fields in the schema. Using
field paths instead of `.`-joined strings has no direct effect
here, but aligns this part of Compass with the recent changes
to nested field editing/dots and dollars support.

##### chore(compass-crud): exclude unindexed encrypted fields in update query COMPASS-5809

When updating unindexed encrypted fields, exclude them from the
find part of the update operation, since they are by definition
not queryable.

This modifies the HadronDocument query building methods to take
options arguments instead of key lists, and updates these options
to be passed as field path lists rather than a plain array of strings
(where `.` was ambiguous between literal field names and nested
fields, which was not a problem previously because the field paths
were always coming from the context of shard keys, where `.` is
unambiguous but now is with FLE1 nested fields).

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-e2e-tests/tests/in-use-encryption.test.ts
```diff
@@ -382,50 +392,66 @@ describe('FLE2', function () {
...
+            await footer.waitForDisplayed({ reverse: true, timeout: 1000 });
+          } catch (err) {
+            if (
+              mode === 'unindexed' &&
+              (await footer.getText()) ===
+                'Found indexed encrypted fields but could not find __safeContent__'
+            ) {
+              return; // This version is affected by SERVER-68065 where updates on unindexed fields in QE are buggy.
+            }
+          }
```
alenakhineika, line 442: So if the field is encrypted (listed in `fleEncryptedFieldsMap`), but does not have `queries: { queryType: 'equality' }` property specified in `fleEncryptedFieldsMap` the filed should be editable but not searchable?

packages/compass-crud/src/stores/crud-store.js
```diff
@@ -513,38 +518,58 @@ const configureStore = (options = {}) => {
...
+            knownSchemaForCollection.hasSchema
+          ) {
+            // SERVER-66662 blocks writes of __safeContent__ for queryable-encryption-enabled
+            // collections. We remove it unless it was edited, in which case we assume that the
+            // user really knows what they are doing.
+            delete object.__safeContent__;
+          }
         }
 
+        const query = doc.getQueryForOriginalKeysAndValuesForSpecifiedKeys(
```
alenakhineika, line 568: I sometimes think about whether or not I can use long function names, but omg this is such a beautiful and self-explanatory name! Won't hesitate in the future :)

packages/compass-e2e-tests/tests/in-use-encryption.test.ts
addaleax: > should be editable but not searchable

Yes :+1:  a field without `queryType: 'equality'` (or, in the future, `queryType: 'range'`) is not searchable by definition, but the feedback from the QE product team was that it would be better for these fields to be editable (and accept the caveat that updates to these fields won’t be able to verify that the field hasn’t been updated in the background)

alenakhineika (approved):
🔥

alenakhineika:
I have noticed that we don't have tests for 6.0 on evergreen. Did we completely exclude them temporarily or this is some CI issue?

addaleax:
@alenakhineika `test-packaged-app-60x-enterprise` runs on RHEL on Windows, so there is some coverage at least (unless I’m misunderstanding and you’re asking about something else :slightly_smiling_face: )

alenakhineika:
I just don't see this job here: https://spruce.mongodb.com/version/62d15afe562343327cfd33ce/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=%5Erhel%24

I can see all verity of 4x and 5x, but not 60x.

addaleax:
This is what I can see at the link above:

Screenshot from 2022-07-15 15-20-54

The 60x task is the last one in that list (and hopefully my most recent push here actually made things :green_circle: :sweat_smile:)

alenakhineika:
Looks like evergreen hides something from me 😄

Screenshot 2022-07-15 at 15 23 38

alenakhineika:
Lol there are pages! Mistery solved.

PR #3266, opened 2022-07-15
Revert "Revert "chore: Use the new node-runtime-worker-thread from mongosh COMPASS-5767""

lerouxb:
Reverts mongodb-js/compass#3264


addaleax:
Follow-up PR in https://github.com/mongodb-js/compass/pull/3301

PR #3278, opened 2022-07-21
chore(compass-query-bar): styling improvements for feature flagged query bar

Anemy:
Syncing with Claudia on Monday on styles, figured I'd make some of the improvements mentioned in slack first. There will likely be a follow up pr.

Run with feature flag:
```bash
```

This includes:
- Keeping the actions at the top when the filter input is expanded
- Lighter border color
- Better font size setting (it looks like it was previously offsetting the scroll).

In the screenshots keep in mind the background will be white after the crud toolbar changes.

Before:
Screen Shot 2022-07-21 at 6 40 36 PM
Screen Shot 2022-07-21 at 6 40 28 PM
Screen Shot 2022-07-21 at 6 40 43 PM

After (PADDING IN SCREENSHOTS IS OUT OF DATE):
Screen Shot 2022-07-21 at 6 35 10 PM
Screen Shot 2022-07-21 at 6 34 06 PM
Screen Shot 2022-07-21 at 6 34 19 PM


packages/compass-query-bar/src/components/option-editor.tsx
```diff
@@ -144,6 +144,7 @@ export const OptionEditor: React.FunctionComponent<OptionEditorProps> = ({
       options={editorSettings}
       completer={completer.current}
       placeholder={placeholder}
+      fontSize={12}
```
mcasimir, line 147: on a similar note, why this is hardcoded to 12?

packages/compass-components/src/constants/mongodb-ace-theme-query.ts
```diff
@@ -111,6 +111,7 @@ background: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAACCAYAAACZg
 const mongodbAceThemeQueryCssText = `
 .ace-mongodb-query .ace_scroller {
 margin: 0px 6px;
+line-height: 14px;
```
mcasimir, line 114: Are all these hardcoded `px` all right here? No need to worry in this PR but feels a bit weird that we are not using variables, don't we have other components/texts that are meant to have the same fonts, spacings, colors and border-radius?

mcasimir (approved):
Cool, should we also align the inputs? Takes a bit more effort to scan the form when the fields are misaligned like that

packages/compass-components/src/constants/mongodb-ace-theme-query.ts
Anemy: This file has a lot of them, I've added updating the ace themes to the new brand in the tech design. I'm thinking we can update this theming to use variables like palette and spacing then.

packages/compass-query-bar/src/components/option-editor.tsx
Anemy: It's the font size we're using for the query bar ace editor. There's may somewhere in compass-components we could share it from. I'll update in another pr to share it with one other place in the query bar we're using this. (Merging this pr now)

PR #3285, opened 2022-07-25
chore: add BSON doc editor component, use for FLE2-related fields COMPASS-5949

addaleax:
This partially addresses COMPASS-5949 and might be useful as a starting
point for what approach we want to take here.

Currently, this creates a new `BSONDocumentEditor` component
specifically for editor instances that are targeting BSON documents
(which most of our editors do), adds an EJSON/Shell syntax switch
to it, and centralizes parsing/JS serialization in a single location.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-components/package.json
+++ b/packages/compass-components/package.json
@@ -77,6 +77,7 @@
     "ace-builds": "^1.4.3",
     "mongodb-ace-autocompleter": "^1.1.1",
     "mongodb-ace-mode": "^1.8.0",
+    "mongodb-query-parser": "^2.4.6",
     "react": "^16.14.0",
     "react-ace": "^9.5.0"
   },
--- a/packages/compass-components/src/components/bson-document-editor.tsx
+++ b/packages/compass-components/src/components/bson-document-editor.tsx
@@ -1,0 +1,180 @@
+import React, { useRef, useState } from 'react';
+import { css } from '@leafygreen-ui/emotion';
+import { spacing } from '@leafygreen-ui/tokens';
+import type { EditorProps } from './editor';
+import { Editor } from './editor';
+import queryParser from 'mongodb-query-parser';
+import { EJSON } from 'bson';
+import { useElementParentHoverState } from '../utils/use-element-parent-hover-state';
+import { Button } from './leafygreen';
+
+type BSONValue = any;
+
+type BSONDocumentEditorProps = Omit<EditorProps, 'variant'> & {
+  variant?: never;
+  onChangeValue?: (
+    value: BSONValue,
+    error: Error | null,
+    variant: 'EJSON' | 'Shell',
+    text: string,
+    event?: any
+  ) => void;
+} & (
+    | {
+        initialValue?: never;
+      }
+    | {
+        text?: never;
+        initialValue?: BSONValue;
+      }
+  );
+
+const actionsGroupContainer = css({
+  position: 'absolute',
+  display: 'flex',
+  gap: spacing[2],
+  width: '100%',
+  top: spacing[3],
+  right: spacing[3],
+  paddingLeft: spacing[3],
+  paddingRight: spacing[3],
+  pointerEvents: 'none',
+  zIndex: 100,
+});
+
+const actionsGroupItem = css({
+  flex: 'none',
+  pointerEvents: 'all',
+});
+
+function BSONDocumentEditor({
+  initialValue,
+  onChangeValue,
+  text,
+  onChangeText,
+  ...otherProps
+}: BSONDocumentEditorProps): React.ReactElement {
+  const containerRef = useRef<HTMLDivElement | null>(null);
+  const isHovered = useElementParentHoverState(containerRef);
+
+  if (!text && initialValue) {
+    text = queryParser.toJSString(initialValue);
+  }
+
+  const [currentVariant, setCurrentVariant] = useState<'EJSON' | 'Shell'>(
+    'Shell'
+  );
+  const [lastAlternateVariantRepr, setLastAlternateVariantRepr] = useState<{
+    text: string;
+    value: BSONValue;
+  } | null>(null);
+  const [currentText, setCurrentText] = useState(text ?? '');
+
+  function onChangeTextWrapper(text: string, event?: any) {
+    setCurrentText(text);
+    onChangeText?.(text, event);
+    if (onChangeValue) {
+      let value;
+      let error: Error | null = null;
+      try {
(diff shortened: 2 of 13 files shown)


packages/compass-components/src/components/bson-document-editor.tsx
```diff
@@ -0,0 +1,180 @@
...
+import { Button } from './leafygreen';
+
+type BSONValue = any;
+
+type BSONDocumentEditorProps = Omit<EditorProps, 'variant'> & {
+  variant?: never;
+  onChangeValue?: (
+    value: BSONValue,
+    error: Error | null,
+    variant: 'EJSON' | 'Shell',
```
addaleax, line 18: Logically speaking, we have another way to edit BSON documents, namely the `Document` component from compass-crud. I went back and forth a bit about integrating that one as well, but I don’t think our component structure is currently really laid out for that (e.g. `EditableDocument` itself contains the document actions group, rather than being a component that could be used as an editor straight from the beginning). I’m also not sure if this is something that would make sense from a UX perspective, but ultimately these *are* different representations of the same thing.

packages/compass-components/src/components/bson-document-editor.tsx
```diff
@@ -0,0 +1,180 @@
...
+          aria-label={
+            currentVariant === 'EJSON'
+              ? 'Switch to Shell BSON syntax'
+              : 'Switch to Extended JSON syntax'
+          }
+          data-testid="toggle-ejson-shell-button"
+          onClick={onToggle}
+          className={actionsGroupItem}
+        >
+          {currentVariant === 'EJSON' ? '()' : '$'}
```
addaleax, line 167: It’s pretty hard to concisely communicate what this button does, esp. since for a lot of inputs the EJSON and Shell syntaxes are actually the same.

packages/compass-components/src/components/bson-document-editor.tsx
```diff
@@ -0,0 +1,180 @@
...
+        </Button>
+      </div>
```
addaleax, line 169: This is inspired by the `DocumentActionsGroup` used in compass-crud.

packages/compass-components/src/components/bson-document-editor.tsx
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  const [currentVariant, setCurrentVariant] = useState<'EJSON' | 'Shell'>(
+    'Shell'
+  );
+  const [lastAlternateVariantRepr, setLastAlternateVariantRepr] = useState<{
+    text: string;
+    value: BSONValue;
+  } | null>(null);
+  const [currentText, setCurrentText] = useState(text ?? '');
```
addaleax, line 71: Never quite sure about whether I’m doing the right things state-management-wise. :)

packages/compass-components/package.json
```diff
@@ -77,6 +77,7 @@
     "ace-builds": "^1.4.3",
     "mongodb-ace-autocompleter": "^1.1.1",
     "mongodb-ace-mode": "^1.8.0",
+    "mongodb-query-parser": "^2.4.6",
```
addaleax, line 80: This pulls in acorn, amongst other things. Are we okay with adding it as a dependency of this package? Should this be put into a separate package?

gribnoysup: I think it's sorta unavoidable and would be a thing at some point anyway. Also the benefits of having this stuff in one place outweigh any possible downsides

packages/compass-components/src/components/bson-document-editor.tsx
```diff
@@ -0,0 +1,180 @@
...
+          })}
+        ></span>
+        <Button
+          size="xsmall"
+          title={
+            currentVariant === 'EJSON'
+              ? 'Switch to Shell BSON syntax'
+              : 'Switch to Extended JSON syntax'
+          }
+          aria-label={
```
gribnoysup, line 158: Maybe add a `title` in addition to the label so that you can also see the hint when you hover over the button

packages/compass-components/src/components/bson-document-editor.tsx
gribnoysup: The editor part of it is in compass-components and separate from the whole card: https://github.com/mongodb-js/compass/blob/main/packages/compass-components/src/components/document-list/document.tsx

There is something there that is sort of an issue: as this was modeled around previous editor, there is no component interface for edits, no `onChange` callback prop, only hadron-document events to subscribe to (but if we wanted to, I think this could be integrated to the current flow)

packages/compass-components/src/components/bson-document-editor.tsx
gribnoysup: Oh, sorry, don't know how I missed that it's literally one line above 🤦

packages/compass-components/src/components/bson-document-editor.tsx
```diff
@@ -0,0 +1,180 @@
...
+          value = EJSON.deserialize(parsed, { relaxed: false });
+        }
+      } catch (err) {
+        error = err as Error;
+      }
+      onChangeValue(value, error, currentVariant, text, event);
+    }
+  }
+
+  function onToggle() {
```
gribnoysup, line 104: An interesting corner case here: if I first break syntax, switch, then fix the syntax and switch again, the switcher goes "out of sync" with content and transforms the value into an unexpected format

Kapture 2022-08-04 at 14 00 10

packages/compass-components/src/components/bson-document-editor.tsx
gribnoysup: I think this looks alright, maybe one thing worth noting is that the way it's currently implemented this `text` property is more like `initialText` (or `defaultText` if we want to follow React controlled / uncontrolled components patterns)

packages/compass-components/src/components/bson-document-editor.tsx
```diff
@@ -0,0 +1,180 @@
...
+const actionsGroupItem = css({
+  flex: 'none',
+  pointerEvents: 'all',
+});
+
+function BSONDocumentEditor({
+  initialValue,
+  onChangeValue,
+  text,
+  onChangeText,
```
gribnoysup, line 54: Do we use this anywhere (or can we see ourselves using it)? I am wondering if we can simplify the interface by only having one `onChange` here instead of two separate ones

PR #3295, opened 2022-07-28
feat(sidebar): COMPASS_SHOW_NEW_SIDEBAR feature flag COMPASS-5959 COMPASS-5960

lerouxb:
This renames the existing sidebar components dir and makes a new one. Then it has a new feature flag `COMPASS_SHOW_NEW_SIDEBAR`. If set to `true` it will use the new folder, otherwise the old one.

Then I moved the ResizableSidebar from compass-connections to compass-components.

Finally the new sidebar uses ResizableSidebar (so it is the same between the main sidebar and the connections sidebar) and I started copying across the parts that are already leafygreenified. Which turned out to only be compass-databases-navigation for now.

Oh and I typescripted the new sidebar (which is really just an outline) and the sidebar-databases-navigation.ts file.

The plan here on out is to do one PR per TODO (ie. one for each thing that goes into the sidebar). By either copying it from components-legacy to components and changing it to be using leafygreen or re-implementing it to use leafygreen. Whichever is easiest.

By the end we'll also have to port the sidebar functionality where it can collapse as ResizableSidebar (and therefore the connections sidebar) doesn't have that functionality.


packages/compass-sidebar/src/typings.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module '@mongodb-js/mongodb-redux-common/app-registry';
```
lerouxb, line 1: Copy/pasted this from elsewhere, but it is probably about time that we type that module?

packages/compass-components/src/components/resizeable-sidebar.tsx
```diff
@@ -1,10 +1,7 @@
 import React, { useCallback, useState } from 'react';
-import {
-  ResizeHandle,
-  ResizeDirection,
-  uiColors,
-  css,
-} from '@mongodb-js/compass-components';
+import { ResizeHandle, ResizeDirection, css } from '../index';
+
+import { uiColors } from '@leafygreen-ui/palette';
```
lerouxb, line 4: I got the weirdest errors when trying to import this from ../index along with everything else where it gets re-exported. That suddenly went away when I imported straight from leafygreen-ui/palette..

packages/compass-sidebar/src/typings.d.ts
mcasimir: Amazing how new modules still pops up after years .. does someone know why is that not part of this repo?

Also it seems to only have 1 file, maybe can be moved in another package that already exists? https://github.com/mongodb-js/redux-common/blob/master/app-registry.js

PR #3298, opened 2022-07-28
feat: remove in the background setting and use editor for collation COMPASS-5952 COMPASS-5980

alenakhineika:
## Description

- [COMPASS-5952](https://jira.mongodb.org/browse/COMPASS-5952) Remove the "Build index in the background" setting
- [COMPASS-5980](https://jira.mongodb.org/browse/COMPASS-5980) Replace collation options dropdowns with an editor

Screenshot 2022-07-28 at 12 53 07

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


mcasimir:
@alenakhineika If I read the docs correctly, the "build index in background" had an effect till 4.2, which means users on 4.0 won't be able to use it. @mmarcon should we still show that if `serverVersion < 4.2` or are you fine dropping that setting entirely?

alenakhineika:
I think during kick-off, we discussed removing it entirely, but @mmarcon please confirm that.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.jsx
```diff
@@ -513,7 +500,7 @@ const mapStateToProps = (state) => ({
   isPartialFilterExpression: state.isPartialFilterExpression,
   partialFilterExpression: state.partialFilterExpression,
   isCustomCollation: state.isCustomCollation,
-  collation: state.collation,
+  collationString: state.collationString,
```
mcasimir, line 503: Inside the state that is not really a string anyway or?

I'm not sure this renaming is necessary, to me, it doesn't make it too much clearer than before, and if the value of that is not really a string it makes it a bit confusing.

I like that the action is called `collationStringChanged` though, that is perfect, but I'd never expect `state.collationString` not to be a string.

What about we keep the module name as `collation.js`, the state as `.collation` and the action as `COLLATION_STRING_CHANGED`?

packages/compass-indexes/src/modules/create-index/index.js
```diff
@@ -165,14 +160,13 @@ export const createIndex = () => {
...
     const options = {};
-    options.background = state.isBackground;
     options.unique = state.isUnique;
     // The server will generate a name when we don't provide one.
     if (state.name !== '') {
       options.name = state.name;
     }
     if (state.isCustomCollation) {
-      options.collation = state.collation;
+      options.collation = state.collationString.value;
```
mcasimir, line 169: I'm not sure about storing the `.value` in the state, is not really used as a react state, you can just `parse(state.collationString.text)` and avoid introducing statefulness here.

To me the same goes for `isValid`: we can simplify everything and store only the string, parse the option here and handle the parsing exception nicely.

We stopped doing this kind of on-the-fly validations everywhere. It simplifies the code and having them is not really helpful cause it would trigger multiple times while the user is typing and give a negative signal while there is nothing actually bad going on.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.jsx
addaleax: Just for some extra context, https://github.com/mongodb-js/compass/pull/3285 would make it so that the BSON document editor component would update its user by passing the current value, not (only) the current text content, so it might be good to keep that in mind here

alenakhineika: I have aligned it with what we did in aggregations: https://github.com/mongodb-js/compass/blob/main/packages/compass-aggregations/src/modules/collation-string.ts

There are a couple of places already where this collation is handled differently. I think we should use one way everywhere and then improve naming.

packages/compass-indexes/src/modules/create-index/index.js
alenakhineika: Same as above.

mcasimir: Actually thinking it twice I'm not too sure, cause really none of the state that was here seems a fit for redux (we probably don't want to change it right away?). Is all of a volatile form state that doesn't seem very global to me. Regardless, probably none of what I'm saying makes sense, @gribnoysup recently reviewed a ton of best practices etc, maybe he can give us a pointer.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.jsx
mcasimir: ah nice, does that keep the string as internal value?

alenakhineika: It will be a server error shown to a user since you specified a custom collation, but the validated string returned null. Something like `The field 'collation' must be an object, but got null`.

alenakhineika: Renamed `collationString` to `collationInfo` to reduce confusion about what it stores.

mmarcon:
> I think during kick-off, we discussed removing it entirely, but @mmarcon please confirm that.

@alenakhineika @mcasimir 4.0 went EOL last April and looking at the analytics, it seems that only 0.4% of the users who have connected to a 4.0.x server within 1 hour before creating an index created it with `background=true`. I am ok with dropping that checkbox.

@jaoppenheim FYI

mcasimir (approved):

packages/compass-indexes/src/modules/create-index/index.js
gribnoysup: > We stopped doing this kind of on-the-fly validations everywhere

Is it something we agreed on and have documented somewhere? This seems like an important thing to be aware of, but it's the first time I hear about this 😅 Should we add this to "component conventions" or something like that? Also I wonder how this aligns with what we heard on the last design sync where people were asking for syntax errors highlights right when they are typing in code editor inputs, is it something that should apply only to very explicit forms like "create / edit" modals and connection form?

> Is all of a volatile form state that doesn't seem very global to me

That would always depend on the use case, but yes, Redux docs mention that form state [should not be part of the store](https://redux.js.org/style-guide/#avoid-putting-form-state-in-redux). FWIW this only applies to very straightforward cases where you have an actual form that is totally isolated from everything else and which state can be completely discarded on submit or close / cancel. Which does match this modal case, but would require a bunch of additional refactoring: moving only one value to the component state and keeping all the other input state still in the store doesn't make much sense to me and will create a bigger mess from my perspective.

I do think though that we can avoid producing the parsed value on every edit and keeping it in the state and instead only keep the collation string in the redux store. I see that this was modeled following the aggregation plugin code, but the compass-aggregation case is way different: we do show validation error there right away when you type and so we have to parse / validate on edit _and_ this state is used in multiple different actions across the plugin code and so it needs to be "globally" available. I see that here you don't use the validation information at all and the parsed value is only used when form is submitted, so that's probably the only place where you would want to parse / validate the value. Also other form values here seem to [dispatch a `handleError` action](https://github.com/mongodb-js/compass/blob/a364e83f811047ede797cdee5dc3d9e29cce1ac4/packages/compass-indexes/src/modules/create-index/index.js#L156) when validation fails, it's probably something you would want to do in this action for collation too.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.jsx
gribnoysup: The error in this case does look a bit weird though 🤔 I have a string value but am shown that there is an issue with collation because it was `null` which totally would prompt me to open an issue and ask why my collation is not being sent to the server

image

We definitely can decide that we stop doing our own validation and rely on the server, but then we should also stop modifying payloads that we pass to server commands probably, otherwise the errors can be confusing

packages/compass-indexes/src/modules/create-index/index.js
```diff
@@ -165,14 +160,13 @@ export const createIndex = () => {
...
     const options = {};
-    options.background = state.isBackground;
     options.unique = state.isUnique;
     // The server will generate a name when we don't provide one.
     if (state.name !== '') {
       options.name = state.name;
     }
     if (state.isCustomCollation) {
-      options.collation = state.collation;
+      options.collation = state.collationInfo.value;
```
gribnoysup, line 169: As I mentioned in https://github.com/mongodb-js/compass/pull/3298#discussion_r932990110 you probably don't need to keep the value in the state and always parse / validate the value on edit, this is the only place where value is used and so it would make sense to do the parsing / validation part here. You probably might want to add validation action dispatch similar to all the other fields in the form here too, so something like this:

```suggestion
      options.collation = isCollationValid(collationString);
      if (!options.collation) {
        dispatch(handleError('Collation is not valid'))
        return;
      }
```

gribnoysup (commented):
Feel free to merge as-is, but I do think that you can simplify this a bit

packages/compass-indexes/src/modules/create-index/index.js
```diff
@@ -151,12 +147,19 @@ export const createIndex = () => {
...
-    // check for errors
+    // Check for field errors.
     if (state.fields.some((field) => field.name === '' || field.type === '')) {
       dispatch(handleError('You must select a field name and type'));
       return;
     }
 
+    // Check for collaction errors.
+    const collation = queryParser.isCollationValid(state.collationString);
+    if (state.isCustomCollation && collation === false) {
```
gribnoysup, line 158: Seems like the return types here are pretty messed up and we should fix that upstream, but probably would make sense to handle this here too for the time being. Return values that indicate that collation is not valid are not just `false`, but falsy it seems:

image

So something like that should probably cover it a bit better:

```suggestion
    if (state.isCustomCollation && !collation) {
```

gribnoysup (approved):
One small thing, otherwise lgtm

PR #3306, opened 2022-08-01
chore(compass-indexes): css in js COMPASS-5940

mabaasit:
chore(compass-indexes): css in js

This PR removes `.less` styles and styles in js. In addition to that, it
1. uses LG badge to represent index type and its properties. 
2. uses LG tooltip.

Before:
Before

After:
After

Video:
https://user-images.githubusercontent.com/1305718/182066122-a8816a52-38bf-4afd-a02d-f437b1050ca0.mov

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-components/src/components/index-keys.spec.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { render, cleanup, screen, within } from '@testing-library/react';
+import { expect } from 'chai';
+import type { IndexDirection } from 'mongodb';
```
gribnoysup, line 4: AFAIU the idea of compass-components is not to know about driver, mongodb, server, etc explicitly. This is also not used anywhere outside of compass-indexes package. Is there any particular reason we are putting this in compass-components?

/cc @Anemy any thoughts?

packages/compass-indexes/src/components/index-header-column/index-header-column.jsx
```diff
@@ -1,11 +1,15 @@
...
-
-import classnames from 'classnames';
-import styles from './index-header-column.module.less';
+import { css, cx, uiColors } from '@mongodb-js/compass-components';
 
 import { DEFAULT, DESC, ASC } from '../../modules/indexes';
 
+const sortStyles = css({
+  display: 'inline-block',
+  fontSize: '13px',
```
alenakhineika, line 9: Would it be possible to avoid a hardcoded number here? Maybe to use something from typography or value from `compass-font-sizes`.

packages/compass-indexes/src/components/index-header/index-header.jsx
```diff
@@ -1,9 +1,14 @@
...
 import PropTypes from 'prop-types';
+import { css } from '@mongodb-js/compass-components';
 import IndexHeaderColumn from '../index-header-column';
 
-import classnames from 'classnames';
-import styles from './index-header.module.less';
+const containerStyles = css({
+  textAlign: 'left',
+  'th': {
+    padding: '12px 24px',
```
alenakhineika, line 9: Also maybe here we could replace these hardcoded paddings? Like use spacing or something.

packages/compass-components/src/components/index-keys.spec.tsx
mcasimir: It makes more sense to me to move this inside the index plugin unless we are already using it somewhere else, not sure if we have similar code already, in which case, we could probably first see if we can decompose this component in parts that are a bit more flexible to reuse.

gribnoysup: Yeah, this sounds about right to me. At the very least I would really tried to avoid having a driver as a dependency here: there is not reason for component interface to be directly attached to the driver implementation

Anemy: I'd agree let's avoid the driver implementation specifics here and the driver import if possible. As a rule of thumb I think unless we know something will be used in more packages than one we probably don't want it in Compass components. I'm thinking for this the index badge might also be used in things like explain output. That being said this component isn't much more than a wrapper around a badge with an icon and text so I think it would be find to build it in both places with those building blocks.

packages/compass-indexes/src/components/index-header-column/index-header-column.jsx
mabaasit: i ended up removing this as it didn't have an effect. plus, in the next PR as I'll use LG table for this, this whole will be gone.

packages/compass-components/src/components/index-keys.tsx
```diff
@@ -0,0 +1,47 @@
...
+import Icon from '@leafygreen-ui/icon';
+import { spacing } from '@leafygreen-ui/tokens';
+import type { IndexDirection } from 'mongodb';
+
+type IndexKeysProps = {
+  keys: Record<string, IndexDirection>;
+};
+
+const containerStyles = css({
+  marginTop: spacing[1],
```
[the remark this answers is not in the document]
mabaasit: Yes! great catch! i removed that and let the components that use this component, style it as needed.

packages/compass-components/src/components/index-keys.spec.tsx
mabaasit: This component is now being used in 2 places: here (in indexes) and aggregation explain. it will also be used in the Explain tab as well. Not certain if it makes sense to have it in three different places. 
I kept it in compass-components while removing the dependency. Is there a better way to handle this?

packages/compass-components/src/components/index-keys.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { spacing } from '@leafygreen-ui/tokens';
+
+type IndexDirection = -1 | 1 | '2d' | '2dsphere' | 'text' | 'geoHaystack' | number
+
+type IndexKeysProps = {
+  keys: Record<string, IndexDirection>;
+};
+
+const containerStyles = css({
+  '*:not(:last-child)': {
```
gribnoysup, line 14:
```suggestion
  '&:not(:last-child)': {
```

packages/compass-components/src/components/icon-badge.tsx
```diff
@@ -0,0 +1,51 @@
...
+}: IconBadgeProps): React.ReactElement {
+  return (
+    <Badge data-testid={dataTestId} variant={variant}>
+      <div className={contentStyles}>
+        {text}
+        {
+          <button className={iconButtonStyles} onClick={onClick}>
+            <Icon glyph={icon} />
+          </button>
+        }
```
gribnoysup, line 45: There is no label for this button

packages/compass-components/src/components/icon-badge.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+type IconBadgeProps = {
+  text: string;
+  icon: string;
+  dataTestId?: string;
+  onClick?: () => void;
+  variant?: Variant;
+};
+
+function IconBadge({
```
gribnoysup, line 30: If this is a shared, reusable component, should we use it in other places where IconBadge is used? For example in collections list?

image

Or is this very specific `BadgeWithIconButton` or something?

packages/compass-components/src/components/icon-badge.tsx
```diff
@@ -0,0 +1,51 @@
...
+  display: 'flex',
+  alignItems: 'center',
+  gap: spacing[1],
+});
+
+type IconBadgeProps = {
+  text: string;
+  icon: string;
+  dataTestId?: string;
+  onClick?: () => void;
```
gribnoysup, line 26: What's the use-case of having an optional onClick if this component always renders a button?

packages/compass-components/src/components/icon-badge.tsx
gribnoysup, line 45:
```suggestion
        <button className={iconButtonStyles} onClick={onClick}>
          <Icon glyph={icon} />
        </button>
```

packages/compass-components/src/components/icon-badge.tsx
mabaasit: I ended up removing this component to keep everything simple

packages/compass-components/src/components/index-keys.tsx
```diff
@@ -0,0 +1,54 @@
...
+};
+
+const IndexKeys: React.FunctionComponent<IndexKeysProps> = ({ keys }) => {
+  return (
+    <div className={containerStyles}>
+      {Object.entries(keys).map(([keyName, direction], listIndex) => (
+        <Badge
+          data-testid={`${keyName}-key`}
+          variant={Variant.LightGray}
+          key={listIndex}
```
gribnoysup, line 43: This can probably be the keyName instead, no need to use the index here, this also better represent the items you are passing here

packages/compass-indexes/src/components/drop-column/drop-column.jsx
```diff
@@ -52,15 +55,15 @@ class DropColumn extends PureComponent {
...
-      <td className={classnames(styles['drop-column'])}>
+      <td className={containerStyles}>
         {this.isDroppable() ? (
-          <button
-            className="drop-btn btn btn-default btn-sm"
-            type="button"
+          <IconButton
+            aria-label="Delete Index"
             onClick={this.clickDropHandler.bind(this)}
+            data-testId={`drop-index-button-${this.props.indexName}`}
```
gribnoysup, line 63:
```suggestion
            data-testid={`drop-index-button-${this.props.indexName}`}
```

packages/compass-indexes/src/components/index-header/index-header.jsx
```diff
@@ -1,9 +1,17 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
+import { css, spacing } from '@mongodb-js/compass-components';
 import IndexHeaderColumn from '../index-header-column';
 
-import classnames from 'classnames';
-import styles from './index-header.module.less';
+const containerStyles = css({
+  textAlign: 'left',
+  th: {
```
gribnoysup, line 8: Why are these styles here and not in `IndexHeaderColumn`?

packages/compass-indexes/src/components/name-column/name-column.jsx
```diff
@@ -1,38 +1,38 @@
...
-   * Render the name column.
-   *
-   * @returns {React.Component} The name column.
-   */
   render() {
+    const indexKeys = {};
+    const indexName = this.props.index.name;
+    this.props.index.fields.serialize().forEach(({ field, value }) => {
+      indexKeys[field] = value;
+    });
```
gribnoysup, line 28: Why are we converting existing array to object to use it in the component as array again?

packages/compass-indexes/src/components/property-column/property-column.jsx
```diff
@@ -2,19 +2,37 @@ import map from 'lodash.map';
...
 
-const TOOLTIP_ID = 'index-property';
+const containerStyles = css({
+  width: 'auto',
+  paddingLeft: spacing[4],
+  paddingRight: spacing[4],
+  marginTop: spacing[1],
+  '*:not(:last-child)': {
+    marginRight: spacing[1],
+  },
```
gribnoysup, line 25: Targeting children like that is error prone and especially with css-in-js should be an exception rather than the rule. This should either be applied to children or done with a gap css property that is specifically a parent prop

packages/compass-indexes/src/components/type-column/type-column.jsx
```diff
@@ -22,70 +39,50 @@ class TypeColumn extends PureComponent {
...
-        {helpLink && (
-          <InfoSprinkle
-            helpLink={helpLink}
-            onClickHandler={this.props.openLink}
-          />
-        )}
-      </div>
-    );
+    const items = map(info, (v, k) => {
+      return <Body>{format('%s: %j', k, v)}</Body>;
```
gribnoysup, line 57: Would be nice if we could remove dependency on `util` here as this is a node.js package that is only available in the browsers through polyfills and this will be in the way for our plans to eventually disable node modules in electron renderer

packages/compass-indexes/src/components/type-column/type-column.jsx
```diff
@@ -22,70 +39,50 @@ class TypeColumn extends PureComponent {
...
+              {children}
+              <Badge variant={BadgeVariant.DarkGray}>
+                {this.props.index.type}&nbsp;
+                <button
+                  aria-label="Index type docs"
+                  className={iconButtonStyles}
+                  onClick={() => this.props.openLink(helpLink)}
+                >
+                  <Icon glyph="InfoWithCircle" />
+                </button>
```
gribnoysup, line 79: This can (and should) be a link, not a button

packages/compass-components/src/components/index-keys.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { spacing } from '@leafygreen-ui/tokens';
+
+type IndexDirection =
+  | -1
+  | 1
+  | '2d'
+  | '2dsphere'
+  | 'text'
+  | 'geoHaystack'
+  | number;
```
alenakhineika, line 14: The index-model describes one indexes: https://github.com/mongodb-js/compass/tree/main/packages/index-model#index-field-types

But there is a bit different list in the create modal: https://github.com/mongodb-js/compass/blob/main/packages/compass-indexes/src/components/create-index-field/create-index-field.jsx#L17

@mcasimir which one is correct? to they can be different?

packages/compass-indexes/src/components/property-column/property-column.jsx
mabaasit: thank you! i really like the idea of gap! let me try that.

packages/compass-components/src/components/index-keys.spec.tsx
mcasimir: This is just a thought. If this component is reused then there is nothing wrong with keeping the code in a shared place like this. However, when we decide to generalize a solution, we may want to try to reframe it in a way that lets us share what is actually reused and that won't change with context.

I'm not suggesting doing anything specific in this case, but just to give an example: I would consider sharing just sharing a single `IndexKey` component instead of a "list" as is pretty straightforward to build the list where is used and the layout may depend on the context where is used. 

By the same token, there is probably not much that is special about the badge, so .. again is just an example, we could consider only having an "IndexTypeIcon". 

>  it will also be used in the Explain tab

I really hope that that won't be a 3rd place cause all the code for explains should be in the same package eventually.

packages/compass-indexes/src/components/indexes/indexes.jsx
```diff
@@ -14,7 +14,28 @@ import IndexHeader from '../index-header';
...
+  width: '100%',
+  padding: spacing[3],
+  paddingTop: 0,
+  flexGrow: 1,
+  overflow: 'auto',
+});
+const indexesTableStyles = css({
+  tableLayout: 'auto',
+  borderCollapse: 'separate',
+  borderSpacing: '0 3px',
```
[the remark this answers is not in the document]
mabaasit: as mentioned yesterday in standup, i've a follow up PR to use LG table component, where i'll take care of these styles. cc @gribnoysup

packages/compass-indexes/src/components/name-column/name-column.jsx
mabaasit: due to aggregation explain where we have an object, i tried to align this with that implementation. however, after your comment, i made `keys` an array of object and aligned aggregation explain with that.

gribnoysup:
Seems like in the previous version we were doing a better job of allowing table columns to collapse resize for the small window size:

image

In this implementation you get a very wide first column, and very small ones after it (with more content not fitting on the screen)

image

Looks like a bug to me

packages/compass-components/src/components/index-keys.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const IndexDirectionIcon = ({ direction }: { direction: IndexDirection }) => {
+  return direction === 1 ? (
+    <Icon glyph="ArrowUp" />
+  ) : direction === -1 ? (
+    <Icon glyph="ArrowDown" />
+  ) : (
+    <>({String(direction)})</>
+  );
+};
```
gribnoysup, line 36: Without a proper label / title, VO in browser will read this as "Arrow Up Icon" (default title that lg icon will have) which is really not helpful if you are using assistive tech:

image (5)

At the very least let's add a proper label here. Even better would be to use aria-hidden in combination with VisuallyHidden label to provide a meaningful description for this badge (or the whole cell even), something like this:

image (6)

packages/compass-components/src/components/index-keys.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const containerStyles = css({
+  display: 'flex',
+  gap: spacing[1],
+});
+
+const IndexDirectionIcon = ({ direction }: { direction: IndexDirection }) => {
+  return direction === 1 ? (
+    <Icon glyph="ArrowUp" />
```
gribnoysup, line 30: This looks big compared to the font, we probably want to use small size icon here?
```suggestion
    <Icon glyph="ArrowUp" size="small" />
```

packages/compass-components/src/components/index-keys.tsx
mcasimir: As I was saying on slack, this is one of those situations where being more tolerant with types could be better. 

The driver type (which we copy/pasted here) is meant for the `createIndex` option and while the values returned by the server should be aligned, that may not be true with a more recent server, which could return new unexpected values. 

Restricting the set of possible values like this is more dangerous than anything cause we would rely on something that is not actually true.

I recommend changing this type to `IndexDirection = number | string` and handling possible corner cases appropriately: for example just render the index name without any icon if we get nonsensical values like `2` or an empty string.

Another comment that I have here is that this type could be named a bit better since this is not really only a direction, for example we could consider calling it `IndexTypeOrDirection`

packages/compass-indexes/src/components/drop-column/drop-column.jsx
mcasimir: Should we consider adding something like this? https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/consistent-data-testid.md

mcasimir: CC: @gribnoysup

mabaasit: created a follow up: https://jira.mongodb.org/browse/COMPASS-6012

packages/compass-indexes/src/components/name-column/name-column.jsx
```diff
@@ -1,39 +1,56 @@
...
-          >
-            {this.props.index.name}
-          </div>
-          <IndexDefinitionType index={this.props.index} />
-        </div>
+      <td>
+        <Accordion
+          text={indexName}
+          data-testid="index-field-name"
+          aria-label={`${indexName}-index`}
```
Anemy, line 37: I think `aria-label` is something readable, maybe we can remove the `-index` or make it `{Index ${indexName}}`?
```suggestion
          aria-label={indexName}
```
```suggestion
          aria-label={`Index ${indexName}`}
```

packages/compass-indexes/src/components/property-column/property-column.jsx
```diff
@@ -24,113 +42,78 @@ class PropertyColumn extends PureComponent {
...
-  /**
-   * Render the property column.
-   *
-   * @returns {React.Component} The property column.
-   */
   render() {
-    const properties = map(this.props.index.properties, (prop) => {
-      return this.renderProperty(prop);
-    });
+    const properties = map(
```
Anemy, line 109: Maybe not the best time for this so feel free to ignore, but we might be able to remove lodash map here and use default js.
```suggestion
    const properties = this.props.index.properties?.map(
```
Looks like lodash map is used in a few other parts of this package so we can't remove it as a dependency just yet, but reducing usage will get us there eventually.

packages/compass-indexes/src/components/type-column/type-column.jsx
```diff
@@ -22,70 +39,52 @@ class TypeColumn extends PureComponent {
...
-        {this.renderType()}
-        <ReactTooltip id={TOOLTIP_ID} data-testid="tooltip-index-field-type" />
+      <td data-testid="index-field-type">
+        <Tooltip
+          data-testid="tooltip-index-field-type"
+          enabled={this.canRenderTooltip()}
+          trigger={({ children, ...props }) => (
+            <span {...props}>
+              {children}
+              <Badge variant={BadgeVariant.DarkGray} className={badgeStyles}>
```
Anemy, line 72: Previously we were showing different colors for the badges, are we planning to re-add them?
before:
Screen Shot 2022-08-15 at 4 50 09 PM

after:
Screen Shot 2022-08-15 at 4 50 04 PM

packages/compass-indexes/src/components/usage-column/usage-column.jsx
```diff
@@ -43,43 +28,30 @@ class UsageColumn extends PureComponent {
...
-          </div>
-          {this.renderSince()}
-        </span>
+      <td>
+        <Tooltip
+          trigger={({ children, ...props }) => (
+            <span {...props}>
+              {children}
+              <Body>
+                {usage}&nbsp;({this.renderSince()})
```
Anemy, line 48: Do we want to hide the `()` around the `renderSince` date when there isn't any usage:
Screen Shot 2022-08-15 at 4 52 50 PM

Anemy (commented):
Nice, looking good, left a couple comments, I think the only possible blockers are the usage date label and the badge colors.

packages/compass-indexes/src/components/type-column/type-column.jsx
mabaasit: its be design that we are removing coloured badges.

Anemy: Ah gotcha, I'll miss them 😿 
Looks good though.

Anemy (approved):
lgtm! nice

PR #3311, opened 2022-08-01
fix: create index combobox input field crashes on regex characters COMPASS-5992

alenakhineika:
## Description

As a workaround for https://jira.mongodb.org/browse/PD-2250 we copy the Combobox component to `compass-componets` to allow users to create indexes with special characters eg [wildcard indexes](https://www.mongodb.com/docs/compass/current/indexes/#wildcard-indexes) (field.$**).

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-components/src/components/leafy-green/combobox/combobox-option.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+    const handleOptionClick = useCallback(
+      (e: React.SyntheticEvent) => {
+        e.stopPropagation();
+        // If user clicked on the option, or the checkbox
+        // Ignore extra click events on the checkbox wrapper
+        if (
+          e.target === optionRef.current ||
+          (e.target as Element).tagName === 'INPUT' ||
+          (e.target as Element).tagName === 'SPAN' // TODO(alenakhineika): investigate this workaround later.
```
alenakhineika, line 99: Here for some reason, the event did not happen on the span name of the li element.

mcasimir: Seems fine, would it be possible to provide a link to the original code in the comment?

packages/compass-components/src/components/leafy-green/combobox/chip.tsx
```diff
@@ -0,0 +1,223 @@
+import React, { useContext, useEffect, useMemo, useRef } from 'react';
```
mcasimir, line 1: Just a thought .. as someone that has to work on this code without much context I think I would benefit from either a README inside the folder or even just a comment on top of all of the files here that could provide some backstory and point to the ticket and to the original code we copied

packages/compass-components/src/components/leafy-green/combobox/util.tsx
```diff
@@ -0,0 +1,121 @@
...
+ * @param element
+ * @returns `JSX.Element`
+ */
+export const wrapJSX = (
+  str: string,
+  wrap?: string,
+  element?: string
+): JSX.Element => {
+  if (wrap && element) {
+    const regex = new RegExp(escapeRegExp(wrap), 'gi');
```
mcasimir, line 34: Should we add a comment here to mark that we changed something from the original code?

mcasimir (approved):
Awesome, thanks for the super-quick fix! Would it be possible to also add a unit test in compass-indexes as part of this to make sure fancy index names won't break the UI again?

mcasimir (approved):

PR #3315, opened 2022-08-02
feat(compass-query-bar): update export to language button and icon

Anemy:
Syncing w/ Claudia before merging, cc @mcasimir 

Now we're using a button with the `Code` icon from https://www.mongodb.design/component/icon/example/

Before
Screen Shot 2022-08-01 at 3 23 56 PM

After
Screen Shot 2022-08-01 at 10 40 05 PM


mcasimir (approved):
Ah nice!

Anemy:
@mcasimir Ah good call updated the aggregations one as well. Still waiting to hear back from Claudia before merging.

PR #3320, opened 2022-08-03
feat(sidebar) New sidebar navtree COMPASS-5960	

lerouxb:
With COMPASS_SHOW_NEW_SIDEBAR="true", the sidebar you get after connecting will now have the light background and compass-databases-navigation (the only thing in there) will have the colours and spacing more in line with the sidebar Figma designs. There is now also always a database icon in line with the collection icons. Without the feature flag the sidebar should be unchanged.

Well. except for one thing: The little active tree item indicator escaped, but I think that's OK. This is just because I changed compass-databases-navigation in place rather than making a copy of it like I did with the rest of the sidebar.

Screenshot 2022-08-04 at 16 55 53

Screenshot 2022-08-05 at 11 54 19

### DIFF
--- a/packages/compass-components/src/components/placeholder.tsx
+++ b/packages/compass-components/src/components/placeholder.tsx
@@ -4,6 +4,7 @@ import React, { useMemo } from 'react';
 import { css, cx, keyframes } from '@leafygreen-ui/emotion';
 import { spacing } from '@leafygreen-ui/tokens';
 import { uiColors } from '@leafygreen-ui/palette';
+import { withTheme } from '../hooks/use-theme';
 
 // Ratio of showing a highlight passing through the placeholder to background color
 const scale = 4;
@@ -47,7 +48,7 @@ function getBoundRandom(min: number, max: number) {
   return Math.random() * (max - min) + min;
 }
 
-export const Placeholder: React.FunctionComponent<
+const UnthemedPlaceholder: React.FunctionComponent<
   Omit<
     React.HTMLProps<HTMLDivElement>,
     'minChar' | 'maxChar' | 'width' | 'height'
@@ -81,3 +82,7 @@ export const Placeholder: React.FunctionComponent<
     ></div>
   );
 };
+
+const Placeholder = withTheme(UnthemedPlaceholder);
+
+export { Placeholder };
--- a/packages/compass-components/src/components/resizeable-sidebar.tsx
+++ b/packages/compass-components/src/components/resizeable-sidebar.tsx
@@ -2,6 +2,7 @@ import React, { useCallback, useState } from 'react';
 import { ResizeHandle, ResizeDirection, css, cx } from '../index';
 
 import { uiColors } from '@leafygreen-ui/palette';
+import { spacing } from '@leafygreen-ui/tokens';
 
 const containerStyles = css({
   display: 'flex',
@@ -51,14 +52,17 @@ const containerStylesLight = css({
   backgroundColor: 'var(--bg-color)',
 });
 
+const initialSidebarWidth = spacing[6] * 4 - spacing[1]; // 252px
+const minSidebarWidth = spacing[4] * 9; // 216px
+
 const ResizableSidebar = ({
-  initialWidth,
-  minWidth,
+  initialWidth = initialSidebarWidth,
+  minWidth = minSidebarWidth,
   children,
   darkMode,
 }: {
-  initialWidth: number;
-  minWidth: number;
+  initialWidth?: number;
+  minWidth?: number;
   children: JSX.Element;
   darkMode: boolean;
 }): JSX.Element => {
--- a/packages/compass-components/src/components/resizeable-sidebar.tsx
+++ b/packages/compass-components/src/components/resizeable-sidebar.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useState } from 'react';
-import { ResizeHandle, ResizeDirection, css } from '../index';
+import { ResizeHandle, ResizeDirection, css, cx } from '../index';
 
 import { uiColors } from '@leafygreen-ui/palette';
 
@@ -11,8 +11,16 @@ const containerStyles = css({
   maxWidth: '80%',
   height: '100%',
   position: 'relative',
+});
+
+const containerStylesDark = css({
   background: uiColors.gray.dark3,
   color: 'white',
+});
+
+const containerStylesLight = css({
+  background: uiColors.gray.light3,
+  color: uiColors.gray.dark3,
 });
 
 const ResizableSidebar = ({
@@ -30,9 +38,15 @@ const ResizableSidebar = ({
     return Math.max(minWidth, window.innerWidth - 100);
   }, [minWidth]);
 
+
+  const useNewSidebar = process?.env?.COMPASS_SHOW_NEW_SIDEBAR === 'true';
+
   return (
     <div
-      className={containerStyles}
+      className={cx(
+      containerStyles,
+      useNewSidebar ? containerStylesLight : containerStylesDark
+      )}
       style={{
         minWidth: minWidth,
         width: width,
--- a/packages/compass-components/src/components/resizeable-sidebar.tsx
+++ b/packages/compass-components/src/components/resizeable-sidebar.tsx
@@ -1,5 +1,12 @@
 import React, { useCallback, useState } from 'react';
-import { ResizeHandle, ResizeDirection, css, cx } from '../index';
+import {
+  ResizeHandle,
+  ResizeDirection,
+  useTheme,
+  Theme,
+  css,
+  cx,
+} from '../index';
 
 import { uiColors } from '@leafygreen-ui/palette';
 import { spacing } from '@leafygreen-ui/tokens';
@@ -59,12 +66,10 @@ const ResizableSidebar = ({
   initialWidth = initialSidebarWidth,
   minWidth = minSidebarWidth,
   children,
-  darkMode,
 }: {
   initialWidth?: number;
   minWidth?: number;
   children: JSX.Element;
-  darkMode: boolean;
 }): JSX.Element => {
   const [width, setWidth] = useState(initialWidth);
 
@@ -72,11 +77,13 @@ const ResizableSidebar = ({
     return Math.max(minWidth, window.innerWidth - 100);
   }, [minWidth]);
 
+  const { theme } = useTheme();
+
   return (
     <div
       className={cx(
         containerStyles,
-        darkMode ? containerStylesDark : containerStylesLight
+        theme === Theme.Dark ? containerStylesDark : containerStylesLight
       )}
       style={{
         minWidth: minWidth,
--- a/packages/compass-components/src/components/resizeable-sidebar.tsx
+++ b/packages/compass-components/src/components/resizeable-sidebar.tsx
@@ -14,13 +14,41 @@ const containerStyles = css({
 });
 
 const containerStylesDark = css({
-  background: uiColors.gray.dark3,
-  color: 'white',
+  '--color': 'white',
+  '--bg-color': uiColors.gray.dark3,
+
+  '--icon-color': 'white',
+  // NOTE: This is for icons that only display when hovering over an item,
+  // not the color used for the icon when hovering over the icon itself.
+  '--hover-icon-color': 'white',
+
+  '--item-color': 'white',
+  '--item-color-hover': 'white',
+  '--item-color-active': 'white',
+  '--item-bg-color': uiColors.gray.dark2,
+  '--item-bg-color-hover': uiColors.gray.dark3,
+  '--item-bg-color-active': uiColors.gray.dark1,
+
+  color: 'var(--color)',
+  backgroundColor: 'var(--bg-color)',
 });
 
 const containerStylesLight = css({
-  background: uiColors.gray.light3,
-  color: uiColors.gray.dark3,
+  '--color': uiColors.gray.dark3,
+  '--bg-color': uiColors.gray.light3,
+
+  '--icon-color': uiColors.gray.dark3,
+  '--hover-icon-color': uiColors.gray.dark1,
+
+  '--item-color': uiColors.gray.dark3,
+  '--item-color-hover': uiColors.gray.dark3,
+  '--item-color-active': uiColors.green.dark2,
+  '--item-bg-color': uiColors.gray.light3,
+  '--item-bg-color-hover': uiColors.gray.light2,
+  '--item-bg-color-active': uiColors.green.light3,
+
+  color: 'var(--color)',
+  backgroundColor: 'var(--bg-color)',
 });
 
 const ResizableSidebar = ({
--- a/packages/compass-components/src/components/resizeable-sidebar.tsx
+++ b/packages/compass-components/src/components/resizeable-sidebar.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useState } from 'react';
-import { ResizeHandle, ResizeDirection, css, cx } from '../index';
+import { ResizeHandle, ResizeDirection, css } from '../index';
 
 import { uiColors } from '@leafygreen-ui/palette';
 
@@ -11,16 +11,8 @@ const containerStyles = css({
   maxWidth: '80%',
   height: '100%',
   position: 'relative',
-});
-
-const containerStylesDark = css({
   background: uiColors.gray.dark3,
   color: 'white',
-});
-
-const containerStylesLight = css({
-  background: uiColors.gray.light3,
-  color: uiColors.gray.dark3,
 });
 
 const ResizableSidebar = ({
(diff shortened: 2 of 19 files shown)


packages/compass-databases-navigation/src/collection-item.tsx
```diff
@@ -126,12 +141,15 @@ export const CollectionItem: React.FunctionComponent<
...
+      className={cx(
+        collectionItem,
+        useNewSidebar ? collectionItemNewSpacing : collectionItemOldSpacing
+      )}
       style={style}
       {...hoverProps}
     >
       <CollectionIcon type={type} />
-      <ItemLabel className={collectionItemLabel}>{name}</ItemLabel>
+      <ItemLabel className={cx(collectionItemLabel)}>{name}</ItemLabel>
```
gribnoysup, line 152:
```suggestion
      <ItemLabel className={collectionItemLabel}>{name}</ItemLabel>
```

packages/compass-databases-navigation/src/icon-button.tsx
```diff
@@ -6,22 +6,46 @@ import {
...
+} from './constants';
 
 const iconContainer = css({
-  color: uiColors.white,
   display: 'block',
   flex: 'none',
   fontSize: 0,
 });
 
+export type IconMode = 'normal' | 'hovered';
```
lerouxb, line 23: Trying to differentiate between the items that go straight on the sidebar background and area always visible like the expand/collapse one and the database or collection items vs the ones that only show up when you hover over the tree item (like the ... next to collections and add or delete next to databases).

This is because in the designs they have different colours.

So far these are the best names I could come up with. Maybe variant rather than mode? Suggestions welcome and then I can find/replace them.

packages/compass-databases-navigation/src/icon-button.tsx
```diff
@@ -6,22 +6,46 @@ import {
...
+  }
+}
+
 export const SmallIcon: React.FunctionComponent<
-  { glyph: string } & React.HTMLProps<HTMLSpanElement>
-> = ({ glyph, className, ...props }) => {
+  { glyph: string; mode: IconMode } & React.HTMLProps<HTMLSpanElement>
+> = ({ glyph, mode, className, ...props }) => {
+  const useNewSidebar = process?.env?.COMPASS_SHOW_NEW_SIDEBAR === 'true';
+  const darkMode = !useNewSidebar; // for now assume the old sidebar is dark and the new one light
```
lerouxb, line 43: My guess is that once we properly implement darkmode and the both the connetions and normal sidebars are light when darkmode is off and dark when it is on then we'd be pulling darkMode out of a context or something as low as we need it and not pass it down as a prop?

packages/compass-sidebar/src/components/sidebar.tsx
```diff
@@ -10,8 +10,8 @@ import { toggleIsGenuineMongoDBVisible } from '../modules/is-genuine-mongodb-vis
 import { changeFilterRegex } from '../modules/databases';
 import { updateAndSaveConnectionInfo } from '../modules/connection-info';
 
-// TODO: this should either change for this sidebar or be shared somewhere
-const initialSidebarWidth = spacing[4] * 10 + spacing[2]; // 248px
+// TODO: this should probably be shared be shared somewhere
+const initialSidebarWidth = spacing[6] * 4 - spacing[1]; // 252px
```
gribnoysup, line 14: Should we just align it right now and have this number in `ResizableSidebar` component? Is it that different between connection screen and this one at the moment?

packages/compass-databases-navigation/src/collection-item.tsx
```diff
@@ -126,12 +141,15 @@ export const CollectionItem: React.FunctionComponent<
       isHovered={isHovered}
       isTabbable={isTabbable}
       onDefaultAction={onDefaultAction}
-      className={collectionItem}
+      className={cx(
+        collectionItem,
+        useNewSidebar ? collectionItemNewSpacing : collectionItemOldSpacing
```
gribnoysup, line 146: Something is off here, placeholders are rendered without enough padding on the left

image

packages/compass-components/src/components/resizeable-sidebar.tsx
```diff
@@ -11,18 +11,28 @@ const containerStyles = css({
   maxWidth: '80%',
   height: '100%',
   position: 'relative',
+});
+
+const containerStylesDark = css({
   background: uiColors.gray.dark3,
   color: 'white',
 });
```
gribnoysup, line 19: If you're feeling fancy here, you can define your palette for the modes in this one place with custom variables:


```suggestion
const containerStylesDark = css({
  '--background-color': uiColors.gray.dark3,
  '--color': 'white',
  '--icon-color': 'white',
  '--icon-color-hover': 'white',
  '--item-color': 'white',
  '--item-color-hover': 'white',
  '--item-color-active': 'white'
  '--item-bg-color': uiColors.gray.dark2,
  '--item-bg-color-hover': uiColors.gray.dark3,
  '--item-bg-color-active': uiColors.gray.dark1,
  backgroundColor: 'var(--background-color)',
  color: 'var(--color)',
});
```

... and then just use these properties inside the components with no need to do the javascript logic for this in multiple places

packages/compass-databases-navigation/src/collection-item.tsx
lerouxb: Damnit I missed that one.

packages/compass-databases-navigation/src/icon-button.tsx
gribnoysup: IIRC we already have a dark mode context and `useDarkMode` hook that you should be able to use here

packages/compass-sidebar/src/components/sidebar.tsx
lerouxb: I'm still not sure if I'm allowed or expected to change the connection sidebar in this scope :)

packages/compass-databases-navigation/src/icon-button.tsx
gribnoysup: My first instinct would be to call this IconState, but both mode and variant sounds good to me 👍

alenakhineika: I think that mode fits pretty well here.

packages/compass-components/src/components/resizeable-sidebar.tsx
lerouxb: Makes sense. Didn't think of using css vars. Will change. Although I'm not 100% sure if this will work with the icon and svg path thingies?

packages/compass-databases-navigation/src/icon-button.tsx
gribnoysup: Ah, [it's called `useTheme`](https://github.com/mongodb-js/compass/blob/main/packages/compass-components/src/hooks/use-theme.tsx)

packages/compass-components/src/components/resizeable-sidebar.tsx
lerouxb: It gets a little bit hairy with some things having darkMode passed in as an attribute. I'll see.

packages/compass-databases-navigation/src/icon-button.tsx
lerouxb: Nice I'll look into it. Although.. right now we're in this weird state where there is only a dark version of the connections sidebar regardless of whether were in dark mode or not. Which complicates matters a bit.

packages/compass-databases-navigation/src/collection-item.tsx
lerouxb: Should be better now. Thanks for pointing it out.

packages/compass-components/src/components/resizeable-sidebar.tsx
lerouxb: Turns out you _can_ pass className to icons. Also weirdly it is using `color`, not `fill`. I guess the fill react prop is used as the `color` property internally somewhere.

packages/compass-components/src/components/resizeable-sidebar.tsx
```diff
@@ -11,18 +11,56 @@ const containerStyles = css({
   maxWidth: '80%',
   height: '100%',
   position: 'relative',
-  background: uiColors.gray.dark3,
-  color: 'white',
+});
+
+const containerStylesDark = css({
```
lerouxb, line 16: Once we just have darkmode support everywhere and both sidebars are only dark when darkmode is true we should probably move these colours even higher up so the whole app can use it?

packages/compass-sidebar/src/components/sidebar.tsx
Anemy: I think having one variable in `resizable-sidebar.tsx` like `defaultInitialSidebarWidth` and referencing that from both sidebars would be a nice simple way to do it. I think it's fine to update other things when we do things like this.

packages/compass-databases-navigation/src/item-action-controls.tsx
```diff
@@ -123,6 +124,7 @@ export const ActionControls: React.FunctionComponent<{
           <IconButtonSmall
             key={action}
             glyph={icon}
+            mode="hovered"
```
Anemy, line 127: Is there a case there this `IconButtonSmall` is used that isn't hovered?

packages/compass-components/src/components/resizeable-sidebar.tsx
Anemy: I think there are some trade offs using css vars, like ensuring you have the right name for a variable, and ensuring the vars you're creating are actually used. We might want to investigate how we can better do those things before using them all over just yet. Looks useful for this implementation though.

packages/compass-databases-navigation/src/item-action-controls.tsx
lerouxb: Aah good point. I removed it. Only used on the icons themselves.

packages/compass-databases-navigation/src/icon-button.tsx
```diff
@@ -6,22 +6,31 @@ import {
...
+const icons = {
+  normal: css({
+    color: 'var(--icon-color)',
+  }),
+  hovered: css({
+    color: 'var(--hover-icon-color)',
+  }),
+} as const;
+
+type IconMode = 'normal' | 'hovered';
```
Anemy, line 26: Small thing, fine to ignore, since `normal` and `hovered` are defined above, we might be able to do
```suggestion
type IconMode = keyof typeof icons;
```

Anemy (approved):
lgtm! One comment/small suggestion on sharing the width between the sidebars, and one on possible typing improvement. Nice

packages/compass-sidebar/src/components/sidebar.tsx
lerouxb: OK I added defaults to resizable sidebar and removed the others.

gribnoysup (approved):

packages/compass-databases-navigation/src/icon-button.tsx
```diff
@@ -78,7 +79,7 @@ export const IconButtonSmall = forwardRef<
       // TODO: we should probably explicitly set our own colors given that we're
       // setting all the colours around it. Rather than rely on leafygreen's
       // default dark or light hover background colour.
-      darkMode={darkMode}
+      darkMode={theme === Theme.Dark}
```
gribnoysup, line 82: IIRC all leafygreen components that we re-export from compass-components are already connected to theme provider, so I think there should be no need to pass this here explicitly

packages/compass-databases-navigation/src/placeholder-item.tsx
```diff
@@ -38,9 +45,10 @@ export const PlaceholderItem: React.FunctionComponent<{
 }> = ({ type = 'collection', style }) => {
   const useNewSidebar = process?.env?.COMPASS_SHOW_NEW_SIDEBAR === 'true';
   const variant = useNewSidebar ? 'new' : 'old';
+  const { theme } = useTheme();
   return (
     <div className={cx(placeholderItem, padding[variant][type])} style={style}>
-      <Placeholder darkMode></Placeholder>
+      <Placeholder darkMode={theme === Theme.Dark}></Placeholder>
```
gribnoysup, line 51: Maybe would make sense to make this component "connected" to theme provider by default, similar to how we do this for lg components. Then you would not need to pass this here explicitly and can remove useTheme call

packages/compass-home/src/components/home.tsx
```diff
@@ -256,7 +257,7 @@ function ThemedHome(
 ): ReturnType<typeof Home> {
   const appRegistry = useAppRegistryContext();
 
-  const [theme, setTheme] = useState<Theme>({
+  const [theme, setTheme] = useState<ThemeState>({
```
gribnoysup, line 260: 👍

packages/compass-sidebar/src/components/sidebar.tsx
```diff
@@ -48,10 +53,18 @@ export function Sidebar({}: Props) {
   // TODO: filter
   // TODO: create database
   // TODO: non genuine warning label
+
+  // For now the new sidebar is always light
+  const [theme] = useState<ThemeState>({
+    theme: Theme.Light,
+  });
```
gribnoysup, line 60: This is super not important for now I guess, but it seems to be that you can just completely remove the provider here and then the new sidebar will just correctly follow the user theme (if this COMPASS_LG_DARKMODE feature flag is set of course), right? Or am I missing something?

packages/compass-connections/src/components/connections.tsx
```diff
@@ -83,30 +86,37 @@ function Connections({
...
-            void connect(connectionInfo);
-          }}
-          removeAllRecentsConnections={() => {
-            void removeAllRecentsConnections();
-          }}
-          removeConnection={removeConnection}
-          duplicateConnection={duplicateConnection}
-        />
-      </ResizableSidebar>
+      <ThemeProvider theme={theme}>
```
gribnoysup, line 101: I very like this pattern of using provider to force / invert the theme where needed, means that you can safely change anything lower in the react tree and don't need to always keep in mind which parts of compass are inverted / always dark mode, etc. Really good stuff!

packages/compass-sidebar/src/components/sidebar.tsx
lerouxb: Good point..

packages/compass-databases-navigation/src/placeholder-item.tsx
lerouxb: good idea

packages/compass-components/src/hooks/use-theme.tsx
```diff
@@ -48,7 +49,9 @@ const withTheme = function <
   ) => {
     const theme = useTheme();
 
-    const applyTheme = global?.process?.env?.COMPASS_LG_DARKMODE === 'true';
+    const applyTheme =
+      theme.enabled === true ||
```
lerouxb, line 53: @Anemy , @gribnoysup The thinking about this change is so we don't have to hardcode the darkMode prop in compass-databases-navigation. In the old sidebar and in the connections sidebar there's now a new provider to always provide the Dark theme.

(At least until we have both light and dark themed versions of both sidebars and they follow the theme as set by the provider in home)

Then those components automatically pick up dark or light.

The easiest places to see this are the placeholders or the background circle hover effect on the action icons.

So to get the new light sidebar you only have to set the `COMPASS_SHOW_NEW_SIDEBAR` feature flag and not use it in conjunction with the darkmode one. And the changes we have to make when this feature lands should also be smaller.

packages/compass-components/src/components/placeholder.tsx
```diff
@@ -81,3 +82,7 @@ export const Placeholder: React.FunctionComponent<
     ></div>
   );
 };
+
+const Placeholder = withTheme(UnthemedPlaceholder);
```
lerouxb, line 86: ☝️ Placeholder is now connected to the theme provider.

packages/compass-sidebar/src/components-legacy/sidebar/sidebar.module.less
```diff
@@ -22,6 +20,24 @@
   position: relative;
   z-index: 50;
 
+  --color: @pw;
+  --bg-color: @compass-sidebar-base-background-color;
+
+  --icon-color: @pw;
```
lerouxb, line 26: ☝️ temporarily including the new css vars on the old sidebar as well so the changes in compass-databases-navigation can pick it up.

packages/compass-components/src/hooks/use-theme.tsx
gribnoysup: Love this change, makes the theme provider already useful and reduces the amount of clean up we will need to do after new sidebar or leafygreen based darkmode will roll out 👍

PR #3321, opened 2022-08-03
feat(compass-query-history): Add interactive popover, open query history as popover in new query-bar COMPASS-5678

Anemy:
Feature flagged:
```
```

This adds and `InteractivePopover` component to `compass-components`. It uses a focus trap for accessibility. (Version of this package, `focus-trap-react`, is shared with leafygreen).

When the new toolbars are active the query history is opened as a popover. No change to the existing query history.

This pr was separated out of https://github.com/mongodb-js/compass/pull/3312 which had some incorporated feedback on the new component.

https://user-images.githubusercontent.com/1791149/182718628-5b3f823d-e6fc-4754-abff-ad9631679169.mp4


packages/compass-collection/src/stores/context.tsx
```diff
@@ -257,6 +262,71 @@ const setupScopedModals = ({
...
+      isTimeSeries,
+      isClustered,
+      isFLE,
+      actions: queryBarActions,
+      query,
+      aggregation,
+    });
+  }
+
+  const queryHistoryRole = globalAppRegistry.getRole('Query.QueryHistory')?.[0];
```
mcasimir, line 305: will this change be required also in DE?

mcasimir (approved):

packages/compass-query-bar/src/components/option-editor.tsx
```diff
@@ -35,7 +35,7 @@ const editorStyles = cx(
     border: `1px solid ${uiColors.gray.light2}`,
     borderRadius: '4px',
```
mcasimir, line 36: Is likely that we would want to have this radius aligned everywhere we use one, could be a good thing if we can have variables for this too, along the font sizes, spacings, etc ..

packages/compass-collection/src/stores/context.tsx
Anemy: No, DE doesn't use query history so it's disabled and this role fetching never happens on their end.  Down the line if we add query history to DE we'll need to register this role and initialize it separately.

mcasimir: awesome thanks!

PR #3345, opened 2022-08-12
fix(compass): include node-forge in unpacked node_modules COMPASS-5802

addaleax:
Since it is a dependency of win-export-certificate-and-key,
node-forge should also be unpacked, like its dependent.

This adds an e2e test for verifying that the shell works
when connecting to Atlas using tlsUseSystemCA (since that
is an environment which always uses TLS and should always
match the system CA suite).

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-e2e-tests/tests/connection.test.ts
+++ b/packages/compass-e2e-tests/tests/connection.test.ts
@@ -429,6 +429,33 @@ describe('Connection screen', function () {
     expect(result).to.have.property('ok', 1);
   });
 
+  it('can connect to an Atlas with tlsUseSystemCA', async function () {
+    if (!hasAtlasEnvironmentVariables()) {
+      return this.skip();
+    }
+
+    const username = process.env.E2E_TESTS_ATLAS_USERNAME ?? '';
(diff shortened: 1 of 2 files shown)


packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -429,6 +429,36 @@ describe('Connection screen', function () {
     expect(result).to.have.property('ok', 1);
   });
 
+  it('can connect to an Atlas with tlsUseSystemCA', async function () {
+    if (!hasAtlasEnvironmentVariables()) {
+      //return this.skip();
```
alenakhineika, line 434: Is this part commented intentionally? If yes maybe `if` can be deleted.

addaleax: No, it’s not commented intentionally, thanks for pointing this out!

alenakhineika (approved):

PR #3352, opened 2022-08-15
feat(compass-aggregations): Update saved aggregations to open as popover COMPASS-5852

Anemy:
COMPASS-5852

This PR updates the saved aggregations view to open as a popover, previously it would open as a panel from the left.

A few driveby updates:
- Updated the title and button in the panel to leafygreen.
- Added an empty state (screenshot below).
- A bit of renaming: previously the saved aggregations panel was named `save-pipeline.jsx`, now it is `saved-pipelines.tsx`.
- (bugfix) Previously if a user had a lot of saved aggregations the panel would overflow the top level container, now it scrolls nicely

Before:
Screen Shot 2022-08-15 at 3 47 23 PM

After:
Screen Shot 2022-08-15 at 3 47 39 PM

With lots of saved aggregations:
Screen Shot 2022-08-15 at 3 32 10 PM

Video:

https://user-images.githubusercontent.com/1791149/184723351-a8095c53-f8f5-4c73-b048-59f740ce5f42.mp4

Empty state before:
Screen Shot 2022-08-15 at 5 25 17 PM

Empty state after:
Screen Shot 2022-08-15 at 5 25 26 PM


Anemy:
Noticed the popover closes when the modal confirmation opens, going to bring this with Claudia and look at what behavior we want here. Maybe we can remove the confirmation all together.

mmarcon:
@Anemy @Sgrinfy given that we are touching this component, should we consider making it clear that those are the pipelines for the current collection and not all of them, maybe by tweaking the copy a bit?

lerouxb (approved):

packages/compass-aggregations/src/modules/saved-pipeline.ts
```diff
@@ -0,0 +1,168 @@
...
+      dataService?.dataService?.getConnectionString?.().hosts.join(',') ??
+      null
+  };
+
+  track('Aggregation Saved', {
+    id: stateRecord.id,
+    num_stages: pipeline.length
+  });
+
+  const dirname = getDirectory();
```
mabaasit, line 152: Not sure if my comment is relevant, but we have [this ticket](https://jira.mongodb.org/browse/COMPASS-5893) to use `pipeline-storage.js` class to handle saving of pipelines in storage. Feel free to ignore this comment

Anemy: Thanks for the heads up on that, will keep that work in the separate ticket/pr.

PR #3363, opened 2022-08-18
feat: implement new indexes modal designs COMPASS-5978

alenakhineika:
## Description

Designs:
https://www.figma.com/file/vbjoD4dcisPYRYJ5RRu72k/Compass-LG?node-id=2425%3A61731

- Implement new designs keeping old branding styles.
- Move package to TypeScript and CSS-in-JS.
- Unify the names of props.
- Dismiss the error message when hitting the white X icon [COMPASS-4252](https://jira.mongodb.org/browse/COMPASS-4252)

Planned follow-up improvements:
- Test editors, buttons (close, cancel, add field, remove filed), banners (in progress, error) [COMPASS-5942](https://jira.mongodb.org/browse/COMPASS-5942)
- Refactor create db/collection modal to use CollapsibleFieldSet from `compass-components` (including more tests for collapsibleFieldSet) [COMPASS-6035](https://jira.mongodb.org/browse/COMPASS-6035)

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

Screenshot 2022-08-18 at 17 23 00


packages/compass-indexes/src/typings.d.ts
```diff
@@ -5,3 +5,6 @@ declare module 'mongodb-query-parser' {
   ) => false | null | CollationOptions;
   export { isCollationValid };
 }
+
+declare module 'hadron-react-buttons';
```
Anemy, line 9: It looks like we're removing the last reference of `hadron-react-buttons` in this package with these changes. Can we remove this line and the `hadron-react-buttons` dependency in this package?
```suggestion
```
I think there's one instance of `hadron-react-components` still used in `drop-index-modal`, however it's in a jsx file. Do we need the type declaration for that? Wondering if we can remove the line below as well.

alenakhineika: Removed `hadron-react-buttons`, but kept `hadron-react-components` because it is still in use.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
```diff
@@ -0,0 +1,220 @@
...
+import CreateIndexForm from '../create-index-form';
+import { toggleUseIndexName } from '../../modules/create-index/use-index-name';
+
+const { track } = createLoggerAndTelemetry('COMPASS-IMPORT-EXPORT-UI');
+
+const bannerStyles = css({
+  marginTop: spacing[3],
+});
+
+const createIndexModal = css({
```
Anemy, line 51: Should the modal have a minimum height or bottom padding before the buttons as well? On the default view opening the type select makes the height scroll:
Screen Shot 2022-08-19 at 2 21 08 AM

packages/compass-indexes/src/components/create-index-form/create-index-form.tsx
```diff
@@ -0,0 +1,213 @@
...
+  ttl?: string;
+  partialFilterExpression?: string;
+  wildcardProjection: string;
+  collationString?: string;
+  columnstoreProjection: string;
+
+  namespace: string;
+  serverVersion: string;
+
+  // TODO: Refactor modules to get rid of return any.
```
Anemy, line 58: Can these be `void` instead of `any`? Maybe I'm missing something but it looks like they aren't expected to return any values.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * Render the create index modal.
+   *
+   * @returns {React.Component} The create index modal.
+   */
+  render() {
+    this.handleShow();
+    return (
+      <ConfirmationModal
```
Anemy, line 136: In a few other places in Compass we use `ConfirmationModal` where we probably want to be using a regular `Modal` with our own styles. (I think I introduced it initially oops). Since we're updating this modal is this a good time to use `Modal`? If we make a footer we can use that in the ticket we have for updating the other places to use the regular `Modal`.

Here's a ticket where we're tracking that: https://jira.mongodb.org/browse/COMPASS-5522

packages/compass-indexes/src/components/create-index-form/columnstore-projection.tsx
```diff
@@ -0,0 +1,40 @@
...
+  Editor,
+  EditorVariant,
+  CollapsibleFieldSet,
+} from '@mongodb-js/compass-components';
+
+type ColumnstoreProjection = {
+  useColumnstoreProjection: boolean;
+  toggleUseColumnstoreProjection: (useColumnstoreProjection: boolean) => any;
+  columnstoreProjection?: string;
+  columnstoreProjectionChanged: (columnstoreProjection: string) => any;
```
Anemy, line 12: Does `void` work here?
```suggestion
  toggleUseColumnstoreProjection: (useColumnstoreProjection: boolean) => void;
  columnstoreProjection?: string;
  columnstoreProjectionChanged: (columnstoreProjection: string) => void;
```

packages/compass-indexes/src/modules/create-index/index.js
```diff
@@ -116,15 +116,15 @@ const rootReducer = (state, action) => {
...
       error: ERROR_INITIAL_STATE,
       isUnique: IS_UNIQUE_INITIAL_STATE,
-      isTtl: IS_TTL_INITIAL_STATE,
-      hasWildcardProjection: HAS_WILDCARD_PROJECTION_INITIAL_STATE,
-      hasColumnstoreProjection: HAS_COLUMNSTORE_PROJECTION_INITIAL_STATE,
-      isPartialFilterExpression: IS_PARTIAL_FILTER_EXPRESSION_INITIAL_STATE,
+      useTtl: IS_TTL_INITIAL_STATE,
+      useWildcardProjection: HAS_WILDCARD_PROJECTION_INITIAL_STATE,
+      useColumnstoreProjection: HAS_COLUMNSTORE_PROJECTION_INITIAL_STATE,
+      usePartialFilterExpression: IS_PARTIAL_FILTER_EXPRESSION_INITIAL_STATE,
```
Anemy, line 127: Nice renaming and standardizing 👍

packages/compass-components/src/components/leafy-green/combobox/combobox.tsx
```diff
@@ -885,13 +886,18 @@ export default function Combobox<M extends boolean>({
       }
 
       setInputFocus(cursorPos);
+      // Open menu on click instead of on focus.
+      // https://jira.mongodb.org/browse/PD-2321
+      openMenu();
```
Anemy, line 891: When trying this combobox out I tried to click into the combobox here to open the dropdown to see the field names and it wasn't opening, it looks like the text input takes focus but the suggestions don't appear when nothing is typed. I think this ux was a bit confusing. Clicking on the dropdown arrow or on the border of the combobox, however, had the expected behavior. Should we update it to open on clicking the input?

I think a bigger case for accessibility is that now with a keyboard a user can't open the field name list without first typing part of the name. I'm thinking we either want to have the select field name menu open on enter or space like a select, or revert to the initial behavior with the menu opening on the wrapper focus.

Related jira ticket we were looking at https://jira.mongodb.org/browse/PD-2321

packages/compass-indexes/src/components/create-index-form/partial-filter-expression.tsx
```diff
@@ -0,0 +1,40 @@
...
+}: PartialFilter) => {
+  return (
+    <CollapsibleFieldSet
+      toggled={usePartialFilterExpression}
+      onToggle={(checked: boolean) => toggleUsePartialFilterExpression(checked)}
+      label="Partial Filter Expression"
+      dataTestId="create-index-modal-is-pfe-checkbox"
+      description="Partial indexes only index the documents in a collection that meet a specified filter expression."
+    >
+      <TextInput
```
Anemy, line 27: Did we end up deciding if we wanted to turn this into an editor instead of a text input?

alenakhineika: Yeah there is already a ticket for that: https://jira.mongodb.org/browse/COMPASS-6017
This PR is too big :)

packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
alenakhineika: Replaced with Modal, but it didn't go away, so I set a min height for the modal to fix select.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
alenakhineika: Replaced with Modal! Yeah, I took a look at other placed and thought we intentionally use ConfirmationModal :)

packages/compass-indexes/src/components/create-index-form/columnstore-projection.tsx
alenakhineika: Yes, thanks for pointing at it! I really didn't like these anys myself 🙈

packages/compass-components/src/components/leafy-green/combobox/combobox.tsx
alenakhineika: This copied Commbobox is a temporary workaround until we can switch to the new design and the last version of LG components. Fixing all these accessibility issues for the old component would be tricky and time-consuming and considering that there is already [WIP PR](https://github.com/mongodb-js/compass/pull/3290) for using new styles we can keep it as is for a little longer while and then replace it with a new component where it all has ready fixed.

alenakhineika: Pushed an update though to open the menu on click on any input area.

packages/compass-indexes/src/components/create-index-fields/create-index-fields.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+const createIndexFieldsStyles = css({
+  display: 'flex',
+  justifyContent: 'stretch',
+  marginBottom: spacing[2],
+  marginTop: spacing[1],
+});
+
+const createIndexFieldsNameStyles = css({
+  width: +spacing[7] * 3 + +spacing[6],
```
Anemy, line 55: Do we need this width and the width below on `createIndexFieldsTypeStyles`? It seems to look alright without it. Asking because this calculation looks like something that would have to update if we update `spacing` and if we can avoid that dependency it might be nice. Fine as is too.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
```diff
@@ -0,0 +1,252 @@
...
+    );
+  };
+
+  const renderInProgress = () => {
+    if (!error || !inProgress) {
+      return;
+    }
+
+    return (
+      <Banner className={bannerStyles} variant="info" dismissible>
```
Anemy, line 132:
```suggestion
      <Banner className={bannerStyles} variant="info">
```
I think we probably don't want this to be `dismissible` if we don't supply an `onClose`.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
```diff
@@ -0,0 +1,252 @@
...
+        Create in Progress
```
Anemy, line 133: Should we have an in progress indicator here? Like the `SpinLoader` component along with this text? I'm thinking it might feel too static otherwise.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
```diff
@@ -0,0 +1,252 @@
...
+      data-testid="create-index-modal"
+      className={modalStyles}
+      contentClassName={modalContentWrapperStyles}
+    >
+      <div className={modalContentStyles}>
+        <H3>Create Index</H3>
+        <Disclaimer>{namespace}</Disclaimer>
+        <CreateIndexForm {...props} />
+        {renderError()}
+        {renderInProgress()}
```
Anemy, line 161: Currently when a user encounters an error these errors appear at the bottom of the modal and can be easily missed as they're often hidden by a scroll bar. How do you feel about moving these banner displays into the `ModalFooter`?

Anemy (commented):
Nice, couple comments on banner displays, looking good.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
alenakhineika: Let's sync with Claudia where to put it. We can create a follow-up ticket when we decide!

packages/compass-indexes/src/components/create-index-fields/create-index-fields.tsx
alenakhineika: If do not set width explicitly it changes the size of components when name or type is selected. I changed it `vw` tho so it does not depend on spacing.

Anemy: Gotcha, in that case I think `spacing` might be better here then `vw`, with `vw` the width of these elements could be huge on large/stretched monitors.

alenakhineika: Reverted to spaces :)

Anemy:
Sorry if it's work in progress, getting an error when expanding the index options:
Screen Shot 2022-08-21 at 3 04 51 PM
And clicking on the options in the combo box doesn't seem to select them

Anemy (approved):
lgtm! Nice work, this modal and the controls in it feel a lot nicer to use now. Are we planning to add any indication of the index being created? Currently if you disconnect from wifi and then click create index it doesn’t show any indication something is in progress until the error shows up.

alenakhineika:
Fixed! Thx for noticing this! More testing is coming...  https://jira.mongodb.org/browse/COMPASS-6034 :)

PR #3384, opened 2022-08-24
feat(indexes): refresh indexes COMPASS-5941

mabaasit:
feat(indexes): refresh indexes COMPASS-5941

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
```diff
@@ -84,13 +73,15 @@ export const IndexesTable: React.FunctionComponent<IndexesTableProps> = ({
...
-    ].map((name) => {
+    const _columns = (
+      [
+        'Name and Definition',
+        'Type',
+        'Size',
+        'Usage',
+        'Properties',
+      ] as SortColumn[]
+    ).map((name) => {
```
addaleax, line 84: tiny TS nit, feel free to ignore: `as` does a cast and would accept array entries here that aren’t actually valid values for `SortColumn`. You could avoid that issue through e.g.:

```suggestion
    const sortColumns: SortColumn[] = [
      'Name and Definition',
      'Type',
      'Size',
      'Usage',
      'Properties',
    ];
    const _columns = sortColumns.map((name) => {
```

packages/compass-indexes/src/modules/indexes.ts
```diff
@@ -0,0 +1,209 @@
...
+>;
+export type SortColumn = keyof typeof sortColumnToProps;
+export type SortDirection = 'asc' | 'desc';
+const sortColumnToProps: Record<string, SortField> = {
+  'Name and Definition': 'name',
+  Type: 'type',
+  Size: 'size',
+  Usage: 'usageCount',
+  Properties: 'properties',
+};
```
addaleax, line 27:
```suggestion
const sortColumnToProps = {
  'Name and Definition': 'name',
  Type: 'type',
  Size: 'size',
  Usage: 'usageCount',
  Properties: 'properties',
} as const;
```

to just directly get the exact type?

addaleax (approved):
nice work on the ts conversions! :rocket:

PR #3385, opened 2022-08-24
fix(index-model): text index is not compound COMPASS-5386

mabaasit:
fix(index-model): text index is not compound COMPASS-5386

Currently we are reporting text indexes as compound, which is calculated based on the keys of an index.

By default if a text index is created on a single field, it will always have 2 keys (_FTS & _FTSX). In this PR we are making this check stronger, by ensuring:
1. An index having only one text field, is a single index;
2. An index having multiple text only fields, is a compound index (if `extra.weights.length > 1`);
3. An index having multiple fields (text and 1, -1 ...), is a compound index (if `keys.length > 2`).

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/index-model/lib/model.js
+++ b/packages/index-model/lib/model.js
@@ -148,7 +148,10 @@ var IndexModel = Model.extend({
         // Text index which is created with other fields (that are not text), will have >2 keys and is a compound
         // Text index which is created with other fields (that are text), will have 2 keys + extra.weights > 1 and is a compound
         const isSingleTextIndex = this.text && _.keys(this.key).length === 2 && _.keys(this.extra.weights).length === 1;
-        return this.single || isSingleTextIndex ? 'single' : 'compound';
+        if (isSingleTextIndex) {
+          return 'single';
+        }
+        return this.single ? 'single' : 'compound';
       }
     },
     properties: {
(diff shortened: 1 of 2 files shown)


packages/index-model/lib/model.js
```diff
@@ -144,7 +144,11 @@ var IndexModel = Model.extend({
     cardinality: {
       deps: ['single'],
       fn: function() {
-        return this.single ? 'single' : 'compound';
+        // By default a text index has only 2 keys: _FTS & _FTSX and its a single.
+        // Text index which is created with other fields (that are not text), will have >2 keys and is a compound
+        // Text index which is created with other fields (that are text), will have 2 keys + extra.weights > 1 and is a compound
+        const isSingleTextIndex = this.text && _.keys(this.key).length === 2 && _.keys(this.extra.weights).length === 1;
+        return this.single || isSingleTextIndex ? 'single' : 'compound';
```
Anemy, line 151: On a quick read this looks like this returns `this.single` now where previously it would return `single` or `compound`. Spreading it out a bit might help readability, how do you feel about:

```suggestion
        if (isSingleTextIndex) {
          return 'single';
        }
        return this.single ? 'single' : 'compound';
```

Anemy (approved):

PR #3388, opened 2022-08-25
fix(hadron-document): fix updating arrays with dots in names COMPASS-6011

addaleax:
Diff probably best viewed with whitespace changes hidden.

---

Follow-up to c865b66bd7c4 (#3239), which introduced this bug:
Handle updates to arrays in the case in which the updated array
(or another updated path) contains dots or starts with a dollar sign,
using alternatives to `$getField` and `$setField` which can work on
arrays (`$arrayElemAt` for the former, `$concatArrays` + `$slice` for
the latter).

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/hadron-document/src/object-generator.ts
+++ b/packages/hadron-document/src/object-generator.ts
@@ -28,7 +28,17 @@ function maybeDecorateWithDecryptedKeys(
 }
 
 /** Used to represent missing values, i.e. non-existent fields. */
-const DoesNotExist = Symbol('DidNotExist');
+const DoesNotExist = Symbol('DoesNotExist');
+
+/**
+ * Describe a single property of a document. For us, not only the
+ * field name, but also whether it is an array index or a regular
+ * document property is relevant.
+ */
+type SubfieldDescription = {
+  key: string;
+  isArrayIndex: boolean;
+};
 
 /**
  * Describe a field in a document, with its path and current value.
@@ -36,7 +46,7 @@ const DoesNotExist = Symbol('DidNotExist');
  * `b` of `a` would be described by `{ path: ['a', 'b'], value: 42 }`.
  */
 type FieldDescription = {
-  path: string[];
+  path: SubfieldDescription[];
   value: BSONValue | typeof DoesNotExist;
 };
 
@@ -212,6 +222,7 @@ export class ObjectGenerator {
       keyInclusionOptions.includableEncryptedKeys ?? [];
 
     for (const element of target.elements ?? []) {
+      const isArrayIndex = target.currentType === 'Array';
       // Do not include encrypted fields in the `originalFields` list
       // unless we know that it is okay to include them (i.e. because
       // we know that we can perform equality queries on those fields).
@@ -274,13 +285,16 @@ export class ObjectGenerator {
             );
           for (const { path, value } of nestedResult.originalFields) {
             originalFields.push({
-              path: [String(element.key), ...path],
+              path: [{ key: String(element.key), isArrayIndex }, ...path],
               value,
             });
           }
           for (const { path, value } of nestedResult.newFields) {
             newFields.push({
-              path: [String(element.currentKey), ...path],
+              path: [
+                { key: String(element.currentKey), isArrayIndex },
+                ...path,
+              ],
               value,
             });
           }
@@ -289,7 +303,7 @@ export class ObjectGenerator {
           // the original field's value.
           if (canIncludeOriginalValue) {
             originalFields.push({
-              path: [String(element.key)],
+              path: [{ key: String(element.key), isArrayIndex }],
               value: element.generateOriginalObject(),
             });
           }
@@ -300,7 +314,7 @@ export class ObjectGenerator {
             !element.isRemoved()
           ) {
             newFields.push({
-              path: [String(element.currentKey)],
+              path: [{ key: String(element.currentKey), isArrayIndex }],
               value: element.generateObject(),
             });
           }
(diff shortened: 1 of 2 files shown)


packages/hadron-document/src/object-generator.ts
```diff
@@ -28,15 +28,25 @@ function maybeDecorateWithDecryptedKeys(
 }
 
 /** Used to represent missing values, i.e. non-existent fields. */
-const DoesNotExist = Symbol('DidNotExist');
+const DoesNotExist = Symbol('DoesNotExist');
```
addaleax, line 31: Purely cosmestic, I think I just forgot to change the name while working on #3239.

packages/hadron-document/src/object-generator.ts
```diff
@@ -334,11 +363,27 @@ export class ObjectGenerator {
...
+            wasRemovedAndLaterReplacedByNewElement = true;
+            break;
+          }
+        }
+        if (!wasRemovedAndLaterReplacedByNewElement) {
+          newFields.push({
+            path: [{ key: String(element.key), isArrayIndex }],
+            value: DoesNotExist,
+          });
+        }
```
addaleax, line 386: (This just mirrors the change above.)

packages/hadron-document/src/object-generator.ts
```diff
@@ -313,14 +327,29 @@ export class ObjectGenerator {
...
+            wasRenamedToKeyOfPreviouslyExistingElement = true;
+            break;
+          }
+        }
+        if (!wasRenamedToKeyOfPreviouslyExistingElement) {
+          originalFields.push({
+            path: [{ key: String(element.currentKey), isArrayIndex }],
+            value: DoesNotExist,
+          });
+        }
```
addaleax, line 350: This is kind of an independent issue, but didn’t appear in tests so far because when building the query in the non-dots-and-dollars mode, duplicate entries in `originalFields` would not cause trouble. For example, if property `a` with value `42` was removed and then later another element renamed to `a`, `originalFields` would contain two entries for `a`, one claiming that it does not exist, one claiming that it had the value `42`; the query would still (accidentally correctly) end up being `{ a: 42 }`, since the property in the query doc was simply overwritten.

In dots-and-dollars mode, however, the different values in `originalFields` are turned into multiple `$eq` checks which are then `$and`-ed together, so obviously there would be a conflict between an `$eq` asserting that the field did not exist and an `$eq` asserting that the field had its previous value.

packages/hadron-document/src/object-generator.ts
```diff
@@ -367,34 +412,62 @@ export class ObjectGenerator {
...
+        isArrayIndex
+          ? {
+              $arrayElemAt: [input, +key],
+            }
+          : {
+              $getField: {
+                field: { $literal: key },
+                input,
+              },
+            },
```
addaleax, line 427: In retrospect, it feels a bit dumb to have assumed (or probably just not thought about) that `$getField` and `$setField` don’t work on arrays the way they do on objects.

packages/hadron-document/src/object-generator.ts
```diff
@@ -436,28 +509,29 @@ export class ObjectGenerator {
...
-        query.$expr ??= { $and: [] };
-        query.$expr.$and.push(matchExpr);
+            ? { $eq: [getFieldExpr, { $literal: value }] }
+            : { $eq: [{ $type: getFieldExpr }, 'missing'] }
+        );
+      }
+      if (equalityMatches.length === 1) {
+        query.$expr = equalityMatches[0];
+      } else if (equalityMatches.length > 1) {
+        query.$expr = { $and: equalityMatches };
```
addaleax, line 529: This part of the change is also a little independent, and only makes it so that `$and: [singleEntry]` is reduced to `singleEntry`, with the main purpose being making the test files a *little* more readable

packages/hadron-document/test/document.test.ts
```diff
@@ -1558,65 +1558,373 @@ describe('Document', function () {
...
   describe('array modifications', function () {
-    it('can add array elements', function () {
-      const doc = new Document({
-        a: [1, 2, 3],
-      });
-      doc.get('a')?.insertEnd(3, new Int32(4));
-      expect(doc.generateUpdateUnlessChangedInBackgroundQuery()).to.deep.equal({
-        query: { _id: null, 'a.3': { $exists: false } },
-        updateDoc: { $set: { 'a.3': new Int32(4) } },
+    context('with a plain array name', function () {
```
addaleax, line 1561: These tests remain unchanged, only split into two `context()`s, one for plain names and one for dots-and-dollar names

packages/hadron-document/test/document.test.ts
```diff
@@ -1558,65 +1558,373 @@ describe('Document', function () {
...
+              $replaceWith: {
+                $setField: {
+                  field: { $literal: '$a.b' },
+                  input: '$$ROOT',
+                  value: { $literal: [new Int32(1), new Int32(2)] },
+                },
+              },
+            },
+          ],
+        });
```
addaleax, line 1927: I don’t necessarily mind having these fairly verbose/repetitive test object assertions, but if we feel that this is too noisy, I’m happy to take suggestions on how to improve that. I guess ideally some kind of snapshot testing where it’s easier to do bulk updates to expected results would be a good pick, but I don’t want to introduce entirely new paradigms here.

lerouxb (approved):
Looks good! And nice new tests.

PR #3389, opened 2022-08-26
feat: align toolbar popover styles with new design

mcasimir:
- Add a `<ModalPopover />` component that builds on the Interactive popover and provides a way to reuse the designs
- Align the UI of toolbars popovers to the new designs
- Wire the theme for background and border

Screenshot 2022-08-26 at 11 30 01
Screenshot 2022-08-26 at 11 29 08


packages/compass-components/src/components/modal-popover.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { gray8 } from '../compass-ui-colors';
+
+const borderRadius = spacing[2];
+
+const contentContainerStyles = css({
+  display: 'flex',
+  height: '100%',
+  alignItems: 'center',
+  borderRadius: borderRadius,
+  boxShadow: `0px 2px 4px -1px ${transparentize(0.85, uiColors.black)}`,
```
mcasimir, line 17: got this from LG, i'm assuming is what is used to make sure things will keep looking consistent through changes in the palette

packages/compass-components/src/components/modal-popover.tsx
```diff
@@ -0,0 +1,81 @@
...
+  color: uiColors.gray.dark2,
+});
+
+const contentContainerStylesDark = css({
+  borderColor: uiColors.gray.dark2,
+  backgroundColor: uiColors.gray.dark3,
+  color: uiColors.white,
+});
+
+type ModalPopopverProps = {
```
lerouxb, line 35: s/ModalPopopverProps/ModalPopoverProps probably

lerouxb (approved):

Anemy (commented):
What's the difference between a `ModalPopover` and the `InteractivePopover` components? Would it make sense to move all of the wrappers and styles the `ModalPopover` component is providing directly into the `InteractivePopover` and avoid adding a new component? I'm thinking it is good if we avoid creating components in `compass-components` that will be only consumed once.

mcasimir:
> What's the difference between the `ModalPopover` and the `InteractivePopover` components? Would it make sense to move all of the wrappers and styles the `ModalPopover` component is providing directly into the `InteractivePopover` and avoid adding a new component? I think it is good if we avoid creating components in `compass-components` that will only be consumed once.

That's a great point .. we can also "unexport" `InteractivePopover`? If I understand correctly "InteractivePopover" was a building block for other "Popover" types of components. more or less what the LG `<Popover>` does for tooltips, or did I get it wrong? I wasn't sure why the styles weren't factored and I assumed moving styles there was not a good idea.

Anemy:
@mcasimir When I added `InteractivePopover` I was intending for that to be the part which other plugins consumes, I hadn't really thought of it as a building block. I think it makes sense as a building block if more things use it. If only one place is using the building block I think it's probably adding more complexity than we need. I'd prefer if we either move all of the `InteractivePopover` into the `ModalPopover` or move these styles we're adding by wrapping `ModalPopover` into the `InteractivePopover` and using the `InteractivePopover` in other plugins then.
How do you feel about condensing the components into one?  I'm thinking it would simplify things, fine as is as well, I'll mark as approved!

packages/compass-components/src/components/interactive-popover.tsx
```diff
@@ -1,12 +1,37 @@
...
+import { gray8 } from '../compass-ui-colors';
+
+const borderRadius = spacing[2];
 
 const contentContainerStyles = css({
   display: 'flex',
   height: '100%',
+  alignItems: 'center',
+  borderRadius: borderRadius,
+  boxShadow: `0px 2px 4px -1px ${transparentize(0.85, uiColors.black)}`,
```
Anemy, line 19: Suggestion/though - Do we need to add an extra npm package for changing the color here? Seems like something we can avoid if we aren't going to be doing this color changing often, I think fewer dependencies usually makes things easier in the long run. 
Are these styles from somewhere in particular? Totally good to merge as is I think too.

mcasimir: Buh not sure, that is what LG does internally to keep shadows consistent, I copy-pasted it. I can convert what's there to the shadow in the rebranded `uiColors.black`, but we would have that dependency in any case from leafygreen itself. I wish there would have been a way to just `import {boxShadow1, borderColor1} from 'leafygreen-constants'` or something, just to have some help keeping things 1-1 with the component library.

Btw we have something with a similar style for the "find in page" that we probably should align / color / border wise there too: 

Screenshot 2022-08-31 at 15 11 53

Anemy: Sounds good, I think it's fine for now. We can add a `floatingBoxBoxShadowStyles` to `compass-components` that they both then use if we want.

PR #3394, opened 2022-08-26
feat(sidebar) Connection info modal for the new sidebar COMPASS-5966

lerouxb:
We can do a lot better here, but in the end I just did a straight copy of the logic from the old sidebar.

Screenshot 2022-08-26 at 16 13 08

---

These screenshots are with a size default modal. I've since changed it to small. Looks a bit better.

localhost via mongodb-runner:
Screenshot 2022-08-26 at 16 08 45

sharded via docker:
Screenshot 2022-08-26 at 16 08 02

serverless (ie. load balanced):
Screenshot 2022-08-26 at 16 07 46

data lake (I took the opportunity to rename it to Atlas Data Federation):
Screenshot 2022-08-26 at 16 07 27

compass datasets:
Screenshot 2022-08-26 at 16 07 11

ssh with password:
Screenshot 2022-08-26 at 16 17 31


mcasimir (approved):

packages/data-service/src/instance-detail-helper.spec.ts
```diff
@@ -277,7 +277,7 @@ describe('instance-detail-helper', function () {
         const client = createMongoClientMock({
           hosts: [{ host: 'fakehost.my.server.com', port: 9999 }],
           commands: {
-            atlasVersion: { version: '1.1.1', gitVersion: '1.2.3' },
+            atlasVersion: { atlasVersion: '1.1.1', gitVersion: '1.2.3' },
```
lerouxb, line 280: drive-by fix: the `atlasVersion` command returns the field `atlasVersion`, not `version`.

PR #3399, opened 2022-08-29
feat: add progress badge to the indexes table COMPASS-5944

alenakhineika:
## Description

Keep track of the `db.createIndex` promise. Allow users to close the dialog while the index creating in progress. Once the user pressed the “Cancel” button or closed the modal with X, keep an "In progress…" pillow while the promise is not fulfilled, and either remove it as soon as the promise resolves or replace it with a "Failed" pillow if the promise rejects.

Since several indexes can be created one by one and have "In progress…" pillows, we won't be able to reopen the modal with an error. We add on hover tooltip for the "Failed" pillow so users have an access to the error message.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

Screenshot 2022-08-29 at 15 59 23

Screenshot 2022-08-30 at 19 09 14

Screenshot 2022-08-30 at 19 09 33


packages/compass-indexes/src/modules/create-index/index.ts
```diff
@@ -223,8 +267,8 @@ export const createIndex = () => {
 
     if (state.useColumnstoreProjection) {
       try {
-        // columnstoreProjection is not part of CreateIndexesOptions yet
-        (options as any).columnstoreProjection = EJSON.parse(
+        // @ts-expect-error columnstoreProjection is not a part of CreateIndexesOptions yet.
```
mcasimir, line 270: @lerouxb here https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md#dependency-version would be awesome, we could make this expire on the next version of driver till driver has it and however did the update can fix it. Would you add a link to this comment in the ticket to add the eslint rule?

packages/compass-indexes/src/components/indexes-table/property-field.tsx
```diff
@@ -78,6 +86,12 @@ const PropertyField: React.FunctionComponent<PropertyFieldProps> = ({
           link={getIndexHelpLink(cardinality.toUpperCase() as any) ?? '#'}
         />
       )}
+      {extra.status === 'inprogress' && (
+        <Badge variant={BadgeVariant.Blue}>In Progress...</Badge>
+      )}
+      {extra.status === 'failed' && (
+        <Badge variant={BadgeVariant.Red}>Failed</Badge>
```
mcasimir, line 93: What about the failure reason? Should we add a tooltip with the error message or something?

mcasimir (commented):
To me this looks awesome but i could be biased :), maybe good if others could take a look too.

One thing catches my eyes in the screenshot, the copy in the dialog says 'Create in Progress', should it be "Index creation in progress"?

Few other stuff:
- Maybe we can tell that the dialog can be closed, as a user i'd be afraid that doing so cancel the creation
- We probably want to change the action bar, remove the "Create index" button and then turn "Cancel" into "Close dialog"? What happens now if the user hit cancel is the operation cancelled for real?

lerouxb:
Looks good to me and seems worth doing.

packages/compass-indexes/src/components/indexes-table/property-field.tsx
alenakhineika: Pushed changes according to your comments 👍 so now for in progress modal we hide the "Create index" button and replace the "Cancel" name with "Close". Also replaced "Create in Progress" with "Index creation in progress. The dialog can be closed.".

packages/compass-indexes/src/components/create-index-actions/create-index-actions.tsx
```diff
@@ -90,25 +79,25 @@ function CreateIndexActions({
...
+            Create Index
+          </Button>
+        )}
         <Button
           data-testid="create-index-actions-cancel-button"
-          darkMode={darkMode}
           onClick={onCancel}
         >
-          Cancel
+          {inProgress ? 'Close' : 'Cancel'}
```
mabaasit, line 96: We added Cancel to this button, but we do not actually cancel the op (of index creation).

packages/compass-indexes/src/components/drop-index-modal/drop-index-modal.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
```
mabaasit, line 1: nit: do we need this file? actually name of the file is a bit off. modal is in `drop-index-form` file. (feel free to ignore)

packages/compass-indexes/src/components/create-index-actions/create-index-actions.tsx
alenakhineika: I think we use the "Cancel" everywhere to cancel the modal (eg create db or collection modals), but here I agree with Maurizio, that cancel sounds confusing because we can not cancel the creating index that is in progress, therefore in his case, we replace it with the "Close" button.

packages/compass-indexes/src/components/indexes/indexes.tsx
```diff
@@ -74,16 +80,16 @@ export const Indexes: React.FunctionComponent<IndexesProps> = ({
           localAppRegistry={localAppRegistry}
           isRefreshing={isRefreshing}
           writeStateDescription={description}
-          onRefreshIndexes={() => onRefresh()}
+          onRefreshIndexes={refreshIndexes}
```
mabaasit, line 83: on refresh should we clear `failed` indexes?

image

packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
```diff
@@ -82,35 +79,32 @@ const createIndexHeaderTitleDarkStyles = css({
...
   ...props
 }: React.ComponentProps<typeof CreateIndexForm> & {
-  toggleIsVisible: (isVisible: boolean) => void;
-  resetForm: () => void;
   isVisible: boolean;
   namespace: string;
   error: string | null;
   clearError: () => void;
   inProgress: boolean;
   createIndex: () => void;
```
mabaasit, line 96: nit: should we clear previous error when user clicks on create index button?


https://user-images.githubusercontent.com/1305718/187974176-a3e13103-45d1-4420-a266-3cd8dcdf9571.mov

packages/compass-indexes/src/modules/indexes.ts
```diff
@@ -198,12 +211,11 @@ const _convertToModels = (indexes: Document[]): IndexDefinition[] => {
...
-) => {
-  return _convertToModels(indexes).sort(
-    _getSortFunction(_mapColumnToProp(sortColumn), sortOrder)
-  );
+export const dropFailedIndex = (
+  id: string
+): ThunkAction<void, RootState, void, AnyAction> => {
+  return (dispatch) => {
+    dispatch(inProgressIndexRemoved(id));
+    dispatch(fetchIndexes());
```
mabaasit, line 219: as we are just deleting the failed index from store, should we fetch indexes again?

packages/compass-indexes/src/components/indexes/indexes.tsx
alenakhineika: As a user, I would prefer to keep these raws until I explicitly remove them. But we can discuss it with a team during design sync if you have a strong opinion about it.

packages/compass-indexes/src/modules/indexes.ts
alenakhineika: Why not if we rerender the table anyway? I think Maurizio wanted to run fetch by timer :) but we haven't come to an agreement on this. So I decided to use this event to actualize the list. We can also bring this at the design sync later.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
alenakhineika: Nice catch! 👀

mabaasit (approved):

alenakhineika:
CI is green if open Evergreen, for some reason GitHub does not update it here :/

Issue #3373, opened 2022-08-22
Attributes skipped while importing json


urswiss:
#### Current Behavior

The attribute `frameworkContract` is ignored while importing. There is no error message.
To reproduce: Click on `Add data`, `Import File`, `Select a file...`, open the file which contains the data below. The option `Stop on errors` is not checked.

#### Code/Gist

```
[{
  "_id": "12345678",
  "linkName": "any link name",
  "defaultConfiguration": {
    "frameworkContract": {
      "frameworkContract": "123",
      "memberNumber": "any member number"
    },
    "campaign": null
  }
}
]
```

#### Expected Behavior/Code

Anything in the json should end up in the collection.

#### Environment

- node: 16.16.0 / npm: 8.11.0
- OS: Windows Version 10.0.19044.1889
- MongoDB Compass Version: 1.32.6

urswiss:
works fine in MongoDB Compass version 1.26.1

Anemy:
Hi @urswiss this looks like this bug we're currently tracking in jira: https://jira.mongodb.org/browse/COMPASS-5971
We'll update here when there's a fix out. Thanks for reporting the issue.

RogerMeyer:
It works for / with MongoDB Compass 1.30.1. I had no issues.

Anemy:
@urswiss This is fixed in the newly released version `1.33.1`. Thanks again for reporting this and helping us improve Compass. Going to close this, feel free to reopen if you update and things aren't behaving as expected.

https://github.com/mongodb-js/compass/releases

PR #3401, opened 2022-08-29
fix(import-export): fix import deep json overwriting variables COMPASS-5971

Anemy:
COMPASS-5971
https://github.com/mongodb-js/compass/issues/3373

Previously, when importing json documents with deep nested fields which have values alongside subdocuments in a subdocument, a bug would cause Compass to overwrite the parent. This PR fixes the bug so that all of the fields are correctly accounted for. (Example failing json document below, the `amount` fields are all overridden).

Previously in `dotnotation.ts` we were creating an empty object for the direct parent for a value when a value arrises. This pr updates it so that we create an empty object at the topmost parent and work our way in so that we don't possibly override existing parents. (With the example before, previously `supermarket.fruits.oranges.amount.2022-01-15` would create an empty parent at `supermarket.fruits.oranges.amount`, however this would be overridden by `supermarket.fruits.currency` creating an empty parent at `supermarket.fruits`. These changes make it so that we iteratively create an empty object at each parent stage and don't skip possible parents which cause overriding errors and possible data loss. 

Example test document:
```json
[
  {
    "abc": "123",
    "supermarket": {
      "def": "456",
      "fruits": {
        "updated": 2,
        "oranges": {
          "amount": {
            "2022-01-15": 1.66,
            "2022-02-16": 1.22,
            "2022-03-13": 1.11,
            "2022-04-14": 7.69,
          },
        },
        "apples": {
          "amount": {
            "2022-01-15": 3.47,
            "2022-02-14": 4.18,
            "2022-03-15": 4.18,
          },
        },
        "currency": "usd",
      },
    },
    "test": "123",
  }
]
```


lerouxb (approved):
nice!

PR #3403, opened 2022-08-30
feat(compass-schema): Add views for when no documents or no fields are found in schema analysis

Anemy:
Adding some temporary message displays for when users either:
- Have no documents returned in the schema analysis
- Have no fields returned in the schema analysis

(Previously we showed an empty page.) Not sure if this pr should be a feat or chore, open to any thoughts on proper tags.

Screen Shot 2022-08-30 at 2 32 29 AM

Screen Shot 2022-08-30 at 2 32 09 AM


mcasimir (commented):
I don't know if is ok that this looks so different from the regular 0 state, can't we either use the regular 0 state or just wait for designs?
Screenshot 2022-08-30 at 08 47 25

PR #3406, opened 2022-08-30
feat(sidebar) copy over csfle/non-genuine markers and modals to the new sidebar

lerouxb:
Probably not possible for both to show at once (do non-genuine mongodbs even support in-use encryption?) but this is how it would look. Also the db stats are going away, so that area will look a bit different.

Screenshot 2022-08-30 at 11 42 15

Screenshot 2022-08-30 at 11 54 05
Screenshot 2022-08-30 at 11 53 57

### DIFF
--- a/packages/compass-components/src/components/modal-title.tsx
+++ b/packages/compass-components/src/components/modal-title.tsx
@@ -9,7 +9,7 @@ const modalTitleStyles = css({
 });
 
 function ModalTitle(
-  props: React.ComponentProps<typeof H3>
+  props: Omit<React.ComponentProps<typeof H3>, 'as'>
 ): React.ReactElement {
   return <H3 {...props} className={cx(modalTitleStyles, props.className)} />;
 }
--- a/packages/compass-sidebar/src/components/csfle-marker.tsx
+++ b/packages/compass-sidebar/src/components/csfle-marker.tsx
@@ -37,7 +37,7 @@ export default function CSFLEMarker({
     <div className={badgeContainerStyles}>
       <button
         {...focusRingProps}
-        type="button"
+        type={focusRingProps.type as 'button'}
         data-testid="fle-connection-configuration"
         aria-label="Open connection In-Use Encryption configuration"
         title="Connection In-Use Encryption configuration"
--- a/packages/compass-sidebar/src/components/csfle-marker.tsx
+++ b/packages/compass-sidebar/src/components/csfle-marker.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import {
+  css,
+  Icon,
+  Badge,
+  BadgeVariant,
+  useFocusRing,
+  spacing,
+} from '@mongodb-js/compass-components';
+
+const badgeContainerStyles = css({
+  padding: `0 ${spacing[3]}px ${spacing[3]}px`,
+});
+
+const badgeButtonStyles = css({
+  background: 'inherit',
+  padding: 0,
+  margin: 0,
+  border: 'none',
+  cursor: 'pointer',
+});
+
+export default function CSFLEMarker({
+  csfleMode,
+  toggleCSFLEModalVisible,
+}: {
+  csfleMode?: 'enabled' | 'disabled' | 'unavailable';
+  toggleCSFLEModalVisible: () => void;
+}) {
+  const focusRingProps = useFocusRing<HTMLButtonElement>();
+
+  if (!csfleMode || csfleMode === 'unavailable') {
+    return null;
+  }
+
+  return (
+    <div className={badgeContainerStyles}>
+      <button
+        {...focusRingProps}
+        type="button"
+        data-testid="fle-connection-configuration"
+        aria-label="Open connection In-Use Encryption configuration"
+        title="Connection In-Use Encryption configuration"
+        className={badgeButtonStyles}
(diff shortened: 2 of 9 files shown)


packages/compass-components/src/components/modal-title.tsx
```diff
@@ -9,7 +9,7 @@ const modalTitleStyles = css({
 });
 
 function ModalTitle(
-  props: React.ComponentProps<typeof H3>
+  props: Omit<React.ComponentProps<typeof H3>, 'as'>
```
lerouxb, line 12: Not sure what the right way to do this would be. (Why would H3 even have as?)

packages/compass-sidebar/src/components/non-genuine-marker.tsx
```diff
@@ -0,0 +1,54 @@
...
+  // isGenuine === undefined means we haven't loaded the info yet
+  if (isGenuine !== false) {
+    return null;
+  }
+
+  return (
+    <div className={nonGenuineMarkerContainer}>
+      <button
+        {...focusRingProps}
+        type="button"
```
lerouxb, line 42: There was this weird thing where focusRingProps brought type="button" which then made typescript complain because button is the string "button" and not the literal "button". Which is annoying because type is optional here and the default is fine. Not sure how best/where to fix that but I worked around it by just having type="button" after the focus ring props so that I can specify it again as the literal.

Let me know if there's a better fix.

packages/compass-sidebar/src/components/csfle-marker.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {
+  css,
+  Icon,
+  Badge,
+  BadgeVariant,
+  useFocusRing,
+  spacing,
+} from '@mongodb-js/compass-components';
+
+const badgeContainerStyles = css({
```
lerouxb, line 11: Technically the badge container and button css plus this pattern / concept of a badge that's in a button that's in the sidebar is copy/pasted but when I really started looking at extracting it it felt like overkill. Maybe something to leave until we have a third one?

Let me know where I should draw the line :)

addaleax:
> do non-genuine mongodbs even support in-use encryption?

In-Use encryption, both QE and CSFLE, have explicit and automatic variants. Explicit encryption is community-licensed and, at least in the case of CSFLE, will also work against non-MongoDB servers. However, Compass does not support explicit encryption beyond its embedded shell, so I would not worry about this too much.

packages/compass-sidebar/src/components/non-genuine-marker.tsx
addaleax: Yeah, this feels a bit like a bug in the TS support for JSX/React. 
```suggestion
        type={focusRingProps.type as 'button'}
```

might be a bit better, it keeps the logic of `{...focusRingProps}` and makes it clear that this line is there to silence TS, not because it is actually relevant logic?

packages/compass-components/src/components/modal-title.tsx
addaleax: What error were you getting? It compiles fine for me when leaving it as `React.ComponentProps<typeof H3>` on top of this PR.

addaleax (approved):

packages/compass-sidebar/src/components/sidebar.tsx
```diff
@@ -107,6 +119,24 @@ export function Sidebar({
...
   );
 
+  const showNonGenuineModal = useCallback(() => {
+    toggleIsGenuineMongoDBVisible(true);
+  }, [toggleIsGenuineMongoDBVisible]);
+
+  const [isCSFLEModalVisible, setIsCSFLEModalVisible] = useState(false);
+
+  const toggleCSFLEModalVisible = useCallback(() => {
+    console.log('toggleCSFLEModalVisible');
```
Anemy, line 129:
```suggestion
```

packages/compass-components/src/components/modal-title.tsx
lerouxb: not here but where ModalTitle gets used:

Screenshot 2022-08-30 at 12 47 12

packages/compass-sidebar/src/components/non-genuine-marker.tsx
```diff
@@ -0,0 +1,54 @@
...
+  Badge,
+  BadgeVariant,
+  Icon,
+  css,
+  spacing,
+  useFocusRing,
+} from '@mongodb-js/compass-components';
+
+const nonGenuineMarkerContainer = css({
+  padding: `0 ${spacing[2]}px ${spacing[2]}px`,
```
Anemy, line 13:
```suggestion
  padding: `0 ${spacing[3]}px ${spacing[3]}px`,
```

Let's align this with the fle marker?

I was also looking at it a bit, how do you feel about a little bit of margin on the right of the badges?
Before:
Screen Shot 2022-08-30 at 7 46 47 AM

After:
Screen Shot 2022-08-30 at 7 50 45 AM

```ts

const iconStyles = css({
  marginRight: spacing[1]
});
```

lerouxb: oops. meant to go back and fix it up :)

Anemy (approved):

PR #3407, opened 2022-08-30
feat(indexes): fixed table header COMPASS-6042

mabaasit:
feat(indexes): fixed table header COMPASS-6042

fixed header scrolling on indexes table.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
```diff
@@ -58,20 +59,35 @@ const nameFieldStyles = css({
...
-  darkMode?: boolean;
   indexes: IndexDefinition[];
   canDeleteIndex: boolean;
   onSortTable: (column: SortColumn, direction: SortDirection) => void;
   onDeleteIndex: (name: string) => void;
 };
 
+const getContainerHeight = (): string => {
+  // 300px for the compass tabs, collection info etc
+  return `${window.innerHeight - 300}px`;
```
alenakhineika, line 80: Is it some min form high or from where we know that it should be 300?

mabaasit: its the calculated height from various component on the compass screen like mongosh shell at the bottom, top nav tabs, collection name/info row and then content tabs. This approximately adds upto to 300px

packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
```diff
@@ -100,53 +116,75 @@ export const IndexesTable: React.FunctionComponent<IndexesTableProps> = ({
...
+
+  useEffect(() => {
+    const resizeListener = () => {
+      setContainerHeight(getContainerHeight());
+    };
+    window.addEventListener('resize', resizeListener);
+    return () => {
+      window.removeEventListener('resize', resizeListener);
+    };
+  }, [containerHeight, indexes]);
```
gribnoysup, line 135: Effect depdencies are not matching used properties

packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
```diff
@@ -100,53 +116,75 @@ export const IndexesTable: React.FunctionComponent<IndexesTableProps> = ({
     return _columns;
   }, [canDeleteIndex, onSortTable]);
 
+  useEffect(() => {
+    const container =
+      tableRef.current?.getElementsByTagName('table')[0]?.parentElement;
+    if (container) {
+      container.style.height = containerHeight;
+    }
+  }, [tableRef, indexes, containerHeight]);
```
gribnoysup, line 125: indexes is part of the effect but is not used inside the effect

packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
```diff
@@ -100,53 +116,75 @@ export const IndexesTable: React.FunctionComponent<IndexesTableProps> = ({
     return _columns;
   }, [canDeleteIndex, onSortTable]);
 
+  useEffect(() => {
+    const container =
+      tableRef.current?.getElementsByTagName('table')[0]?.parentElement;
+    if (container) {
+      container.style.height = containerHeight;
```
gribnoysup, line 123: Doesn't seem like it's working when shell is expanded. Is this expected? 

image

packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
mcasimir: What happens if this value gets changed in another pr (for example we shrink the header or remove the tabs)?

mabaasit: the way compass sets up currently, this component is mounted first when the user is not on indexes screen yet.

as a result i had to put some initial height (window.innerHeight - 320px) which works well with the current compass. after the component renders, the height is calculated from the ref container.

alenakhineika: Is it possible maybe to subtract then not 320px but something like `window.innerHeight - window.innerHeight * 0.5`? like some percentage of height (find a proper value here that past).

packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
```diff
@@ -58,20 +60,46 @@ const nameFieldStyles = css({
...
   onSortTable: (column: SortColumn, direction: SortDirection) => void;
   onDeleteIndex: (name: string) => void;
+  globalAppRegistry: AppRegistry;
+};
+
+const getContainerHeight = (
+  containerRef: React.RefObject<HTMLDivElement>
+): string => {
+  const height =
+    containerRef.current?.parentElement?.parentElement?.clientHeight;
```
gribnoysup, line 84: This feels pretty brittle, you move elements around and this will break

packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
```diff
@@ -100,53 +128,81 @@ export const IndexesTable: React.FunctionComponent<IndexesTableProps> = ({
...
+    if (container) {
+      container.style.height = containerHeight;
+    }
+  }, [containerRef, containerHeight]);
+
+  useEffect(() => {
+    globalAppRegistry.on('compass:compass-shell:resized', () => {
+      setContainerHeight(getContainerHeight(containerRef));
+    });
+  }, [globalAppRegistry, containerRef]);
```
gribnoysup, line 143: Feels a bit drastic that we are making this whole plugin depend on another plugin just for this one header to be sticky.

gribnoysup (commented):
The way we depend on way too many global things here make me feel uneasy about approving this fix. If the bug you are trying to address here is coming from the necessity of having explicit size set on the container, I'd would maybe look into using something like [an auto-sizer component](https://github.com/bvaughn/react-virtualized-auto-sizer) that we use for making virtual lists render correctly. At least this feels more explicit and stable than relying on appRegistry events coming from other plugins

As a side note: what is the actual root cause here? Is it because of how leafygreen table is implemented? Is it something that we should highlight to them as a thing to address on their side in the future?

packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
mcasimir: Just wanted to weigh in with a super random comment here:
- having so many indexes to fill an entire screen has a cost, and while several customers may be doing that, it  should not be the case for the majority of them
- this is an optimization that we do not currently have in GA and that can wait
- the content of rows is largely self-explanatory and the table is still usable to a good degree without sticky headers

This is turning out to be a bit hacky, with obscure interactions and multiple plugins involved. If we want to go on with it I'm happy to reconsider it but I feel in this case we may rather want to stay patient, open a ticket for LG and wait for a proper feature on the table component itself, so we avoid creating more troubles for ourselves than those we solve.

mcasimir: Also of course, scrap that if there is a clean solution instead

mabaasit: i just cleaned it up and used AutoResizer in this one. However, we can also reach out to LG for adding fixed header support.

packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
```diff
@@ -102,57 +114,69 @@ export const IndexesTable: React.FunctionComponent<IndexesTableProps> = ({
     return _columns;
   }, [canDeleteIndex, onSortTable]);
 
+  useEffect(() => {
+    const container =
+      containerRef.current?.getElementsByTagName('table')[0]?.parentElement;
+    if (container) {
+      container.style.height = `${scrollHeight}px`;
+    }
```
gribnoysup, line 122: Please let's add a detailed comment here explaining what and why we are doing, as this is kinda obscure workaround for an issue with leafygreen components I worry that this runs the risk of being removed or broken at some point in the future due to lost context

gribnoysup (approved):

PR #3408, opened 2022-08-30
feat(sidebar): move db stats to the connection modal COMPASS-6079

lerouxb:
Screenshot 2022-08-30 at 15 35 18

Screenshot 2022-08-30 at 15 36 49

### DIFF
--- a/packages/compass-sidebar/src/stores/store.js
+++ b/packages/compass-sidebar/src/stores/store.js
@@ -65,6 +65,7 @@ store.onActivated = (appRegistry) => {
         }),
       };
     });
+
     store.dispatch(changeDatabases(dbs));
   }, 300);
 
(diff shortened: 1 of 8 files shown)


packages/compass-sidebar/src/stores/store.js
```diff
@@ -121,7 +122,7 @@ store.onActivated = (appRegistry) => {
       onDatabasesChange(instance.databases);
     });
 
-    instance.on('change:databases.', () => {
+    instance.on('change:databases', () => {
```
lerouxb, line 125: oops.

packages/database-model/lib/model.js
```diff
@@ -118,7 +118,7 @@ const DatabaseModel = AmpersandModel.extend(
       // if possible if we haven't fetched collections yet, but use the real
       // number if real collections info is available
       collectionsLength: {
-        deps: ['collection_count', 'collectionsStatus'],
+        deps: ['collection_count', 'collectionsStatus', 'collections'],
```
lerouxb, line 121: It needs collections.length which is probably a missing dep, but I'm not 100% sure if we can depend on a collection.

addaleax (approved):

PR #3411, opened 2022-08-31
chore: refactor action controls

mcasimir:
This fixes a bug that we would have with the new LG upgrade.

For the fix to propagate everywhere I've refactored the `<ItemActionControls />` to be used everywhere we have that pattern in the UI:
- added a "size" param
- allow specifying the threshold number of actions to collapse to a menu
- extracted 2 components: Menu and Action group, so the menu can be used without being forced to specify icons
- cleaned usage
- add the prefix for test-ids

Follow up:
- replace each <Menu> and action buttons with the components
- add missing tests for these components as they would now be used outside the sidebar too (we have e2e tests for that)

Screenshot 2022-08-31 at 13 51 40

Screenshot 2022-08-31 at 13 54 24

Screenshot 2022-08-31 at 13 54 32


packages/compass-databases-navigation/src/database-item.tsx
```diff
@@ -26,10 +26,7 @@ const buttonReset = css({
 });
 
 const expandButton = css({
-  // Not using leafygreen spacing here because none of them allow to align the
-  // button with the search bar content. This probably can go away when we are
-  // rebuilding the search also
-  padding: 7,
```
gribnoysup, line 32: Is this removed intentionally? This does make the icons misaligned and especially when database is selected makes the icon to jammed between elements:

image

mcasimir: mmm I have to take a second look, may need to re-add it with that alone didn't align though

mcasimir: Looks much better with the padding back

lerouxb (approved):

packages/compass-components/src/components/item-action-controls.tsx
```diff
@@ -1,140 +1,297 @@
...
+  isVisible = true,
+  'data-testid': dataTestId,
+}: {
+  actions: ItemAction<Action>[];
+  onAction(actionName: Action): void;
+  className?: string;
+  iconClassName?: string;
+  iconSize?: ItemActionButtonSize;
+  isVisible?: boolean;
+  collapseToMenuThreshold?: boolean;
```
gribnoysup, line 206: This looks like a leftover, right?

packages/compass-databases-navigation/src/database-item.tsx
```diff
@@ -26,6 +26,7 @@ const buttonReset = css({
 });
 
 const expandButton = css({
+  display: 'flex',
```
gribnoysup, line 29: Fine as-is, but just FYI setting font-size to zero (like the previous component did) would be a more direct fix to the alignment issue that you are seeing. Changing the parent display type to flex fixes it, but as a side-effect of converting children from inline elements (which svg-s are by default) to flex children

gribnoysup (approved):

PR #3417, opened 2022-09-01
chore: replace item actions with ItemAction* components

mcasimir:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
```diff
@@ -134,17 +135,17 @@ export const IndexesTable: React.FunctionComponent<IndexesTableProps> = ({
...
           </Cell>
-          {/* Delete column is conditional */}
-          {index.name !== '_id_' && canDeleteIndex && (
-            <Cell data-testid="index-drop-field" className={cellStyles}>
-              <div className={cx(deleteFieldStyles, 'delete-cell')}>
-                <DropField
-                  name={index.name}
-                  onDelete={() => onDeleteIndex(index.name)}
-                />
+          <Cell data-testid="index-actions-field" className={cellStyles}>
```
mcasimir, line 138: @mabaasit FYI this fixes the missing cell for `_id` too

mcasimir, line 138: @alenakhineika @mabaasit does this cause too many conflicts?

mabaasit: i dont think it will cause any conflicts. 

But this change (to show cell always), will also be rendered even if user can not delete an index (`canDeleteIndex  = false`), which i think will break the table layout.

mabaasit: Linking ticket here to close once this is approved: https://jira.mongodb.org/browse/COMPASS-6083

mcasimir: I see how would the layout break? Like the header won't be visible?

mabaasit: yes exactly. if we split the check, to show cell based on `canDeleteIndex` and then for `_id` just hide the cell contents

mcasimir: This PR can wait, is only required for the rebranding and is not changing anything at the moment, I can rebase from Alena's when it lands

mcasimir: Ok fixed the column issue!

alenakhineika: Can we always have the delete header and the delete table cell? Maybe to now have then only in case whencanDeleteIndex is not set?

alenakhineika: but now when i read your comment again this is exactly what you meant)

mabaasit (approved):

packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
mcasimir: Oh, that makes sense cause if the index creation fails then we need that column, I think we can always keep that yeah. I'll wait for the other PR and rebase then!

PR #3430, opened 2022-09-06
feat(sidebar): Expand/Collapse the new sidebar COMPASS-5967

lerouxb:
I made only the normal sidebar you see once connected resizable because we don't have designs for the connections one yet. Ideally I think for consistency we should have both be collapsible in future.

You can now drag the sidebar to collapse or expand it just like in the old sidebar (the one you see without the feature flag) and you can also press cmd-shift-d to toggle it or toggle it from the view menu.

Once collapsed you can also click on the leaf logo to expand it.
Screenshot 2022-09-06 at 11 36 37

Screenshot 2022-09-06 at 11 34 55

Screenshot 2022-09-06 at 11 37 21


packages/compass-sidebar/src/components/sidebar.tsx
```diff
@@ -134,57 +133,69 @@ export function Sidebar({
     [globalAppRegistryEmit]
   );
 
+  // TODO: how do we get notified about the toggle-sidebar app registry event and then get it here to pass to the resizable sidebar? or in the sidebar somehow?
```
lerouxb, line 136: ☝️ 

I suppose I can manage isExpanded from the outside and pass it in to the component? But that also feels weird.

packages/compass/src/main/menu.ts
```diff
@@ -395,6 +395,14 @@ function viewSubMenu(
         },
       },
       separator(),
+      {
+        label: '&Toggle Sidebar',
```
lerouxb, line 399: Something I noticed is that there's nothing in the menu that currently gets hidden depending on context. ie. disconnect is always visible even when you're not connected. And now Toggle Sidebar is also always visible even when you're not connected and that's kinda non-sensical when you you're looking at the connections sidebar.

alenakhineika (approved):
Looks great! ✨

packages/compass-components/src/components/resizeable-sidebar.tsx
```diff
@@ -57,24 +57,68 @@ const containerStylesLight = css({
...
+      }
+    },
+    [collapsable, expanded, minWidth, setExpanded, width]
+  );
+
+  const renderedWidth = expanded ? boundSidebarWidth(width) : collapsedWidth;
+
+  if (expanded && width === collapsedWidth) {
+    setWidth(prevWidth);
+  }
```
Anemy, line 120:
```suggestion
  useEffect(() => {
    if (expanded && width === collapsedWidth) {
      setWidth(prevWidth);
    }
  }, [ setWidth, expanded, width, collapsedWidth ]);
```
Could this benefit from a `useEffect`?

Anemy (commented):
Now that we have a hotkey for expanding/collapsing the sidebar, would it be possible to show the hotkey within the `aria-label` description of the resize handle? It helps screen readers know what hotkeys can be used for certain actions. https://github.com/mongodb-js/compass/blob/0b6f35b72781a0930cccf5767e328eb631e45bb6/packages/compass-components/src/components/resize-handle.tsx#L95
Maybe this is something that's part of https://jira.mongodb.org/browse/COMPASS-5590 instead. In that case lgtm! Left two comments on possible improvements, small things that are fine as is.

packages/compass-sidebar/src/components/sidebar.tsx
```diff
@@ -108,10 +112,15 @@ export function Sidebar({
         return;
       }
 
+      if (action === 'expand-sidebar') {
+        setIsExpanded(true);
```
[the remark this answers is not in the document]
lerouxb: oops.

packages/compass/src/main/menu.ts
lerouxb: OK so actually there is some precedent: https://github.com/mongodb-js/compass/blob/0b6f35b72781a0930cccf5767e328eb631e45bb6/packages/compass/src/main/menu.ts#L533

called here https://github.com/mongodb-js/compass/blob/0b6f35b72781a0930cccf5767e328eb631e45bb6/packages/compass-home/src/components/home.tsx#L92

packages/compass-components/src/components/resizeable-sidebar.tsx
lerouxb: I changed it, but thinking about it more it might be the rare case where we want something to happen _before_ render and maybe even in place of it rather than after. Or at least not render until the width's been updated. As this will likely cause multiple renders. Oh well.

lerouxb:
@Anemy Thanks. I added your accessibility suggestion on to that ticket for followup.

packages/compass-components/src/components/resizeable-sidebar.tsx
Anemy: There's a `useLayoutEffect` hook that's maybe what you're looking for:
https://reactjs.org/docs/hooks-reference.html#uselayouteffect
It runs before html renders.

PR #3431, opened 2022-09-06
chore: deduplicate storage path access and create compass-utils

addaleax:
In a number of places, we accessed the `@electron/remote` package
for fetching storage paths. However, with the enterprise improvements
project, we will now also require some of the packages which do this
to work in the main process, not just a renderer, so this needs
to be updated to conditionally use the `electron` package itself
rather than `@electron/remote`.

Instead of manually performing this change in every location,
extract this shared code and implement fallback handling there.
Also move the `promisifyAmpersandMethod` utility there, which
(as I understand it) only lived in data-service because we did
not have a better place for it at the time.

I know adding something named `utils` to a codebase can be a
pretty slippery slope, but I think it’s the best choice for
this particular problem.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-preferences-model/lib/model.js
+++ b/packages/compass-preferences-model/lib/model.js
@@ -2,14 +2,8 @@ var Model = require('ampersand-model');
 var storageMixin = require('storage-mixin');
 var get = require('lodash.get');
 var format = require('util').format;
-
-var electronApp;
-try {
-  electronApp = require('@electron/remote').app;
-} catch (e) {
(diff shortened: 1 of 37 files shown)


alenakhineika (approved):

Anemy (approved):
lgtm!

PR #3434, opened 2022-09-06
fix(storage-mixin): do not monkey-patch prototypes in SpliceDiskIpcBackend

addaleax:
It can not have been intentional that this overwrote the prototype
methods, rather than the methods of the specific instances, since
the previous code meant that using `SpliceDiskIpcBackend` would have
affected completely independent storage-mixing backends (even including
other instances of `SpliceDiskIpcBackend` itself).

This code has been broken like this since its inception. I did verify
that saving credentials works as it did before.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/storage-mixin/lib/backends/splice-disk.js
+++ b/packages/storage-mixin/lib/backends/splice-disk.js
@@ -27,19 +27,17 @@ function SpliceDiskBackend(options) {
 
   // patch the serialize methods in both backends
   var condition = options.secureCondition;
-  DiskBackend.prototype.serialize = function(model) {
-    debug('Serializing for disk backend with condition', condition);
-    var res = _.omitBy(model.serialize(), condition);
-    return res;
-  };
   this.diskBackend = new DiskBackend(options);
-
-  SecureBackend.prototype.serialize = function(model) {
-    debug('Serializing for secure backend with condition', condition);
-    var res = _.pickBy(model.serialize(), condition);
-    return res;
+  this.diskBackend.serialize = function(model) {
+    debug('Serializing for disk backend with condition', condition);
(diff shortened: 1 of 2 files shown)


packages/storage-mixin/lib/backends/splice-disk.js
```diff
@@ -27,19 +27,17 @@ function SpliceDiskBackend(options) {
 
   // patch the serialize methods in both backends
   var condition = options.secureCondition;
-  DiskBackend.prototype.serialize = function(model) {
```
Anemy, line 30: I might be missing something, where do we use `splice-disk`? I see that connection storage uses `splice-disk-ipc`.
I'm wondering if this can possibly be removed?

Anemy (commented):
Changes look good, one comment on possibly removing the `splice-disk` backend, I think good to go if we want to keep the backend, I couldn't find where it was used xD.

packages/storage-mixin/lib/backends/splice-disk.js
addaleax: > I might be missing something, where do we use `splice-disk`?

In my upcoming PR(s?) for connection import/export through the CLI :) Not that I planned for that, I also thought about removing this, but it’s working and will be useful later this week so I’d like to keep it for now.

(There might be an argument for sharing more code between this and splice-disk-ipc but I also don’t want to put a lot of effort into making storage-mixin prettier right now.)

Anemy (approved):
👍

PR #3435, opened 2022-09-06
feat: implement new preferences class COMPASS-6060

alenakhineika:
## Description

- Implement the new Preferences class as laid out in the technical design, to replace the Ampersand model as the public API of the `compass-preferences-model` package, and replace the Ampersand model’s usage throughout Compass [COMPASS-6060](https://jira.mongodb.org/browse/COMPASS-6060)
- Remove the unused preferences: agreedToLicense, indexDDL, showAutoUpdateBanner, queryBuilder, showExplainPlanTab, singleDocumentCrud, serverStats, chartView, authMongodb, authKerberos, authLdap, authX509, sslUnvalidated, sslServer, sslAll [COMPASS-6078](https://jira.mongodb.org/browse/COMPASS-6078)

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-e2e-tests/helpers/commands/set-feature.ts
```diff
@@ -1,15 +1,14 @@
...
   name: string,
   value: boolean | string
 ): Promise<void> {
   await browser.execute(
-    (_name, _value) => {
-      const preferences = (global as any)?.hadronApp?.preferences;
-      preferences.set(_name, _value);
-      preferences.save();
+    async (_name, _value) => {
+      await preferences.savePreferences({ [_name]: _value });
```
gribnoysup, line 11: This is running in a different context, you won't be able to pass the preferences model here like that

packages/compass-auto-updates/src/stores/store.js
```diff
@@ -19,8 +19,8 @@ store.onActivated = () => {
     /**
      * Listen to changes in user preferences.
      */
-    preferences.listenToAndRun(preferences, 'change:autoUpdates', () => {
-      if (preferences.isFeatureEnabled('autoUpdates')) {
+    preferences.onPreferenceChanged('autoUpdates', () => {
```
addaleax, line 22: Didn’t realize we were also doing this for auto-update, opened https://jira.mongodb.org/browse/COMPASS-6107 like the one for telemetry :+1:

packages/compass-preferences-model/index.d.ts
```diff
@@ -1,15 +1,29 @@
...
+  trackUsageStatistics: boolean;
+  enableFeedbackPanel: boolean;
+  theme: THEMES;
+}
+
+declare class CompassPreferencesModel {
+  fetchPreferences: () => Promise<void>;
+  savePreferences: (preferences: any) => Promise<void>;
+  getPreferenceValue: any;
+  onPreferenceChanged: (preferenceName: string, callback: () => void) => void;
```
addaleax, line 21:
```suggestion
  onPreferenceChanged: (preferenceName: keyof UserPreferences, callback: () => void) => void;
```

should there be a way to unsubscribe?

packages/compass-preferences-model/index.d.ts
```diff
@@ -1,15 +1,29 @@
...
+  trackErrors: boolean;
+  trackUsageStatistics: boolean;
+  enableFeedbackPanel: boolean;
+  theme: THEMES;
+}
+
+declare class CompassPreferencesModel {
+  fetchPreferences: () => Promise<void>;
+  savePreferences: (preferences: any) => Promise<void>;
+  getPreferenceValue: any;
```
addaleax, line 20: Just leaving a note here to property type these before it’s merged

packages/compass-logging/src/index.ts
```diff
@@ -27,25 +22,30 @@ export function createLoggerAndTelemetry(component: string): {
   debug: ReturnType<typeof createDebug>;
   track: TrackFunction;
 } {
-  // This application may not be running in an Node.js/Electron context.
-  const ipc: HadronIpcRenderer | null = isElectronRenderer
```
alenakhineika, line 31: `hadron-ipc` does this check itself.

packages/compass-e2e-tests/helpers/commands/set-feature.ts
alenakhineika: A very good point 😏

packages/compass-preferences-model/index.d.ts
alenakhineika: Yeah, there are many any :) Don't worry, I will replace them with proper types before converting PR from draft to open.

packages/compass-logging/src/index.ts
```diff
@@ -1,21 +1,16 @@
...
-function emit(
-  ipc: HadronIpcRenderer | null,
-  event: string,
-  data: Record<string, any>
-): void {
+function emit(event: string, data: Record<string, any>): void {
   // We use ipc.callQuiet instead of ipc.call because we already
   // print debugging messages below
-  void ipc?.callQuiet?.(event, data);
+  void (ipc as any)?.callQuiet?.(event, data);
```
addaleax, line 13:
```suggestion
  void ipc?.ipcRenderer?.callQuiet?.(event, data);
```

packages/compass-logging/src/index.ts
addaleax: `hadron-ipc` logs a warning when it’s being used outside of electron, though, and it always expects to run in a Node.js environment currently (accesses `process.env`).

packages/compass-logging/src/index.ts
```diff
@@ -57,10 +57,7 @@ export function createLoggerAndTelemetry(component: string): {
     event: string,
     properties: TrackProps = {}
   ): Promise<void> => {
-    const isTrackingEnabled = (global as any)?.hadronApp?.isFeatureEnabled(
-      'trackUsageStatistics'
-    );
-    if (!isTrackingEnabled) {
+    if (!trackUsageStatistics) {
```
addaleax, line 60: Why are we changing the structure of the code here? I really liked the idea of `getPreferenceValue()` in the technical design to allow specifically accessing one property value, which could replace `isFeatureEnabled()` here one-to-one.

I’d rather not add change listeners to every place in which we consume preferences, i.e. avoid tracking the same state variables in multiple locations.

packages/compass-preferences-model/lib/model.js
```diff
@@ -1,323 +1,273 @@
...
+    );
+  }
+}
+
+/**
+ * API to communicate with preferences from the electron renderer process.
+ */
+const preferencesIpc = {
+  savePreferences(attributes) {
+    if (ipc && ipc.ipcRenderer && typeof ipc.ipcRenderer.invoke === 'function') {
```
addaleax, line 248:
```suggestion
    if (typeof ipc?.ipcRenderer?.invoke === 'function') {
```

packages/compass-preferences-model/lib/model.js
```diff
@@ -1,323 +1,273 @@
...
+      return ipc.ipcRenderer.invoke('compass:save-preferences', attributes);
+    }
   },
-  onAppVersionMismatch: function(lastKnownVersion, currentVersion) {
-    debug('version mismatch detected: was %s, now %s',
-      lastKnownVersion, currentVersion);
+  getPreferences() {
+    if (ipc && ipc.ipcRenderer && typeof ipc.ipcRenderer.invoke === 'function') {
```
addaleax, line 253:
```suggestion
    if (typeof ipc?.ipcRenderer?.invoke === 'function') {
```

packages/compass-preferences-model/lib/model.js
```diff
@@ -1,323 +1,273 @@
...
-   */
-  isFeatureEnabled: function(feature) {
-    // main network switch overwrites all network related features
-    if (['enableMaps', 'trackErrors', 'enableFeedbackPanel',
-      'trackUsageStatistics', 'autoUpdates'].indexOf(feature) !== -1) {
-      return this.networkTraffic &&
-        process.env.HADRON_ISOLATED !== 'true' &&
-        get(this, feature);
+  getConfigurableUserPreferences() {
+    if (ipc && ipc.ipcRenderer && typeof ipc.ipcRenderer.invoke === 'function') {
```
addaleax, line 258:
```suggestion
    if (typeof ipc?.ipcRenderer?.invoke === 'function') {
```

packages/compass-preferences-model/lib/model.js
```diff
@@ -1,323 +1,273 @@
...
+      return ipc.ipcRenderer.invoke('compass:get-configurable-user-preferences');
     }
-    var res = get(this, feature, null);
-    // don't allow asking for unknown features to prevent bugs
-    if (res === null) {
-      throw new Error(format('Feature %s unknown.', feature));
+  },
+  onPreferencesChanged(callback) {
+    if (ipc && ipc.ipcRenderer && typeof ipc.ipcRenderer.on === 'function') {
```
addaleax, line 263:
```suggestion
    if (typeof ipc?.ipcRenderer?.on === 'function') {
```

packages/compass-preferences-model/lib/model.js
```diff
@@ -1,323 +1,273 @@
...
+const preferencesIpc = {
+  savePreferences(attributes) {
+    if (ipc && ipc.ipcRenderer && typeof ipc.ipcRenderer.invoke === 'function') {
+      return ipc.ipcRenderer.invoke('compass:save-preferences', attributes);
+    }
   },
-  onAppVersionMismatch: function(lastKnownVersion, currentVersion) {
-    debug('version mismatch detected: was %s, now %s',
-      lastKnownVersion, currentVersion);
+  getPreferences() {
```
addaleax, line 252: Is the naming mismatch with `Preferences` intentional (getPreferences vs getAllPreferences)?

packages/compass-schema/src/components/field/field.jsx
```diff
@@ -119,7 +120,7 @@ class Field extends Component {
    */
   getSemanticType(type) {
     // check if the type represents geo coordinates, if privacy settings allow
-    if (global.hadronApp.isFeatureEnabled('enableMaps')) {
+    if (this.props.enableMaps) {
```
addaleax, line 123: Similar question here – instead of adding another variable to the store, installing a listener etc., can we just rewrite this check here?

packages/compass/src/app/index.js
```diff
@@ -185,110 +183,86 @@ var Application = View.extend({
...
-  tourClosed: function() {
-    app.preferences.unset('showFeatureTour');
-    app.preferences.save({}, {success: () => {
-      if (!app.preferences.showedNetworkOptIn && process.env.HADRON_ISOLATED !== 'true') {
-        ipc.ipcRenderer.emit('window:show-network-optin');
-      }
-    }});
+  tourClosed: async function() {
+    const { showedNetworkOptIn } = await preferencesIpc.getPreferences();
+    if (!showedNetworkOptIn && process.env.HADRON_ISOLATED !== 'true') {
```
addaleax, line 210:
```suggestion
    // TODO(COMPASS-6065): Remove `HADRON_ISOLATED` usage
    if (!showedNetworkOptIn && process.env.HADRON_ISOLATED !== 'true') {
```

packages/compass/src/app/index.js
```diff
@@ -185,110 +183,86 @@ var Application = View.extend({
...
+    const tourView = new TourView({ force, previousVersion });
     if (tourView.features.length > 0) {
       tourView.on('close', this.tourClosed.bind(this));
       this.renderSubview(tourView, this.queryByHook('tour-container'));
     } else {
       this.tourClosed();
     }
   },
-  tourClosed: function() {
-    app.preferences.unset('showFeatureTour');
```
addaleax, line 209: Where is the unsetting happening now?

packages/compass-preferences-model/index.d.ts
alenakhineika: I was thinking about it. But when and where unsubscribing should happen? I looks at other similar events in Compass and didn't find a pattern how we do that.

packages/compass/src/app/index.js
alenakhineika: With one of recent commits I replied showFeatureTour with the previousVersion session value and the lastKnownVersion preference.

packages/compass-logging/src/index.ts
alenakhineika: Reverted.

packages/compass-logging/src/index.ts
alenakhineika: Updated.

packages/compass-schema/src/components/field/field.jsx
alenakhineika: Also updated this.

packages/compass-connections/package.json
```diff
@@ -28,6 +28,13 @@
     ".": "./src/index.ts"
   },
   "types": "./dist/index.d.ts",
+  "mocha": {
+    "extension": [
+      "ts"
+    ],
+    "file": "./test/mocha-plugins.ts",
+    "recursive": true
+  },
```
gribnoysup, line 37: There is a `.mocharc.js` file in this package that already has mocha configuration, I would have a slight preference for us to have all the configuration in one place instead of multiple (I am also not sure if mocha will actually merge this and not just use this one instead of a shared config)

packages/compass/package.json
```diff
@@ -164,6 +164,7 @@
   "dependencies": {
     "@mongosh/node-runtime-worker-thread": "^1.5.4",
     "clipboard": "^2.0.6",
+    "compass-preferences-model": "^2.0.0",
```
addaleax, line 167: This section only lists externalized dependencies, which `compass-preferences-model` is not – is there a special reason for moving this to regular `dependencies`?

packages/compass-preferences-model/lib/renderer-ipc.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+    return;
+  },
+  getConfigurableUserPreferences() {
+    if (typeof ipc?.ipcRenderer?.invoke === 'function') {
+      return ipc.ipcRenderer.invoke('compass:get-configurable-user-preferences');
+    }
+    return;
+  },
+  onPreferencesChanged2(callback) {
```
addaleax, line 25: Is this a typo?

packages/compass-preferences-model/index.d.ts
```diff
@@ -1,15 +1,31 @@
...
+  enableFeedbackPanel: boolean;
+  theme: THEMES;
+}
+
+export type GlobalPreferences = UserPreferences; // TODO: extend with global preferences.
+
+declare class Preferences {
+  constructor(basepath?: string): void;
+  fetchPreferences(): Promise<GlobalPreferences>;
+  savePreferences(attributes: GlobalPreferences): Promise<GlobalPreferences>;
```
addaleax, line 24:
```suggestion
  savePreferences(attributes: Partial<GlobalPreferences>): Promise<GlobalPreferences>;
```

packages/compass-preferences-model/index.d.ts
```diff
@@ -1,15 +1,31 @@
...
+  getPreferences(): Promise<GlobalPreferences>;
+  getConfigurableUserPreferences(): Promise<UserPreferences>;
+}
+
+export { preferencesIpc };
```
addaleax, line 29: This is untyped … do we maybe want to convert the implementation files to TS after this PR? That would also make it easier to ensure that `preferencesProps` contains the right things.

packages/compass-preferences-model/index.d.ts
```diff
@@ -1,15 +1,31 @@
...
+}
+
+export type GlobalPreferences = UserPreferences; // TODO: extend with global preferences.
+
+declare class Preferences {
+  constructor(basepath?: string): void;
+  fetchPreferences(): Promise<GlobalPreferences>;
+  savePreferences(attributes: GlobalPreferences): Promise<GlobalPreferences>;
+  getPreferences(): Promise<GlobalPreferences>;
+  getConfigurableUserPreferences(): Promise<UserPreferences>;
```
addaleax, line 26: Just a note here, but it’s surprising that this is named `get...` but sets defaults, and we’re not going to be able to keep this signature in the future (because there will be preferences that are fixed through the CLI or the global config and `Promise<UserPreferences>` does not communicate that). We can take care of that when we get to it, though.

addaleax (commented):
Left a few minor comments, but overall this looks great!

packages/compass-preferences-model/lib/renderer-ipc.js
alenakhineika: No.

alenakhineika: I am trying to mock this method for tests and trying different options to check how it is being called and where it fails. I push intermediate results for discussion. Please ignore this kind of things until pr becomes ready for review.

packages/compass/src/main/setup-preferences.ts
```diff
@@ -0,0 +1,27 @@
...
+const { basepath } = getStoragePaths() || {};
+
+const setupPreferences = async() => {
+  const preferences = new Preferences(basepath);
+
+  await preferences.fetchPreferences();
+
+  ipcMain.handle('compass:save-preferences', async (event, attributes) => {
+    const savedPreferencesValues = await preferences.savePreferences(attributes);
+    ipcMain.broadcast('compass:preferences-changed', savedPreferencesValues);
```
addaleax, line 14: Noticed when we were talking that we only emit this event here … ideally we’d also account for changes to the preferences coming from the main process and emit the event in that case. Maybe we could add `onPreferencesChanged` to the `Preferences` class as well, and emit it when `savePreferences()` is called, then register a handler here that does the IPC broadcast?

alenakhineika: i think in design we agreed that global/cli prefs are set only on startup, what case could be that the preferences change happens in main and not via the 'compass:save-preferences' event?

packages/compass-preferences-model/index.d.ts
alenakhineika: Yes!

packages/compass/src/main/setup-preferences.ts
addaleax: I don’t have a concrete answer, and I agree that we’re currently not planning to introduce anything where this would make a difference, it just feels like a small footgun for future changes that preferences can be modified without triggering this event.

packages/compass-preferences-model/src/index.ts
```diff
@@ -0,0 +1,11 @@
+import Preferences from './preferences';
+
+export default Preferences;
```
addaleax, line 3: Nice, you already did the TS conversion! :blue_heart:

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -0,0 +1,318 @@
...
+const debug = createDebug('mongodb-compass:models:preferences');
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const Model = require('ampersand-model');
+
+export enum THEMES {
+  DARK = 'DARK',
+  LIGHT = 'LIGHT',
+  OS_THEME = 'OS_THEME',
+}
```
addaleax, line 17:
```suggestion
type THEMES = 'DARK' | 'LIGHT' | 'OS_THEME';
```

as we usually tend to do today? We only use this in TS code so there shouldn’t be much need for an enum (completely optional though)

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -0,0 +1,318 @@
...
+
+export type GlobalPreferences = UserPreferences; // TODO: extend with global preferences.
+
+export type OnPreferencesChangedCallback = (
+  changedPreferencesValues: Partial<GlobalPreferences>
+) => void;
+
+export type OnPreferenceValueChangedCallback = (
+  preferenceValue: string | boolean
+) => void;
```
addaleax, line 37:
```suggestion
export type OnPreferenceValueChangedCallback<T> = (
  preferenceValue: T
) => void;
```

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -0,0 +1,318 @@
...
+
+    // Fetch user preferences from the Ampersand model.
+    const fetchUserPreferences = promisifyAmpersandMethod(
+      userPreferencesModel.fetch.bind(userPreferencesModel)
+    );
+
+    try {
+      await fetchUserPreferences();
+    } catch (err) {
+      debug('Failed to load preferences, error while fetching models', {
```
addaleax, line 234:
```suggestion
      // TODO: Log this error with @compass/logging
      debug('Failed to load preferences, error while fetching models', {
```

feels like we should do that, not being able to load preferences sounds like something that could be fairly hard to debug for users

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -0,0 +1,318 @@
...
+    const saveUserPreferences: (
+      attributes: Partial<GlobalPreferences>
+    ) => Promise<void> = promisifyAmpersandMethod(
+      userPreferencesModel.save.bind(userPreferencesModel)
+    );
+
+    try {
+      await saveUserPreferences(attributes);
+    } catch (err) {
+      debug('Failed to save preferences, error while saving models', {
```
addaleax, line 259:
```suggestion
      // TODO: Log this error with @compass/logging
      debug('Failed to save preferences, error while saving models', {
```

packages/compass-preferences-model/src/renderer-ipc.ts
```diff
@@ -0,0 +1,52 @@
...
+      return ipc.ipcRenderer.invoke(
+        'compass:get-configurable-user-preferences'
+      );
+    }
+    return {} as UserPreferences;
+  },
+  onPreferenceValueChanged(
+    preferenceName: string,
+    callback: OnPreferenceValueChangedCallback
+  ) {
```
addaleax, line 35:
```suggestion
  onPreferenceValueChanged<K extends keyof GlobalPreferences>(
    preferenceName: K,
    callback: OnPreferenceValueChangedCallback<GlobalPreferences[K]>
  ): () => void {
```

or, if you feel like that’s less involved, without `OnPreferenceValueChangedCallback` at all:

```suggestion
  onPreferenceValueChanged<K extends keyof GlobalPreferences>(
    preferenceName: K,
    callback: (value: GlobalPreferences[K]) => void
  ): () => void {
```

packages/compass-preferences-model/src/renderer-ipc.ts
```diff
@@ -0,0 +1,52 @@
...
+  GlobalPreferences,
+  UserPreferences,
+  OnPreferenceValueChangedCallback,
+} from './preferences';
+
+/**
+ * API to communicate with preferences from the electron renderer process.
+ */
+export const preferencesIpc = {
+  savePreferences(attributes: Partial<GlobalPreferences>) {
```
addaleax, line 12:
```suggestion
  savePreferences(attributes: Partial<GlobalPreferences>): Promise<GlobalPreferences> {
```

packages/compass-preferences-model/src/renderer-ipc.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (typeof ipc?.ipcRenderer?.invoke === 'function') {
+      return ipc.ipcRenderer.invoke('compass:save-preferences', attributes);
+    }
+    return {} as UserPreferences;
```
addaleax, line 16:
```suggestion
    return Promise.resolve({});
```

packages/compass-preferences-model/src/renderer-ipc.ts
```diff
@@ -0,0 +1,52 @@
...
+  },
+  getPreferences() {
```
addaleax, line 18:
```suggestion
  getPreferences(): Promise<GlobalPreferences> {
```

packages/compass-preferences-model/src/renderer-ipc.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (typeof ipc?.ipcRenderer?.invoke === 'function') {
+      return ipc.ipcRenderer.invoke('compass:get-preferences');
+    }
+    return {} as GlobalPreferences;
```
addaleax, line 22:
```suggestion
    return Promise.resolve({});
```

packages/compass-preferences-model/src/renderer-ipc.ts
```diff
@@ -0,0 +1,52 @@
...
+  },
+  getConfigurableUserPreferences() {
```
addaleax, line 24:
```suggestion
  getConfigurableUserPreferences(): Promise<UserPreferences> {
```

packages/compass-preferences-model/src/renderer-ipc.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (typeof ipc?.ipcRenderer?.invoke === 'function') {
+      return ipc.ipcRenderer.invoke(
+        'compass:get-configurable-user-preferences'
+      );
+    }
+    return {} as UserPreferences;
```
addaleax, line 30:
```suggestion
    return Promise.resolve({});
```

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -0,0 +1,318 @@
...
+        backend: 'disk',
+        basepath,
+      },
+    });
+
+    this._onPreferencesChangedCallbacks = [];
+    this._userPreferencesModel = new PreferencesModel();
+  }
+
+  async fetchPreferences() {
```
addaleax, line 223:
```suggestion
  async fetchPreferences(): Promise<GlobalPreferences> {
```

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -0,0 +1,318 @@
...
+    } catch (err) {
+      debug('Failed to load preferences, error while fetching models', {
+        message: (err as Error).message,
+      });
+    }
+
+    return this.getPreferences();
+  }
+
+  async savePreferences(attributes: Partial<GlobalPreferences>) {
```
addaleax, line 242:
```suggestion
  async savePreferences(attributes: Partial<GlobalPreferences>): Promise<GlobalPreferences> {
```

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -0,0 +1,318 @@
...
+    const changedPreferencesValues = pickBy(
+      savedPreferencesValues,
+      (value, key) => Object.keys(attributes).includes(key)
+    );
+    this._callOnPreferencesChanged(changedPreferencesValues);
+
+    return this.getPreferences();
+  }
+
+  getPreferences() {
```
addaleax, line 274:
```suggestion
  getPreferences(): GlobalPreferences {
```

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -0,0 +1,318 @@
...
+    // TODO: merge user, global, and CLI preferences here.
+    return (
+      this._userPreferencesModel.getAttributes({
+        props: true,
+        derived: true,
+      }) || {}
+    );
+  }
+
+  async getConfigurableUserPreferences() {
```
addaleax, line 284:
```suggestion
  async getConfigurableUserPreferences(): Promise<UserPreferences> {
```

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -0,0 +1,318 @@
...
+        showedNetworkOptIn: true,
+        theme: THEMES.LIGHT,
+      });
+    }
+    const preferences = this.getPreferences();
+    return Object.fromEntries(
+      Object.entries(preferences).filter(
+        ([key]) => preferencesProps[key].ui === true
+      )
+    );
```
addaleax, line 302:
```suggestion
    ) as UserPreferences;
```

(always a bit sad that TS can’t really do this for us)

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+  _callOnPreferencesChanged(
+    changedPreferencesValues: Partial<GlobalPreferences>
+  ) {
```
addaleax, line 307:
```suggestion
  ): void {
```

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -0,0 +1,318 @@
...
+    for (const callback of this._onPreferencesChangedCallbacks) {
+      return callback(changedPreferencesValues);
+    }
+  }
+
+  onPreferencesChanged(callback: OnPreferencesChangedCallback) {
```
addaleax, line 313:
```suggestion
  onPreferencesChanged(callback: OnPreferencesChangedCallback): void {
```

packages/compass-user-model/lib/model.js
```diff
@@ -37,8 +37,8 @@ var User = Model.extend(storageMixin, {
   }
 });
 
-User.getOrCreate = function(id, done) {
-  var user = new User({
+User.getOrCreate = (id) => new Promise((resolve, reject) => {
```
addaleax, line 40: Also big :clap: for these cleanups! Definitely feel like these are worth doing when we touch older code :)

packages/compass-preferences-model/src/renderer-ipc.ts
```diff
@@ -0,0 +1,52 @@
...
+  },
+  onPreferenceValueChanged(
+    preferenceName: string,
+    callback: OnPreferenceValueChangedCallback
+  ) {
+    const listener = (_: Event, preferences: GlobalPreferences) => {
+      if (Object.keys(preferences).includes(preferenceName)) {
+        // TODO: find a proper type.
+        return callback((preferences as any)[preferenceName]);
```
addaleax, line 39: This should “just work” with the suggestion above this one:

```suggestion
        return callback(preferences[preferenceName]);
```

packages/compass-auto-updates/src/modules/index.js
```diff
@@ -69,3 +72,41 @@ export const visitReleaseNotes = () => {
     shell.openExternal(RELEASE_NOTES);
   };
 };
+
+/**
+ * The autoUpdates preference value changed.
+ *
+ * @param {Boolean} autoUpdates - The autoUpdates prefernce valuse.
```
addaleax, line 79:
```suggestion
 * @param {Boolean} autoUpdates - The autoUpdates preference value.
```

addaleax (approved):
Looks great! :rocket:  Don’t worry about the # of comments, it’s pretty much all typos/minor TS adjustments like adding missing return types :slightly_smiling_face:

packages/compass-preferences-model/src/index.ts
alenakhineika: Yeah decided to do it sooner than later to catch potential issues since it touches so many parts of the app.

packages/compass-preferences-model/src/preferences.ts
alenakhineika: I was hoping you advise on that 😄❤️

packages/compass-preferences-model/src/preferences.ts
alenakhineika: But then it will be only a type and we won't be able to access it as a value as it is used in code right now, eg:
```ts
const { THEMES } = require('compass-preferences-model');
if (theme === THEMES.OS_THEME
    && remote.nativeTheme.shouldUseDarkColors
  ) { }
```

addaleax: Yes – didn’t realize that there’s one place in which we access it from vanilla JS code :)

packages/compass-preferences-model/src/renderer-ipc.ts
alenakhineika: But we probably still need to set a type here like:
```ts
return Promise.resolve({} as GlobalPreferences);
```
Otherwise, ts will complain about missing properties of the GlobalPreferences type.

addaleax: Hm yeah … good point. I think I’d go with what you’re suggesting here then :+1:

packages/compass-preferences-model/src/preferences.ts
alenakhineika: Yeeeeah these conversions look not very clean in code.

packages/compass-schema/src/components/field/field.jsx
```diff
@@ -110,6 +112,11 @@ class Field extends Component {
     return null;
   }
 
+  async componentDidMount() {
+    const { enableMaps } = await preferencesIpc.getPreferences();
+    this.setState({ enableMaps });
+  }
```
gribnoysup, line 118: Seems like we only set the value once, but don't update it if it was changed. There probably should be a corresponding subscription to preferences change here?

addaleax: @gribnoysup We decided to leave that out of this PR for now, since it didn’t change from how this was behaving before

gribnoysup: Gotcha, thanks for the context. Let's open a ticket for that if we don't have one yet, but otherwise that answers my question

alenakhineika: https://jira.mongodb.org/browse/COMPASS-6127

PR #3443, opened 2022-09-07
background color, not bold

lerouxb:
The reason why we tried bolding the text on hover rather than a background color was because of the hover effect on the icon button on top of the hover effect of the tree item. See videos:

https://user-images.githubusercontent.com/69737/188893757-896967a3-eb88-443f-bf39-a57d26ddafce.mov

https://user-images.githubusercontent.com/69737/188893768-df7049fb-c82d-4f04-a61d-46a70ccfb3c0.mov

https://user-images.githubusercontent.com/69737/188893791-170b87c6-3fbf-4426-86a2-d0d2737692e9.mov

https://user-images.githubusercontent.com/69737/188893793-3d13c3fc-c705-462d-8404-a639bccff756.mov

But.. I now think this is maybe less of a problem than the bold text jiggle effect?

Reference:
leafy green's own sidebar nav has green active color (similar to us) and gray hover effect (same as here): https://www.mongodb.design/component/side-nav/example/

leafy green's icon button also has that gray background hover effect:
https://www.mongodb.design/component/icon-button/example/

The reason why the database navigation tree behaves different to the connections list is a bit technical, but basically in the nav tree case the hover is on the entire \<div\> whereas in the connection case there's a \<button\> in there. So in the nav tree the icon button is a child of the hovered thing and in the connection case it is a sibling.

We can maybe restructure the html? Not sure if we can do that without substantial refactoring.

### DIFF
--- a/packages/compass-components/src/components/resizeable-sidebar.tsx
+++ b/packages/compass-components/src/components/resizeable-sidebar.tsx
@@ -33,6 +33,7 @@ const containerStylesDark = css({
   '--item-color': 'white',
   '--item-color-active': '#71F6BA', // TODO: there is no uiColors.green.light1
   '--item-bg-color': uiColors.gray.dark3,
+  '--item-bg-color-hover': uiColors.gray.dark2,
   '--item-bg-color-active': uiColors.black,
 
   color: 'var(--color)',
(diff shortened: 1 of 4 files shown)


Anemy (approved):

lerouxb:
I'm going with this for now and will file some follow-up tickets to bring the connection list and nav tree a bit more in line with each other. Visually I think we want it to work like the connection list, but the connection list has some of its own issues. cc @gribnoysup

lerouxb:
https://jira.mongodb.org/browse/COMPASS-6112

PR #3451, opened 2022-09-08
fix(telemetry): undefined rather than false if getCloudInfo fails, support SRV URIs COMPASS-6111

lerouxb:
https://mongodb.slack.com/archives/G2L10JAV7/p1662621306480259

It turns out that if `await getCloudInfo()` fails we cast isPublicCloud (which is undefined) to false and public_cloud_name became ''.

This rather leaves those `undefined` when we don't know.

The tests in this area even checked that it set those to false and '' respectively. But it passed in domains that would be looked up and never found for all four/five test cases.

srv URIs will now be resolved in order to find a valid hostname that we can use with getCloudInfo.


lerouxb:
actually now that I think about it, this is not the bug. Those defaults only get used with localhost or digital ocean.

lerouxb:
closing for now. can re-open if/when we get to the ticket

packages/compass-connections/src/modules/telemetry.spec.ts
```diff
@@ -178,21 +175,28 @@ describe('connection tracking', function () {
...
   // eslint-disable-next-line mocha/no-setup-in-describe
   [
+    // domain resolves, so is_public_cloud and public_cloud_name are defined
     {
-      url: 'mongodb://compass-data-sets.e06dc.mongodb.net',
+      url: 'mongodb://compass-data-sets-shard-00-00.e06dc.mongodb.net',
       is_srv: false,
       title: 'is atlas, no srv',
     },
+    // this domain does not resolve. Are there atlas dev domains that we can use to test?
```
lerouxb, line 184: ☝️

packages/compass-connections/src/modules/telemetry.spec.ts
```diff
@@ -237,6 +239,11 @@ describe('connection tracking', function () {
         has_kms_azure: false,
       };
 
+      if (!is_srv) {
```
lerouxb, line 242: Shouldn't valid mongodb+srv urls result in is_public_cloud: true and public_cloud_name to be filled in to be one of the clouds atlas runs on? Right now those all seem to fail dns lookup in mongodb-cloud-info. Before this PR it would set is_public_cloud to false and public_cloud_name to ''. Now it leaves those undefined. Neither are ideal, I think.

packages/compass-connections/src/modules/telemetry.spec.ts
```diff
@@ -286,7 +293,7 @@ describe('connection tracking', function () {
     const trackEvent = once(process, 'compass:track');
     const connectionInfo = {
       connectionOptions: {
-        connectionString: 'mongodb://notlocalhost',
```
lerouxb, line 289: This domain didn't resolve to anything, so I don't think it was a very good test. The code looks up  the IP in some sets to figure out if it is contained in there and obviously nothing that resolves would be in there. But neither of the Atlas tests higher up resolved either.

packages/compass-connections/src/modules/telemetry.spec.ts
addaleax: I think mongodb+srv:// URLs should result in the same telemetry being sent as their resolved mongodb:// equivalents, yes.

lerouxb: OK I added code to resolve those URLs now.

packages/compass-connections/src/modules/telemetry.spec.ts
```diff
@@ -178,26 +175,33 @@ describe('connection tracking', function () {
...
       title: 'is dev atlas, no srv',
     },
+    // domain does not resolve, so is_public_cloud and public_cloud_name are both undefined
     {
       url: 'mongodb+srv://compass-data-sets.e06dc.mongodb.net',
       is_srv: true,
       title: 'is atlas, is srv',
     },
+
+    // TODO: we need a dev atlas domain for testing
```
lerouxb, line 197: ☝️

packages/compass-connections/src/modules/telemetry.ts
```diff
@@ -80,7 +122,7 @@ async function getConnectionData({
   const connectionStringData = new ConnectionString(connectionString, {
     looseValidation: true,
   });
-  const hostName = connectionStringData.hosts[0];
+  //const hostName = connectionStringData.hosts[0];
```
Anemy, line 125:
```suggestion
```

packages/compass-connections/src/modules/telemetry.ts
```diff
@@ -5,28 +5,41 @@ import { getCloudInfo } from 'mongodb-cloud-info';
...
-  };
+type HostInformation = {
+  is_localhost: boolean;
+  is_atlas_url: boolean;
+  is_do_url: boolean;
+  is_public_cloud?: boolean;
+  public_cloud_name?: string;
+};
+
+async function getHostInformation(host: string | null) {
```
addaleax, line 20:
```suggestion
async function getHostInformation(host: string | null): Promise<HostInformation> {
```

addaleax (approved):

PR #3454, opened 2022-09-08
feat(compass): add CLI connection import/export COMPASS-6073 COMPASS-6074

addaleax:
##### ~~chore(compass-e2e-tests): refactor out common favorite operations~~

~~(Split out into https://github.com/mongodb-js/compass/pull/3453 to avoid/limit possible conflicts with sidebar work)~~

##### chore(connection-model): allow accessing from main process

`splice-disk` is the main process equivalent of `splice-disk-ipc`
for renderers. Properly expose it and use it in connection-model.

##### feat(compass): add CLI connection import/export COMPASS-6073 COMPASS-6074

Add support for exporting and importing favorite connections when
using Compass as a CLI. Users can specify `--export-connections=<file>`,
`--import-connections=<file>` and optionally also a passphrase
through `--passphrase=<string>`.

The encryption/decryption and export/import logic are added to the
data-service package. The Compass main entry point is modified
to the degree to which it is necessary to allow running it as
a CLI rather than always in GUI mode.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/data-service/src/import-export-connection.spec.ts
```diff
@@ -0,0 +1,211 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Anemy, line 1:
```suggestion
```
Leftover lint?

packages/data-service/src/import-export-connection.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+        storage: outStorage,
+        passphrase: 'p4ssw0rd',
+      });
+      expect.fail('missed exception');
+    } catch (err) {
+      expect(err.message).to.include(
+        'Cannot decrypt due to corrupt data or wrong passphrase'
+      );
+    }
+  });
```
Anemy, line 210:
```suggestion
  });
  
  it('rejects importing with an incorrect passphrase', async function () {
    const exported = await exportConnections({
      storage: inStorage,
      passphrase: 'p4ssw0rd',
    });
    try {
      await importConnections(exported, {
        storage: outStorage,
        passphrase: 'wr0ng_p4ssw0rd!',
      });
      expect.fail('missed exception');
    } catch (err) {
      expect(err.message).to.include(
        'Cannot decrypt due to corrupt data or wrong passphrase'
      );
    }
  });
```

packages/compass/src/main/import-export-connections.ts
```diff
@@ -0,0 +1,18 @@
+import { promises as fs } from 'fs';
+import type { ExportConnectionOptions, ImportConnectionOptions } from 'mongodb-data-service';
+
+export async function doExportConnections(filename: string, options: ExportConnectionOptions = {}): Promise<void> {
+  console.log(`Exporting connections to "${filename}" (${options.passphrase ? 'with' : 'without'} passphrase)`);
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const { exportConnections } = await import('mongodb-data-service');
```
Anemy, line 7: Question, is there a reason why we dynamically import the data service here and below?

Anemy (approved):
lgtm! Nice work, this would have helped me when I swapped over to the new m1 xD

Left a question and possible test add. No blockers.

packages/compass/src/main/import-export-connections.ts
addaleax: Not a good one probably :) So far we’re always importing data-service in the renderer, I didn’t want to add it and its dependencies to the main process startup sequence since 99 % of the time it’s not going to be used. Can also move this to a regular import if you prefer.

Anemy: Sounds good, I like the optimization, can we leave that comment you've just put as a comment there to help some future readers?

addaleax: Sure, done! :+1:

packages/data-service/src/import-export-connection.ts
```diff
@@ -0,0 +1,203 @@
...
+      'Cannot both specify to remove secrets and provide a passphrase for encrypting secrets'
+    );
+  }
+
+  const allConnections = await storage.loadAll();
+  let exportConnections: ConnectionInfoWithEncryptedData[] = cloneDeep(
+    allConnections.filter(filter)
+  );
+  log.info(
+    mongoLogId(1_001_000_151),
```
alenakhineika, line 56: From where do we get these numbers? Is there somewhere a complete list with descriptions?

Anemy: There's no complete list. The number is incremental so each log gets its own id. There's a lint script to ensure we don't use the same one twice.

alenakhineika: So we just manually search in code which number is the last and add +1?

addaleax: `node scripts/check-logids.js` :)

PR #3456, opened 2022-09-08
chore(indexes): Remove unneeded darkmode prop usage, update components to match designs

Anemy:
We don't need to explicitly color the `H3` component as it will inherit darkmode in the new leafygreen versions.
Also removed a few debugs I noticed in tests.

Mocks: https://www.figma.com/file/vbjoD4dcisPYRYJ5RRu72k/Compass-LG?node-id=2430%3A64950

Before:
Screen Shot 2022-09-08 at 12 55 23 PM

After:
Screen Shot 2022-09-08 at 12 52 46 PM

### DIFF
--- a/packages/compass-indexes/src/components/create-index-fields/create-index-fields.tsx
+++ b/packages/compass-indexes/src/components/create-index-fields/create-index-fields.tsx
@@ -12,6 +12,7 @@ import {
   css,
   Icon,
   uiColors,
+  withTheme,
 } from '@mongodb-js/compass-components';
 
 import type { IndexField } from '../../modules/create-index/fields';
@@ -270,4 +271,4 @@ class CreateIndexFields extends Component<CreateIndexFieldsProps> {
   }
 }
 
-export default CreateIndexFields;
+export default withTheme(CreateIndexFields);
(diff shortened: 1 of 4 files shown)


mabaasit (approved):

packages/compass-indexes/src/components/create-index-fields/create-index-fields.tsx
```diff
@@ -270,4 +271,4 @@ class CreateIndexFields extends Component<CreateIndexFieldsProps> {
   }
 }
 
-export default CreateIndexFields;
+export default withTheme(CreateIndexFields);
```
mcasimir, line 274: why is this needed here, aren't all the components used here already bound inside `compass-components`?

mcasimir (approved):

packages/compass-indexes/src/components/create-index-fields/create-index-fields.tsx
Anemy: It's needed currently to pass the `darkMode` to the `ComboBox` wrapper that's in this file. Thanks for bringing it up, I think it would make more sense to move this into `compass-components` along with the extra wrapper as this is the only place that uses that. It is a bit of refactoring to do that though. Would you prefer I do that here?
Previously this `withTheme` was a level higher up and was being passed to this component, so the changes in this pr are at least bringing us closer to there.

Anemy: Going to merge this, we can clean up the combobox usage a bit down the line, we also might be able to avoid it if we end up using the combobox directly from leafygreen instead of our tweaked port.

PR #3457, opened 2022-09-08
styles: remove spacing between sidebar and tabs COMPASS-6110

Anemy:
COMPASS-6110

Before:
Screen Shot 2022-09-08 at 12 13 10 PM

After:
Screen Shot 2022-09-08 at 12 32 23 PM

We should not merge this until we flip the sidebar feature flag (after we merge COMPASS-6091)


lerouxb:
I was about to do this myself :) I spoke to Claudia and the idea is to make the tab the same height as the title area. 64px, I think. (ie. without the favourite colour bar).

lerouxb:
Oh is that a separate ticket? In that case I don't mind. But I figured we can do it at the same time https://jira.mongodb.org/browse/COMPASS-6114

lerouxb (approved):

PR #3463, opened 2022-09-09
fix(testing): standardise on data-testid over data-test-id COMPASS-6116

lerouxb:
This is just a find/replace and then I fixed the cases where we happened to have a data-testid and a data-test-id with the same value.

I did some grepping magic to extract both sets and get the intersection and there aren't many possibilities that have to be fixed up.


Anemy (approved):
:shipit: 
I ran into this a bunch on the toolbars end to end test updates so thanks for doing this xD

addaleax (approved):

PR #3466, opened 2022-09-11
feat(query-bar): update query options to 4 rows, make filter a link COMPASS-6088

Anemy:
COMPASS-6088
Mocks: https://www.figma.com/file/vbjoD4dcisPYRYJ5RRu72k/Compass-LG?node-id=3325%3A104508

This PR updates the query bar to have 4 rows instead of 3. It removes the previous link at the bottom right and updates the filter label to be a link.
When updating this layout initially @gribnoysup and I added a grid that would layout the query options based on the amount passed in. This PR removes the grid we added and instead gives the consumer the ability to define where things will go with the layout prop, similar to the current, legacy query bar. While this generalized layout prop does increase the risk of improper use of the query bar, I think it simplifies the implementation and consumption of the component. As we work closely with the cloud team I don't foresee improper usage being an issue.

Screenshots of how the query bar looks with the various layouts the Data Explorer uses are below.
Note: Since taking these videos the inputs for the document editors are now aligned.

https://user-images.githubusercontent.com/1791149/189545455-e6e64850-d31a-49e1-a5fa-2018d38d2d07.mp4

`queryOptionsLayout` = `[]`
Screen Shot 2022-09-11 at 3 24 18 PM

`queryOptionsLayout` = `['project', 'sort']`
Screen Shot 2022-09-11 at 3 27 37 PM

 `queryOptionsLayout` = `['project', 'sort', 'collation]`
Screen Shot 2022-09-11 at 3 16 17 PM


packages/compass-query-bar/src/components/query-bar.tsx
```diff
@@ -138,7 +148,9 @@ const UnthemedQueryBar: React.FunctionComponent<QueryBarProps> = ({
             id="query-bar-option-input-filter-label"
             className={filterLabelStyles}
           >
-            Filter
+            <Link href={queryBarDocumentationLink} target="_blank">
```
Anemy, line 151: This puts a link inside a label, need to check on the implications of that, might be something we don't want to use `label` for (then we can hidden label the text area).

PR #3471, opened 2022-09-13
feat(compass): remove feature tour COMPASS-6099

Anemy:
COMPASS-6099

We'll be adding a different, new feature tour in the future with a better onboarding experience. In the interum are removing the current feature tour from the product and the codebase to avoid spending time on the conversion to Leafygreen, and because without converting it it be touring a different looking Compass.


addaleax (approved):
@alenakhineika

addaleax:
> *Is this a feat?

I’d say so, yes :+1:

packages/compass/src/app/index.js
```diff
@@ -184,34 +181,6 @@ var Application = View.extend({
...
-      this.renderSubview(tourView, this.queryByHook('tour-container'));
-    } else {
-      this.tourClosed();
-    }
-  },
-  tourClosed: function() {
-    app.preferences.unset('showFeatureTour');
-    app.preferences.save({}, {success: () => {
-      if (!app.preferences.showedNetworkOptIn && process.env.HADRON_ISOLATED !== 'true') {
-        ipc.ipcRenderer.emit('window:show-network-optin');
```
alenakhineika, line 212: I think we need to keep this part. It shows a settings window after closing the tour.

alenakhineika:
Glad this tour is gone! 🦅 I don't really like it in other applications and it was annoying in tests 😬

Anemy:
Waiting to hear back on designs thread before fixing tests and merging:
https://mongodb.slack.com/archives/GDTJXPHD0/p1663251656188339
Looks like some react bootstrap dependencies which are now unused have jquery dependencies so we'll remove those too.

mcasimir:
> have jquery dependencies so we'll remove those too.

🎻

packages/compass/src/app/index.js
Anemy: Re-added 👍

addaleax (approved):

PR #3475, opened 2022-09-14
feat(sidebar): show icons in the sidebar menus COMPASS-6081

lerouxb:
Screenshot 2022-09-14 at 13 24 57
Screenshot 2022-09-14 at 12 08 13
Screenshot 2022-09-14 at 12 08 04

The connections sidebar isn't using the same pattern as we use elsewhere, yet. But there's now a ticket for that: https://jira.mongodb.org/browse/COMPASS-6125

I also filed a ticket for adding icons to all the other menus: https://jira.mongodb.org/browse/COMPASS-6126


lerouxb:
These screenshots kinda make it obvious that we're inconsistent in how we capitalise text. See https://mongodb.slack.com/archives/G2L10JAV7/p1663158472156239

packages/compass-connections/src/components/connection-list/connection-menu.tsx
```diff
@@ -81,17 +79,19 @@ function ConnectionMenu({
...
           data-testid="copy-connection-string"
+          glyph={<Icon glyph="Copy" />}
           onClick={() => {
             copyConnectionString(connectionString).finally(() => {
               setMenuIsOpen(false);
             });
           }}
         >
-          Copy Connection String
+          Copy connection string
```
alenakhineika, line 89: So it should always be lowercase to not wrap?

lerouxb: It just happens to shave off a pixel or two which helps. But sentence-case is probably the way to go anyway. See the thread https://mongodb.slack.com/archives/G2L10JAV7/p1663158472156239

alenakhineika (approved):

PR #3478, opened 2022-09-15
fix(sidebar): remove old sidebar code COMPASS-6091

lerouxb:


mcasimir:
Why is this a fix and in the release notes? does it cause user-facing changes?

alenakhineika (approved):
🎊

PR #3486, opened 2022-09-16
feat: add support for CLI args/global config file COMPASS-6070 COMPASS-6071

addaleax:
##### feat: add support for CLI args/global config file COMPASS-6070 COMPASS-6071

The settings UI looks like this when settings have been fixed externally:

Screenshot from 2022-09-16 13-13-21

##### feat: define settings descriptions in preferences-model

Currently split out into a separate commit because while I feel this might be nice, I can also see how we might not want to do this. It keeps preference definition and description a bit more closer together, but at the same time makes our UI code a bit less flexible. However, it allows displaying the Compass `--help` text like this:

```
Available options:

  --theme (*)                                Compass UI Theme
  --networkTraffic (*)                       [Not implemented yet]
  --enableMaps (*)                           Enable Geographic Visualizations
  --trackErrors (*)                          Enable Crash Reports
  --enableFeedbackPanel (*)                  Give Product Feedback
  --trackUsageStatistics (*)                 Enable Usage Statistics
  --autoUpdates (*)                          Enable Automatic Updates
  --exportConnections                        Export Favorite Connections
  --importConnections                        Import Favorite Connections
  --passphrase                               Connection Export/Import Passphrase
  --help                                     Show Compass Options
  --version                                  Show Compass Version
  --ignoreAdditionalCommandLineFlags (*)     Allow Additional CLI Flags

(Options marked with (*) are also configurable through the global configuration file.)

Only available in the global configuration file:


The following global configuration file paths will be searched:

  /etc/mongodb-compass.conf

See the MongoDB Compass documentation for more details.
```

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-import-export/src/utils/formatters.spec.js
```diff
@@ -67,9 +67,9 @@ describe('formatters', function () {
 
       it('works for input with Binary data', async function () {
         const binary = new Binary(
-          Buffer.from('56391cc226bc4affbe520f67856c09ec'),
+          Buffer.from('56391cc226bc4affbe520f67856c09ec', 'hex'),
           4
-        );
+        ).toUUID();
```
addaleax, line 72: This was made necessary by the `bson` bump that came with adding the latest version of `bson` to the preferences package

packages/compass/src/main/index.ts
```diff
@@ -21,34 +25,62 @@ if (process.env.APP_ENV === 'webdriverio') {
...
+
+  if (globalPreferences.cli.help) {
+    process.stdout.write(getHelpText());
+    return app.exit(0);
+  }
+
+  if (preferenceParseErrors.length > 0) {
+    process.stderr.write(chalk.yellow(preferenceParseErrorsString) + '\n');
+    process.stderr.write('Use --ignore-additional-command-line-flags to allow passing additional options to Chromium/Electron\n');
+  }
```
gribnoysup, line 47: If we age getting strict with this maybe would make sense to allow at least some flags, like disable-gpu, that we recommended a few times as a workaround for visual glitches on macos and ubuntu crashes?

addaleax: Yeah, I’ve also gone back and forth a bit on this. There seems to be no exhaustive list of Electron and Chromium flags, let alone a way to programmatically access them. Do you know of other flags that we might want to allow here? `--disable-gpu` and `--no-sandbox` are the only obvious candidates for me right now.

gribnoysup: Yeah, this is super fair, I think all autogenerated lists for switches I saw were building it from chromium source, but I don't think it's really worth it for us. I think `disable-gpu` and `no-sandbox` are very good candidates and nothing else comes to mind, let's do them for now maybe and see if there is ever any need to expand when we release this?

addaleax: Sounds good! Done in 13113284239aa2ff6b51cf5df00ea57a7f3d5bfd :+1:

packages/compass-logging/src/index.ts
```diff
@@ -58,7 +61,22 @@ export function createLoggerAndTelemetry(component: string): {
...
+      // @ts-ignore-error Types from the dependency may not be available in early bootstrap.
+      preferencesIpc ??= (await import('compass-preferences-model'))
+        .preferencesIpc;
+    } catch {
+      preferencesIpc ??= {
+        getPreferences() {
+          return Promise.resolve({ trackUsageStatistics: true });
+        },
+      };
+    }
```
gribnoysup, line 78: A nit that I think might be more important later down the road when we are exploring how to make things like logger or preferences behave more like a injected services and so totally feel free to ignore this. I wonder how do you feel about already making this part a more explicit dependency injection mechanism, allowing to somehow set this interface for the logger from outside instead of having a direct dependency here on the `compass-preferences-model`. Maybe we can have some sort of `setIsTrackingEnabled(getIsEnabled: () => Promise<boolean>)` interface exported out of the compass-logging here?

mabaasit (approved):

packages/compass-logging/src/index.ts
addaleax: So … that sounds like a really nice thing to do, but I’m not sure how to go about it practically speaking. What would ensure that `setIsTrackingEnabled()` is called for all instances of `compass-logging`? We could manually call it in the main process and renderer process setup, that _should_ cover it, but I’m not sure that it’s good to spread the setup logic to a fairly independent place right now …

packages/compass-preferences-model/src/renderer-ipc.ts
```diff
@@ -29,7 +30,13 @@ export const makePreferencesIpc = (ipc: typeof hadronIpc) => ({
         'compass:get-configurable-user-preferences'
       );
     }
-    return Promise.resolve({} as UserPreferences);
+    return Promise.resolve({} as UserConfigurablePreferences);
+  },
+  getPreferenceStates(): Promise<PreferenceStateInformation> {
```
alenakhineika, line 35: What is getPreferenceStates used for?

packages/compass/src/main/logging.ts
```diff
@@ -15,6 +15,13 @@ import type { EventEmitter } from 'events';
 
 const debug = createDebug('mongodb-compass:main:logging');
 
+// Queue up events that happen before logging is initialized.
+const earlyLogEvents: any[] = [];
+const earlyLoggingListener = (ev: any) => earlyLogEvents.push(ev);
```
alenakhineika, line 20: can we use Event type here?

packages/compass-preferences-model/src/renderer-ipc.ts
addaleax: Mh yeah, I think all of these methods could use doc comments. Added in a0f5bc0baa07a54f5ca5cb14f57f15e904ecee77!

packages/compass-logging/src/index.ts
gribnoysup: On a module level our build setup should make sure that only one instance of `compass-logging` is being imported in the application, so using an interface to set something in the module scope once should work I think, but I also get your point, it's not the cleanest interface and it might be hard to find a good place for setting it, so then it's something that is better left for later.

I might be making a wrong assumption here, so please tell me if I'm saying something what was discussed before or just silly, but giving logging / telemetry code another look it seems like `CompassTelemetry` in `packages/compass` is currently a place where all the telemetry is being passed through and setup initially and might be a better place to depend on `compass-preferences-model` directly in general, especially now that main electron process holds the model and not the renderer. Also being the application entrypoint, the code in `packages/compass` always feels like the best place to wire up together pieces that should stay fairly independent for me, no DI required, what do you think?

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -309,19 +497,32 @@ class Preferences {
...
+   * Return the subset of preferences that can be edited through the UI.
+   * If this is the first call to this method, this sets the defaults for
+   * user preferences.
+   *
+   * @returns The currently active set of UI-modifiable preferences.
+   */
   async getConfigurableUserPreferences(): Promise<UserConfigurablePreferences> {
     // Set the defaults and also update showedNetworkOptIn flag.
-    if (!this.getPreferences().showedNetworkOptIn) {
+    if (!(await this.fetchPreferences()).showedNetworkOptIn) {
```
alenakhineika, line 525: 👀

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -341,6 +542,22 @@ class Preferences {
     ) as UserConfigurablePreferences;
   }
 
+  /**
+   * Report which preferences were set through external sources, i.e.
+   * command line or global configuration file.
+   *
+   * @returns A map of preference names to 'set-cli' or 'set-global' if the preference has been set in the respective source.
+   */
+  getPreferenceStates(): PreferenceStateInformation {
```
alenakhineika, line 551: Thx for commenting methods, very helpful!

alenakhineika (approved):

packages/compass-logging/src/index.ts
gribnoysup: As I mentioned I think for now we can totally even keep it as-is, especially as it seems we would want more time to think this through.

FWIW addressing the tight coupling of logging and preferences overweights relying on certain part of the application init code doing its job, I am certain we would have to change this to [make tracking pluggable](https://jira.mongodb.org/browse/COMPASS-5799) for cloud team for example. I think it's just ultimately unavoidable in an application like Compass, we already make a bunch of assumptions like that in the init code and we should rather test for this working as expected (which we already do), instead of trying to avoid it

addaleax: Yeah, let’s maybe chat about this in our 1:1 tomorrow (if you’re around for it :slightly_smiling_face:)

packages/compass-settings/src/components/settings/privacy.tsx
```diff
@@ -34,86 +44,43 @@ const checkboxStyles = css({
...
+      This setting cannot be modified as it has been set at Compass startup.
+    </Banner>
+  ),
+  'set-global': (
+    <Banner variant={BannerVariant.Info} data-testid="set-global-banner">
+      This setting cannot be modified as it has been set in the global Compass
+      configuration file.
+    </Banner>
+  ),
+  '': <></>,
```
gribnoysup, line 72: Just in case rendering fragments feels weird, null is a valid React child type

```suggestion
  '': null,
```

packages/compass-logging/src/index.ts
```diff
@@ -58,7 +61,22 @@ export function createLoggerAndTelemetry(component: string): {
...
   ): Promise<void> => {
-    const { trackUsageStatistics } = await preferencesIpc.getPreferences();
+    // Avoid circular dependency between compass-logging and compass-preferences-model
+    // Note that this is mainly a performance optimization, since the main process
+    // telemetry code also checks this preference value, so it is safe to fall back to 'true'.
+    try {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore-error Types from the dependency may not be available in early bootstrap.
+      preferencesIpc ??= (await import('compass-preferences-model'))
+        .preferencesIpc;
```
gribnoysup, line 71: Small q for my understanding: as compass-logging is something we should be able to use both in main process and renderer process, shouldn't this somehow fallback to reading settings from actual model? Seems like `preferencesIpc` will just always return empty preferences object when used outside of renderer, or the idea here that tracking in main should just always be disabled? 🤔

packages/compass-preferences-model/src/global-config.ts
```diff
@@ -0,0 +1,277 @@
...
+function getCliPropNamesByType(type: AmpersandType<any>): string[] {
+  return cliProps
+    .filter(([, definition]) => definition.type === type)
+    .map(([key]) => key);
+}
+
+const yargsOptions: YargsOptions = {
+  string: [...getCliPropNamesByType('string')],
+  boolean: [...getCliPropNamesByType('boolean')],
+  number: [...getCliPropNamesByType('number')],
```
gribnoysup, line 102: Super tiny nit, doesn't look like we need to spread here?

```suggestion
  string: getCliPropNamesByType('string'),
  boolean: getCliPropNamesByType('boolean'),
  number: getCliPropNamesByType('number'),
```

gribnoysup (approved):

packages/compass-logging/src/index.ts
addaleax: Ugh yes, good point. For now, I’ve simply defaulted this to `true`, but it *is* really annoying that there is no unified access to the preferences model that works both in the renderer and the main process, in the way that e.g. compass-logging works well in both. Don’t want to bloat this PR further since it’s not entirely trivial to get access to the main preferences instance, but I wrote down some notes in COMPASS-6149 as a TODO ticket here.

packages/compass-settings/src/components/settings/privacy.tsx
addaleax: Thanks, done in 57e7ff75dfaf2ba2268fd651f41eaef16d9ba858!

packages/compass-preferences-model/src/global-config.ts
addaleax: Thank you, yes :slightly_smiling_face: (Originally I had put some options like `--help` not as entries in the preferences description list, but only added them to the parser here, so it was something like `[...getCliPropNamesByType('boolean'), 'help'], then decided the special handling would introduce more complexity then it saves, removed them again, and forgot to remove the array spreading here :slightly_smiling_face:)

Removed in 57e7ff75dfaf2ba2268fd651f41eaef16d9ba858!

packages/compass-logging/src/index.ts
gribnoysup: Thanks for context, good to know we will be addressing this later!

One thing I still don't really follow is how this defaults to `true` (sorry, I am probably missing something here 🙇). AFAIU this import will not fail in node or electron main environment and the `preferencesIpc` fallback currently will return an empty object:

image

so all places where we check for `trackUsageStatistics` will evaluate to `false`, right?

addaleax: It explicitly defaults to `true` here: https://github.com/mongodb-js/compass/blob/57e7ff75dfaf2ba2268fd651f41eaef16d9ba858/packages/compass-logging/src/index.ts#L79

I think this is currently the only place where it’s potentially accessed outside of the renderer process.

PR #3493, opened 2022-09-19
feat: make mongodb-client-encryption optional VSCODE-342

alenakhineika:
## Description

Make the `mongodb-client-encryption optional` package optional in `data-service` so it can be used by VSCode OS-independent.

Also made this and and `os-dns-native` dependencies throw a warning instead of an error because the error looks annoying in logs.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/data-service/src/data-service.ts
+++ b/packages/data-service/src/data-service.ts
@@ -79,10 +79,12 @@ import type { CSFLECollectionTracker } from './csfle-collection-tracker';
 import { CSFLECollectionTrackerImpl } from './csfle-collection-tracker';
 
 import * as mongodb from 'mongodb';
+import type { ClientEncryption as ClientEncryptionType } from 'mongodb-client-encryption';
 
-// TODO: replace any with proper types when the Node bundles native binding distributions
+// TODO: remove try/catch and refactor encryption related types
+// when the Node bundles native binding distributions
 // https://jira.mongodb.org/browse/WRITING-10274
-let ClientEncryption: any;
+let ClientEncryption: typeof ClientEncryptionType;
 try {
   // eslint-disable-next-line @typescript-eslint/no-var-requires
   const { extension } = require('mongodb-client-encryption');
--- a/packages/data-service/src/data-service.ts
+++ b/packages/data-service/src/data-service.ts
@@ -79,18 +79,21 @@ import type { CSFLECollectionTracker } from './csfle-collection-tracker';
 import { CSFLECollectionTrackerImpl } from './csfle-collection-tracker';
 
 import * as mongodb from 'mongodb';
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore no types for 'extension' available
-import { extension } from 'mongodb-client-encryption';
-import type {
-  ClientEncryption as ClientEncryptionType,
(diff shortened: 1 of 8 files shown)


packages/data-service/src/data-service.ts
```diff
@@ -79,18 +79,21 @@ import type { CSFLECollectionTracker } from './csfle-collection-tracker';
...
-const ClientEncryption: typeof ClientEncryptionType =
-  extension(mongodb).ClientEncryption;
+
+// TODO: replace any with proper types when the Node bundles native binding distributions
+// https://jira.mongodb.org/browse/WRITING-10274
+let ClientEncryption: any;
+try {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const mongodbClientEncryption = require('mongodb-client-encryption');
+  const extension = mongodbClientEncryption.extension;
```
addaleax, line 89:
```suggestion
  const { extension } = require('mongodb-client-encryption');
```

packages/data-service/src/data-service.ts
```diff
@@ -2549,8 +2552,8 @@ export class DataServiceImpl extends EventEmitter implements DataService {
   }
 
   async createDataKey(
-    provider: ClientEncryptionDataKeyProvider,
-    options?: ClientEncryptionCreateDataKeyProviderOptions
+    provider: any,
+    options?: any
```
addaleax, line 2556: Let’s maybe still leave comments here about what the types are intended to be?

```suggestion
    provider: any /* ClientEncryptionDataKeyProvider */,
    options?: any /* ClientEncryptionCreateDataKeyProviderOptions */
```

packages/data-service/src/data-service.ts
```diff
@@ -2570,7 +2573,11 @@ export class DataServiceImpl extends EventEmitter implements DataService {
     return result;
   }
 
-  _getClientEncryption(): ClientEncryptionType {
+  _getClientEncryption(): any {
+    if (!ClientEncryption) {
+      return;
```
addaleax, line 2578: Can we throw an exception instead? That’s almost certainly always better than returning `undefined` and letting the caller crash.

packages/data-service/src/data-service.ts
```diff
@@ -79,18 +79,20 @@ import type { CSFLECollectionTracker } from './csfle-collection-tracker';
...
-  ClientEncryptionCreateDataKeyProviderOptions,
-} from 'mongodb-client-encryption';
-// mongodb-client-encryption only works properly in a packaged
-// environment with dependency injection
-const ClientEncryption: typeof ClientEncryptionType =
-  extension(mongodb).ClientEncryption;
+
+// TODO: replace any with proper types when the Node bundles native binding distributions
+// https://jira.mongodb.org/browse/WRITING-10274
+let ClientEncryption: any;
```
addaleax, line 85: This one can actually be properly typed, right?

```suggestion
import type { ClientEncryption as ClientEncryptionType } from 'mongodb-client-encryption';
let ClientEncryption: ClientEncryptionType;
```

PR #3498, opened 2022-09-20
chore(typography): Add Source Code Pro COMPASS-6096

lerouxb:

PR #3502, opened 2022-09-20
chore(typography): Add Source Code Pro COMPASS-6096

lerouxb:
Replaces https://github.com/mongodb-js/compass/pull/3498


gribnoysup:
Seems to work just fine (the new mono font is a bit more condensed compared to Menlo)

|Before|After|
|---|---|
|image|image|

Also `document.fonts.check('12px "Source Code Pro"')` returns `true`

gribnoysup (approved):

PR #3503, opened 2022-09-20
feat: implement HADRON_ISOLATED through preferences COMPASS-6065

addaleax:
- Implement `HADRON_ISOLATED` through the `networkTraffic` preference.
- Derive the value of other preferences that are shadowed by `networkTraffic` through a generic mechanism for doing so (we will also need this for `readOnly`/`enableShell`)
- Add an e2e test that runs on Linux and verifies via `strace` that Compass does not actually perform external network I/O when `--no-network-traffic` has been specified.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass/src/main/window-manager.ts
```diff
@@ -103,8 +103,13 @@ function showConnectWindow(
   };
 
   debug('creating new main window:', windowOpts);
+  const { networkTraffic } = compassApp.getPreferences().getPreferences();
 
   let window: BrowserWindow | null = new BrowserWindow(windowOpts);
+  if (networkTraffic !== true) {
+    // https://github.com/electron/electron/issues/22995
+    window.webContents.session.setSpellCheckerDictionaryDownloadURL('http://127.0.0.1:0/');
```
addaleax, line 111: Turned out not to be safebrowsing after all, “only” fetching an English dictionary from Google servers on first run :slightly_smiling_face:

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -503,16 +556,68 @@ class Preferences {
...
+      return (
+        descriptor.deriveValue as DeriveValueFunction<GlobalPreferences[K]>
+      )(
+        // `as unknown` to work around TS bug(?) https://twitter.com/addaleax/status/1572191664252551169
+        (k) =>
+          (k as unknown) === key ? originalValues[k] : deriveValue(k).value,
+        (k) =>
+          (k as unknown) === key ? originalStates[k] : deriveValue(k).state
+      );
+    }
```
addaleax, line 606: @alenakhineika I don’t think we specified in the tech design _how_ `networkTraffic` interacts with the individual preference values. I chose to use a generic mechanism for this rather than hardcoding the preference names and their interactions, also considering that for `enableShell` and `readOnly` we are going to have to do something similar again.

packages/compass-preferences-model/src/preferences.spec.ts
```diff
@@ -91,4 +91,80 @@ describe('Preferences class', function () {
...
+
+    const fetchResult = await preferences.fetchPreferences();
+    expect(fetchResult.autoUpdates).to.equal(false);
+    const saveResult = await preferences.savePreferences({ autoUpdates: true });
+    expect(saveResult.autoUpdates).to.equal(false); // (!)
+    expect(calls).to.deep.equal([{ autoUpdates: false }]); // (!)
+
+    const preferences2 = new Preferences(tmpdir);
+    const fetchResult2 = await preferences2.fetchPreferences();
+    expect(fetchResult2.autoUpdates).to.equal(true); // (!)
```
addaleax, line 139: Left `// (!)` comments here because the behavior might be surprising, I _think_ it’s the least surprising one of all different options though.

packages/compass/src/main/menu.ts
```diff
@@ -67,10 +67,10 @@ function quitItem(label: string): MenuItemConstructorOptions {
   };
 }
 
-function networkOptInDialogItem(): MenuItemConstructorOptions {
+function settingsDialogItem(): MenuItemConstructorOptions {
   return {
     label: '&Settings',
-    accelerator: 'Command+,',
+    accelerator: 'CmdOrCtrl+,',
```
addaleax, line 73: I also want to be able to use a shortcut on non-macOS platforms :relieved:

packages/compass/src/main/window-manager.ts
gribnoysup: Should we just make it a default for all versions? Do we get anything from having this enabled? 🤔

packages/compass/src/main/application.ts
```diff
@@ -70,15 +78,8 @@ class CompassApplication {
...
-    if (process.env.HADRON_ISOLATED !== 'true') {
-      // This is done asyncronously so that webpack can completely remove
-      // autoupdater from the application bundle during compilation
-      const { CompassAutoUpdateManager } = await import(
-        './auto-update-manager'
-      );
-      CompassAutoUpdateManager.init();
-    }
+  private static setupAutoUpdate(): void {
+    CompassAutoUpdateManager.init();
```
gribnoysup, line 82: Should we check for `networkTraffic` / `autoUpdates` here before enabling autoupdates in the application, or should a check be added in the `CompassAutoUpdateManager` itself? I think right now calling init here would just unconditionally enable autoupdates which we don't want in isolated as far as I can tell

packages/compass/src/app/setup-plugin-manager.js
```diff
@@ -44,9 +44,7 @@ marky.mark('Loading plugins');
 // eslint-disable-next-line no-nested-ternary
 const COMPASS_PLUGINS = process.env.HADRON_READONLY === 'true'
   ? require('./plugins/readonly')
-  : process.env.HADRON_ISOLATED === 'true'
-    ? require('./plugins/isolated')
```
gribnoysup, line 48: The side-effect of doing this would be that the plugin is mounted on the app start: https://github.com/mongodb-js/compass/blob/b659e8aeb048e2404ce277ca0f86a51d2dcdfefa/packages/compass/src/app/index.js#L171-L177

Should we add a check for the user preferences there now that we are removing a separate plugins list for isolated version?

packages/compass/src/main/application.ts
addaleax: I think what we’re aiming for here with the `HADRON_ISOLATED` and `HADRON_READONLY` is to load the necessary code always, and only alter runtime behavior depending on the settings values (and settings changes). In particular, the situation after this PR is the same as in the regular edition with the `autoUpdates` option turned off – and if auto updates still behave as if they were enabled in that case, that’s a bug either way.

We only call the underlying Electron auto-updater here: https://github.com/mongodb-js/compass/blob/6ba206d37b0a9a51372bbccede1dd66efdfbb595/packages/hadron-auto-update-manager/index.js#L136

if `AutoUpdateManager.enable()` has been called here: https://github.com/mongodb-js/compass/blob/6ba206d37b0a9a51372bbccede1dd66efdfbb595/packages/hadron-auto-update-manager/index.js#L183

which in turn is called only from here: https://github.com/mongodb-js/compass/blob/6ba206d37b0a9a51372bbccede1dd66efdfbb595/packages/compass/src/main/auto-update-manager.ts#L77-L78

and finally the call site of that event is guarded by taking in the actual value of the `autoUpdates` preference.

We _are_ going to simplify this call tree in COMPASS-6107.

packages/compass/src/app/setup-plugin-manager.js
addaleax: The plugin is mounted, but it never actually reaches the state in which its contents are displayed if `autoUpdates` is `false`, right? We don’t check for updates, so we don’t ever get the `app:update-available` event and consequently `isVisible` remains `false` in the plugin?

packages/compass/src/main/window-manager.ts
addaleax: I don’t know, it’s probably all hypothetical since we don’t really have human-readable text input fields, other than maybe the intercom surveys?

packages/compass/src/main/application.ts
gribnoysup: Got it, thanks for context! Without this it was hard to figure out whether we are changing the behavior or not. It's good to know that we are going to address the flow here, I kinda wish we did these in a different order so that reviewing this change would be easier 😅

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -89,25 +89,53 @@ export type AmpersandType<T> = T extends string
...
     ? false
     : K extends keyof CliOnlyPreferences
     ? false
     : boolean;
+  /** A description used for the --help text and the Settings UI */
   description: K extends keyof InternalUserPreferences
     ? null
     : { short: string; long?: string };
+  /** A method for deriving the current semantic value of this option, even if it differs from the stored value */
+  deriveValue?: DeriveValueFunction<GlobalPreferences[K]>;
```
gribnoysup, line 117: If we are allowing to have derived values, should we also account for this in the `onPreferenceValueChanged` method? It seems like right now I can be subscribed to changes of e.g., `trackErrors` value without getting an update if `networkTraffic` was changed

addaleax: Yes, great point! Done in 8a2e563cc71845710af445f7863ce06bc499f08e

packages/compass/src/app/setup-plugin-manager.js
gribnoysup: Yeah, that's true that it doesn't do anything, I was wondering if we want to completely preserve the old behavior where you wouldn't be able to find any references for the autoupdate in the application. But I guess this would also require us to check the `state` to know whether this is just dynamically disabled by the user through options or completely overriden by config and this sounds complex enough to just keep the plugin mounted even if it doesn't do anything

addaleax: Yeah – I’m okay with Isolated/Readonly being implemented that way. Especially Readonly is really just UIs being missing rather than actual backend code being disabled.

packages/compass/src/main/window-manager.ts
gribnoysup: Oh, another thing that comes to mind here, if we disable it conditionally, should we also listen to the option change here in case it was disabled in the application runtime, or download happens only once when the window is spawned and so it doesn't matter afterwards?

packages/compass-preferences-model/src/global-config.ts
```diff
@@ -260,7 +261,9 @@ export function getHelpText(): string {
     text += formatSingleOption(key, 'cli');
   }
   text +=
-    '\n(Options marked with (*) are also configurable through the global configuration file.)\n';
+    '\nOptions marked with (*) are also configurable through the global configuration file.\n';
+  text +=
+    'Boolean options can be negated by using a `--no` prefix, e.g. `--no-network-traffic`.\n';
```
gribnoysup, line 266: When I'm running the app locally and print help if shows switches in camelCase:

image

should this additional help match this and be `no-networkTraffic` to make it easier to connect the two?

packages/compass/src/main/window-manager.ts
addaleax: Yeah, this method is only a setter and does not trigger a download: https://github.com/electron/electron/blob/8c3c0f0b50634bd62c1a4c9a826d8c7aad329c11/patches/chromium/feat_add_support_for_overriding_the_base_spellchecker_download_url.patch#L29-L31

I don’t think there’s an explicit API to trigger a download.

gribnoysup (commented):
One thing caught my attention when trying out the feature locally: I see that in tech design `networkTraffic` is marked as available in settings UI, are you planning to add it there in the same PR or in a separate one? (I looked through the tickets, but couldn't find a separate ticket for something like that)

packages/compass-preferences-model/src/global-config.ts
gribnoysup: Ahh, now I get why we are listing them like that, thanks! Hmm, yeah, I do like the look of dashed arguments more too, maybe it's worth making the \* description more verbose so that we can have them listed in kebab case 😅

packages/compass/src/main/window-manager.ts
gribnoysup: Gotcha, thanks for source pointers. I kinda feel like we might just disable it for every page as this feels very unnecessary, but also don't have a strong preference here, so this looks good to me as-is, I'll resolve this one

addaleax:
@gribnoysup Just had a conversation with Claudia and Alena about that, we’re not planning to explicitly add a checkbox to the settings UI for `networkTraffic` at this point, thinking that being able to modify all the individual privacy options will give users what they need to achieve (almost) the same effect.

gribnoysup (approved):
Worth marking in the tech design maybe just so we have this decision documented? Otherwise looks good to me, that's an amazing feature we are building here!

One thing that does slightly bother me is that our preferences class is now re-implementing some of the ampersand models behavior, which I think is okay as a one-off thing here, but if we find ourselves doing this again, I would maybe give it another thought if we want to just use them for now instead of re-implementing them 🤔

Issue #3516, opened 2022-09-23
Increase "import document" editor focus size


nmoreaud:
When using "Add Data" > "Import document", the grey editor zone is quite large, but when clicking in it we cannot easily clear the example content.
Ex if we click in the green rectangle, and drag the mouse, the text is not selected.

![image](https://user-images.githubusercontent.com/2413781/191982265-703eb35f-9dce-4bae-a325-0d2b220fc681.png)

PR #3519, opened 2022-09-25
fix(crud): improve selection area for insert document editor

Anemy:
Fixes https://github.com/mongodb-js/compass/issues/3516
We'll be improving this modal and the status message it shows in the leafygreen work, figured I'd open this as it's quick and not connected to that work.

We could also add variant sizes to the editor, then the aggregations and other places could avoid these things.


lerouxb (approved):

alenakhineika (approved):
Tried locally, works smooth 👍

PR #3535, opened 2022-09-29
feat: Welcome modal COMPASS-6150

lerouxb:
* Renamed 'window:show-network-optin' to 'window:show-settings'.
* The welcome modal will only ever be included in the page (and shown) if we determine that it should be shown. The logic for it is the same as the settings modal before. ie. if it has not been shown and network traffic is enabled it will be shown.
* The settings modal can only be shown from  home. There we ensure that the defaults are set before it is shown. This meant I could factor out that code from getConfigurableUserPreferences.
* The settings modal doesn't listen for the ipc event directly anymore.

Screenshot 2022-09-29 at 16 02 43


packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -434,6 +434,7 @@ async function startCompass(opts: StartCompassOptions = {}): Promise<Compass> {
   process.env.DEBUG = `${process.env.DEBUG ?? ''},mongodb-compass:main:logging`;
   process.env.MONGODB_COMPASS_TEST_LOG_DIR = path.join(LOG_PATH, 'app');
   process.env.CHROME_LOG_FILE = chromedriverLogPath;
+  process.env.SHOW_WELCOME = 'false';
```
addaleax, line 437: If we do this, can we pick a Compass-specific environment variable name?

Can we maybe test the actual flow here instead, where we just close the welcome modal if it appears at startup?

packages/compass-welcome/src/typings.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.png' {
```
addaleax, line 1: Are we adding a .png somewhere here? (If so, I'm not sure where)

lerouxb: I was going to. Just forgot to clean that up.

packages/compass-e2e-tests/helpers/compass.ts
lerouxb: We'd have to wait to see if it appears which means we have to timeout. In the past 5 or 10 seconds wasn't enough because in CI there are random pauses and generally no real consistency in how long it takes to run tests. Which meant every test where we open compass gained 10+ seconds of waiting which quickly added up and was really painful when working on.

Plan is to have just one test that tests the actual flow - in time to first query. Working on it.

lerouxb: Another problem with it is that it changes test behaviour / timing based on the order of tests. We either have to clear everything always (and then how do we test things like favourites being stored?) and then always have the first run experience on every test and not be able to test the usual flow (important for things like time to first query) or you have things like the order matters so suddenly you focus a test or file and you get the welcome modal where you normally never do.

Being able to turn it on/off was the best I could come up with when we had the tour.

packages/compass-e2e-tests/tests/time-to-first-query.test.ts
```diff
@@ -59,7 +60,12 @@ describe('Time to first query', function () {
 
   it('can open compass, connect to a database and run a query on a collection (second run onwards)', async function () {
     // start compass inside the test so that the time is measured together
-    compass = await beforeTests();
+
+    // TODO: this should work with firstRun left as undefined, but even though the
+    // preferences gets saved to disk, it gets the old value when opening
+    // compass again. The only way to get the saved value in a new compass seems
+    // to be if the whole process is relaunched.
+    compass = await beforeTests({ firstRun: false });
```
lerouxb, line 68: I feel like I'm going crazy. The idea with these tests is that the first one makes a new user directory, so it is guaranteed to be the first run and therefore you see the welcome modal. That works as planned.

Then the second test (this one) would reuse the same user directory. ie, it should see the saved settings and therefore the app should know that the user already opted in. But even though the preferences got saved and the file on disk says that the user opted in, the welcome modal pops up. 

I've logged `console.log('shouldShowWelcomeModal', {showedNetworkOptIn, networkTraffic});` in shouldShowWelcomeModal and you can see that showedNetworkOptIn is false. But if you look at the file in the user dir it is true.

My last vague hunch left is that it is somehow cached in the test process? Seems crazy. But I get something similar when testing compass manually: remove the preferences file and it doesn't take effect until you kill the process and start over.

Most likely I'm just missing something dumb, though.

lerouxb: I can work around it by always passing --showed-network-opt-in, but would love to understand what's going on.

packages/compass/src/app/index.js
```diff
@@ -71,6 +71,14 @@ const { log, mongoLogId, debug, track } =
     'COMPASS-APP'
   );
 
+function shouldShowWelcomeModal(showedNetworkOptIn, networkTraffic) {
+  if (!showedNetworkOptIn && networkTraffic) {
```
Anemy, line 75: Do we check for `networkTraffic` here so that we don't show the welcome modal in isolated? How do you feel about still showing the welcome modal but changing the wording around network opt/telemetry in for that distribution?

Anemy (approved):
lgtm! nice. Left one question and one suggestion for possible test case improvements. 
Tried it manually by editing the `Preferences.json` in `/Users/rhys/Library/Application Support/Electron/AppPreferences` and it worked smooth.
Totally not a big thing, but I also feel this would probably be a bit simpler as a modal inside of `compass-home` as it is the only place it's used. I missed that slack discussion as I was out that week xD

packages/compass/src/app/index.js
lerouxb: Hmm. I just kept what we had without thinking much. We probably don't even have to change the wording. I can open a PR to remove that and bring it up in design.

PR #3538, opened 2022-09-30
feat: align backgrounds and uniform spacings COMPASS-6148

mcasimir:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-collection/src/components/collection/collection.tsx
+++ b/packages/compass-collection/src/components/collection/collection.tsx
@@ -2,13 +2,7 @@ import type AppRegistry from 'hadron-app-registry';
 import { createLoggerAndTelemetry } from '@mongodb-js/compass-logging';
 import type { Document } from 'mongodb';
 import React, { useCallback, useEffect } from 'react';
-import {
-  TabNavBar,
-  css,
-  withTheme,
-  uiColors,
-  cx,
-} from '@mongodb-js/compass-components';
+import { TabNavBar, css } from '@mongodb-js/compass-components';
 
 import CollectionHeader from '../collection-header';
 import type { CollectionStatsObject } from '../../modules/stats';
@@ -71,7 +65,6 @@ type CollectionProps = {
 };
 
 const Collection: React.FunctionComponent<CollectionProps> = ({
-  darkMode,
   namespace,
   isReadonly,
   isTimeSeries,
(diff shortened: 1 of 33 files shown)


packages/compass-schema/src/components/compass-schema/compass-schema.module.less
```diff
@@ -98,8 +98,8 @@
 
       .schema-field {
         // second level
-        border-bottom: 3px solid @gray8;
-        border-left: 24px solid @gray8;
+        margin-bottom: 4px;
+        // border-left: 24px solid @gray8;
```
Anemy, line 102:
```suggestion
```

packages/compass-schema/src/components/compass-schema/compass-schema.module.less
```diff
@@ -116,11 +116,11 @@
         }
 
         &:first-child {
-          border-top: 3px solid @gray8;
+          // border-top: 3px solid @gray8;
```
Anemy, line 119:
```suggestion
```

Anemy (approved):
lgtm! Once tests and checks pass. Left two small comments on commented code to remove.

Also before merging, do we want to update the backgrounds of the `InteractivePopover` to white, and the aggregations settings background color?

Pipeline settings:


Popover before/after:
Screen Shot 2022-10-03 at 4 10 17 PM
Screen Shot 2022-10-03 at 4 11 04 PM

packages/compass-collection/src/components/collection/collection.tsx
```diff
@@ -160,6 +146,7 @@ const Collection: React.FunctionComponent<CollectionProps> = ({
           pipeline={pipeline}
           sourceName={sourceName}
           stats={stats}
+          darkMode={darkMode}
```
Anemy, line 149: As this `Collection` component is no longer the user of the `darkMode` prop, it's passed down, it's probably a bit cleaner to remove the property from this component and the `withTheme` wrapper and instead have the `withTheme` wrapper inside of `CollectionHeader.tsx`

mcasimir:
> Also before merging, do we want to update the backgrounds of the InteractivePopover to white, and the aggregations settings background color?

That's a good point, actually doesn't look bad in white at all .. maybe with some adjustments to the components there .. would definitely need some more design input though. We can do that when we convert those components to LG

PR #3543, opened 2022-10-03
chore: remove unused mms-icon less font COMPASS-5982

Anemy:
COMPASS-5982

The last use was in the sidebar that we recently updated.


mcasimir (approved):
Nice! there may be files here: `@{compass-fonts-path}/MMSIcons-*` to be deleted, and some references in [fonts.less](https://github.com/mongodb-js/compass/blob/COMPASS-5982-remove-mms-icon-font/packages/compass/src/app/styles/fonts.less#L138), is that intentional?

Anemy:
Ah thanks for catching @mcasimir - removed.

PR #3545, opened 2022-10-05
chore(compass-collection): Add more types to tab reducers

Anemy:
To prep for possible bug fixes for [COMPASS-6146](https://jira.mongodb.org/browse/COMPASS-6146) I wanted to introduce more types to reduce chances of introducing more bugs.

Created COMPASS-6162 to add more types down the line (will take some webpack/ts config work though).

Removed usage of `semver` in `compass-collection` - it was used for checking the `minimumServerVersion` in plugin roles, which currently no plugins use. I'll leave a comment to highlight this one.

### DIFF
--- a/packages/compass-collection/src/stores/context.tsx
+++ b/packages/compass-collection/src/stores/context.tsx
@@ -52,7 +52,7 @@ const setupActions = (role: any, localAppRegistry: AppRegistry) => {
   return actions;
 };
 
-type ContextProps = {
+export type ContextProps = {
   tabs?: string[];
   views?: JSX.Element[];
   role?: any;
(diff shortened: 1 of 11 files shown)


packages/compass-collection/src/stores/context.tsx
```diff
@@ -360,19 +371,7 @@ const createContext = ({
...
-  // Filter roles for feature support in the server.
-  const filteredRoles = roles.filter((role: any) => {
-    if (
-      ['Indexes', 'Validation', 'Explain Plan'].includes(role.name) &&
-      isDataLake
-    ) {
-      return true;
-    }
-    if (!role.minimumServerVersion) return true;
-    return semver.gte(serverVersion, role.minimumServerVersion);
```
Anemy, line 372: Here's where we were previously using `semver` which we don't need anymore. No plugins were using `minimumServerVersion` in their role definition.

lerouxb (approved):

PR #3554, opened 2022-10-05
chore: replace named colors and rgb with hex colors COMPASS-6163

mcasimir:


packages/compass-explain-plan/src/components/explain-json/explain-json.module.less
```diff
@@ -6,10 +6,10 @@
   .document-list {
     padding: 0;
     position: relative;
-    background: white;
-    border: 1px solid #ddd;
+    background: #ffffff;
+    border: 1px solid #dddddd;
     border-radius: 4px;
-    box-shadow: 0 1px 1px rgb(0 0 0 / 5%);
```
Anemy, line 12: I never know you could do something like this, I don't think I would though.

Anemy (approved):
nice

packages/compass-explain-plan/src/components/explain-json/explain-json.module.less
mcasimir: yeah ... that was new for me too! that's when I discovered that VScode can actually convert between any color format by clicking on the color box, saved me 1 hour of copy/paste

Screenshot 2022-10-06 at 10 02 20

PR #3557, opened 2022-10-06
feat: align all the colors with rebranded colors COMPASS-6121

mcasimir:
- replaces all the colors with leafygreen colors
- replaces hardcoded colors in .less with variables

Follow up:
- replace `palette.less` with an import from leafygreen
- replace `uiColors` with `palette` and replace hardcoded colors inside jsx / tsx with palette

Screenshot 2022-10-07 at 15 14 31
Screenshot 2022-10-07 at 15 14 00


packages/compass/src/app/styles/palette.less
```diff
@@ -1,151 +1,46 @@
...
-@alertRed: #ef4c4c;
-@alertRedBg: #feeded;
-@alertRedText: #A94442;
-@alertBlue: #43b1e5;
-
-/* Semantic Colors */
-
-@inactiveIcon: @gray4;
-
+@palette__white: #ffffff;
```
mcasimir, line 1: i'll move this to an import from LG in a separate PR, importing here directly from `@leafygreen-ui` would trigger the check since we are outside of compass-components

PR #3563, opened 2022-10-07
chore: migrate to leafygreen palette and use color vars everywhere 

mcasimir:
- remove compassUiColors
- replace uiColors with palette
- use color vars everywhere


Anemy (approved):
lgtm!

Anemy (approved):
lgtm, tried it out, looks good. I do think usage of the `fade` function in less files might introduce a slight bit more complexity for when folks move things to css in js, definitely small though so all good.

mcasimir:
> lgtm, tried it out, looks good. Nice work, lots of styles involved! I do think usage of the `fade` function in less files might introduce a slight bit more complexity for when folks move things to css in js, definitely small though so all good.

That's a good point, fun stuff we are using `transparentize` for `rgbas` in css-in-js which has the opposite "fraction", like `rgba(..., 0.6)` would translate to `transparentize(0.4, ...)`, which would be an headache for whoever ports that to css-in-js, but i feel the issue is more with `transparentize` itself than `fade`.  `polished` has a better function that translates 1-1 and I have no idea why we and LG don't use that instead. Is called `rgba` and `fade(@palette--black, 60%)` would convert to  `rgba(palette.black, .6)`. I'll add a point to my follow up notes, would take a minute to switch to that.

PR #3573, opened 2022-10-11
chore: move aggregation settings and replace buttons COMPASS-6184

mcasimir:
Screenshot 2022-10-11 at 17 05 46

### DIFF
--- a/packages/compass-aggregations/src/components/settings/settings.jsx
+++ b/packages/compass-aggregations/src/components/settings/settings.jsx
@@ -1,9 +1,8 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import { Label, Description } from '@mongodb-js/compass-components';
+import { Label, Description, Button } from '@mongodb-js/compass-components';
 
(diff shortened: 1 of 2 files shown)


packages/compass-aggregations/src/components/settings/settings.jsx
```diff
@@ -154,19 +153,8 @@ class Settings extends PureComponent {
...
-            />
-
-            <TextButton
-              id="aggregation-settings-apply"
-              className="btn btn-primary btn-xs"
-              text="Apply"
-              clickHandler={this.onApplyClicked.bind(this)}
-            />
+            <Button id="aggregations-settings-cancel" size="xsmall" onClick={this.onCancelClicked.bind(this)}>Cancel</Button>
+            <Button id="aggregation-settings-apply" size="xsmall" variant="primary" onClick={this.onApplyClicked.bind(this)}>Apply</Button>
```
Anemy, line 157: If we want we could update the `id` usage to `data-testid` and update the selector in the end to end tests and the `settings.spec.jsx` test.

Anemy (approved):
lgtm! feel free to do nothing on the comment, is super minor and probably not worth time

PR #3574, opened 2022-10-11
fix(crud): menu not fully showing for field actions, remove old backgrounds COMPASS-6186

Anemy:
COMPASS-6186

Will remove the release notes label before merging as this isn't a released bug. 
This PR fixes the menus not showing in the crud view. We're fixing it by removing the `overflow: none` in the `KeylineCard` component. I think this is better than changing the menu `usePortal={true}` for now. Currently when an element that a portal is referring to is removed from the dom, the menu teleports to the top left while it collapses. We could work around this, but I think `usePortal={false}` is a cleaner approach.
Did some drive by cleanup, removing old backgrounds.
Also added another 16px of margin to the bottom of the document list for a little better ux with the menus.

Because of these changes the previous box shadow on document cards being edited is now shown (it was hidden by the hidden overflow before). Do we want to remove this shadow? cc @mcasimir 

| Before      | After |
| ----------- | ----------- |
| Screen Shot 2022-10-11 at 2 34 39 PM |  Screen Shot 2022-10-11 at 2 35 16 PM |


mcasimir:
Ditching the overflow hidden and the background makes sense to me, good call! Let's also remove the shadow, if we want them back we can re-add them later together with design.

packages/compass-crud/src/components/document-json-view.less
```diff
@@ -11,12 +11,11 @@
     margin-bottom: 5px;
 
     .json-document-is-editing {
-      box-shadow: 2px 5px 8px fade(@palette__black, 20%);
+      border-radius: inherit;
```
mcasimir, line 14: nice trick, didn't know that we could have done this!

mcasimir (approved):

PR #3576, opened 2022-10-11
chore(hadron-react-components): remove unused ModalCheckbox component

Anemy:


mcasimir (approved):
What is the advantage of removing things from `hadron-react-components`? isn't it better to replace all the components where they are used and then just drop this package?

Anemy:
> What is the advantage of removing things from hadron-react-components? isn't it better to replace all the components where they are used and then just drop this package?

@mcasimir  I think removing them gives a better sense of when we can drop the package entirely, it isn't necessary, but at least it gives a sense of how much we have left. I think it's fine either way though so if you have a strong preference we can stick to it.

PR #3577, opened 2022-10-11
feat(crud, hadron-react-components): update error styles in crud, remove unused status row component

Anemy:
Screen Shot 2022-10-11 at 5 03 13 PM
Screen Shot 2022-10-11 at 5 04 02 PM


mcasimir (approved):
Nice, i've always wondered what was that status-row thing

PR #3581, opened 2022-10-12
chore: fix transparent dropdown scrollbar

mcasimir:
Fixes this: 

Screenshot 2022-10-12 at 09 46 53

Overrides:

``` css
::-webkit-scrollbar-track {
  background-color: none;
  &:hover {
    background-color: fade(@palette__black, 16%);
  }
}
```

Question: how is this not causing issues also in other places?


Anemy (approved):
> Question: how is this not causing issues also in other places?

Not sure, this is the only place where these dropdown styles are used. Might be something in the z indexes or background in the dropdown styles that cause this behavior.

mcasimir:
> > Question: how is this not causing issues also in other places?
> 
> Not sure, this is the only place where these dropdown styles are used. Might be something in the z indexes or background in the dropdown styles that cause this behavior.

Ahh .. I think that is the dropdown background that is transparent now

PR #3582, opened 2022-10-12
chore: FormModal, ModalHeader, ModalContent COMPASS-5522

lerouxb:
**The following are now using FormModal:**

* create-view-modal
* duplicate-view-modal
* import-pipeline
* saving-pipeline-modal
* insert-document-dialog
* drop-index-modal
* edit-item-modal
* open-item-modal
* save-connection-modal
* create-collection-modal
* create-database-modal
* drop-collection-modal
* drop-database-modal

Most of them were using ConfirmationModal before, but some were just using Modal.

For FormModal there's now a ModalHeader (knows about required title and optional subtitle) and ModalBody (knows about max-height and overflow scrolling).

FormModal and ModalBody can also take scroll={false} for the rare case where a select has to be able to hang out of the content. See open-item-modal for an example. Or minScrollHeight in FormModal (or minHeight in ModalBody) for when you need to reserve space but you still want scrolling (see insert-document-dialog).

**The following are still using ConfirmationModal:**

* confirm-import-pipeline
* confirm-new-pipeline
* saved-pipeline-card
* delete-item-modal
* confirm-edit-connection-string

That's because none of them contain forms.

**The following are now using InfoModal:**

* pipeline-explain
* create-index-modal
* shell info-modal
* connection-info-modal
* csfle-connection-modal
* non-geniune-warning-modal

**create-index-modal is left alone for now because it is weird**

InfoModal has a showCloseButton boolean, defaulting to true. This is because pipeline explain and a bunch of others had close buttons for whatever reason. Arguably the X close button top-right is good enough as it does the same thing, but this way we can leave that as a design decision for another time.

**The following are left for a future task as part of milestone 4b:**

* export-modal (export to language)
* export-modal (import export)
* import-modal

**The following are left unchanged:**

* settings modal
* connecting modal

Only change to settings modal is that I got rid of ModalTitle and hardcoded it there since it was the only thing left using it. Everything else uses ModalHeader (extracted from ConfirmationModal) to make the header via FormModal or InfoModal.

I just left connecting modal alone for now. It doesn't seem to match any of the other patterns.

---

**Things to look out for:**

* We still have lots of different form element components / libraries / patterns / CSS so the spacing inside modals is still not consistent. Maybe something for later in this epic? (see incoming screenshots)


packages/compass-components/src/components/form-modal.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { ModalHeader } from './modal-header';
+
+export const Variant = {
+  Default: ButtonVariant.Primary,
+  Danger: ButtonVariant.Danger,
+} as const;
+
+export type Variant = typeof Variant[keyof typeof Variant];
+
+const baseModalStyle = css`
```
lerouxb, line 18: Lots of CSS throughout that I copy/pasted from ConfirmationModal and will hopefully be able to clean up and get up to our standards.

lerouxb:
I did a quick audit of our e2e test coverage of these modals:

covered:

create-view-modal
import-pipeline-modal
saving-pipeline-modal
insert-document-dialog
drop-index-modal
save-connection-modal
create-connection-modal
create-database-modal
drop-collection-modal
drop-database-modal
pipeline-explain-modal
create-index-modal
connection-info-modal
settings-modal

not covered:

duplicate-view-modal
edit-item-modal
open-item-modal
confirm-import-pipeline
confirm-new-pipeline
restore-pipeline-modal (in saved-pipeline-card)
delete-item-modal
confirm-edit-connection-string
shell-info-modal
csfle-connection-modal
non-genuine-warning-modal

Might be worth at least just opening these modals in e2e tests which should be quick  to do. Then we can leverage the e2e tests for taking screenshots which can automate the tedious before/after comparison process.

packages/compass-crud/src/components/insert-document-dialog.jsx
```diff
@@ -212,15 +212,17 @@ class InsertDocumentDialog extends React.PureComponent {
...
         open={this.props.isOpen}
-        onConfirm={this.handleInsert.bind(this)}
+        onSubmit={this.handleInsert.bind(this)}
         onCancel={this.props.closeInsertDocumentDialog}
-        buttonText="Insert"
+        submitButtonText="Insert"
         submitDisabled={this.hasErrors()}
         trackingId="insert_document_modal"
+        data-testid="insert-document-modal"
+        scrollClassName={css({ minHeight: spacing[6] * 2 })} // make sure there is enough space for the menu
```
lerouxb, line 225: We could have a default minHeight for ModalContent, but then the ones that are designed to look like ConfirmationModal look weird.

But then without minHeight or overflow: 'visible' if you have a menu or similar popover thing that doesn't fit in the available area it will get cut off. Maybe there's a third option I'm missing?

packages/compass-saved-aggregations-queries/src/components/open-item-modal.tsx
```diff
@@ -135,13 +135,14 @@ const OpenItemModal: React.FunctionComponent<OpenItemModalProps> = ({
...
+    <FormModal
       open={isModalOpen}
       onCancel={onClose}
-      onConfirm={onSubmit}
+      onSubmit={onSubmit}
       title="Select a Namespace"
-      buttonText="Open"
+      submitButtonText="Open"
       submitDisabled={isSubmitDisabled}
+      scrollClassName={css({ overflow: 'visible' })} // this is so that the selects can hang over the footer and out of the modal
```
lerouxb, line 145: Here the modal is very short (like ConfirmationModal) but contains a Select which should be able to hang out. So we probably don't want the maxHeight and overflow: auto/scroll behaviour.

lerouxb:
add-collection-modal-basic
add-collection-modal-capped
add-collection-modal-clustered
add-collection-modal-custom-collation
add-collection-modal-timeseries
add-database-modal-basic
aggregation-explain-modal
confirm-import-pipeline-modal
confirm-new-pipeline-modal
connection-info-modal
create-index-modal-basic
create-index-modal-wildcard
create-view-modal
drop-collection-modal
drop-database-modal
drop-index-modal
export-modal
export-to-language-modal
new-pipeline-from-text-modal
save-favorite-modal-new
save-favorite-modal
save-pipeline-modal
welcome-modal

packages/compass-components/src/components/modal.tsx
```diff
@@ -1,21 +1,48 @@
...
+import { css, cx } from '@leafygreen-ui/emotion';
+import { spacing } from '@leafygreen-ui/tokens';
 import { createLoggerAndTelemetry } from '@mongodb-js/compass-logging';
 const { track } = createLoggerAndTelemetry('COMPASS-UI');
 
 import { Modal as LeafyGreenModal } from './leafygreen';
 
+const variantStyles = {
+  'with-footer': css({
+    width: '600px',
```
mcasimir, line 11: width, padding and letterSpacing are the same, maybe just `const baseStyles = {padding: 0, letterSpacing ...}` and then `{...baseStyles, paddingBottom}`?

Do we need these variants? we should have alternatives (playing around with removing the padding top from the footer for example), but the most direct question is, isn't this needed only for Info modals? we could put that only there and avoid variants for the content.

packages/compass-e2e-tests/helpers/commands/add-database.ts
```diff
@@ -1,10 +1,17 @@
 import type { CompassBrowser } from '../compass-browser';
 import * as Selectors from '../selectors';
 
+type AddDatabaseOptions = {
+  // Intentionally empty. Just here as placeholder to make it the same as addCollection
```
mcasimir, line 5: what is the advantage of this?

lerouxb: I have to call the method with parameters after options.

packages/compass-components/src/components/modal.tsx
lerouxb: I was trying to avoid having another modal class just for one line of css.

packages/compass-components/src/components/form-modal.tsx
```diff
@@ -0,0 +1,76 @@
...
+export type Variant = typeof Variant[keyof typeof Variant];
+
+type FormModalProps = React.ComponentProps<typeof Modal> & {
+  variant?: Variant;
+  title: string;
+  subtitle?: string;
+  submitButtonText: string;
+  cancelButtonText?: string;
+  submitDisabled?: boolean;
+  scrollClassName?: string;
```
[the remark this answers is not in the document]
mcasimir: Ah damn, that's even more confusing. If there is an ambiguity and content refers to 2 different things, what about renaming `ModalContent` to `ModalBody` and clear the confusion completely?

packages/compass-e2e-tests/helpers/commands/add-database.ts
mcasimir: i'm sorry .. i still don't get it, what's wrong with:
 
```
  await browser.addDatabase(
      dbName,
      collectionName,
      'add-database-modal-basic.png'
    );
 ```
 
 btw super non-important, i'm just curious

File: packages/compass-components/src/components/form-modal.tsx (hunk 6 above)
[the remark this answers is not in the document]
Comment (lerouxb) [reply packages/compass-components/src/components/form-modal.tsx:24]: Yeah I'm strongly considering that :)

File: packages/compass-indexes/src/components/create-index-actions/create-index-actions.tsx [hunk 7]
```diff
@@ -1,5 +1,10 @@
 import React from 'react';
-import { css, Banner, spacing, Button } from '@mongodb-js/compass-components';
+import {
+  css,
+  Banner,
+  spacing,
+  ModalFooterButton,
```
mcasimir, line 6: why do we need this?

lerouxb: The spacing of ConfirmationModal (and now FormModal) buttons is different than what we have in create index modal.

packages/compass-components/src/components/modal-footer-button.tsx
```diff
@@ -0,0 +1,35 @@
...
+      margin: '0 5px 0 0',
+    },
+  }),
+  [Theme.Dark]: css({
+    margin: '0 2px',
+    '&:first-of-type': {
+      margin: '0 0 0 4px',
+    },
+    '&:last-of-type': {
+      margin: '0 4px 0 0',
```
mcasimir, line 23: does the margin changing with the theme here?

lerouxb: Yes that's copied from confirmation modal. It is all their CSS.

packages/compass-indexes/src/components/create-index-actions/create-index-actions.tsx
mcasimir: I see, is it only limited to this dialog? Since this is a special case I'd keep it styled "manually" here, and we can tackle this in a follow-up PR rather than introducing this concept right away, feels like something that can be better thought through independently.

packages/compass-components/src/components/modals/modal-footer-button.tsx
```diff
@@ -0,0 +1,35 @@
...
+    },
+    '&:last-of-type': {
+      margin: '0 4px 0 0',
+    },
+  }),
+};
+
+function UnthemedModalFooterButton({ className, ...props }: ButtonProps) {
+  const { theme } = useTheme();
+  return <Button {...props} className={css(className, buttonStyle[theme])} />;
```
gribnoysup, line 30:
```suggestion
  return <Button {...props} className={cx(className, buttonStyle[theme])} />;
```

packages/compass-components/src/components/modals/modal-header.tsx
```diff
@@ -0,0 +1,93 @@
...
+        </div>
+      )}
+
+      <h1
+        className={cx(titleStyle, {
+          [css({
+            fontWeight: 'bold',
+            lineHeight: '25px',
+            color: uiColors.gray.light2,
+          })]: darkMode,
```
gribnoysup, line 81: I would suggest moving this outside the component as the css call result here never changes

lerouxb: Hmm I should have noticed that. Just copy/pasted from leafygreen somewhere.

gribnoysup (approved):
Disclaimer that I didn't manage to go though all the changed modals in the running app 😅  but I think the code looks good and the tests are passing so, awesome work updating all this stuff!

lerouxb:
Good that I checked because a bug did slip in at the last minute 😬

PR #3587, opened 2022-10-13
chore: use rgba instead of transparentize

mcasimir:
Tiny quality of life improvement: replaces `transparentize` with `rgba` from polished. `transparentize` uses an inverted form for the opacity compared to `rgba` (1 - `opacity`) and is a bit confusing when we need to convert from figma or from less ([fade](https://lesscss.org/functions/#color-operations-fade)).

`fade(@color1, 30%)` --> `rgba(color1, 0.3)` --> `transparentize(0.7, color1)`


Anemy (approved):
👍

PR #3588, opened 2022-10-13
feat: add theme as regular setting COMPASS-6067 COMPASS-5284

addaleax:
- Modify and read `theme` as a regular preference value, and use it to set the Electron theme settings instead of using custom listeners and IPC ping-pong for changing its value.
- Add the leafygreen darkmode flag to the preferences list.
- Add an entry to the settings modal for changing the selected theme.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-components/src/components/file-input.tsx
```diff
@@ -194,7 +195,7 @@ function FileInput({
     );
   };
 
-  const applyTheme = global?.process?.env?.COMPASS_LG_DARKMODE === 'true';
+  const applyTheme = preferences.getPreferences().lgDarkmode;
```
addaleax, line 198: Maurizio and I had a conversation recently about the idea of moving feature flags to also be communicated as preferences recently, lmk if y'all have any thoughts on it.

packages/compass-home/package.json
```diff
@@ -81,6 +81,9 @@
     "sinon": "^8.1.1",
     "xvfb-maybe": "^0.2.1"
   },
+  "optionalPeerDependencies": {
+    "@electron/remote": "^2.0.8"
+  },
```
addaleax, line 86: Our depcheck scripts fail if this is added to `dependencies` and/or `peerDependencies`, unfortunately.

packages/compass-settings/src/components/settings/theme.tsx
```diff
@@ -0,0 +1,111 @@
...
+        >
+          Light Theme
+        </RadioBox>
+        <RadioBox
+          id="theme-selector-dark"
+          data-testid="theme-selector-dark"
+          value="DARK"
+          disabled={!!preferenceStates.theme || themeValue === 'OS_THEME'}
+        >
+          Dark Theme
```
addaleax, line 95: This should ideally look more like https://www.figma.com/file/bSnDxjrKFf9EROX7X6FJoB/Compass-Nav-tree%2BSettings?node-id=25%3A3098, my Figma skills just aren't good enough to extract the relevant parts as SVG.

packages/compass/src/app/theme.ts
```diff
@@ -0,0 +1,32 @@
...
+    } else {
+      darkreader.disable();
+    }
+  }
+}
+
+export function setupTheme() {
+  preferences.onPreferenceValueChanged('theme', value => {
+    const electronThemeSource = compassThemeToElectronTheme[value];
+    remote.nativeTheme.themeSource = electronThemeSource;
```
addaleax, line 27: It seems a bit more natural that this is what we should do here. Notably, this also affects other theme properties such as the window's own theme, not just its contents.

packages/compass-home/package.json
gribnoysup: What error are you seeing?

addaleax: @gribnoysup When in `dependencies` or `peerDependencies` only:

```
Errors found in the @mongodb-js/compass-home package.json dependencies description

Package `peerDependencies` should match `dependencies`
```

When in both:

```
Errors found in the @mongodb-js/compass-home package.json dependencies description

Following dependencies should not be in peerDependencies:

  @electron/remote
```

packages/compass-components/src/components/file-input.tsx
gribnoysup: I think it would be worth it if we ever consider exposing these to the user (like github feature preview for example), otherwise I would probably prefer them to be just env vars to distinct them from the actual options. But also don't really have a very strong opinion here, wouldn't mind if you feel like it's a good idea to move them to prefs

packages/compass-home/package.json
addaleax: Well, I’m still not entirely clear around the expected semantics here (esp. regarding why `peerDependencies` duplicates `dependencies`), so I’m not quite sure what the error message should actually be :) But yeah, for this problem, I guess moving it to dev deps works. I think I was just confused by the fact that the `devDependencies` list looked like actual dev dependencies, rather than dev dependencies + what ends up being bundled.

packages/compass-components/src/components/file-input.tsx
addaleax: I don't have super strong opinions either, but I do like that it makes it potentially a bit easier to ask users to try something out and that it provides us with a single mechanism for reading configurable properties of the application

gribnoysup: Ahh, okay, so those will not be exposed in the UI, but you'd still be able to set them with command args, yeah, this makes sense

packages/compass-settings/src/components/settings/theme.tsx
addaleax: Thanks to @gribnoysup the SVGs are included now :slightly_smiling_face: There’s a slight bit of awkwardness in that the darkreader dark mode turns the light mode icon dark when it’s on, but that might be okay, esp. since it’s not a forever solution (hopefully):

LG dark mode:
Screenshot from 2022-10-26 17-02-33

Darkreader:
Screenshot from 2022-10-26 17-01-40

(and sorry for the inconsistent cropping)

packages/compass-components/src/hooks/use-theme.tsx
```diff
@@ -50,8 +51,7 @@ const withTheme = function <
     const theme = useTheme();
 
     const applyTheme =
-      theme.enabled === true ||
-      global?.process?.env?.COMPASS_LG_DARKMODE === 'true';
+      theme.enabled === true || preferences.getPreferences().lgDarkmode;
```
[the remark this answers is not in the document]
addaleax: @gribnoysup Yes, all of this absolutely makes sense! Can you take a look at 4cd9bf2bf0b9ed6e62c4ba242ee4f6e4e748d6b8 and see if that matches what you were expecting? I’m just feeling a little confused about it because I couldn’t find anything setting the `enabled` flag before, but maybe there really wasn’t anything.

packages/compass/src/app/theme.ts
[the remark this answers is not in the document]
addaleax: Yes and yes, done! :)

packages/compass-components/src/hooks/use-theme.tsx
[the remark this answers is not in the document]
gribnoysup: I am in the same boat and it also took me a moment to grasp how these flags should interact 😄 But yeah, at least I can definitely say that our understanding is the same here, this commit looks good to me and what I would expect!

I think `enabled` is currently exclusively used in the Tooltip component to invert the theme for anything that is rendered inside the tooltip, because they are dark in the light mode and vice verca:

https://github.com/mongodb-js/compass/blob/1326e454c989aabae8f884b9e3a81fb2fc836f5c/packages/compass-components/src/components/tooltip.tsx#L68-L76

packages/compass-settings/src/components/settings/theme.tsx
gribnoysup: Oh, darkreader changing our light icon to the dark one is very meta 😆 Depending on how much time you feel like spending on this, you can probably convert this svg to data-url and use an img tag instead of svg directly, I would hope that in this case darkreader will not alter the content.

But also I agree that it's probably fine for now

packages/compass/src/main/menu.ts
```diff
@@ -528,19 +443,15 @@ class CompassMenu {
...
 
-    ipcMain.respondTo({
-      'window:theme-loaded': (_, theme) => {
-        if (this.themeState.theme !== theme) {
-          this.themeState.theme = theme;
-          this.saveThemeAndRefreshMenu({
-            theme
-          });
+    preferences.onPreferenceValueChanged('theme', newTheme => {
+      CompassTelemetry.track({
```
gribnoysup, line 447: Asking because I'm adding tracking in main process in another PR and I am doing this through the usual `createLoggerAndTelemetry`:

https://github.com/mongodb-js/compass/blob/b0cb4fc4940d55871ea7dffe261b22c7d580cc98/packages/compass/src/main/auto-update-manager.ts#L156

Should I use `CompassTelemetry` directly instead?

gribnoysup (approved):
Left one comment about how the switch looks in dark mode, but I generally agree that we can ignore it and otherwise this looks good to me!

packages/compass-settings/src/components/modal.tsx
```diff
@@ -54,7 +55,10 @@ const footerStyles = css({
   paddingBottom: 0,
 });
 
-const settings: Settings[] = [{ name: 'Privacy', component: PrivacySettings }];
+const settings: Settings[] = [
+  { name: 'Privacy', component: PrivacySettings },
+  { name: 'Theme', component: ThemeSettings },
```
gribnoysup, line 60: Oh, one small thing here. In the menu we mention that this is still a "Preview" feature, should we add it to the label here too?

image

packages/compass/src/main/menu.ts
addaleax: No, actually, let me fix this while I’m here :+1:

packages/compass-settings/src/components/modal.tsx
addaleax: Yes! :+1:

gribnoysup: Looks good, thanks for updating!

PR #3590, opened 2022-10-13
chore(telemetry): track event on fetching attribute failure, add tests COMPASS-6189

Anemy:
COMPASS-6189
Doesn't look like anything is wrong with telemetry. While investigating I figured it might be good to still track an event when fetching the attributes for the event fails. This would ensure if something breaks in a function like the fetching connection details one here: https://github.com/mongodb-js/compass/blob/27e900f99a6608b77068ba74cc4171fe898a1106/packages/compass-connections/src/modules/telemetry.ts#L125 that we still track the event, however with an `error_fetching_properties: true`  attribute so we know something is up.

Before merging I'm going to check with product that emitting the same event with that attribute is okay. We might instead want to send a different, unique, event with the event name in the properties.
Also I'll cleanup the tests a bit so they aren't doing a possibly flaky await timeout.

### DIFF
--- a/packages/compass-logging/src/index.spec.ts
+++ b/packages/compass-logging/src/index.spec.ts
@@ -94,9 +94,9 @@ describe('createLoggerAndTelemetry', function () {
 
     expect(trackingLogs).to.deep.equal([
       {
-        event: 'Test Event3',
+        event: 'Error Fetching Attributes',
         properties: {
-          error_fetching_properties: true,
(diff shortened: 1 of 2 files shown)


Anemy:
Chatted with Julia, will update to make it a separate event for the failures.

mcasimir (approved):
FYI @mmarcon

PR #3593, opened 2022-10-14
fix(import-export): fix tooltip overflow in export select fields header, use lg tooltip COMPASS-6202

Anemy:
COMPASS-6202

Also did a few drive by updates to leafygreen components.

| Before | After |
| --- | --- |
| Screen Shot 2022-10-14 at 1 19 52 PM | Screen Shot 2022-10-14 at 1 20 26 PM |


mcasimir (approved):
Great catch!

PR #3602, opened 2022-10-17
chore(components): add ListEditor component, align input list components COMPASS-6120

Anemy:
COMPASS-6120

Adds a `ListEditor` component to `compass-components`. Uses it in `connection-form` (hosts, url options) and `compass-indexes` (field inputs).

As the url options are mapped directly to the connection string and we can't add placeholders into the connection string its behavior is a little different. We are in this pr removing the table view there, checking with design that's cool. We can always merge the other two if we want.

Did some converting to function components in `compass-indexes` and removed some code around combobox dark mode, as that is now supported out of the box.

Also noticed some improper use of `aria-labelledby` in these components, not addressing in this pr. Figured it's worth mentioning though as we were chatting about this sort of thing today.

Screen Shot 2022-10-17 at 7 26 38 PM
Screen Shot 2022-10-17 at 7 27 00 PM
Screen Shot 2022-10-17 at 7 27 24 PM
Screen Shot 2022-10-17 at 7 27 30 PM
Screen Shot 2022-10-17 at 7 27 59 PM


packages/compass-indexes/src/components/create-index-form/create-index-form.tsx
```diff
@@ -70,135 +74,134 @@ export type CreateIndexProps = {
...
-          <ColumnstoreProjectionCollapsibleFieldSet
-            useColumnstoreProjection={this.props.useColumnstoreProjection}
-            toggleUseColumnstoreProjection={
-              this.props.toggleUseColumnstoreProjection
-            }
-            columnstoreProjection={this.props.columnstoreProjection}
-            columnstoreProjectionChanged={
-              this.props.columnstoreProjectionChanged
-            }
+        <Body weight="medium" className={indexFieldsHeaderStyles}>
```
mcasimir, line 132: Shouldn't this be a label?

For the other fields we do this: `<Label htmlFor={labelId}>Create unique index</Label>`

mcasimir (approved):
Awesome!

packages/compass-indexes/src/components/create-index-fields.tsx
```diff
@@ -0,0 +1,165 @@
...
+});
+
+const createIndexFieldsNameStyles = css({
+  flexGrow: 1,
+  textTransform: 'none',
+  whiteSpace: 'nowrap',
+});
+
+const createIndexFieldsTypeStyles = css({
+  flexGrow: 1,
```
mcasimir, line 41: Probably better to set a fixed `width` for this one. Flex causes some weird layout as the type field is not always the same size:

Screenshot 2022-10-18 at 09 17 14

mcasimir (changes requested):

packages/compass-indexes/src/components/create-index-fields.tsx
Anemy: Ah nice catch. Updated.

Anemy:
> About aria-labelledby, if you want it should be safe to convert them to aria-label.

This would result in a typescript error with leafygreen's types:
https://github.com/mongodb/leafygreen-ui/blob/main/packages/text-input/src/types.ts#L147
As it's unrelated to these changes I'd like to fix that in seperate pr.

packages/compass-indexes/src/components/create-index-form/create-index-form.tsx
mcasimir: Oh, I only meant that style-wise it would have been nice to use `<Label>` instead of trying to reproduce one by using a `Body` with the same weight and margin of a label, I think is fine for now, just something that shouldn't be necessary.

mcasimir (approved):
🔥

PR #3609, opened 2022-10-19
feat: implement readOnly option COMPASS-6064

alenakhineika:
## Description

Use `readOnly` from settings instead of having a separate list of plugins for the Compass read-only edition.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-crud/src/stores/crud-store.js
```diff
@@ -117,6 +118,17 @@ export const setIsReadonly = (store, isReadonly) => {
   store.onReadonlyChanged(isReadonly);
 };
 
+/**
+ * Set the isPreferencesReadonly flag in the store.
+ *
+ * @param {Store} store - The store.
+ */
+export const setIsPreferencesReadonly = (store) => {
```
addaleax, line 126: I think it might be good to change the name and description to reflect the fact that this adds a listener, and doesn’t set anything on its own.

packages/compass-e2e-tests/tests/global-preferences.test.ts
```diff
@@ -168,6 +169,26 @@ describe('Global preferences', function () {
     }
   });
 
+  it('allows setting readOnly: true and reflects that on the crud view', async function () {
+    await fs.writeFile(path.join(tmpdir, 'config'), 'readOnly: true\n');
+    // No settings modal opened for Isolated edition
+    const compass = await beforeTests();
```
addaleax, line 175: … but this isn’t isolated, so there _should_ be a settings modal, right? I’m not sure, maybe we just need to drop the comment(s) here after the recent welcome modal refactor.

packages/compass-e2e-tests/tests/read-only.test.ts
```diff
@@ -0,0 +1,196 @@
...
+  const disabled = await checkbox.getAttribute('disabled');
+  const value = await checkbox.getAttribute('aria-checked'); // .getValue() always returns 'on'?
+  const banner = await browser.$(
+    `${settingSelector} [data-testid="set-cli-banner"], ${settingSelector} [data-testid="set-global-banner"]`
+  );
+  const bannerText = (await banner.isExisting())
+    ? await banner.getText()
+    : null;
+  return { disabled, value, bannerText };
+}
```
addaleax, line 25: Can this be deduplicated with the function from global-preferences.test.ts?

packages/compass-home/src/components/workspace.tsx
```diff
@@ -61,7 +62,11 @@ export default function Workspace({
             <WorkspaceContent namespace={namespace} />
           </div>
         </div>
-        <div>{GlobalShellComponent && <GlobalShellComponent />}</div>
+        <div>
+          {!preferences.getPreferences().readOnly && GlobalShellComponent && (
```
addaleax, line 66: Do we want to add a listener for changes here as well?

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -139,6 +141,18 @@ function deriveNetworkTrafficOptionState<K extends keyof AllPreferences>(
...
 }
+
+/** Helper for defining how to derive value/state for readOnly-affected preferences */
+function deriveReadOnlyOptionState<K extends keyof AllPreferences>(
+  property: K
+): DeriveValueFunction<boolean> {
+  return (v, s) => ({
+    value: v(property) && !v('readOnly'),
+    state:
+      s(property) ?? s('readOnly') ?? (v('readOnly') ? undefined : 'derived'),
```
addaleax, line 152:
```suggestion
      s(property) ?? s('readOnly') ?? (!v('readOnly') ? undefined : 'derived'),
```

to match the networkTraffic option logic? Otherwise (as in the tests above) `enableShell: 'derived'` shows up when it shouldn’t and vice versa

packages/compass-schema-validation/src/stores/store.js
```diff
@@ -74,6 +81,9 @@ const configureStore = (options = {}) => {
     // set the initial value
     changeEditMode();
 
+    // set hadronReadOnly if readOnly value has cganged in Compass settings.
```
addaleax, line 84:
```suggestion
    // set hadronReadOnly if readOnly value has changed in Compass settings.
```

(but also similarly here, it might be good to rename things, esp. since there is also nothing called …hadron… anymore related to readonly settings)

packages/compass-preferences-model/src/preferences.ts
alenakhineika: Yes, I am aware of this, fix is not pushed yet.

packages/compass-crud/src/stores/crud-store.js
alenakhineika: This is similar to `setIsReadonly` in the same file, I think it means here that it sets this value in the state of the store, not on the collection or settings themselves.

addaleax: But `setIsReadonly` actually sets a value in the state of the store, and this one doesn’t do so directly, it ‘only’ installs a listener that does?

packages/compass-aggregations/src/modules/preferences-readonly.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * The preferencesReadOnly changed action name.
+ */
+export const PREFERENCES_READONLY_CHANGED =
+ 'aggregations/preferences-readonly/PREFERENCES_READONLY_CHANGED';
+
+/**
+ * The initial state of the preferencesReadOnly attribute.
+ */
+export const INITIAL_STATE = false;
```
addaleax, line 12: Afaict we always initialize these as soon as the store is loaded, so is there any reason not to use the safe default of `true` here?

packages/compass-aggregations/src/utils/stage.js
```diff
@@ -57,8 +57,9 @@ export const filterStageOperators = ({ serverVersion, env, isTimeSeries, sourceN
     sourceName ? VIEW :
     COLLECTION;
 
+  const isEditable = !isReadonly && !preferencesReadOnly;
   return STAGE_OPERATORS
-    .filter(disallowOutputStagesOnCompassReadonly)
+    .filter((op) => disallowOutputStagesOnCompassReadonly(op, isEditable))
```
addaleax, line 62: :+1:

packages/compass-shell/src/stores/store.js
```diff
@@ -19,6 +22,18 @@ export default class CompassShellStore {
...
+
+    this.reduxStore.dispatch(preferencesReadOnlyChanged(!!readOnly));
+    preferences.onPreferenceValueChanged('readOnly', (changedReadOnly) => {
+      this.reduxStore.dispatch(preferencesReadOnlyChanged(changedReadOnly));
+    });
+
+    this.reduxStore.dispatch(preferencesReadOnlyChanged(!!enableShell));
+    preferences.onPreferenceValueChanged('enableShell', (changedEnableShell) => {
+      this.reduxStore.dispatch(enableShellChanged(changedEnableShell));
+    });
```
addaleax, line 35: Why do we track both of these, and not just `enableShell`?

packages/compass-settings/src/components/settings/privacy.tsx
```diff
@@ -29,6 +29,8 @@ const privacyFields = [
   'autoUpdates',
   'enableMaps',
   'trackErrors',
+  'readOnly',
+  'enableShell',
```
addaleax, line 33: Should these be in the privacy section? They aren’t really privacy-related, and I think it would be fine to add a new section if we want to (although, heads up, #3588 also adds one).

packages/compass-sidebar/src/components/sidebar-databases-navigation.ts
```diff
@@ -28,7 +29,7 @@ function mapStateToProps(state: any) {
     ])
   );
   const isReadOnly =
-    process.env.HADRON_READONLY === 'true' ||
+    preferencesReadOnly ||
```
addaleax, line 32: Had a good conversation with @gribnoysup about these cases where we keep track of the preference value in the stores in this PR, and how we could avoid having to do that – not blocking for this PR, but more in the “it might be nice not to have to add tracking to the stores” kind of way:

In most cases, accessing preferences falls into two categories:
1. Affecting how something is displayed in the UI
2. Affecting how an action is performed in response to a UI event

In the first case, the `usePreference` hook should get us fairly far, in the second case, it should be fine to access the current value of the preference in the place in which it is actually used via `getPreferences().<name>`.

Taking this as an example, I think it would be pretty okay to drop the readonly checks from this package entirely, and instead do a check inside the target component itself, something along the lines of:

```diff
diff --git a/packages/compass-databases-navigation/src/databases-navigation-tree.tsx b/packages/compass-databases-navigation/src/databases-navigation-tree.tsx
index edd5a27a6311..44cad773beaa 100644
--- a/packages/compass-databases-navigation/src/databases-navigation-tree.tsx
+++ b/packages/compass-databases-navigation/src/databases-navigation-tree.tsx
@@ -84,12 +84,13 @@ const NavigationItem = memo<{
   // eslint-disable-next-line @typescript-eslint/unbound-method
   const {
     items,
-    isReadOnly,
     activeNamespace,
     currentTabbable,
     onDatabaseExpand,
     onNamespaceAction,
   } = data;
+  const preferencesReadOnly = usePreference('readOnly');
+  const isReadOnly = preferencesReadOnly || data.isReadOnly;
 
   const itemData = items[index];
 ```

File: packages/compass/src/main/menu.ts [hunk 12]
```diff
@@ -452,6 +452,14 @@ class CompassMenu {
       this.refreshMenu();
     });
 
+    preferences.onPreferenceValueChanged('readOnly', newReadOnly => {
+      track('Read-Only Changed', {
+        readOnly: newReadOnly
+      });
```
addaleax, line 458: Might be good to double-check with product that it makes sense to track this event

packages/compass-shell/src/components/compass-shell/compass-shell.jsx
```diff
@@ -250,6 +251,7 @@ export default connect(
         state.appRegistry.globalAppRegistry.emit('compass:compass-shell:opened');
       }
     },
-    runtime: state.runtime ? state.runtime.runtime : null
+    runtime: state.runtime ? state.runtime.runtime : null,
+    enableShell: state.enableShell ? state.enableShell : false
   })
 )(CompassShell);
```
addaleax, line 257: Another example of a possible simplification – you can avoid having to track `enableShell` in the state fairly directly by adding it to the props:

```suggestion
  })
)(withPreferences(CompassShell, ['enableShell'], React));
```

packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -425,7 +460,12 @@ describe('Connection screen', function () {
 
     const connectionString = parsedString.toString();
 
+    await browser.setFeature('enableShell', true);
     await browser.connectWithConnectionString(connectionString);
+
+    const shellSection = await browser.$(Selectors.ShellSection);
+    await shellSection.waitForDisplayed();
```
addaleax, line 467: Do you know why we need these changes? `enableShell` should default to `true` when Compass is started normally, right?

alenakhineika: I think I made this change while was debugging particular test suits and this was required in case I run `describe.only()`, because initially, settings apply only when all tests are run.

alenakhineika: Yeah, it is still the case. if I run `describe.only('Connection screen', function () {})` the shell is hidden by default.

Screenshot 2022-11-10 at 13 14 04

packages/compass-settings/src/components/settings/features.tsx
```diff
@@ -0,0 +1,100 @@
...
+> = ({ checkboxValues, preferenceStates, handleChange }) => {
+  const handleCheckboxChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    handleChange(event.target.name as FeaturesFields, event.target.checked);
+  };
+
+  return (
+    <div data-testid="features-settings">
+      <Body>
+        To enhance the user experience, Compass can enable or disable particular
+        featues. Please choose from the settings below:
```
addaleax, line 63:
```suggestion
        features. Please choose from the settings below:
```

packages/compass-e2e-tests/tests/connection.test.ts
addaleax: Yeah, it’s this: https://github.com/mongodb-js/compass/blob/a83dab45ca4da6c7597406d8d02e1133fa312044/packages/compass-e2e-tests/helpers/compass.ts#L405-L408

We should definitely not be doing that.

packages/compass-e2e-tests/tests/shell.test.ts
```diff
@@ -27,6 +27,8 @@ describe('Shell', function () {
   });
 
   it('has an info modal', async function () {
+    await browser.setFeature('readOnly', false);
+    await browser.setFeature('enableShell', true);
```
addaleax, line 31: These should be the default options now, right?

packages/compass-e2e-tests/tests/shell.test.ts
```diff
@@ -42,6 +44,8 @@ describe('Shell', function () {
   });
 
   it('shows and hides shell based on settings', async function () {
+    await browser.setFeature('readOnly', false);
+    await browser.setFeature('enableShell', true);
```
addaleax, line 48: Ditto here :)

packages/compass-indexes/src/components/indexes-toolbar/indexes-toolbar.tsx
```diff
@@ -36,30 +36,30 @@ type IndexesToolbarProps = {
...
   writeStateDescription,
   onRefreshIndexes,
+  readOnly, // preferences readOnly.
 }) => {
   const onClickCreateIndex = useCallback(() => {
     localAppRegistry.emit('toggle-create-index-modal', true);
   }, [localAppRegistry]);
   const showCreateIndexButton =
-    !isReadonly && !isReadonlyView && !preferencesReadOnly && !errorMessage;
+    !isReadonly && !isReadonlyView && !readOnly && !errorMessage;
```
addaleax, line 62: Totally non-blocking, but if you’d rather not pass this as a prop (and you might want to do it for easier testing or similar), you can do 

```js
  const preferencesReadOnly = usePreference('readOnly', React);
  const showCreateIndexButton = 
    !isReadonly && !isReadonlyView && !preferencesReadOnly && !errorMessage;
```

(because I totally see that it’s confusing to have `isReadonly` and `readOnly` as prop names)

packages/compass-indexes/src/components/indexes/indexes.tsx
```diff
@@ -30,29 +31,29 @@ type IndexesProps = {
...
   isReadonlyView,
-  preferencesReadOnly,
   description,
   error,
   localAppRegistry,
   isRefreshing,
   sortIndexes,
   refreshIndexes,
   dropFailedIndex,
+  readOnly, // preferences readOnly.
```
addaleax, line 56: Similarly here, it’s totally optional but for function components `usePreference()` might be a bit easier to read than `withPreferences()`, although in the end they pretty much boil down to the same thing anyway

packages/compass-e2e-tests/tests/shell.test.ts
alenakhineika: I thought so! And locally it works, but for some reason fails on CI: https://spruce.mongodb.com/task/10gen_compass_main_macos_arm64_test_packaged_app_60x_enterprise_patch_e2a6e6fe8cc1d4b767455ada36dc112ce2df69f2_636e61d13066154e7bb66841_22_11_11_14_53_06/logs?execution=0

I added this explicitly here as POC to verify and as we can see CI is green now. Any ideas what is different for the shell suit?

packages/compass-e2e-tests/tests/read-only.test.ts
```diff
@@ -0,0 +1,421 @@
...
+import { createNumbersCollection } from '../helpers/insert-data';
+
+describe('readOnly: true / Read-Only Edition', function () {
+  let tmpdir: string;
+  let i = 0;
+
+  beforeEach(async function () {
+    tmpdir = path.join(
+      os.tmpdir(),
+      `compass-global-preferences-${Date.now().toString(32)}-${++i}`
```
addaleax, line 16:
```suggestion
      `compass-read-only-${Date.now().toString(32)}-${++i}`
```

packages/compass-e2e-tests/tests/read-only.test.ts
```diff
@@ -0,0 +1,421 @@
...
+
+  afterEach(async function () {
+    await fs.rmdir(tmpdir, { recursive: true });
+  });
+
+  it('hides and shows the plus icon on the siderbar to create a database', async function () {
+    const compass = await beforeTests();
+    try {
+      const browser = compass.browser;
+      await browser.setFeature('readOnly', true);
```
addaleax, line 29: Should this maybe be set back to false in the `finally` part?

packages/compass-indexes/src/components/indexes-toolbar/indexes-toolbar.tsx
alenakhineika: Yeah, I found it handy for unit testing. But you would still replace it with usePreference? How then to tweak readOnly for testing?

packages/compass-e2e-tests/tests/read-only.test.ts
alenakhineika: Yeah maybe it would be more explicit, but all tests start with setting it to false, but I can see your concern, will add.

addaleax: I’m more worried about this potentially affecting the next test that runs, if only this test + another test run, for example

packages/compass-e2e-tests/tests/read-only.test.ts
alenakhineika: But this dir is for global preferences, right? Maybe it is ok to call it the same? I have seen it also named like this here: https://github.com/mongodb-js/compass/blob/COMPASS-6064-implement-readonly-option/packages/compass-e2e-tests/tests/no-network-traffic.test.ts#L20

So I assumed we wanted to call it preferences :)

addaleax: > I have seen it also named like this here: https://github.com/mongodb-js/compass/blob/COMPASS-6064-implement-readonly-option/packages/compass-e2e-tests/tests/no-network-traffic.test.ts#L20

Yeah that just means I was too lazy and copied it without fixing it :see_no_evil: 

The name is `compass-global-preferences-...` because it is called that way in the `global-preferences.test.ts` file. It’s used to avoid conflicts between tests (since the `i` variable is local to each test), in the rare case in which `Date.now()` might not be enough to disambiguate directories (very unlikely for e2e tests, more likely for other tests).

So, no, this is not because it’s about storing global preferences, it’s about the test name.

alenakhineika: Got it! :) will update two places then!

packages/compass-e2e-tests/tests/shell.test.ts
addaleax: Yeah, confirmed this is happening because this test doesn’t reset the settings after it’s run: https://github.com/mongodb-js/compass/blob/07e3a93dd9461cf56aca667550e070faa2e60d7a/packages/compass-e2e-tests/tests/read-only.test.ts#L390

I do see how this cross-test interaction is annoying, and I’m not opposed to going one step further than 9e71cfd883bf91df4 and starting Compass with a fresh user data directory for e2e tests by default.

packages/compass-indexes/src/components/indexes-toolbar/indexes-toolbar.tsx
alenakhineika: I decided to keep it as it is for now to get this PR merged, and revisit later as a separate chunk of work together with refactoring of `compass-aggregations` and `compass-crud` stores that still keep using `preferencesReadOnly` and listen for readOnly changed events.

packages/compass-aggregations/src/components/aggregations/aggregations.jsx
```diff
@@ -83,6 +83,8 @@ const mapStateToProps = (state) => ({
   setIsNewPipelineConfirm: state.setIsNewPipelineConfirm,
   updateViewError: state.updateViewError,
   workspace: state.workspace,
+  isReadonly: state.isReadonly,
+  preferencesReadOnly: state.preferencesReadOnly,
```
addaleax, line 87: I tried to find the place where these two are being consumed but couldn’t – what am I missing? `Aggregations` forwards its props as-is to `Pipeline`, but `Pipeline` doesn’t seem to do anything with them?

packages/compass-aggregations/src/modules/preferences-readonly.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Action creator for preferencesReadOnly changed events.
+ *
+ * @param preferencesReadOnly - Is preferencesReadOnly.
+ *
+ * @returns The preferencesReadOnly changed action.
+ */
+export const preferencesReadOnlyChanged = (preferencesReadOnly: boolean) => ({
+  type: PREFERENCES_READONLY_CHANGED,
+  preferencesReadOnly,
+});
```
addaleax, line 39: Is there a TODO ticket for removing this from the state? I understand the tradeoff here, but we should get on removing this code again as soon as we can.

packages/compass-aggregations/src/modules/update-view.js
```diff
@@ -94,6 +94,7 @@ export const updateView = () => {
         const metadata = {
           namespace: viewNamespace,
           isReadonly: true,
+          preferencesReadOnly: state.preferencesReadOnly,
```
addaleax, line 97: Where is this being used/why is this being added here?

packages/compass-aggregations/src/utils/stage.js
```diff
@@ -37,11 +37,11 @@ export function isAtlasOnly(operatorEnv) {
   return operatorEnv?.every(env => env === ATLAS);
 }
 
-function disallowOutputStagesOnCompassReadonly(operator) {
+function disallowOutputStagesOnCompassReadonly(operator, isEditable) {
   if (operator?.outputStage) {
     // NOTE: this should be innocuous in Data Explorer / Web
     // and just always return `false`
```
addaleax, line 43:
```suggestion
```

packages/compass-e2e-tests/tests/shell.test.ts
addaleax: (I’m not sure why this conversation was marked resolved)

packages/compass-indexes/src/components/indexes-toolbar/indexes-toolbar.tsx
alenakhineika: The follow up ticket: https://jira.mongodb.org/browse/COMPASS-6300

packages/compass-aggregations/src/modules/preferences-readonly.ts
alenakhineika: Created one: https://jira.mongodb.org/browse/COMPASS-6300

packages/compass-aggregations/src/components/aggregations/aggregations.jsx
alenakhineika: You are right, removed.

packages/compass-aggregations/src/modules/update-view.js
alenakhineika: Removed.

addaleax (approved):

PR #3595, opened 2022-10-14
chore(components): upgrade leafygreen-ui package versions, add darkmode to pipeline

Anemy:
Figured it's good to keep these up to date when we can.
Doesn't look like too many visual changes. `badge`, `menu`, `banner`, `toast`, and `pipeline` all got major bumps.
Labelled this pr a `chore` as the changes are fairly small. Happy to change if anyone thinks a better label is more appropriate.

There were two type issues which I'll leave comments by, asked qs in the leafygreen channel about.

Now the `Pipeline` component supports darkmode:
Screen Shot 2022-10-14 at 2 35 52 PM

PR #3611, opened 2022-10-19
chore: undo the 2px hack

lerouxb:
I added it here https://github.com/mongodb-js/compass/pull/3551 and then when we bumped leafygreen here https://github.com/mongodb-js/compass/pull/3595 it became unnecessary and in fact now moves it too far.


lerouxb:
Closed in favor of https://github.com/mongodb-js/compass/pull/3602

PR #3612, opened 2022-10-19
chore: Make sure MenuItems are type button, not type submit

lerouxb:
If a leafygreen Menu (and therefore MenuItems) makes its way into a `<form>` somehow then clicking on a menu item will submit that form. This is because it uses a button tag without specifying a type and buttons by default have `type="submit"`.

### DIFF
--- a/packages/compass-components/src/components/leafygreen.tsx
+++ b/packages/compass-components/src/components/leafygreen.tsx
@@ -27,6 +27,9 @@ import {
 } from '@leafygreen-ui/logo';
 import { Menu, MenuSeparator, MenuItem } from '@leafygreen-ui/menu';
 
+// If a leafygreen Menu (and therefore MenuItems) makes its way into a <form>,
+// clicking on a menu item will submit that form. This is because it uses a button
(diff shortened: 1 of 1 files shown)


packages/compass-components/src/components/leafygreen.tsx
```diff
@@ -26,6 +26,12 @@ import {
   MongoDBLogo,
 } from '@leafygreen-ui/logo';
 import { Menu, MenuSeparator, MenuItem } from '@leafygreen-ui/menu';
+
```
Anemy, line 29: Can we add a comment with the description to help future readers?
```suggestion

// If a leafygreen Menu (and therefore MenuItems) makes its way into a <form>,
// clicking on a menu item will submit that form. This is because it uses a button
// tag without specifying a type and buttons by default have type="submit".
```

Anemy (approved):

kraenhansen:
I created https://jira.mongodb.org/browse/LG-5586 to track fixing this upstream 🤞

PR #3620, opened 2022-10-20
feat(pipeline-builder): pipeline text editor

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/modules/pipeline-builder/builder-helpers.ts
```diff
@@ -1,31 +1,30 @@
...
   state: RootState,
   pipelineBuilder: PipelineBuilder
 ) {
-  if (state.pipelineBuilder.pipelineMode === 'builder-ui') {
-    return pipelineBuilder.stages.map((stage) => stage.toBSON());
-  } else {
-    // TODO
-    return [];
+  if (state.pipelineBuilder.pipelineMode === 'as-text') {
+    pipelineBuilder.sourceToStages();
```
gribnoysup, line 25: I am happy to chat through this a bit more if you want to, from my perspective it would be a bit cleaner if we don't change / commit builder state unless the view type is switched. For that reason I would suggest to use `getPipelineFromSource` here instead of committing source to stages and then returning stage operators from stages. It also means that you don't need to go through the source > stages > stage to bson flow, you will just immediately go from source to bson / parsed pipeline

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-mode.ts
```diff
@@ -26,9 +29,27 @@ const reducer: Reducer<State> = (state = INITIAL_STATE, action) => {
...
+    } else {
+      pipelineBuilder.sourceToStages();
+      dispatch(changeStages(pipelineBuilder.stages));
+    }
+
+    // Toggle the view (and state)
+    dispatch({
+      type: ActionTypes.PipelineModeToggled,
+      mode: newMode,
+    });
```
gribnoysup, line 51: I'd maybe suggest making this one event instead of three separate ones, something like:

```ts
if (newMode === 'as-text') {
  pipelineBuilder.stagesToSource();
} else {
  pipelineBuilder.sourceToStages();
}
dispatch({
  type: ActionTypes.PipelineModeToggled,
  source: pipelineBuilder.source,
  stages: pipelineBuilder.stages,
});
```

I think that way the UI flow is clearer in the reducers because you are avoiding actions that are too generic and lack context, this aligns with [redux style guide](https://redux.js.org/style-guide/#model-actions-as-events-not-setters) too

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  return (
+    <div className={documentListStyles}>
+      <DocumentListView {...listProps} />
+    </div>
+  );
+};
+
+export const PipelineAsTextWorkspace: React.FunctionComponent<
```
gribnoysup, line 115: Should this be `PipelinePreview`?

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
```diff
@@ -0,0 +1,143 @@
...
+    );
+  }
+
+  if (isLoading) {
+    return (
+      <div className={centerStyles}>
+        <SpinLoader size="24px" />
+      </div>
+    );
+  }
```
gribnoysup, line 97: Something weird is going on here. Looks like component completely remounts every time I type a letter causing the animation on the spinner to restart:

Kapture 2022-10-21 at 09 21 35

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
```diff
@@ -0,0 +1,143 @@
...
+    return (
+      <div className={centerStyles}>
+        <DocumentIcon size={spacing[6]} />
+        <Body className={messageStyles}>
+          Preview results to see a sample of the aggregated results from this
+          pipeline.
+        </Body>
+      </div>
+    );
+  }
```
gribnoysup, line 89: Shouldn't we completely hide this panel when autopreview is off like we do in stage by stage editor?

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-mode.ts
```diff
@@ -26,9 +34,28 @@ const reducer: Reducer<State> = (state = INITIAL_STATE, action) => {
...
+    if (newMode === 'as-text') {
+      pipelineBuilder.stagesToSource();
+    } else {
+      pipelineBuilder.sourceToStages();
+    }
+
+    dispatch({
+      type: ActionTypes.PipelineModeToggled,
+      mode: newMode,
+      stageOperators: getStageOperatorsFromPipelineSource(getState(), pipelineBuilder),
```
mabaasit, line 51: stageOperators are part of `as-text` slice and used to show stage list in toolbar.

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
```diff
@@ -0,0 +1,132 @@
...
+        <Overline>Pipeline Output</Overline>
+        {!isLoading && docCount ? (
+          <Body>{`Sample of ${docCount} ${docText}`}</Body>
+        ) : (
+          <></>
+        )}
+      </div>
+      <PreviewResults isLoading={isLoading} previewDocs={previewDocs} />
+    </div>
+  );
```
gribnoysup, line 120: Tiny nit: you don't really need to return empty fragments if you don't want to render anything:

```tsx
        {!isLoading && docCount && (
          <Body>{`Sample of ${docCount} ${docText}`}</Body>
        )}
```

Additionally as 0 is falsy you are running into corner cases here with the way the code structured. What I'd suggest you to do here is to move the condition into a correctly typed variable just so it's clear what's the expectation for what you are trying to achieve here:

```suggestion
  const docCount = previewDocs?.length ?? 0;
  const shouldShowCount = !isLoading && docCount > 0 // or don't check docCount if we want to render always even when count is zero
  const docText = docCount === 1 ? 'document' : 'documents';
  return (
    <div className={containerStyles} data-testid="pipeline-as-text-preview">
      <div className={previewHeaderStyles}>
        <Overline>Pipeline Output</Overline>
        {shouldShowCount && (
          <Body>{`Sample of ${docCount} ${docText}`}</Body>
        )}
      </div>
      <PreviewResults isLoading={isLoading} previewDocs={previewDocs} />
    </div>
  );
```

That way whoever is working with this code in the future knows exactly what's the intention here was

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-settings/pipeline-extra-settings.tsx
```diff
@@ -105,11 +113,18 @@ export const PipelineExtraSettings: React.FunctionComponent<
...
-}: RootState) => ({
-  isAutoPreview: autoPreview,
-  pipelineMode,
-});
+  pipelineBuilder: { pipelineMode, textEditor, stageEditor },
+}: RootState) => {
+  const isPipelineModeDisabled =
+    pipelineMode === 'as-text'
+      ? textEditor.syntaxErrors.length > 0
+      : stageEditor.stages.some((x) => x.syntaxError);
```
gribnoysup, line 121: Should we use builder helper here? Is the concern that we don't want to disable on network errors? If that's the case it might be a good idea to not account for them anywhere, not only here and so I think that using a helper is still warranted

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-parser/stage-parser.ts
```diff
@@ -36,13 +36,14 @@ export function assertStageNode(node: t.Node): asserts node is StageLike {
...
     return;
   }
-  throw new SyntaxError(
-    node.type === 'ObjectExpression'
-      ? (node.properties[0] as t.ObjectProperty | undefined)?.key == null
-        ? 'A pipeline stage specification object must contain exactly one field.'
-        : 'Stage value can not be empty'
-      : 'Each element of the pipeline array must be an object'
-  );
+  const { column, line } = node.loc?.start ?? { line: 1, column: 1 };
```
gribnoysup, line 39: Can't say I have a strong preference, but I wonder whether we even want a fallback if there is not loc info available on the node here 🤔 Is 1:1 a meaningful fallback also? Seems like babel when fails to parse for example always defaults to the end of file

packages/compass-aggregations/src/modules/pipeline-builder/text-editor.ts
```diff
@@ -0,0 +1,209 @@
...
+
+export const changeEditorValue = (
+  value: string
+): PipelineBuilderThunkAction<void, EditorValueChangeAction> => {
+  return (dispatch, getState, { pipelineBuilder }) => {
+    pipelineBuilder.changeSource(value);
+    const stageOperators = getStageOperatorsFromPipelineSource(
+      getState(),
+      pipelineBuilder
+    );
```
gribnoysup, line 197: As we are doing this on every source change, I am starting to think that maybe it makes sense to keep the parsed pipeline and the logic for getting it on source update on the pipeline builder instance

packages/compass-aggregations/src/modules/pipeline-builder/text-editor.ts
```diff
@@ -0,0 +1,209 @@
...
+  EditorPreviewFetchAction |
+  EditorPreviewFetchSuccessAction |
+  EditorPreviewFetchErrorAction
+> => {
+  return async (dispatch, getState, { pipelineBuilder }) => {
+    const {
+      autoPreview
+    } = getState();
+
+    const canRunPipeline = pipelineBuilder.syntaxError.length === 0;
```
gribnoysup, line 141: This should also check whether there is an out stage or not, similar to stage editor. Otherwise we are trying to run an out pipeline it seems (and failing with a server error luckily):

image

gribnoysup (commented):
Seems like this empty state has broken styles: text is not aligned correctly

image

gribnoysup (commented):
I think in the designs the error message has some padding around it, looks kinda messy without it:

image

gribnoysup:
Not sure if we have this in designs or not, but I would think that syntax error should also be displayed at the bottom

image

Otherwise you need to hover over the lines to see the error message. Feels like seeing the message right away is quicker in a lot of cases

gribnoysup:
There seems to be some weird 1px spacing at the top of the card and the background of the editor is falling through the card component: 

image

packages/compass-aggregations/src/modules/pipeline-builder/text-editor.ts
mabaasit: added it in [`2f5a100` (#3620)](https://github.com/mongodb-js/compass/pull/3620/commits/2f5a10089d3170c4050a6bcc46d2dc3c2a377985)

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-parser/stage-parser.ts
mabaasit: removed the fallback option. 
now that we are showing syntax errors as warnings on screen, this should not be an issue.

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-settings/pipeline-extra-settings.tsx
mabaasit: Yes, used `getIsPipelineInvalidFromBuilderState` to do that. I extended it to ignore the server errors, because the toggle should only be disabled if there's a parse error.

mabaasit:
> There seems to be some weird 1px spacing at the top of the card and the background of the editor is falling through the card component:
> 
> image

this is broken across and needs fix in editor. will create a follow up for this.

packages/compass-aggregations/src/modules/pipeline-builder/text-editor.ts
```diff
@@ -0,0 +1,221 @@
...
+const INITIAL_STATE: TextEditorState = {
+  pipelineText: '',
+  stageOperators: [],
+  syntaxErrors: [],
+  serverError: null,
+  loading: false,
+  previewDocs: null,
+};
+
+const reducer: Reducer<TextEditorState> = (state = INITIAL_STATE, action) => {
```
gribnoysup, line 64: Seems like actions like loading or creating new empty pipeline (or pipeline from text, but this matters less as we are planning to remove it anyway) are not handled in the reducer, leading to weird behavior

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
```diff
@@ -0,0 +1,131 @@
...
+const documentListStyles = css({
+  overflow: 'scroll',
+});
+
+type PipelinePreviewProps = {
+  isLoading: boolean;
+  previewDocs: Document[] | null;
+};
+
+const PreviewResults = ({
```
gribnoysup, line 49: I think for the empty pipeline (no stages) we should show the input documents, at least this is what I see [in this figma design](https://www.figma.com/file/5dqmdz5h9tqZ4Ypi2VRMHM/Product-Design%3A-AggEx%3A-See-the-whole-pipeline-in-editor-mode?node-id=120%3A26975)

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-mode.ts
gribnoysup, line 51: Redux docs recommend to leave deriving state from actions to reducers, so in this case I would suggest to instead of doing this "fallback to current state" inside the `getStageOperatorsFromPipelineSource` method, just return the stageOperators from the builder here directly:

```suggestion
      pipeline: pipelineBuilder.pipeline,
```

and handle this in the reducer of the text-editor slice directly with something like

```js
return {
  ... other state
  stageOperators: action.pipeline ? getOperators(action.pipeline) : state.stageOperators
}
```

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
mabaasit: there are two design screens for an empty pipeline in the figma link. I'll check with Claudia and if we need to show the documents, i'll create a follow up.

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-builder.ts
```diff
@@ -23,13 +27,32 @@ export class PipelineBuilder {
...
+    return this._source;
+  }
+
+  private parseSourceToPipeline() {
+    try {
+      this.pipeline = parseEJSON(this.source, { mode: ParseMode.Loose });
+    } catch (e) {
+      // After setting source, its either validated or mapped to stages
+      // where the whole parsing takes places and errors are handled.
+    }
```
gribnoysup, line 45: This really should be set to something meaningful like null when parsing fails, otherwise you end up with pipeline builder state being not in sync

packages/compass-aggregations/src/modules/pipeline-builder/text-editor.ts
```diff
@@ -0,0 +1,225 @@
...
+    ) ||
+    isAction<PipelineModeToggledAction>(
+      action,
+      PipelineModeActionTypes.PipelineModeToggled
+    ) ||
+    action.type === RESTORE_PIPELINE ||
+    action.type === CONFIRM_NEW ||
+    action.type === NEW_PIPELINE
+  ) {
+    const stageOperators = action.pipeline.length > 0
```
gribnoysup, line 80: See my other comment in pipeline-builder. Because you don't have an explicit value for a pipeline source that has failed to parse, you are not distinguishing correctly this case from just the empty pipeline, allowing to put UI (and builder) into a unsynced state:

image

gribnoysup (approved):

PR #3626, opened 2022-10-21
chore(e2e): Modal e2e tests

lerouxb:
Added E2E tests for:

* shell info modal
* edit connection string confirmation modal
* open, rename and delete saved item (query/aggregation) modals
* duplicate view, modify view

The only still-untested modal I know of is the non-genuine warning modal, but I can't think of a quick and easy way to do that and it is probably not worth it.


mcasimir (approved):
❤️

PR #3635, opened 2022-10-25
chore: remove the line height and font size globals

Anemy:
Part of COMPASS-5579

This will require more work in moving more text components to LeafyGreen before we can merge, I figure it's a useful branch to be able to jump to to test the font sizes. I'll open as draft.


packages/compass/src/app/styles/deprecated-bootstrap-styles/globals.less
```diff
@@ -2,9 +2,6 @@
 // Compass. These styles are globals from bootstrap which we have removed
 // from the global namespace and are isolated to this plugin.
 .with-global-bootstrap-styles {
-  line-height: 1.428571429; // 20/14
```
mcasimir, line 5: Just an idea, this seems to be the same base line-height/font size as LeafyGreen's, maybe we don't need to remove it, we could just move it here: https://github.com/mongodb-js/compass/blob/abddc77eafe958e59e0fc393b4fb9e937bd4374e/packages/compass-home/src/components/home.tsx#L396

Another idea would be to wrap the entire `<Home>` within a `<Body as="div">`

PR #3644, opened 2022-10-26
chore: Convert export to language to the new config COMPASS-6207

lerouxb:
This just switches compass-export-to-language over to the new shared config. I then did the minimal amount of work necessary to get check and test to pass.


configs/mocha-config-compass/register/jsdom-global-register.js
```diff
@@ -1,5 +1,6 @@
 if (typeof window === 'undefined') {
-  require('global-jsdom')(undefined, { pretendToBeVisual: true });
+  require('global-jsdom')(undefined, { pretendToBeVisual: true, runScripts: 'dangerously' });
+  console.log('window', window, window === 'undefined');
```
lerouxb, line 3: It looks fine here, btw. But who knows which `window` object leafygreen sees.

packages/bson-transpilers/codegeneration/javascript/Visitor.js
```diff
@@ -604,9 +604,7 @@ module.exports = (CodeGenerationVisitor) => class Visitor extends CodeGeneration
       Buffer: Buffer,
       __result: {}
     };
-    const ctx = new Context(sandbox);
-    const res = ctx.evaluate('__result = ' + input);
-    ctx.destroy();
+    const res = vm.runInContext('__result = ' + input, vm.createContext(sandbox));
```
lerouxb, line 607: On node this is all that [context-eval](https://github.com/amasad/context-eval) does. For the browser (ie. data explorer) the plan is to use a [polyfill](https://github.com/browserify/vm-browserify).

context-eval broke for us in this upgrade for kinda tricky reasons:

Before this package was using jsdom via [jsdom-global](https://github.com/rstacruz/jsdom-global) which has a [hardcoded list of properties](https://github.com/rstacruz/jsdom-global/blob/master/keys.js) to copy over to the global (window) object. The shared config (ie. the new way) uses [global-jsdom](https://github.com/modosc/global-jsdom) which uses [Object.getOwnPropertyNames(window)](https://github.com/modosc/global-jsdom/blob/5b90b116426869054cfd551daa7d83b733a35395/commonjs/index.js#L38).

For whatever reason bson-transpiler executes some of the javascript it transpiles and it does this in a context. That requires the jsdom `runScripts: "dangerously"` option to be set. Unfortunately this causes (for reasons we're not 100% clear on, I think) `Object.getOwnPropertyNames(window)` to not return things like `Element` which breaks leafygreen.

So to recap: Without runScripts: "dangeriously" tests would fail with ObjectId not defined. With it leafygreen would fail.

Looking at context-eval, it looks for the existance of the `window` object to know if it is running in node (where it can use the vm module) or in the browser (where it falls back to using an iframe). Since the tests are running in node could just be using `vm`. In electron we should have access to it as well. Elsewhere we shouldn't probably be polyfilling it.

So the cleanest solution seems to be to just not use context-eval and polyfill node's vm module if necessary.

packages/compass/package.json
```diff
@@ -182,7 +182,7 @@
     "@mongodb-js/compass-database": "^3.2.0",
     "@mongodb-js/compass-databases-collections": "^1.2.0",
     "@mongodb-js/compass-explain-plan": "^6.2.0",
-    "@mongodb-js/compass-export-to-language": "^8.2.0",
+    "@mongodb-js/compass-export-to-language": "^0.1.0",
```
gribnoysup, line 185: We probably want the old version to be preserved

lerouxb: aah good point. I think I miss that every time :)

gribnoysup: Yeah, trips me up all the time too 😅

addaleax (approved):

packages/compass-export-to-language/src/modules/count-aggregation-stages-in-string.js
```diff
@@ -13,17 +13,33 @@ export function countAggregationStagesInString(str) {
   if (!dummySandbox) {
     dummySandbox = vm.createContext(Object.create(null), {
       codeGeneration: { strings: false, wasm: false },
-      microtaskMode: 'afterEvaluate'
+      microtaskMode: 'afterEvaluate',
     });
-    vm.runInContext([
```
lerouxb, line 18: I just noticed afterwards that here we were already using vm.runInContext directly anyway.

addaleax: … to be fair, I only used this because it was quick and reliable. This is not a good long-term solution! :)

lerouxb: Aaah now I remember what this was. OK so probably not releveant, then. Couldn't tell from the diff.

lerouxb:
From mms's perspective they can just change this line https://github.com/10gen/mms/blob/c5b032ada223f54e4eabba79a1e10d5bf6e5c5bb/client/scripts/webpack/createWebpackConfigs.js#L196 to `vm: 'vm-browserify'`

(via https://mongodb.slack.com/archives/G2L10JAV7/p1666882029693469?thread_ts=1666881513.720759&cid=G2L10JAV7)

packages/bson-transpilers/codegeneration/javascript/Visitor.js
mcasimir: I may be missing something do we already have the polyfill in this PR? Does something change for Data Explorer?

lerouxb: We don't seem to need the polyfill at this point because electron seems to make it work for us. Data Explorer will have to add it. We're testing to see if it will work.

Btw we might eventually need the polyfill as well if we make the changes to not rely on electron and use the same stuff they do.

gribnoysup:
Just checked with @lerouxb and this works with a polyfill correctly:

image

PR #3649, opened 2022-10-27
fix(settings): Make the settings modal a form modal

lerouxb:
I missed this in https://github.com/mongodb-js/compass/pull/3582

At some point that was going to be too different from FormModal and it would still be a Modal and then I addressed some feedback and didn't check it again and it turns out it is totally broken.

So this PR just makes the settings modal a FormModal. Doesn't seem to require any special overrides.

I noticed that the settings modal's sidebar scrolls with the main area, but that was already the case before so no harm done by this PR, I think.

No release notes because the thing I'm fixing hasn't been released.

Screenshot 2022-10-27 at 09 21 31

### DIFF
--- a/packages/compass-components/src/components/modals/modal-header.tsx
+++ b/packages/compass-components/src/components/modals/modal-header.tsx
@@ -77,8 +77,13 @@ function UnthemedModalHeader({
           <Icon glyph="Warning" fill={palette.red.base} role="presentation" />
         </div>
       )}
-
-      <h1 className={cx(titleStyle, darkMode && titleStyleDark)}>{title}</h1>
+      <h1
+        className={cx(titleStyle, darkMode && titleStyleDark)}
+        data-testid="modal-title"
+        id="modal-title"
+      >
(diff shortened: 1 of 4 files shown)


packages/compass-components/src/components/modals/modal-header.tsx
```diff
@@ -77,8 +77,7 @@ function UnthemedModalHeader({
           <Icon glyph="Warning" fill={palette.red.base} role="presentation" />
         </div>
       )}
-
-      <h1 className={cx(titleStyle, darkMode && titleStyleDark)}>{title}</h1>
+      <h1 className={cx(titleStyle, darkMode && titleStyleDark)} data-testid="modal-title" id="modal-title">{title}</h1>
```
lerouxb, line 80: id here is for area-labelledby in the settings sidebar. Only used there, but I think that's OK. Allows it to just be a normal form modal.

gribnoysup (approved):

PR #3651, opened 2022-10-27
feat(preview-menu): add pipeline preview options menu COMPASS-6265

mabaasit:
feat(preview-menu): add expand/collapse all documents menu on pipeline preview and pipeline results COMPASS-6265

https://user-images.githubusercontent.com/1305718/198258199-7ce95181-4cd0-45d7-979e-4a179923265f.mov

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
```diff
@@ -20,9 +22,11 @@ const containerStyles = css({
 });
 
 const previewHeaderStyles = css({
-  paddingTop: spacing[3],
-  paddingBottom: spacing[3],
-  paddingLeft: spacing[3],
+  padding: spacing[3],
+  display: 'flex',
+  flexDirection: 'row',
+  justifyContent: 'space-between',
```
gribnoysup, line 28: I will really never stop suggesting moving things to the end of the flex parent with margin instead because it's a more predictable behavior when somebody adds another element in the flex container 😄

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
```diff
@@ -38,7 +42,10 @@ const centerStyles = css({
 const messageStyles = css({ marginTop: spacing[3] });
 
 const documentListStyles = css({
-  overflow: 'scroll',
+  overflow: 'auto',
+  '.document-list': {
```
gribnoysup, line 46: Maybe let's allow to pass className to the Document list if this is required?

packages/compass-aggregations/src/components/documents-disclosure-menu.tsx
```diff
@@ -0,0 +1,58 @@
...
+}) => {
+  const [option, setOption] = useState<DocumentsDisclosureOption>("collapsed");
+  const [isMenuOpen, setIsMenuOpen] = useState(false);
+  const onMenuItemClick = useCallback(
+    (evt) => {
+      const newOption = evt.currentTarget.dataset.action as DocumentsDisclosureOption;
+      evt.stopPropagation();
+      setIsMenuOpen(false);
+      setOption(newOption);
+      onChange(newOption);
```
gribnoysup, line 24: This looks to me like 3 states (2 states and 1 prop kinda) that are all controlling the same thing, I might be missing the reason for doing thins, but it looks to me that you can change the interface to something like `{option: string, onOptionChange: (newOption) => void}` and this should be enough?

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
mabaasit: something like [this](https://github.com/mongodb-js/compass/pull/3651/files#diff-c7c6e400cf3d002df3f2b8665c059d582d1e5f5a592fb6dfa0864424c7b156aaR47)?

packages/compass-aggregations/src/components/documents-disclosure-menu.tsx
mabaasit: clean it up and passed down the selected option as a prop.

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
mabaasit: I removed this code to align document list with the PipelineOutputMenu when we have scrollbar. Not sure if we need to pass the class prop to DocumentList now

packages/compass-aggregations/src/components/pipeline-results-workspace/pipeline-results-list.tsx
```diff
@@ -16,18 +16,20 @@ const containerStyles = css({
 
 const PipelineResultsList: React.FunctionComponent<{
```
gribnoysup, line 17: Can't leave a comment there directly, but the type of `DocumentJsonView` as defined in [index.d.ts here](https://github.com/mongodb-js/compass/blob/main/packages/compass-crud/index.d.ts#L3) is not correct, can you fix this while you are working a relevant part of the code: from how it's implemented `DocumentJsonView` only allows hadron document instances to be passed

packages/compass-crud/src/components/json-editor.jsx
```diff
@@ -77,7 +77,8 @@ class EditableJson extends React.Component {
 
     if (
       prevState.editing !== this.state.editing &&
-      this.state.editing === false
+      this.state.editing === false &&
+      !this.props.isExpanded
```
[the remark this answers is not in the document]
mabaasit: Well, this `if` folds editor when user simply switches from editing to non-editing mode. And if user had expanded the editor using dropdown options, then why would we want to fold all the code?

I [removed it now](https://github.com/mongodb-js/compass/pull/3651/commits/c30ef05692476029f5407f8b5f880cbf0869b79f), but honestly I really don't know why we fold the code on this switch.

https://user-images.githubusercontent.com/1305718/199207395-0a664c72-ab48-42a6-a00c-0ed8c11073f2.mov

packages/compass-aggregations/src/components/pipeline-results-workspace/pipeline-results-list.tsx
mabaasit: Added in [5ccb375](https://github.com/mongodb-js/compass/pull/3651/commits/5ccb375f88a69caa5ab56b55b3f353c5c42e9bb9)

packages/compass-aggregations/src/components/pipeline-results-workspace/index.tsx
```diff
@@ -99,6 +99,7 @@ type PipelineResultsWorkspaceProps = {
   documents: Document[];
   isLoading?: boolean;
   isError?: boolean;
+  allDocsExpanded: boolean;
```
gribnoysup, line 102: Should this be an optional?

packages/compass-aggregations/src/components/pipeline-results-workspace/pipeline-results-header.tsx
```diff
@@ -1,44 +1,69 @@
...
+  isMergeOrOutPipeline,
+  onChangePipelineOutputOption,
+  pipelineOutputOption,
+}) => {
+  if (isMergeOrOutPipeline) {
+    return null;
+  }
+  return (
+    <div className={containerStyles} data-testid="pipeline-results-header">
+      {process?.env?.COMPASS_ENABLE_AS_TEXT_PIPELINE === 'true' ? (
```
gribnoysup, line 46: Does it need to be behind a feature flag here?

packages/compass-aggregations/src/components/pipeline-results-workspace/pipeline-results-header.tsx
```diff
@@ -1,44 +1,69 @@
...
+  onChangePipelineOutputOption: (val: PipelineOutputOption) => void;
+  pipelineOutputOption: PipelineOutputOption;
 };
 
-const controlsStyles = css({
+const containerStyles = css({
   display: 'flex',
-  gap: spacing[2],
-  justifyContent: 'flex-end',
+  justifyContent: 'space-between',
```
gribnoysup, line 23: `margin-left: auto` instead of this is exactly something that allows you not to render weird empty divs to preserve the element alignment

packages/compass-aggregations/src/components/pipeline-results-workspace/pipeline-results-header.tsx
```diff
@@ -1,44 +1,69 @@
...
+      {process?.env?.COMPASS_ENABLE_AS_TEXT_PIPELINE === 'true' ? (
+        <div className={groupStyles}>
+          <Overline>All Results</Overline>
+          <PipelineOutputOptionsMenu
+            option={pipelineOutputOption}
+            onChangeOption={onChangePipelineOutputOption}
           />
         </div>
+      ) : (
+        <div />
```
gribnoysup, line 55: The usual pattern in react for this `{condition $$ <Component />}`, you don't need to render a fallback if fallback is nothing

packages/compass-aggregations/src/components/pipeline-results-workspace/pipeline-results-header.tsx
```diff
@@ -1,44 +1,69 @@
...
-    return (
-      <div data-testid="pipeline-results-header">
-        <div className={controlsStyles}>
-          <PipelinePagination />
-          <PipelineResultsViewControls
-            value={resultsViewType}
-            onChange={onChangeResultsView}
+const groupStyles = css({
+  display: 'flex',
+  gap: spacing[2],
```
gribnoysup, line 29: You probably still want the gap in parent even if you are wrapping stuff in another container so that your menu has a spacing between it and other controls

image

packages/compass-aggregations/src/components/pipeline-results-workspace/pipeline-results-list.tsx
```diff
@@ -16,13 +16,14 @@ const containerStyles = css({
...
+}> = ({ documents, allDocsExpanded, view }) => {
+  const listProps = useMemo(
     () => ({
       docs: documents.map((doc) => new HadronDocument(doc)),
       isEditable: false,
-      copyToClipboard(doc) {
+      copyToClipboard(doc: HadronDocument) {
         const str = doc.toEJSON();
         void navigator.clipboard.writeText(str);
       },
```
gribnoysup, line 29: Actually while we are at it, what's the reason to keep these in the memo? Both are not changing at all no matter what happens to the document list, right? 🤔

gribnoysup (approved):
Left a few small comments, otherwise looks good

packages/compass-crud/src/components/json-editor.jsx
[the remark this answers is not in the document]
gribnoysup: I think the idea is that when you switch back and forth between editing we sorta want to restore the document into it's "original" state, assumption being that you probably will expand a lot while editing, but might not want this state anymore after you are done editing. But I'm kinda guessing here, this is obviously not something explicitly documented

packages/compass-aggregations/src/components/pipeline-results-workspace/pipeline-results-header.tsx
mabaasit: helped to remove the empty div. thanks!

packages/compass-aggregations/src/components/pipeline-results-workspace/pipeline-results-list.tsx
mabaasit: currently it will just avoid mapping of docs, when `isExpanded` changes. i'd leave it as is.

PR #3661, opened 2022-10-28
feat: replace react-select-plus with Combobox

mcasimir:
Replaces `react-select-plus` with the `Combobox` component:

https://user-images.githubusercontent.com/334881/198715479-dd7db2de-8db2-427d-9361-3d4066cb255a.mp4

https://user-images.githubusercontent.com/334881/198715985-578eba22-a272-4ebd-8633-d31280ebd657.mp4

I've accidentally replaced the select on the table view as well as initially I thought that was the library that is not compatible with react 17 

Screenshot 2022-10-28 at 21 07 01


packages/compass/src/app/styles/deprecated-bootstrap-styles/globals.less
```diff
@@ -21,7 +21,7 @@
   .checkbox input[type="checkbox"] {
     position: absolute;
     margin-left: -20px;
-    margin-top: 4px \\9;
+    margin-top: 4px;
```
mcasimir, line 24: drive-by fix

before:
Screenshot 2022-10-28 at 20 53 44

after:
Screenshot 2022-10-28 at 20 53 14

packages/compass-export-to-language/package.json
```diff
@@ -54,6 +54,7 @@
     "reformat": "npm run prettier -- --write ."
   },
   "peerDependencies": {
+    "@mongodb-js/compass-components": "*",
```
gribnoysup, line 57: Star versions lead to weird behavior with npm@8, I cleaned it up awhile ago, but seems like this is getting introduced again. Can you clean these up please in this package while you are adding new dependencies here?

mcasimir: Sure!

Anemy: If we find this happens often it may be worth adding something to our `depalign` script.

packages/compass/src/main/window-manager.ts
```diff
@@ -43,7 +43,7 @@ const DEFAULT_HEIGHT = (() => {
   return height;
 })();
 
-const MIN_WIDTH = process.env.COMPASS_MIN_WIDTH ?? 1024;
+const MIN_WIDTH = process.env.COMPASS_MIN_WIDTH ?? 1025;
```
Anemy, line 46: How do you feel about leaving a comment about why we have the min width set at that amount?
```suggestion
// We set the min width to 1025 so that the screensize breakpoints of leafygreen
// components are not hit. The breakpoints make the styles of the Select component
// change significantly at widths of 1024 and less.
const MIN_WIDTH = process.env.COMPASS_MIN_WIDTH ?? 1025;
```

packages/compass/src/app/styles/index.less
```diff
@@ -12,7 +12,6 @@
 @import './deprecated-bootstrap-styles/grid.less';
 
 // Dependencies
-@import './react-select.less';
```
Anemy, line 15: Can we also remove this file now? Or are we going to clean up all of the files later on?

packages/compass-crud/src/components/table-view/types-dropdown.jsx
```diff
@@ -0,0 +1,44 @@
+import React, { useCallback } from 'react';
+import PropTypes from 'prop-types';
+import TypeChecker from 'hadron-type-checker';
+import { Select, Option, css, spacing } from '@mongodb-js/compass-components';
+
+const CASTABLE_TYPES = TypeChecker.castableTypes(true);
+
+const selectStyles = css({ minWidth: spacing[3] * 10 });
+
+const TypesDropdown = ({ element }) => {
```
Anemy, line 10: Is there a reason we don't want this file in typescript? `compass-crud` is on the new config. We might be able to use some of the types from `hadron-document` as well.  Same q with the `types-dropdown.spec.jsx`

packages/compass-aggregations/src/components/stage-editor-toolbar/stage-operator-select.jsx
```diff
@@ -1,61 +1,82 @@
...
-          value={this.props.selectedStage}
-          onChange={this.onStageOperatorSelected}
-        />
-      </div>
-    );
-  }
+function comboboxOptionStyles(stage) {
+  return css({
+    '&::after': {
+      content: JSON.stringify(
```
Anemy, line 26: Just a thought, it might be worth leaving a comment why we show the combobox description inside of css. Nice workaround btw!

Anemy (approved):
lgtm! A couple ideas/suggestions, no blockers.

packages/compass-aggregations/src/components/stage-editor-toolbar/stage-operator-select.jsx
```diff
@@ -1,61 +1,82 @@
...
+}) => {
+  const onStageOperatorSelected = useCallback((name) => {
+    onChange(index, name);
+  }, [onChange, index]);
+
+  const optionStyleByStageName = useMemo(() => {
+    return Object.fromEntries(stages.map((stage) => [stage.name, comboboxOptionStyles(stage)]))
+  }, [stages])
+
+  return <Combobox value={selectedStage}
```
Anemy, line 55: I think it's fine as is, however if we want to make this scroll a bit nicer with the aggregations content we could pass a `scrollContainer` reference to the aggregation pipeline window.
https://www.mongodb.design/component/combobox/documentation/

PR #3673, opened 2022-10-31
chore(connections): remove logo on connections page sidebar COMPASS-6265

Anemy:
COMPASS-6265

We're removing or updating this as the previous one wasn't okay with brand.

Also un-skips a skipped test.

| current ga | before | after |
| - | - | - |
| Screen Shot 2022-10-31 at 4 41 25 PM | Screen Shot 2022-10-31 at 4 46 23 PM | Screen Shot 2022-10-31 at 4 42 20 PM |

We had a sketch from Claudia on what it might look like:
https://www.figma.com/file/vbjoD4dcisPYRYJ5RRu72k/Compass-LG?node-id=4053%3A78365
Since she's out this week Maurizio and I weren't sure if it should be added yet. We were leaning towards removing it for now and then adding it back in when she's back (maybe next week before the ga).


gribnoysup (approved):

packages/compass-e2e-tests/helpers/commands/select-connection-menu-item.ts
```diff
@@ -11,7 +11,7 @@ export async function selectConnectionMenuItem(
   await browser.$(selector).waitForDisplayed();
 
   // workaround for weirdness in the ItemActionControls menu opener icon
-  await browser.clickVisible(Selectors.ConnectionsTitle);
+  await browser.clickVisible(Selectors.SidebarNewConnectionButton);
```
lerouxb, line 14: This has a side-effect that wasn't there before. Clicking the ConnectionsTitle did nothing other than move focus somewhere else. (was to give hover another chance of maybe doing something on windows, if I remember correctly.)

Clicking new connection clears the connect form. Maybe nothing that calls this command cares. I don't know.

lerouxb:
Is there any plan for bringing back some logo? Looks a bit weird now.

packages/compass-e2e-tests/helpers/commands/select-connection-menu-item.ts
Anemy: Gotcha, makes sense. I'll make it click something that isn't interactive 👍

addaleax: Is this solving the same problem that https://github.com/mongodb-js/compass/pull/3660/files#diff-93fd67837e378b3f719a92f9fde522278291f129ab8fc56ace9b36d514cbba0aR19 is aiming to solve?

Anemy:
Closing and updating to the one from newest designs in: https://github.com/mongodb-js/compass/pull/3686
Planning to sync when Claudia is back before releasing GA.

PR #3675, opened 2022-10-31
chore(import-export): use RadioBoxGroup for import/export file type selection

Anemy:
Part of COMPASS-6220

We'll wait to merge this until we've updated the `Modal` to LeafyGreen's `Modal` as this one makes the new padding from the `FormFieldContainer` look off in my opinion.

I'm also thinking we should reword `Select export file type` to `Export file type` as the select is implicit.

| before | after |
| - | - |
| Screen Shot 2022-10-31 at 6 16 37 PM | Screen Shot 2022-10-31 at 6 14 50 PM |


packages/compass-import-export/src/components/select-file-type.tsx
```diff
@@ -0,0 +1,60 @@
...
+          data-testid="select-file-type-json"
+          value={FILE_TYPES.JSON}
+          checked={fileType === FILE_TYPES.JSON}
+        >
+          JSON
+        </RadioBox>
+        <RadioBox
+          data-testid="select-file-type-csv"
+          value={FILE_TYPES.CSV}
+          checked={fileType === FILE_TYPES.CSV}
```
addaleax, line 51: Since we’re already using string unions for the component’s types, do we also want to update these to use `value="csv"`, `checked={fileType === 'csv'}` etc. rather than the enums?

Anemy: I like it! Updated.

addaleax (approved):

PR #3681, opened 2022-11-01
feat(pipeline-preview): handle output stages COMPASS-6257

mabaasit:
feat(pipeline-preview): handle output stages COMPASS-6257

Show preview banner for output stages (`$out` and `$merge`). 

Pipeline that has an output stage, we will show a stage specific info banner in preview section along with pipeline preview (for all pipeline except for output stage, which is always last in pipeline).

For atlas, the info banner has an action to run the pipeline as well.

<details>
<summary>Preview</summary>

https://user-images.githubusercontent.com/1305718/199228452-c4a68ee4-e21d-4486-91eb-d52b0c3b588f.mov

https://user-images.githubusercontent.com/1305718/199228425-4c47817f-9d89-4674-b026-8a708e1b2184.mov

</details>

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/utils/stage.js
```diff
@@ -86,20 +95,20 @@ export function getStageOperator(stage) {
...
   const { database } = toNS(namespace);
   if (stageOperator === '$merge') {
-    const ns = typeof stage === 'string' ? stageValue : stageValue.into;
-    if (ns.atlas) {
+    if (stageValue.atlas) {
       // TODO: Not handled currently and we need some time to figure out how to
       // handle it so just skipping for now
       return null;
     }
+    const ns = typeof stageValue === 'string' ? stageValue : stageValue.into;
```
gribnoysup, line 111: I am not sure why this was moved down. From the documentation, `atlas` is the key on `into` property, not on the stage description itself, [see the documentation linked in the method description](https://www.mongodb.com/docs/atlas/data-federation/supported-unsupported/pipeline/merge/#syntax)

packages/compass-aggregations/src/modules/pipeline-builder/index.ts
```diff
@@ -2,11 +2,13 @@ import { combineReducers } from 'redux';
 import stageEditor from './stage-editor';
 import textEditor from './text-editor';
 import pipelineMode from './pipeline-mode';
+import outputStage from './text-editor-output-stage';
 
 const reducer = combineReducers({
   pipelineMode,
   stageEditor,
   textEditor,
+  outputStage,
```
gribnoysup, line 11: Feel free to keep as-is, but from my perspective putting it outside of text editor even when text editor is the only user of this slice makes it a bit confusing: makes it look like something else might rely on this state, like stage editor, where in reality it is not

packages/compass-aggregations/src/modules/pipeline-builder/builder-helpers.ts
```diff
@@ -22,7 +23,7 @@ export function getStagesFromBuilderState(
   pipelineBuilder: PipelineBuilder
 ) {
   if (state.pipelineBuilder.pipelineMode === 'builder-ui') {
-    return pipelineBuilder.stages.map((stage) => stage.toBSON());
+    return pipelineBuilder.stages.map((stage) => stage.toBSON() as Document);
```
gribnoysup, line 26: Why are you asserting here? Return type of toBSON is Document or null, not just Document

packages/compass-aggregations/src/modules/pipeline-builder/text-editor-output-stage.ts
```diff
@@ -0,0 +1,167 @@
...
+export type OutputStageState = {
+  isLoading: boolean,
+  serverError: MongoServerError | null,
+  isComplete: boolean;
+};
+
+const INITIAL_STATE: OutputStageState = {
+  isLoading: false,
+  isComplete: false,
+  serverError: null,
```
gribnoysup, line 43: Based on usage this is the same as serverError in text editor slice, why are we keeping this one separate?

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-preview-manager.ts
```diff
@@ -30,21 +30,32 @@ export interface PreviewOptions extends AggregateOptions {
...
     PreviewOptions,
-    'sampleSize' | 'previewSize' | 'totalDocumentCount'
+    'sampleSize' | 'previewSize' | 'totalDocumentCount' | 'filterOutputStage'
   > = {}
 ) {
+  const _pipeline = [...pipeline];
+  if (options.filterOutputStage) {
+    const isOutput = isOutputStage(getLastStageOperator(_pipeline));
+    isOutput && _pipeline.splice(_pipeline.length - 1, 1);
+  }
```
gribnoysup, line 51: I don't think this logic belongs in the preview manager, this is only used by text editor view and so should be handled there explicitly

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-preview-manager.ts
```diff
@@ -67,7 +78,7 @@ export function createPreviewAggregation(
 
 export class PipelinePreviewManager {
   private queue = new Map<number, AbortController>();
-  constructor(private dataService: DataService) {}
+  constructor(private dataService: DataService) { }
```
gribnoysup, line 81:
```suggestion
  constructor(private dataService: DataService) {}
```

packages/compass-aggregations/src/modules/pipeline-builder/index.ts
mabaasit: I started that way, but then separated the two. I did not want to mix preview pipeline and actually running of pipeline. 

With another point you raised regarding `serverError` prop,  I kinda feel like i should put this in `textEditor` slice.

packages/compass-aggregations/src/utils/stage.js
mabaasit: ah damn, i was following [docs on aggregation page](https://www.mongodb.com/docs/manual/reference/operator/aggregation/merge/).

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-preview-manager.ts
mabaasit: I kinda disagree here. Even though its only used in `textEditor`, i think it belongs in preview manager. 

For the output stages, preview is not an option (at least yet). We should instead always filter the last stage if it's an  output stage.

packages/compass-aggregations/src/modules/pipeline-builder/index.ts
gribnoysup: Hmmm, if you want to keep them split, maybe would make sense to restructure stuff a little and make text-editor a combined reducer itself? Something like `{ textEditor: { pipeline: <pipeline editor state>, output: <preview and out / merge state> } }` maybe?

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-preview-manager.ts
gribnoysup: I completely agree that previews are not (and can not be) supported by preview manager, but FWIW if we want to make it explicit that preview manager doesn't support out stages we should throw in that case, not implement the logic for removing the stage before getting a preview

packages/compass-aggregations/src/utils/stage.js
mabaasit: fixed.

packages/compass-aggregations/src/modules/pipeline-builder/builder-helpers.ts
mabaasit: cleaned

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-preview-manager.ts
mabaasit: Implemented [a98e02e](https://github.com/mongodb-js/compass/pull/3681/commits/a98e02ec84825495a1af5fccd92485b45f98ae26)

*Note:* we agreed on throwing error from the preview manager if the last stage of pipeline is an output stage. and the filtering of pipeline to remove last output stage will take place in `PipelineBuilder.getPreviewForPipeline`

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-builder.ts
```diff
@@ -120,7 +121,11 @@ export class PipelineBuilder {
     namespace: string,
     options: PreviewOptions
   ): Promise<Document[]> {
-    const pipeline = this.getPipelineFromSource();
+    // For preview we ignore $out/$merge stage.
+    const pipeline = [...this.getPipelineFromSource()];
+    if (isLastStageOutputStage(pipeline)) {
+      pipeline.splice(pipeline.length - 1, 1);
+    }
```
gribnoysup, line 126: I woulds still put it behind an option: we need to be explicit that this happens because it's still only text editor that should do this, if this somehow starts affecting stage editor, this should throw

packages/compass-aggregations/src/modules/pipeline-builder/text-editor-pipeline.ts
```diff
@@ -50,7 +49,7 @@ export type TextEditorState = {
   stageOperators: string[];
   syntaxErrors: PipelineParserError[];
   serverError: MongoServerError | null;
-  loading: boolean;
+  isLoading: boolean;
   previewDocs: Document[] | null;
```
gribnoysup, line 53: Feel free to keep as is, but that's not really what I suggested, what confuses me about this store structure is that you have now two slices that are running actions and keeping state for the same UI which makes it harder to reason about (e.g., what does it mean if both loading flags are true? what does it mean if both completed flags return true? both error? what if there is still an error in this slice but not one in another, what if it's the other way around?). You have a clear piece of the UI that is dedicated to output, it can be either preview results or output result, it can't be both, why not move all relevant state and actions into this new slice you added if you want to keep them separate?

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
```diff
@@ -115,17 +125,24 @@ export const PipelinePreview: React.FunctionComponent<PipelinePreviewProps> = ({
         {shouldShowCount && <Body>{`Sample of ${docCount} ${docText}`}</Body>}
       </div>
       <PreviewResults isLoading={isLoading} previewDocs={previewDocs} />
```
gribnoysup, line 127: This seems to have some weird padding at the bottom, would be nice not to have it:

image

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
```diff
@@ -115,17 +125,24 @@ export const PipelinePreview: React.FunctionComponent<PipelinePreviewProps> = ({
         {shouldShowCount && <Body>{`Sample of ${docCount} ${docText}`}</Body>}
```
gribnoysup, line 125:
```suggestion
        {shouldShowCount && <Body>Sample of ${docCount} ${docText}</Body>}
```

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
```diff
@@ -115,17 +125,24 @@ export const PipelinePreview: React.FunctionComponent<PipelinePreviewProps> = ({
...
       </div>
       <PreviewResults isLoading={isLoading} previewDocs={previewDocs} />
+      <div className={outputStageStyles} data-testid="output-stage-preview">
+        {isOutStage && <OutputStageBanner stageOperator="$out" />}
+        {isMergeStage && <OutputStageBanner stageOperator="$merge" />}
```
gribnoysup, line 130: Tiny nit: looks kinda weird that it's two separate props, it can be either one or another, right? Maybe something like `outStage: '$out' | '$merge' | null` will be a better interface here?

gribnoysup (approved):
Overall seems to work fine. I left a comment about store structure: from my perspective it doesn't make sense to have multiple slices for the state that can be only one thing or another, it makes it hard to change / inspect / reason about, but I'll leave it to you if you want to address that or not

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-builder.ts
```diff
@@ -118,7 +119,11 @@ export class PipelineBuilder {
     namespace: string,
     options: PreviewOptions
   ): Promise<Document[]> {
-    const pipeline = this.getPipelineFromSource();
+    // For preview we ignore $out/$merge stage.
+    const pipeline = [...this.getPipelineFromSource()];
+    if (isLastStageOutputStage(pipeline)) {
+      pipeline.splice(pipeline.length - 1, 1);
```
gribnoysup, line 125: This can be just a pop
```suggestion
      pipeline.pop();
```

packages/compass-aggregations/src/modules/pipeline-builder/text-editor-pipeline.ts
mabaasit: maybe i am missing something, but the two slices are completely independent. have their own state, which is consumed separately on UI (except for showing server error).

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
mabaasit: I'll be handling that in the next PR to handle Atlas specific operators. If i make changes here, it may create conflicts with my other PR, as I'll have to add classname prop to DocumentList view.

packages/compass-aggregations/src/modules/pipeline-builder/text-editor-pipeline.ts
gribnoysup: They could be independent, but this is not how UI code is written. Here's an example of the broken state:

image

PR #3688, opened 2022-11-02
feat(export-to-language): replace export-to-language with leafygreen components COMPASS-6219

lerouxb:
Here's how it looks so far:
Screenshot 2022-11-04 at 09 51 51


packages/compass-export-to-language/src/typings.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module 'bson-transpilers';
+declare module 'js-beautify';
```
Anemy, line 2: Looks like there's a typings library for this we might be able to use:
https://www.npmjs.com/package/@types/js-beautify

mcasimir (commented):
Can we move this change here? https://github.com/mongodb-js/compass/pull/3661/files#diff-3b7fb2df98fc806098a80cbedeab1add744c87ca6481aa52dce4a35b935af81aR57-R63

i'm going to close the other PR, thanks!

packages/compass-e2e-tests/tests/collection-documents-tab.test.ts
```diff
@@ -297,7 +297,7 @@ describe('Collection documents tab', function () {
       useBuilders: true,
     });
 
-    expect(text).to.equal(`import static com.mongodb.client.model.Filters.eq;
+    expect(text).to.equal(`import org.bson.Document;
```
lerouxb, line 300: This was actually a bug before. The UI code and runTranspiler had  two different copies of the logic around whether the useBuilders checkbox should be there and taken into account. So we displayed the use builders syntax checkbox and the test clicked it, but the code that was generated didn't use builders syntax.

packages/compass-export-to-language/src/modules/transpiler.ts
```diff
@@ -0,0 +1,104 @@
...
+        : inputExpression.filter;
+    output = compiler.shell[outputLanguage].compile(
+      toCompile,
+      useBuilders,
+      false
+    );
+  }
+
+  if (includeImports) {
+    // bson-transpilers appears to be stateful, so getImports() has to be called after compile()
```
lerouxb, line 97: ☝️

packages/compass-export-to-language/src/components/modal.tsx
```diff
@@ -0,0 +1,287 @@
...
+} from '../modules/languages';
+import type { OutputLanguage } from '../modules/languages';
+import { getInputExpressionMode, runTranspiler } from '../modules/transpiler';
+import type { InputExpression } from '../modules/transpiler';
+
+import { createLoggerAndTelemetry } from '@mongodb-js/compass-logging';
+import { countAggregationStagesInString } from '../modules/count-aggregation-stages-in-string';
+const { track } = createLoggerAndTelemetry('COMPASS-EXPORT-TO-LANGUAGE-UI');
+
+// TODO: bring back the tracking that the old export modal had around opening the modal and copying the text
```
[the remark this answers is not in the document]
lerouxb: I just forgot to remove the TODO. Since the old tracking had more than just trackId I just brought back the old code in a useEffect.

packages/compass-export-to-language/src/components/modal.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+const codeStyles = css({
+  alignItems: 'start',
+  height: `${spacing[6] * 4 - spacing[5]}px`,
+});
+
+const checkboxStyles = css({
+  marginTop: spacing[2],
+  // prevent an extra couple of pixels that always causes the modal to scroll unnecessarily
+  overflow: 'hidden',
```
Anemy, line 81:
```suggestion
  width: 'auto',
```
Looks like this unintentionally cuts off the focus state view of the checkbox:
Screen Shot 2022-11-07 at 12 53 03 PM
Using `width: auto` overrides LeafyGreen's width while keeping things in view:
Screen Shot 2022-11-07 at 12 53 56 PM

Anemy (commented):
A couple logs/todos I left questions/comments on. Tried it out, works smooth and looks nice. Nice typescripts. lgtm

Not related, I was thinking functionality wise how you felt about if by default the exported code would `protectConnectionStrings`=`true` and not show the password. When the preference isn't set to true the modal would have a checkbox to show it or hide it. Not for this, only curious if you feel that would be helpful or extra.

packages/compass-export-to-language/src/components/modal.tsx
lerouxb: Thanks!

packages/compass-export-to-language/src/stores/store.js
```diff
@@ -96,10 +79,6 @@ const configureStore = (options = {}) => {
     setNamespace(store, options.namespace);
   }
 
-  if (options.copyToClipboardFn) {
-    store.dispatch(copyToClipboardFnChanged(options.copyToClipboardFn));
```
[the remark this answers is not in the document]
lerouxb: https://mongodb.slack.com/archives/G2L10JAV7/p1667900171512839?thread_ts=1667899974.480019&cid=G2L10JAV7

Anemy (approved):
lgtm!
Looks good.

PR #3691, opened 2022-11-02
feat: use leafygreen combobox to select stages

mcasimir:
Extracted from https://github.com/mongodb-js/compass/pull/3661


packages/compass-aggregations/src/components/stage-editor-toolbar/stage-operator-select.jsx
```diff
@@ -1,61 +1,88 @@
...
+}) => {
+  const onStageOperatorSelected = useCallback((name) => {
+    onChange(index, name);
+  }, [onChange, index]);
+
+  const optionStyleByStageName = useMemo(() => {
+    return Object.fromEntries(stages.map((stage) => [stage.name, comboboxOptionStyles(stage)]))
+  }, [stages])
+
+  return <Combobox value={selectedStage}
```
Anemy, line 55: Fine as is, however if we want to make this scroll a bit nicer with the aggregations content we could pass a scrollContainer reference to the aggregation pipeline window.
https://www.mongodb.design/component/combobox/documentation/

I was thinking a `ScrollContainerProvider` with a scroll container in the context might be a useful thing for places like this, and with the menus for editable documents. I'll probably add it for https://jira.mongodb.org/browse/COMPASS-6271 so maybe it can get some use there.

Anemy (approved):
One idea on scroll container, not blocker. One thing we also might be able to do is update the `stage-operator-select` file to typescript as most of it is being updated.
 lgtm!

packages/compass-e2e-tests/helpers/commands/select-stage-operator.ts
```diff
@@ -1,32 +1,40 @@
...
-  // the "select" should now blur and the ace textarea become focused
   await browser.waitUntil(async () => {
     const textareaElement = await browser.$(textareaSelector);
     const isFocused = await textareaElement.isFocused();
     return isFocused === true;
   });
+
+  // NOTE: even if the focus moved away from the combobox interactions are still
+  // targeting the combobox
+  await delay(1000);
```
mcasimir, line 39: @lerouxb any suggestion to avoid this here?

lerouxb: wait for something (like the options popover thingy) to not be displayed or exist (see our usage of reverse: true) or wait for something to be not focused in a waitUntil?

lerouxb: I don't think this is used in the time to first query test, but anywhere it comes up where we might measure the time these delays will skew results. sometimes it is pretty hard to avoid, though.

mcasimir: is so strange, it is like there is some event listener still attached for a while after the dropdown is hidden, I'm already waiting for the focus to go somewhere else and for the dropdown to be hidden, no idea what else I can wait for, I'll check the `{reverse: true}`

lerouxb: waitForExist({ reverse: true }) on something, maybe?

PR #3692, opened 2022-11-02
feat: replace types dropdown with LG select

mcasimir:
Extracted from https://github.com/mongodb-js/compass/pull/3661


packages/compass/src/main/window-manager.ts
```diff
@@ -43,7 +43,7 @@ const DEFAULT_HEIGHT = (() => {
   return height;
 })();
 
-const MIN_WIDTH = process.env.COMPASS_MIN_WIDTH ?? 1024;
+const MIN_WIDTH = process.env.COMPASS_MIN_WIDTH ?? 1025;
```
Anemy, line 46: Maybe a comment on why we have the min width set at that amount?
```suggestion
// We set the min width to 1025 so that the screensize breakpoints of leafygreen
// components are not hit. The breakpoints make the styles of the Select component
// change significantly at widths of 1024 and less.
const MIN_WIDTH = process.env.COMPASS_MIN_WIDTH ?? 1025;
```

packages/compass-crud/src/components/table-view/types-dropdown.jsx
```diff
@@ -0,0 +1,44 @@
+import React, { useCallback } from 'react';
+import PropTypes from 'prop-types';
+import TypeChecker from 'hadron-type-checker';
+import { Select, Option, css, spacing } from '@mongodb-js/compass-components';
+
+const CASTABLE_TYPES = TypeChecker.castableTypes(true);
+
+const selectStyles = css({ minWidth: spacing[3] * 10 });
+
+const TypesDropdown = ({ element }) => {
```
Anemy, line 10: Is there a reason we don't want this file in typescript? compass-crud is on the new config. We might be able to use some of the types from hadron-document as well. Same q with the types-dropdown.spec.jsx

Anemy (approved):
Nice, copied the two comments from the last pr over. Looks good, not blockers, ideas/suggestions.

packages/compass-crud/src/components/table-view/types-dropdown.jsx
mcasimir: I'll give it a try here seems quick enough. let's remind though that we listed this specifically as a non-goal in the scope cause in this case it adds an effort that is not conducive to the other goals

PR #3699, opened 2022-11-03
feat(pipeline-text): preview atlas stage COMPASS-4763

mabaasit:
add support for atlas only operators in pipeline text mode.

<details>
<summary>Preview</summary>

With Atlas Operator
image

With Atlas and Output Operator
image

</details>

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-preview.tsx
```diff
@@ -167,12 +190,20 @@ export const PipelinePreview: React.FunctionComponent<PipelinePreviewProps> = ({
...
   const pipeline = getPipelineStageOperatorsFromBuilderState(state);
   const lastStage = pipeline[pipeline.length - 1] ?? '';
-  const { isLoading, previewDocs } = state.pipelineBuilder.textEditor.pipeline;
+  const { isLoading, previewDocs, serverError, stageOperators } =
+    state.pipelineBuilder.textEditor.pipeline;
+  const isMissingAtlasSupport = isMissingAtlasStageSupport(
+    state.env,
+    serverError
+  );
+  const atlasOperator = getAtlasOnlyOperators(stageOperators)[0] ?? '';
```
gribnoysup, line 199: Is it a safe assumption that the first atlas-only stage is the one that cause the error? Maybe there is a way to get this from error somehow? 🤔

packages/compass-aggregations/src/utils/stage.js
```diff
@@ -190,4 +196,29 @@ export function getStageInfo(namespace, stageOperator, stageValue) {
...
+      31082,
+    ].includes(Number(serverError.code))
+  );
+};
+
+/**
+ * Return a list of atlas only operators
+ * @param {string[]} operators 
+ */
+export const getAtlasOnlyOperators = (operators) => {
```
gribnoysup, line 222: Based on how you are using this wouldn't it make more sense to make this a `find` instead of `filter`?

gribnoysup (approved):

mcasimir (commented):
Awesome stuff! few little things that we may want to discuss with design: 
- it may be nicer to also add an explanation for that screen, especially if someone would paste an entire pipeline, looks like a sort of an Ad, may be better to give some context: "Unable to display the preview: ..."
- The text says "This stage", which made sense in the stage-by-stage screen, in this case it would be better to have it say "The $searchMeta stage ..."
- That logo seems deprecated at this point, we removed it from the stage select, and we may need an alternative

gribnoysup:
> it may be nicer to also add an explanation for that screen, especially if someone would paste an entire pipeline, looks like a sort of an Ad, may be better to give some context: "Unable to display the preview: ..."

Similar to the stages editor, we are still showing an error when the atlas cta is shown, it's the same behavior as for stages editor:

image

gribnoysup:
> That logo seems deprecated at this point, we removed it from the stage select, and we may need an alternative

We discussed this when adding the initial CTA screen, that's the only one we have. Do we need to decide what to do with this in this scope?

PR #3701, opened 2022-11-03
chore: reduce height of collection tab, reduce collection header size

Anemy:
Quick poc for giving more room back to data. There are other things, I figure this probably helps folks though. I don't see the tab name of where people are being all that useful in the tab title so that is removed here.

| before | after |
| - | - |
| Screen Shot 2022-11-03 at 10 07 32 AM | Screen Shot 2022-11-03 at 10 05 25 AM |
| Screen Shot 2022-11-03 at 10 07 24 AM | Screen Shot 2022-11-03 at 10 05 45 AM |


addaleax (approved):

Anemy:
Closing this as we've made some progress in reducing the collection header size with the tab changes. Still a bit we can do, but this pr is beyond stale at this point.

PR #3704, opened 2022-11-04
feat: add connection import/export UI

addaleax:
Working on adding tests at this point, but this should be ready for playing around with the UI if you like.

https://user-images.githubusercontent.com/899444/200596888-5588e57d-e82a-497c-9d96-c970041c0ee3.mov

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-connection-import-export/src/hooks/use-import.ts
```diff
@@ -0,0 +1,156 @@
...
+          };
+        });
+        return;
+      }
+      finish('succeeded');
+    })();
+  }, [connectionList, fileContents, passphrase, finish]);
+
+  useEffect(() => {
+    void loadFile({ filename, passphrase }, importConnections).then(
```
addaleax, line 123: This probably needs to be debounced and/or `importConnections` needs to be extended with the ability to take an `AbortSignal` or similar to cancel in-flight operations

packages/compass-connection-import-export/src/components/file-input.tsx
```diff
@@ -0,0 +1,51 @@
...
+    });
+  }, [mode]);
+
+  return (
+    <CompassFileInput
+      disabled={disabled}
+      label={label}
+      onChange={onChangeFiles}
+      id="conn-import-export-file-input"
+      accept=".json"
```
gribnoysup, line 45: Not sure how this is working internally / what's causing this, but seems like I am not able to select any file on import: 

image

packages/compass-connections/src/components/connection-list/connection-list.tsx
```diff
@@ -141,11 +147,31 @@ function ConnectionList({
...
+      action: 'import-favorites',
+      label: 'Import saved connections',
+      icon: 'Download',
+    });
+
+    actions.push({
+      action: 'export-favorites',
+      label: 'Export saved connections',
+      icon: 'Export',
+    });
```
gribnoysup, line 170: Any reason we are pushing into the array here in a memo inside the render method? These seem to be constant, this can be defined outside of the component I think, unless I'm missing something

packages/compass-connection-import-export/src/components/export-modal.tsx
```diff
@@ -0,0 +1,135 @@
...
+import { Passphrase } from './passphrase';
+import { SelectTable } from './select-table';
+import type { ImportExportResult } from '../hooks/common';
+import { useExportConnections } from '../hooks/use-export';
+import type { ConnectionInfo } from 'mongodb-data-service';
+import { usePreference } from 'compass-preferences-model';
+
+const TOAST_TIMEOUT_MS = 5000;
+const tableStyles = css({
+  maxHeight: '24vh',
```
gribnoysup, line 21: Two small comments here, feel free to ignore.

This `maxHeight` value makes it so that on the default compass screen size you start already with two scroll bars in the modal:

image


Maybe it's worth reducing it a little so that there is only table that is scrollable in this case?

I'd maybe also suggest dropping these `FormFieldContainer`s here (due to their margin that you can't disable), changing this to `minHeight` instead and making parent container a grid with the following styles:

```css
display: grid;
grid-template-columns: 100%;
grid-template-rows: 1fr repeat(3, max-content);
gap: 24px;
```

This makes this a grid container where table will take as much space as possible (but not less than the minHeight here) and lead to less scroll on bigger screens:

|image|image|
|---|---|

packages/compass-connections/src/components/connection-list/connection-list.tsx
addaleax: Just the fact that I lazily copied this from `connections.tsx` (and I imagine there it was copied from other places where a prop _did_ influence the action list). :slightly_smiling_face: I took the liberty to apply your suggestion in both places, totally makes sense.

packages/compass-connection-import-export/src/components/file-input.tsx
gribnoysup: Aha, good to know! Kinda wish we could just use browser api here, the interface there seems a bit more straightforward to me

addaleax: I mean, yeah, it’s just that there is no browser API that covers what we need here :confused:

gribnoysup (approved):
Other than a few questions / suggestions I left, looks good to me and works well after the fix for the import

packages/compass-connection-import-export/src/components/export-modal.tsx
addaleax: Yeah, that also seems like what we’d ideally do, I couldn’t get this to work just by following the “instructions” here – if I only change `maxHeight` to `minHeight`, wrap the modal contents in a div with those css properties and remove the FormFieldContainers, it just results in the table taking up its full height (or `maxHeight` if I add one) – this is most likely just because I misunderstood something here, though, because I believe that your screenshots are real :sweat_smile: 

<details><summary>(diff on top of this PR + screenshot)</summary>

```diff
diff --git a/packages/compass-connection-import-export/src/components/export-modal.tsx b/packages/compass-connection-import-export/src/components/export-modal.tsx
index 195c2487b975..284ba2f7a57a 100644
--- a/packages/compass-connection-import-export/src/components/export-modal.tsx
+++ b/packages/compass-connection-import-export/src/components/export-modal.tsx
@@ -3,8 +3,8 @@ import {
   Banner,
   Checkbox,
   css,
-  FormFieldContainer,
   FormModal,
+  spacing,
   ToastVariant,
   useToast,
 } from '@mongodb-js/compass-components';
@@ -18,10 +18,17 @@ import { usePreference } from 'compass-preferences-model';
 
 const TOAST_TIMEOUT_MS = 5000;
 const tableStyles = css({
-  maxHeight: '24vh',
+  minHeight: '20vh',
   overflow: 'auto',
 });
 
+const modalContainerStyles = css({
+  display: 'grid',
+  gridTemplateColumns: '100%',
+  gridTemplateRows: '1fr repeat(3, max-content)',
+  gap: spacing[4]
+})
+
 export function ExportConnectionsModal({
   open,
   setOpen,
@@ -93,6 +100,7 @@ export function ExportConnectionsModal({
       }
       data-testid="connection-export-modal"
     >
+      <div className={modalContainerStyles}>
       <SelectTable
         className={tableStyles}
         items={connectionList}
@@ -100,7 +108,6 @@ export function ExportConnectionsModal({
         disabled={inProgress}
         onChange={onChangeConnectionList}
       />
-      <FormFieldContainer>
         <FileInput
           label="Target File"
           mode="save"
@@ -108,8 +115,6 @@ export function ExportConnectionsModal({
           onChange={onChangeFilename}
           value={filename}
         />
-      </FormFieldContainer>
-      <FormFieldContainer>
         <Checkbox
           checked={removeSecrets}
           disabled={inProgress || protectConnectionStrings}
@@ -118,8 +123,6 @@ export function ExportConnectionsModal({
           onChange={onChangeRemoveSecrets}
           data-testid="connection-export-remove-secrets"
         />
-      </FormFieldContainer>
-      <FormFieldContainer>
         <Passphrase
           label="Encryption Password"
           description="Optional passphrase to encrypt secrets such as passwords and access tokens"
@@ -128,8 +131,8 @@ export function ExportConnectionsModal({
           onChange={onChangePassphrase}
           value={passphrase}
         />
-      </FormFieldContainer>
       {error && <Banner variant="danger">Error: {error}</Banner>}
+      </div>
     </FormModal>
   );
 }
```

Screenshot from 2022-11-08 17-34-02


</details>

packages/compass-connections/src/components/connection-list/connection-list.tsx
```diff
@@ -180,6 +210,13 @@ function ConnectionList({
           >
             Saved connections
           </H3>
+          <ItemActionControls<FavoriteAction>
```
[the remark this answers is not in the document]
addaleax: sgtm! done in e1a4f7cc1805ac90e2b7e3716946e1cb716dab03

Now that I’m thinking about it, I’m not sure if the idea really was to only show the menu opener button on hover… it would feel pretty inconsistent to me to not follow what we do for the rest of the connections list sidebar, though

packages/compass-connection-import-export/src/hooks/use-import.ts
```diff
@@ -0,0 +1,166 @@
...
+    const fileContents = await fs.readFile(filename, 'utf8');
+    const connectionList: ConnectionShortInfo[] = [];
+    await importConnections(fileContents, {
+      passphrase,
+      saveConnections(list: ConnectionInfo[]) {
+        for (const info of list) {
+          if (info.favorite?.name) {
+            connectionList.push({
+              name: info.favorite.name,
+              id: info.id,
```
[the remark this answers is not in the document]
addaleax: @gribnoysup No yeah, totally valid concern. 69207d56f4216c37d744dba2a89a0fa8343f384e is what I’ve done for now, I’m not sure if it overloads the UI too much, but this is what it looks like right now:

- Existing connections are deselected by default
- Existing connections are identified with a badge, and if that connection is selected, that badge uses the warning variant
- If any existing connection is selected for re-import, display a warning banner:

Screenshot from 2022-11-08 22-54-13
Screenshot from 2022-11-08 22-54-32

packages/compass-connection-import-export/src/components/export-modal.tsx
gribnoysup: Ah, that's on me, I was trying this out just in the browser devtools, not in the source. For this to work you'd need to make the form body into a grid (the result you are seeing is because the form body has overflow auto, which allows the inner div that you added to stretch to full height when browser is composing the layout) and this `FormModal` component doesn't allow any additional class names to be passed. I am assuming it's not allowed for a reason, so tbh not sure we'd want to change this, sorry for the detour!

packages/compass-connection-import-export/src/hooks/use-import.ts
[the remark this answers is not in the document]
gribnoysup: Oh, that looks amazing!

packages/compass-connections/src/components/connection-list/connection-list.tsx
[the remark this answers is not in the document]
gribnoysup: Yeah, we do the same for "Clear all" button that shows up in "Recents" list, so even for list headers the pattern seems to be there. For a feature like this it's nice to expose it in more accessible places I think, so thanks for adding this!

PR #3724, opened 2022-11-09
chore(import-export): Update import reducer to ts

Anemy:
Part of COMPASS-6220

Code changes:
- Adds a `bulkWrite` function to the `dataService` so that import is not relying on the internals of `DataService`.
- Moves the `insertOne` handler to use `insertOne` from DataService instead of using _collection. 
I'll leave a comment on another change or two which typescript raised.

This PR helps prep for moving the import modal to LeafyGreen components. (hopefully the upcoming import/export epic as well). 

**Notes for reviewers**:
Be sure to open the [`src/modules/import.ts`](packages/compass-import-export/src/modules/import.ts) file, in github it's hidden by default to me.


packages/compass-import-export/src/modules/import.ts
```diff
@@ -827,9 +875,9 @@ const reducer = (state = INITIAL_STATE, action) => {
 
   if (action.type === FINISHED) {
     const isComplete = state.status !== PROCESS_STATUS.CANCELED;
-    const hasErrors = state.errors.length > 0;
+    const hasErrors = (state.errors || []).length > 0;
 
-    let status = state.stats;
```
Anemy, line 832: Updated possible typo, doesn't seem likely to impact anything as a bug as this variable is pretty much always set to something else.

packages/compass-import-export/src/utils/collection-stream.ts
```diff
@@ -67,27 +136,29 @@ class WritableCollectionStream extends Writable {
...
   }
 
-  async _executeBatch(callback) {
+  async _executeBatch(callback: (err?: Error) => void) {
     const documents = this.batch;
 
     this.batch = [];
 
-    let result;
-    let error;
```
Anemy, line 79: Looked unused, it was never set to anything.

packages/compass-import-export/src/utils/collection-stream.ts
```diff
@@ -67,27 +136,29 @@ class WritableCollectionStream extends Writable {
...
+    let result: BulkOpResult;
 
     try {
-      result = await this._collection().bulkWrite(
-        documents.map((doc) => ({ insertOne: doc })),
+      result = await this.dataService.bulkWrite(
```
mcasimir, line 150: definitely not for this PR, but how does this gets canceled? Would it be worth creating a follow-up ticket in the import/export epic?

mcasimir (approved):

packages/data-service/src/data-service.ts
```diff
@@ -645,6 +647,12 @@ export interface DataService {
     callback: Callback<Document | UpdateResult>
   ): void;
 
+  bulkWrite(
```
mcasimir, line 650: 👏 👏 👏

packages/compass-import-export/src/constants/file-types.ts
```diff
@@ -1,4 +1,8 @@
-const FILE_TYPES = {
+export type AcceptedFileType = 'csv' | 'json';
+
+const FILE_TYPES: {
+  [AcceptedFileType: string]: string;
+} = {
```
addaleax, line 5: Just curious, this is making the type less specific than it previously was with just `as const` – any particular reason for that (now it’s `{ [key: string]: string }`, it used to be `{ CSV: 'csv', JSON: 'json' }`)?

addaleax (approved):

packages/compass-import-export/src/constants/file-types.ts
Anemy: Nope, thanks for catching! Updated to be more strict:
```ts
const FILE_TYPES: {
  [fileType: string]: AcceptedFileType;
} = {
```
Eventually we'll remove this variable once everything is using the `AcceptedFileType`.

PR #3728, opened 2022-11-09
feat(compass-import): update import modal to LeafyGreen components COMPASS-6220

Anemy:
Part of COMPASS-6220

The pr updates the modal and buttons of the import modal to LeafyGreen components.
There are still components not updated (mostly in `import-options.js` still to do, I figure smaller prs are a bit easier to look at so I'll do those in another pr.)

One difference in the modals is that now the content scrolls while before the content would always show and the whole screen with the modal would scroll. This may hinder how easy it is to see and acknowledge when there are errors importing. 
Asked about in `#compass-designs`: https://mongodb.slack.com/archives/GDTJXPHD0/p1668017573654519

LG Modal: https://www.mongodb.design/component/modal/example/
This is the last Modal in Compass to be updated to LeafyGreen ✨ 

Screen Shot 2022-11-09 at 12 59 37 PM
Screen Shot 2022-11-09 at 12 59 50 PM
Screen Shot 2022-11-09 at 1 00 30 PM


packages/compass-import-export/src/components/export-modal.tsx
```diff
@@ -378,10 +378,6 @@ function ExportModal({
 
 /**
  * Map the state of the store to component properties.
- *
- * @param {Object} state - The state.
- *
- * @returns {Object} The mapped properties.
```
addaleax, line 384: :blue_heart:

packages/compass-import-export/src/components/import-modal.tsx
```diff
@@ -0,0 +1,290 @@
...
+          </Button>
+        ) : (
+          <>
+            <Button
+              data-testid="import-button"
+              onClick={handleImportBtnClicked}
+              disabled={!fileName || status === STARTED}
+              variant="primary"
+            >
+              {status === STARTED ? 'Importing...' : 'Import'}
```
addaleax, line 237: *obviously*, feel free to ignore:

```suggestion
              {status === STARTED ? 'Importing\u2026' : 'Import'}
```

addaleax (approved):

packages/compass-import-export/src/components/import-modal.tsx
```diff
@@ -0,0 +1,290 @@
...
+  [COMPLETED_WITH_ERRORS]: 'Import completed with following errors:',
+  [FAILED]: 'Failed to import with the following error:',
+  [UNSPECIFIED]: '',
+} as const;
+
+const closeButtonStyles = css({
+  marginRight: spacing[2],
+});
+
+type ImportModalProps = {
```
lerouxb, line 61: Wow this thing takes a lot of props. Will be super handy to have this as reference for the upcoming project.

lerouxb (approved):
Looks good. Diancheng's suggestions in the thread in dev-tools-designs also makes sense. Don't have to make those changes in this PR anyway.

PR #3729, opened 2022-11-09
chore(databases-collections): update drop database and collection modal components to LeafyGreen COMPASS-6223

Anemy:
COMPASS-6223
Updates the `TextInput`, `Label`, and `Banner`/status message components in the drop collection modal and the drop database modal.

Before merging, going to sync on the ux and the text shown while the operation is in progress:
Previously we would show a blue banner with a not moving in progress icon. Now we're showing the loading spinner. Both before and previously we don't disable the drop button while it's in progress. I'm thinking we might want to change the in progress indicator to be part of the button (disabled with loading spinner while in progress). Going to chat with design before doing that change, as well as the text shown (`Drop in Progress...` vs `Dropping Collection...`).

| before | after |
| - | - |
| Screen Shot 2022-11-09 at 2 50 59 PM | Screen Shot 2022-11-09 at 2 33 08 PM |
| Screen Shot 2022-11-09 at 2 52 20 PM | Screen Shot 2022-11-09 at 2 28 56 PM |


packages/databases-collections/src/modules/drop-database/index.js
```diff
@@ -6,9 +6,6 @@ import isRunning, {
 import isVisible, {
   INITIAL_STATE as IS_VISIBLE_INITIAL_STATE
 } from '../is-visible';
-import name, {
-  INITIAL_STATE as NAME_INITIAL_STATE
-} from '../drop-database/name';
```
Anemy, line 11: This reducer looked unused, the name would be set once, but the `updateDatabaseName` event was never used.

packages/databases-collections/src/components/drop-collection-modal.tsx
```diff
@@ -0,0 +1,129 @@
...
+  name,
+  nameConfirmation,
+  error,
+  changeCollectionNameConfirmation,
+  dropCollection,
+  toggleIsVisible,
+}: DropCollectionModalProps) {
+  const onNameConfirmationChange = useCallback((evt: React.ChangeEvent<HTMLInputElement>) => {
+    changeCollectionNameConfirmation(evt.target.value);
+  }, [ changeCollectionNameConfirmation ]);
```
addaleax, line 48: Maybe a dumb question, but would it make sense to move this into the UI state along the lines of `const [ nameConfirmation, changeCollectionNameConfirmation ] = useState('');` (and similarly in the drop-datbase-modal)? Obviously not something that’s really being introduced here but it would seem like something that makes sense to me, since the confirmation string is really just affecting UI state.

packages/databases-collections/src/components/drop-collection-modal.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  return (
+    <FormModal
+      title="Drop Collection"
+      open={isVisible}
+      onSubmit={onFormSubmit}
+      onCancel={onHide}
+      submitButtonText="Drop Collection"
+      variant="danger"
+      submitDisabled={name !== nameConfirmation}
```
addaleax, line 71: This would be new, but should this be

```suggestion
      submitDisabled={name !== nameConfirmation || isRunning}
```

to prevent clicking the button twice?

addaleax (approved):

packages/databases-collections/src/components/drop-database-modal.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const onNameConfirmationChange = useCallback((evt: React.ChangeEvent<HTMLInputElement>) => {
+    changeDatabaseNameConfirmation(evt.target.value);
+  }, [ changeDatabaseNameConfirmation ]);
+
+  const onHide = useCallback(() => {
+    toggleIsVisible(false);
+  }, [ toggleIsVisible ]);
+
+  /**
+   * When user hits enter to submit the form we prevent the default behaviour.
```
lerouxb, line 57: Not technically true. FormModal already prevents default behaviour :)

lerouxb (approved):
Looks good. Will be interesting to see what the design team says. Your suggestions make sense.

packages/databases-collections/src/components/drop-collection-modal.tsx
Anemy: Makes sense to me, those are better in the state for a component like this. How they are currently in the store does make it so if the component unmounts and is remounted it can persist the name confirmation. I don't see how that would help for these components thought. Updating 👍

PR #3737, opened 2022-11-10
feat(query-history): convert compass query history to new components COMPASS-6221

lerouxb:
The recent queries' date format is awful, but I'll address that in a follow-up PR -I'll probably refactor some aggregation stuff into compass-components so we can use it here.

Screenshot 2022-11-10 at 16 35 57

Screenshot 2022-11-10 at 16 35 55

Screenshot 2022-11-10 at 16 36 01


packages/compass-query-history/src/components/query-history/query-history.jsx
```diff
@@ -14,15 +14,8 @@ const componentStyle = css({
...
-  width: '335px',
-  height: '100%',
-});
-
-const innerStyle = css({
-  display: 'flex',
-  flexDirection: 'column',
-  flex: 1,
-  height: '100%',
+  width: '388px',
```
lerouxb, line 17: Copied from the design.

packages/compass-query-history/src/components/query/query.tsx
```diff
@@ -0,0 +1,131 @@
...
+        customHeading
+      ) : (
+        <QueryHeading title={title}>
+          <div
+            className={cx(
+              childrenStyles,
+              isHovered ? childrenHoveredStyles : undefined
+            )}
+          >
+            {children}
```
lerouxb, line 124: The basic idea I landed on was to specify the actions as children as that's what differs between recents and favorites. But maybe better to just pass the buttons as a prop? Not sure.

packages/compass-query-history/src/components/query/query.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  return (
+    <KeylineCard
+      onClick={runQuery}
+      data-testid={props['data-testid']}
+      className={cx(queryStyles, isHovered ? queryHoveredStyles : undefined)}
+      {...hoverProps}
+    >
+      {customHeading ? (
+        customHeading
```
lerouxb, line 115: This is how the save form works now - just pass in a customHeading.

addaleax (approved):

PR #3739, opened 2022-11-11
fix(compass-preferences-model): account for changes in preferences while fetching

addaleax:
Bug fixes to account for the fact that not only explicitly updating the value of a preference can change its value, but also fetching preferences from disk, if another Compass instance has concurrently modified the preferences model file.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-preferences-model/src/preferences.spec.ts
+++ b/packages/compass-preferences-model/src/preferences.spec.ts
@@ -54,12 +54,63 @@ describe('Preferences class', function () {
     expect(result.enableMaps).to.equal(true);
   });
 
-  it('notifies callers of preferences changes', async function () {
+  it('notifies callers of preferences changes after savePreferences', async function () {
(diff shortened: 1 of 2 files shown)


addaleax:
@gribnoysup So, yeah, I did have that same thought. I didn’t end up implementing this (but can do so if you prefer) because:

- This feels like it’s going into the internals of how we store data, and semantically should probably be addressed in storage-mixin, rather than on a case-by-case basis, if that’s possible
- This PR here is about ensuring data consistency inside the preferences package (i.e. that when the value of a preference as reported by `getPreferences()` changes, the on-changed listener is called and in particular the renderer caches are updated – so that `getPreferences()` doesn’t return different values in main and renderer, for example), not about ensuring that all data is in sync with the disk or with other Compass instances

PR #3743, opened 2022-11-14
feat(query-history): use the same date hook in query history as in saved aggregations COMPASS-6221

lerouxb:
Followup from https://github.com/mongodb-js/compass/pull/3737 to use the same date format as in aggregation queries. See https://mongodb.slack.com/archives/GDTJXPHD0/p1668075550174579
Screenshot 2022-11-14 at 11 28 35


packages/compass-query-history/src/stores/recent-list-store.js
```diff
@@ -71,7 +71,7 @@ const configureStore = (options = {}) => {
       });
 
       /* Ignore duplicate queries */
-      const existingQuery = this.state.items.find((item) =>
+      const existingQuery = filtered.find((item) =>
```
lerouxb, line 74: Drive-by: When we filtered duplicate queries we didn't take the namespace into account. So if you've run filter `{ foo: 1 }` against collection A and now run it against B it wouldn't add it to B's recent query list. Oops.

addaleax (approved):

PR #3748, opened 2022-11-14
feat: add --username and --password for auto-connect COMPASS-6216

addaleax:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass/src/app/auto-connect.ts
+++ b/packages/compass/src/app/auto-connect.ts
@@ -10,8 +10,8 @@ function applyUsernameAndPassword(
   { username, password }: Pick<AllPreferences, 'username' | 'password'>
 ): ConnectionInfo {
   const connectionString = new ConnectionString(connectionInfo.connectionOptions.connectionString);
-  if (username) connectionString.username = username;
-  if (password) connectionString.password = password;
+  if (username) connectionString.username = encodeURIComponent(username);
+  if (password) connectionString.password = encodeURIComponent(password);
   return {
     ...connectionInfo,
     connectionOptions: {
--- a/packages/compass/src/app/auto-connect.ts
+++ b/packages/compass/src/app/auto-connect.ts
@@ -3,9 +3,26 @@ import { importConnections } from 'mongodb-data-service';
 import type { ConnectionInfo } from 'mongodb-data-service';
 import { promises as fsPromises } from 'fs';
 import { UUID } from 'bson';
+import { ConnectionString } from 'mongodb-connection-string-url';
+
+function applyUsernameAndPassword(
+  connectionInfo: Readonly<ConnectionInfo>,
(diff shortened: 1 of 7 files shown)


packages/compass/src/app/auto-connect.ts
```diff
@@ -3,9 +3,26 @@ import { importConnections } from 'mongodb-data-service';
 import type { ConnectionInfo } from 'mongodb-data-service';
 import { promises as fsPromises } from 'fs';
 import { UUID } from 'bson';
+import { ConnectionString } from 'mongodb-connection-string-url';
+
+function applyUsernameAndPassword(
```
Anemy, line 8: What happens if a user passes a `username` or `password` that errors in a connection string? Should we provide some error protection here? I haven't tested locally, but it seems to me like it would fail the `render` inside of the ampersand view which might make things get into an error state a user would be confused by.

packages/compass/src/app/auto-connect.ts
```diff
@@ -3,9 +3,26 @@ import { importConnections } from 'mongodb-data-service';
...
+  connectionInfo: Readonly<ConnectionInfo>,
+  { username, password }: Pick<AllPreferences, 'username' | 'password'>
+): ConnectionInfo {
+  const connectionString = new ConnectionString(connectionInfo.connectionOptions.connectionString);
+  if (username) connectionString.username = username;
```
Anemy, line 13: Currently in the username and password inputs in the connection form we `encodeURIComponent` the value that's inputted. Should we also escape these? Or better that the users do that themselves?
https://github.com/mongodb-js/compass/blob/main/packages/connection-form/src/utils/connection-string-helpers.ts#L68

packages/compass/src/app/auto-connect.ts
addaleax: Apart from the fact that your comment below needs to be addressed, what would an invalid `username` or `password` look like? The only invalid values I can think of are zero-length strings, but then again we’re already not applying those here

packages/compass/src/app/auto-connect.ts
addaleax: Sigh yeah, I tend to forget that `URL` doesn’t actually do this itself. Yes, we should escape these :)

packages/compass/src/app/auto-connect.ts
Anemy: Can't think of one, probably not a possibility if we're encoding it. 👍

packages/compass/src/app/auto-connect.ts
addaleax: Alright, done, and also applied for `--forceConnectionOptions`

Anemy (approved):

PR #3755, opened 2022-11-15
chore(schema, aggregations): remove react-tooltip usage

Anemy:
Removes the `react-tooltip` usage and defaults to using the html `title` on hover. We aren't removing the last `react-tooltip` usage in Compass in `hadron-react-components` just yet as it will take a bit of work in the table view in compass crud. That'll happen in LINK_TICKET_WHEN_JIRA_UP

Alternatively we could use LeafyGreen's tooltip. I'm leaning towards not using them as they felt a bit odd in both of these places. In aggregations only one button had a tooltip, and in compass-schema the tooltips were already not that aligned and giving limited information on a sample of data. If folks feel it's better with the tooltips I'll bring them back in.

| before | after |
| - | - |
| Screen Shot 2022-11-14 at 8 44 31 PM | Screen Shot 2022-11-14 at 8 44 20 PM |
| Screen Shot 2022-11-14 at 8 48 07 PM | Screen Shot 2022-11-14 at 8 44 07 PM |

### DIFF
--- a/packages/compass-schema/src/components/type/type.jsx
+++ b/packages/compass-schema/src/components/type/type.jsx
@@ -3,10 +3,14 @@ import PropTypes from 'prop-types';
 import sortBy from 'lodash.sortby';
 import find from 'lodash.find';
 import numeral from 'numeral';
-import ReactTooltip from 'react-tooltip';
-import { Disclaimer } from '@mongodb-js/compass-components';
+import { Disclaimer, css } from '@mongodb-js/compass-components';
 
-const schemaTooltipClass = 'schema-probability-percent';
+const fieldTypeLabelStyles = css({
+  textTransform: 'lowercase',
+  overflow: 'hidden',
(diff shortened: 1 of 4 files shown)


packages/compass-schema/src/components/type/type.jsx
```diff
@@ -107,39 +111,26 @@ class Type extends Component {
...
     return (
-      <button
-        {...tooltipOptions}
-        type="button"
-        className={cls}
-        style={style}
-        onClick={handleClick}
-      >
-        <ReactTooltip id={schemaTooltipClass} />
+      <button type="button" className={cls} style={style} onClick={handleClick}>
```
mcasimir, line 133: Shouldn't we keep the tooltip here? is a bit different from the other buttons as is not a description for the icon, is more an actual content

mcasimir (approved):

PR #3757, opened 2022-11-15
chore(compass-instance): update to shared config

Anemy:
This plugin renders all of the instance tab plugins (role = `Instance.Tab`). Currently that's the saved queries, databases, and performance plugins. It might make sense down the line to combine this into compass-home, I'm thinking for now having it in typescript makes that change a bit easier.


packages/compass-instance/src/components/instance.tsx
```diff
@@ -56,21 +62,21 @@ const InstanceComponent = ({
           return true;
       }
     });
-  }, [isDataLake]);
```
Anemy, line 59: Adding `tabs` to this list for the eslint rule for `useMemo` dependencies. There's an update in the store as well so that on the `reset` event, the tabs are kept, previously they became `undefined`: https://github.com/mongodb-js/compass/pull/3757/files#diff-8628da8182f8f61f17f937e945e794263a1c82cd1398384bb7e5e68da5d277ebR60

addaleax (approved):

PR #3764, opened 2022-11-15
feat: Kerberos password field setting COMPASS-5950

alenakhineika:
## Description

Turn the Kerberos password field display into the global Compass setting.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-preferences-model/src/preferences.ts
```diff
@@ -392,6 +393,21 @@ const modelPreferencesProps: Required<{
...
+  showKerberosPasswordField: {
+    type: 'boolean',
+    required: false,
+    default: false,
+    ui: true,
+    cli: true,
+    global: true,
+    description: {
+      short: 'Show Kerberos Password Field',
+      long: 'Shows the field in those rare situations in which Kerberos requires a password.',
```
addaleax, line 408: I think it might be a bit confusing to refer to “those rare situations” without actually saying what those situations are, and “in which Kerberos requires a password” might make it sound as if the protocol itself requires a password sometimes (which it does not). How about this?

```suggestion
      long: 'Show a password field for Kerberos authentication. Typically only useful when attempting to authenticate as another user than the current system user.',
```

if that isn’t too verbose?

addaleax (approved):

packages/compass-preferences-model/src/preferences.ts
alenakhineika: Sounds good, thx!

PR #3768, opened 2022-11-16
feat: add maxTimeMS as setting COMPASS-6063

addaleax:
~~The first commits are from #3767, so feel free to review that one first!~~

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-e2e-tests/helpers/commands/focus-stage-operator.ts
+++ b/packages/compass-e2e-tests/helpers/commands/focus-stage-operator.ts
@@ -11,7 +11,13 @@ export async function focusStageOperator(
   await browser.waitUntil(async () => {
     const inputElement = await browser.$(comboboxInputSelector);
     const isFocused = await inputElement.isFocused();
-    return isFocused === true;
+    if (isFocused === true) {
+      return true;
+    } else {
+      // try to click again
+      await inputElement.click();
+      return false;
+    }
   });
 
   const stageSelectorListBoxElement = await browser.$(
--- a/packages/compass-query-bar/src/components/query-option.tsx
+++ b/packages/compass-query-bar/src/components/query-option.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React from 'react';
 import {
   Label,
   TextInput,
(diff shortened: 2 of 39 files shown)


packages/compass-e2e-tests/helpers/commands/focus-stage-operator.ts
```diff
@@ -11,7 +11,13 @@ export async function focusStageOperator(
...
     const inputElement = await browser.$(comboboxInputSelector);
     const isFocused = await inputElement.isFocused();
-    return isFocused === true;
+    if (isFocused === true) {
+      return true;
+    } else {
+      // try to click again
+      await inputElement.click();
+      return false;
+    }
```
alenakhineika, line 20: 👍

packages/compass-query-bar/src/components/query-option.tsx
```diff
@@ -154,25 +173,30 @@ const UnthemedQueryOption: React.FunctionComponent<QueryOptionProps> = ({
...
-          <TextInput
-            aria-labelledby={`query-bar-option-input-${queryOption}-label`}
-            id={id}
-            data-testid="query-bar-option-input"
-            className={cx(
-              darkMode
-                ? numericTextInputDarkStyles
-                : numericTextInputLightStyles,
-              hasError && optionInputWithErrorStyles
+          <WithOptionDefinitionTextInputProps definition={optionDefinition}>
```
alenakhineika, line 176: Interesting, so TS was complaining about extra props? When looking at their docs I thought we can pass any props we want on top of the standard set. https://www.mongodb.design/component/text-input/documentation/

addaleax: > Interesting, so TS was complaining about extra props? 

Not that I am aware of – the comment above the definition of `WithOptionDefinitionTextInputProps` has some context around why the wrapper is being used rather than just passing the props through directly:

> ```js
> // Helper component to allow flexible computation of extra props for the TextInput
> // component if the query option definition suggests it. In particular,
> // using a separate component allows those extra props to use React hooks in their definition.
> ```

> When looking at their docs I thought we can pass any props we want on top of the standard set. https://www.mongodb.design/component/text-input/documentation/

I’m not sure but I’d read that page as saying that we can pass and props that are supported by the native HTML `input` element, not any props, but in any case, I don’t see why we would want to pass props that are not part of the standard set (and the type used here is `Partial<React.ComponentProps<typeof TextInput>>` to reflect that).

alenakhineika (approved):

PR #3769, opened 2022-11-16
feat: track application launched telemetry COMPASS-6077

alenakhineika:
## Description

Add Application Launched telemetry event.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass/src/app/index.js
+++ b/packages/compass/src/app/index.js
@@ -58,13 +58,6 @@ const { setupTheme } = require('./theme');
 
 const { setupIntercom } = require('./intercom');
 
-ipc.once('app:launched', function () {
-  console.log('in app:launched');
-  if (process.env.NODE_ENV === 'development') {
-    require('debug').enable('mon*,had*');
-  }
-});
-
 const { log, mongoLogId, debug, track } =
   require('@mongodb-js/compass-logging').createLoggerAndTelemetry(
     'COMPASS-APP'
--- a/packages/compass/src/app/index.js
+++ b/packages/compass/src/app/index.js
@@ -251,8 +251,6 @@ app.extend({
             process.env.HADRON_PRODUCT_NAME
           );
           setupIntercom(state.user);
-          // Signal to main process that app is ready.
-          ipc.call('window:renderer-ready');
           // Catch a data refresh coming from window-manager.
           ipc.on('app:refresh-data', () =>
             global.hadronApp.appRegistry.emit('refresh-data')
--- a/packages/compass/src/app/index.js
+++ b/packages/compass/src/app/index.js
@@ -136,6 +136,15 @@ const Application = View.extend({
     webvitals.getCLS(trackPerfEvent);
   },
   /**
+   * Enable all debug output for the development mode.
+   */
+  preRender: function () {
+    if (process.env.NODE_ENV === 'development') {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      require('debug').enable('mon*,had*');
+    }
+  },
+  /**
    * Pre-load into the require cache a bunch of expensive modules while the
    * user is choosing which connection, so when the user clicks on Connect,
    * Compass can connect to the MongoDB instance faster.
@@ -193,7 +202,7 @@ const Application = View.extend({
       lastKnownVersion
     } = preferences.getPreferences();
 
-    // The main process ensured that `telemetryAnonymousId` contains the id of the User model
+    // The main process ensured that `telemetryAnonymousId` contains the id of the User model.
     const user = await User.getOrCreate(telemetryAnonymousId);
 
     this.user.set(user.serialize());
@@ -220,9 +229,9 @@ app.extend({
 
     async.series(
       [
-        // check if migrations are required
+        // Check if migrations are required.
         migrateApp.bind(state),
-        // get user
+        // Get user.
         state.fetchUser.bind(state),
       ],
       function (err) {
@@ -234,6 +243,7 @@ app.extend({
         setupTheme();
 
         Action.pluginActivationCompleted.listen(() => {
+          state.preRender();
           global.hadronApp.appRegistry.onActivated();
           global.hadronApp.appRegistry.emit(
             'application-initialized',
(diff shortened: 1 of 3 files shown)


packages/compass/src/app/index.js
```diff
@@ -58,13 +58,6 @@ const { setupTheme } = require('./theme');
 
 const { setupIntercom } = require('./intercom');
 
-ipc.once('app:launched', function () {
-  console.log('in app:launched');
-  if (process.env.NODE_ENV === 'development') {
-    require('debug').enable('mon*,had*');
-  }
-});
-
```
addaleax, line 67: Nice!

packages/compass/src/main/application.ts
```diff
@@ -54,6 +62,34 @@ class CompassApplication {
...
+      context: mode,
+      launch_connection: launchConnection(file, positionalArguments),
+      protected: protectConnectionStrings,
+      readOnly,
+      autoUpdates,
+      // TODO: COMPASS-6063
+      // maxTimeMS,
+      has_global_config: !!Object.keys(globalPreferences.global).length,
+      has_cli_config: !!Object.keys(globalPreferences.cli).length,
+    });
```
addaleax, line 87: Do you want to split this out into a separate function, like the ones above?

packages/compass/src/main/application.ts
```diff
@@ -12,12 +11,21 @@ import { CompassMenu } from './menu';
...
 
-const debug = createDebug('mongodb-compass:main:application');
+import createLoggerAndTelemetry from '@mongodb-js/compass-logging';
+
+const { debug, track } = createLoggerAndTelemetry('COMPASS-MAIN');
 
 type ExitHandler = () => Promise<unknown>;
 type CompassApplicationMode = 'CLI' | 'GUI';
 
+const launchConnection = (file?: string, positionalArguments?: string) => {
```
addaleax, line 23: using `launch` as a verb here sounds a bit like this would be the function that actually launches a connection somehow, might be nice to indicate in the function name that this only computes a value based on preferences / provides a result for telemetry

alenakhineika: Yeah, i was going to rename it, absolutely agree.

packages/compass/src/app/index.js
```diff
@@ -241,24 +243,25 @@ app.extend({
...
           global.hadronApp.appRegistry.onActivated();
           global.hadronApp.appRegistry.emit(
             'application-initialized',
             APP_VERSION,
             process.env.HADRON_PRODUCT_NAME
           );
           setupIntercom(state.user);
-          // signal to main process that app is ready
+          // Signal to main process that app is ready.
           ipc.call('window:renderer-ready');
```
addaleax, line 255: This is unused now, right?

packages/compass/src/main/application.ts
addaleax: (again, not sure why this conversation was resolved – not a big deal and definitely not something blocking for this PR, but since you gave this a :+1: reaction, I assumed that you actually did want to make this change, so something’s messed up with your git client again maybe? :confused:)

packages/compass/src/main/application.ts
```diff
@@ -54,6 +72,27 @@ class CompassApplication {
...
+    } = preferences.getPreferences();
+
+    debug('application launched');
+    track('Application Launched', {
+      context: getContext(mode),
+      launch_connection: getLaunchConnectionSource(file, positionalArguments),
+      protected: protectConnectionStrings,
+      readOnly,
+      // TODO: replace with maxTimeMS from preferences COMPASS-6063.
+      maxTimeMS: DEFAULT_MAX_TIME_MS,
```
addaleax, line 92: The default for the preference value is going to be `undefined`, though, since this isn’t the maxTimeMS value that’s used in queries, it’s the _upper limit_ for that value (even though it’s also called `maxTimeMS`).

packages/compass/src/main/application.ts
alenakhineika: This time I am pretty sure I changed it! :)

https://github.com/mongodb-js/compass/blob/COMPASS-6077-track-application-launched/packages/compass/src/main/application.ts#L93-L94

alenakhineika: Screenshot 2022-11-17 at 11 55 25

packages/compass/src/main/application.ts
alenakhineika: So maxTimeMS in settings is the default value in input and DEFAULT_MAX_TIME_MS in query / aggregations / schema is the upper limit? do you think maybe it would be nice to rename this DEFAULT_MAX_TIME_MS for readability?

packages/compass/src/main/application.ts
addaleax: Ah sorry, I was referring to this entire block, both the `getPreferences()` and the `track()` call, which could be treated like the functions immediately above this block (the `this.setup...()` ones), but yeah, again, not important or blocking, so if you’re good with this I am too :slightly_smiling_face:

packages/compass/src/main/application.ts
addaleax: Err, no, it’s the other way around – maxTimeMS from settings is the upper limit, because the goal is that admins are supposed to be able to restrict users from running long-running queries.

packages/compass/src/main/application.ts
alenakhineika: Aaaaah got it! Yeah, makes sense, something like `this.trackApplicationLaunched()`.

packages/compass/src/main/application.ts
alenakhineika: Maybe in the setting then it can be something like maxTimeMSUpperLimit?

alenakhineika: In any way, I got this is wrong here and remove it!

addaleax: It could be `maxMaxTimeMS` :stuck_out_tongue_winking_eye: I mean, yeah, happy to rename it in #3768, I feel like `maxTimeMSUpperLimit` is pretty verbose since it already mentions that it’s a maximum in the name, but I do see your point

addaleax (approved):

PR #3776, opened 2022-11-17
feat: implement enableDevTools option COMPASS-6061, COMPASS-5615

alenakhineika:
## Description

By default devtools should not be accessible in Compass. Users can enable or disable devtools from settings.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-preferences-model/src/preferences.ts
```diff
@@ -168,6 +169,19 @@ function deriveReadOnlyOptionState<K extends keyof AllPreferences>(
...
+function deriveProtectConnectionStringsOptionState<
+  K extends keyof AllPreferences
+>(property: K): DeriveValueFunction<boolean> {
+  return (v, s) => ({
+    value: v(property) && !v('protectConnectionStrings'),
+    state:
+      s(property) ??
+      s('protectConnectionStrings') ??
+      (v('protectConnectionStrings') ? 'derived' : undefined),
+  });
```
addaleax, line 182: I’m wondering if this should account for more than just `protectConnectionStrings`, since devtools can effectively be used to do anything with Compass (i.e. basically everything in the `Features` tab)? On the other hand, we probably still want to be able to inspect code while debugging.

The tech design lists this as an option that cannot be set in the UI. If we go that way, we can use the default of `false` here as it is, and require users (like us) to explicitly opt into enabling dev tools through the command line and/or global config file. Is that an approach we want to take? (This might be a separate concern, and we probably still want to keep this logic here, since `protectConnectionStrings: true` in the global config file should probably not be something a user can override through `--enableDevTools` on the command line.)

packages/compass/src/main/menu.ts
```diff
@@ -472,6 +477,15 @@ class CompassMenu {
       this.refreshMenu();
     });
 
+    preferences.onPreferenceValueChanged('enableDevTools', (enableDevTools: boolean) => {
+      this.refreshMenu();
+      if (enableDevTools) {
+        BrowserWindow.getFocusedWindow()?.webContents.openDevTools();
```
addaleax, line 483: Would this open dev tools every time the option is being changed?

packages/compass-preferences-model/src/preferences.ts
alenakhineika: Yes, I missed this point for enableShell as well, the setting should not be editable from UI according to scope. But is it so bad tho? I mean if these features also like readOnly should not be editable for users they can be disabled via global setting or cli args. And if they are not disabled at those levels then we can still keep enableDevTools as default false as it is right now.

packages/compass/src/main/menu.ts
alenakhineika: Yes, I thought it can be handy for feature discoverability. If someone switched it on, they probably want to use it. That being said there is a way to not show it right away.

addaleax: Fair enough, that makes sense :)

packages/compass-preferences-model/src/preferences.ts
addaleax: Yeah, I don’t think it’s bad for enableShell really and it’s probably not bad here either. It just means we need to be careful about disabling this option when _any_ other option that is used to lock Compass down is set.

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -157,6 +158,29 @@ function deriveNetworkTrafficOptionState<K extends keyof AllPreferences>(
...
+      !v('protectConnectionStrings') &&
+      !v('readOnly'),
+    state:
+      s(property) ??
+      s('protectConnectionStrings') ??
+      (v('protectConnectionStrings') ? 'derived' : undefined) ??
+      s('readOnly') ??
+      (v('readOnly') ? 'derived' : undefined) ??
+      s('enableShell') ??
+      (v('enableShell') ? undefined : 'derived'),
```
addaleax, line 180: Thinking about this, I probably slightly messed up the other definitions here – this should maybe be more like


```suggestion
      s(property) ??
      (v('protectConnectionStrings') ? (s('protectConnectionStrings') ?? 'derived') : undefined) ??
      (v('readOnly') ? (s('readOnly') ?? 'derived') : undefined) ??
      (v('enableShell') ? undefined: (s('enableShell') ?? 'derived')),
```

(and similarly for `deriveNetworkTrafficOptionState` / `deriveReadOnlyOptionState` …)

Because somebody can of course force the default value of an option through the command line or global config file, i.e. `--no-protect-connection-strings`, which doesn’t change Compass’s behavior compared to the default and therefore probably also shouldn’t affect `enableDevTools`, but does set the state of `protectConnectionStrings` to `set-cli`.

packages/compass-e2e-tests/tests/global-preferences.test.ts
```diff
@@ -194,7 +193,7 @@ describe('Global preferences', function () {
         expect(value).to.equal('false');
         expect(disabled).to.equal(''); // null = missing attribute, '' = set
         expect(bannerText).to.include(
-          'This setting cannot be modified as it has been set at Compass startup.'
+          'This setting cannot be modified as it has been set in the global Compass configuration file.'
```
addaleax, line 196: These changes are fine ofc, but was there anything that made them necessary? They don’t seem to be related to the rest of the PR, so I’m curious :)

packages/compass/src/main/menu.ts
```diff
@@ -455,7 +461,7 @@ class CompassMenu {
       'window:hide-collection-submenu': this.hideCollection.bind(this),
     });
 
-    preferences.onPreferenceValueChanged('theme', (newTheme) => {
+    preferences.onPreferenceValueChanged('theme', (newTheme: THEMES) => {
```
addaleax, line 464: Also just out of curiosity, not a request to change anything, was something complaining about this? Typescript should be smart enough to infer the type of `newTheme`, because `onPreferenceValueChanged` is defined as a template method

addaleax (approved):

packages/compass-e2e-tests/tests/global-preferences.test.ts
alenakhineika: Like nothing important for the test result :) just noticed that similar tests in this file do it in another way and aligned them.

packages/compass/src/main/menu.ts
alenakhineika: TS was complaining that newTheme is implicitly any type. But it was before I restarted the computer on Friday, and now lots of weird stuff is gone.

PR #3777, opened 2022-11-17
chore(import-export): update export select fields table to LeafyGreen components COMPASS-6220

Anemy:
Part of COMPASS-6220

| before | after |
| - | - |
| Screen Shot 2022-11-17 at 6 13 18 PM | Screen Shot 2022-11-17 at 5 57 02 PM |
| Screen Shot 2022-11-17 at 6 13 25 PM | Screen Shot 2022-11-17 at 5 56 51 PM |


packages/compass-import-export/src/components/export-select-fields.tsx
```diff
@@ -0,0 +1,261 @@
...
+});
+
+const enterToAddStyles = css({
+  display: 'inline-flex',
+  alignItems: 'center',
+  marginLeft: spacing[2],
+});
+
+type ExportSelectFieldsProps = {
+  fields: Record<string, number>;
```
Anemy, line 66: I feel like this should be a boolean, it's only ever 1 or 0. It touches a bit more code than this though, so if folks like it I can update in the other places in the package. https://github.com/mongodb-js/compass/blob/b9d3ece378f90236697a6a3fbe84a820072af15f/packages/compass-import-export/src/modules/load-fields.ts#L61

packages/compass-import-export/src/components/export-select-fields.tsx
```diff
@@ -0,0 +1,261 @@
...
+                    isEveryFieldChecked
+                      ? 'Deselect all fields'
+                      : 'Select all fields'
+                  }
+                  title={
+                    isEveryFieldChecked
+                      ? 'Deselect all fields'
+                      : 'Select all fields'
+                  }
+                  checked={isEveryFieldChecked}
```
addaleax, line 198: I think you might want to set `indeterminate` here if a subset of the fields are checked?

https://github.com/mongodb-js/compass/blob/7984c2abefd94e1d215275be2ae07b6b53338a2d/packages/compass-connection-import-export/src/components/select-table.tsx#L67

(Fwiw, that SelectTable component was written with the intent to be somewhat generically usable if we decide so, not tied to connection import/export. I’m not sure how easy it is to actually re-use it, though, esp. with the logic for adding additional fields here.)

packages/compass-import-export/src/components/export-select-fields.tsx
```diff
@@ -0,0 +1,261 @@
...
+                      onChange={handleFieldCheckboxChange}
+                    />
+                  </div>
+                </Cell>
+                <Cell className={cellContainerStyles}>
+                  <Body>{field.field}</Body>
+                </Cell>
+              </Row>
+              {field.index === fieldKeys.length - 1 && (
+                <Row key=".__add-new-field">
```
addaleax, line 229: Took me a while to get that this is actually returning two `Row`s for one `datum`! Clever! :)

addaleax (approved):

packages/compass-import-export/src/components/export-select-fields.tsx
Anemy: Good idea, going to look into it more in the upcoming import/export work. Having the add new field column is a bit of a hack of the table.

packages/compass-import-export/src/components/export-select-fields.tsx
Anemy: Ended up going fully for boolean, I think it's more consistent with how we do variables in the rest of the app.

PR #3786, opened 2022-11-18
feat(compass-explain-plan): update explain plan components COMPASS-6236

lerouxb:
I just updated exactly the things mentioned in the ticket. We'll have to do more work to get this updated when we do dark mode.

Screenshot 2022-11-18 at 15 48 53


packages/compass-explain-plan/src/components/explain-stage/explain-stage.jsx
```diff
@@ -227,56 +228,51 @@ class ExplainStage extends Component {
...
           zIndex: this.state.detailsOpen ? this.getNewZIndex() : 'initial',
           top,
           left,
         }}
       >
-        <h3 className={classnames(styles['stage-header'])}>
+        <Subtitle className={styles['stage-header']}>
           {this.props.name}
-        </h3>
-        <ul className={classnames(styles.core)}>
```
addaleax, line 240: Just for my own understanding, why are the `classnames()` calls only removed in some places?

addaleax (approved):
Code LGTM, commit message should probably be something like `feat(compass-explain-plan): update Explain plan components COMPASS-6236` (package name & start with lowercase verb)?

packages/compass-explain-plan/src/components/explain-stage/explain-stage.jsx
lerouxb: I just noticed they were redundant. Essentially all it does is `args.join(' ')`. So I just kept the ones where it actually joined stuff. For a moment there I got excited, thinking maybe I can remove the dep :) It usually goes away in these PRs.

lerouxb: (So yes the changes are out of scope)

PR #3802, opened 2022-11-22
fix(compass-preferences-model): ignore squirrel options

addaleax:
### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-preferences-model/src/global-config.spec.ts
+++ b/packages/compass-preferences-model/src/global-config.spec.ts
@@ -44,6 +44,18 @@ describe('Global config file handling', function () {
     });
   });
 
+  it('ignores startup with --squirrel-firstrun', async function () {
+    const result = await parseAndValidateGlobalPreferences({
+      globalConfigPaths: [],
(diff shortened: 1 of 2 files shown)


gribnoysup (commented):
Before `App.init` was the first thing called and `electron-squirrel-startup` would be the first thing handled inside of it ending with the app exiting by doing nothing, should we move the call to `electron-squirrel-startup` to the new entry point of the application now that we moved the app init logic higher?

addaleax:
@gribnoysup I don’t think it would affect this bug or the changes in this PR (or really make any practical difference) since `--squirrel-firstrun` and `--squirrel-updated` aren’t being handled by that package, but you’re probably right that semantically it’s the right thing to do. I’d like to keep this PR minimal and just get the fix out, so: https://github.com/mongodb-js/compass/pull/3803

gribnoysup (approved):

PR #3804, opened 2022-11-22
feat(compass-aggregations): update Compass aggregations modals COMPASS-6286

lerouxb:
Couldn't find any mention of ModalStatusMessage, but replaced the last uses of ModalInput. Removed the aggregation modal notes css (for confirmation modals) or replaced with banner.

Screenshot 2022-11-22 at 12 38 16

Screenshot 2022-11-22 at 12 38 34

Screenshot 2022-11-22 at 12 38 39


packages/compass-aggregations/src/components/saving-pipeline-modal/saving-pipeline-modal.jsx
```diff
@@ -62,13 +62,12 @@ class SavingPipelineModal extends PureComponent {
         trackingId="save_pipeline_modal"
         data-testid="save-pipeline-modal"
       >
-        <input
```
lerouxb, line 65: Drive-by: This wasn't using ModalInput or TextInput before. Only case of this anywhere I could find. Since it is a near-identical aggregations-related modal to the others I changed it with them to work the same.

packages/compass-aggregations/src/components/create-view-modal/create-view-modal.jsx
```diff
@@ -73,11 +72,12 @@ class CreateViewModal extends PureComponent {
...
       >
-        <ModalInput
-          id="create-view-name"
-          name="Enter a View Name"
+        <TextInput
+          data-testid="create-view-name"
           value={this.props.name || ''}
-          onChangeHandler={this.onNameChange}
+          onChange={this.onNameChange}
+          label="Name"
```
lerouxb, line 79: We often have <Label> as a separate element or we jump through hoops to hide the label and use a placeholder or have some nearby heading act as a implicit label or sometimes we even use TextInput's builtin label like here.

I just standardised on using TextInput's label for these modals.

PR #3805, opened 2022-11-22
feat: remove trackErrors setting COMPASS-5708

alenakhineika:
## Description

Remove trackErrors setting and the checkbox in the privacy settings.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-preferences-model/src/preferences.ts
```diff
@@ -336,22 +335,6 @@ const modelPreferencesProps: Required<{
...
-    default: false,
-    ui: true,
-    cli: true,
-    global: true,
-    description: {
-      short: 'Enable Crash Reports',
-      long: 'Allow Compass to send crash reports containing stack traces and unhandled exceptions.',
-    },
-    deriveValue: deriveNetworkTrafficOptionState('trackErrors'),
-  },
```
addaleax, line 354: We only removed bugsnag earlier this year. Removing this from the model means that users who go back from a current Compass version to an older one won’t have this setting preserved – that’s probably fine?

addaleax (approved):
(The JIRA ticket isn’t In Code Review yet but I assume this is ready to be approved :+1:)

packages/compass-preferences-model/src/preferences.ts
alenakhineika: Compass that supports bugsnag is already a decent amount of versions ago. There is actually another preference that we already deleted from the model in the same way - agreedToLicense. I would say this is ok. I guess we could add a migration to always set these settings to the strict value, but I am not sure about that.

PR #3807, opened 2022-11-22
chore(compass-components): add xsmall icon size COMPASS-6306

lerouxb:
Alias size 11px to xsmall.


mcasimir (approved):
There is also another one here: https://github.com/mongodb-js/compass/blob/f1d2d575e66cf18470818caa6446e0528f672e9e/packages/compass-components/src/components/document-list/element-actions.tsx#L23

PR #3810, opened 2022-11-23
chore(compass-crud): convert to ts

addaleax:
[second commit applies changes from #3801 in order to be prepared for the merge conflict]

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-crud/package.json
```diff
@@ -91,7 +91,6 @@
     "electron": "^15.5.7",
     "enzyme": "^3.11.0",
     "eslint": "^7.25.0",
-    "fast-json-parse": "^1.0.3",
```
addaleax, line 94: Drive-by-fix; the package doesn’t provide any types, or any value at all really (“from Node 7 and forward this module is not useful anymore” in its README)

packages/compass-crud/src/components/insert-document-dialog.tsx
```diff
@@ -36,20 +37,48 @@ const documentViewContainer = css({
   marginTop: spacing[3],
 });
 
+export type InsertDocumentDialogProps = InsertCSFLEWarningBannerProps & {
+  closeInsertDocumentDialog: () => void;
+  toggleInsertDocumentView: (view: 'JSON' | 'List') => void;
+  toggleInsertDocument: (view: 'JSON' | 'List') => void;
```
alenakhineika, line 43: Should here also be a table? Like `view: 'List' | 'JSON' | 'Table'`

packages/compass-crud/src/utils/cancellable-queries.ts
```diff
@@ -119,23 +141,23 @@ export async function fetchShardingKeys(
...
+    void cursor.close();
   };
   signal.addEventListener('abort', abort, { once: true });
 
   let configDocs;
 
   try {
     configDocs = await raceWithAbort(cursor.toArray(), signal);
-  } catch (err) {
+  } catch (err: any) {
```
alenakhineika, line 152: Is it really `any` here or do we expect some kind of error format? when we check below for `err.message` we know that err is not undefined or something.

packages/compass-crud/test/aggrid-helper.ts
```diff
@@ -189,4 +189,4 @@ module.exports = {
   NUM_DOCS: NUM_DOCS,
   expectedDocs: expectedDocs,
   checkPageRange: checkPageRange,
-};
+} as any;
```
alenakhineika, line 192: Why do we want to export it as any? Does it brake something where it is being imported?

packages/compass-crud/src/components/insert-document-dialog.tsx
addaleax: We don’t have a table view in the insert modal, so `'JSON' | 'List'` is the right type

packages/compass-crud/test/aggrid-helper.ts
addaleax: It’s a test helper that contains mocks for use in tests, so this isn’t really expected to match the types in the places where it’s being used

packages/compass-crud/src/utils/cancellable-queries.ts
addaleax: So, yeah, we're definitely expecting that `err` is an `Error` instance here. That seems reasonable, because `signal` is under our control, and `cursor.toArray()` is a driver method, and the driver team is a team of good JS engineers who won’t write code that throws `undefined` or `null` as exceptions :)

It’s good that you point this out, though. I do think it’s much nicer how the aggregations package handles this, using an `Error` subclass with a proper `name` property. Since the code is mostly duplicated anyway and this PR might be a good opportunity to clean this up, I’ve decided to move this to `compass-utils` in 6cff480fe9be73c9d96394b3ffd227c031560a38, and then followed up with adding a bit more web compat in 548e098e663e80a9ba261251fe6c6bf50a11953c, feel free to take a look :)

alenakhineika: Nice! so we don't need any type here anymore since isCancelError expects any?

addaleax: The interesting part of the `isCancelError` type definition is the `error is ...` return type, which means it returns `true` if the error is a cancel/abort error and `false` otherwise, and TS can use that information at the call site (like here) to know "ah yes, the type was checked here".

alenakhineika (approved):

PR #3811, opened 2022-11-23
fix(aggregations): fix the saved pipelines popover's scrolling COMPASS-6277

lerouxb:
Screenshot 2022-11-23 at 15 31 20


mcasimir (approved):
Shouldn't this be in the release notes? or is not affecting the current version?

PR #3816, opened 2022-11-24
fix(compass-settings): ignore non-digits in number input COMPASS-6326

addaleax:
### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


mabaasit (approved):
If we change the input type to number, that i think is also an option. plus, it will enable increase/decrease controls on the input

mcasimir:
> If we change the input type to number, that i think is also an option. plus, it will enable increase/decrease controls on the input

Yup: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number

PR #3817, opened 2022-11-24
fix: fix modals in LG darkmode

mcasimir:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-components/src/components/modals/modal-body.tsx
+++ b/packages/compass-components/src/components/modals/modal-body.tsx
@@ -3,6 +3,7 @@ import { css, cx } from '@leafygreen-ui/emotion';
 import { palette } from '@leafygreen-ui/palette';
 import { fontFamilies, spacing } from '@leafygreen-ui/tokens';
 import { Variant as ButtonVariant } from '@leafygreen-ui/button';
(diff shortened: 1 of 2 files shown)


packages/compass-components/src/components/modals/modal-body.tsx
```diff
@@ -33,6 +34,10 @@ const variantStyle = {
   }),
 };
 
+const darkModeStyle = css({
+  paddingBottom: spacing[3],
```
addaleax, line 38: … but why do we want padding to differ between light mode and dark mode? This does look a bit odd in dark mode now because the bottom and top paddings aren’t the same anymore

mcasimir: that's actually a fix for that, LG adds a border on the footer in darkmode

addaleax (approved):

mcasimir:
@addaleax turns out that broke the index dialog, added a new commit to fix it

addaleax (approved):

PR #3824, opened 2022-11-25
fix(export): Speed up export COMPASS-6332

lerouxb:
Two things:

1) We weren't only updating the progress every 250ms like the code was asking for.

2) There was a flood of debug() calls that killed the dev console at least in development.

This speeds up exports in development on my local machine by at least 1000x.

### DIFF
--- a/packages/compass-import-export/src/modules/export.ts
+++ b/packages/compass-import-export/src/modules/export.ts
@@ -506,7 +506,6 @@ export const startExport =
         totalNumberOfDocuments: numDocsToExport,
       });
 
-      let times = 0;
       const throttledProgress = throttle(
         ({
           percentage,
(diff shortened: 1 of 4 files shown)


packages/compass-import-export/src/modules/export.ts
```diff
@@ -518,13 +506,28 @@ export const startExport =
...
+        250
+      );
+
       exporter.on('progress', function (transferred) {
         let percent = 0;
         if (numDocsToExport !== null && numDocsToExport > 0) {
           percent = (transferred * 100) / numDocsToExport;
         }
         numDocsActuallyExported = transferred;
-        progress.emit('progress', {
```
addaleax, line 527: yeah lol we've been using this all wrong anyway 😅

addaleax (approved):

PR #3827, opened 2022-11-25
feat(insert): Convert more insert dialog code to compass components & leafygreen COMPASS-6285

lerouxb:
This replaces the footer thing with a banner and it adds darkmode support for the editor container.

The ticket also mentions "Document action button icons to leafygreen icons (currently fontawesome)." but I can't find any more fontawesome icons. Maybe I'm not looking in the right places?


packages/compass-crud/src/components/table-view/cell-editor.tsx
```diff
@@ -483,14 +481,6 @@ class CellEditor
     return (
       <div>
         <span className={this.wrapperStyle()}>
-          <Tooltip
```
lerouxb, line 486: While looking for things that still use hadron-react-components I noticed that this is (I think) the last Tooltip. But I can't get it to show up. So it seems to not be doing anything anymore? If so I guess we can just remove it. Otherwise I suppose we'll have to replace it.

If it is still used, please tell me how to trigger it ;)

lerouxb: The reason why I'm reasonably sure that the tooltip doesn't work is because the corresponding text input doesn't have any of data-tip, data-place or data-for.

There's actually still one left in aggregations that will be replaced shortly: https://github.com/mongodb-js/compass/blob/4b5384b621dc280fa8e616257311124ba2decf9c/packages/compass-aggregations/src/components/stage-editor-toolbar/add-after-stage.jsx#L33-L47

And I realised something in compass schema as well: https://github.com/mongodb-js/compass/blob/4b5384b621dc280fa8e616257311124ba2decf9c/packages/compass-schema/src/components/type/type.jsx#L135-L142

Anemy: We're updating those other two tooltips in this pr: https://github.com/mongodb-js/compass/pull/3792 so if we remove/update that one in crud here (or another pr) we can remove `react-tooltip` 🙌

Anemy:
> The ticket also mentions "Document action button icons to leafygreen icons (currently fontawesome)." but I can't find any more fontawesome icons. Maybe I'm not looking in the right places?

Updated them in this pr:
https://github.com/mongodb-js/compass/pull/3672
Forgot to update the ticket - sorry for the confusion!

packages/compass-crud/src/components/insert-json-document.tsx
```diff
@@ -1,24 +1,48 @@
...
+  borderLeft: `3px solid ${palette.gray.light2}`,
+});
+
+const editorContainerStylesDark = css({
+  backgroundColor: palette.gray.dark3,
+  borderLeft: `3px solid ${palette.gray.dark2}`,
+});
+
+const editorStyles = css({
+  // NOTE: this height is coupled with the padding and height of the .editor-container.
```
Anemy, line 28: Looks like this comment and the one on line 8 are now outdated. I'm thinking now because these two are reliant on each other for their height, how do you feel about sharing a variable for the height so there's one source of truth?
Something like `const editorHeight = 280;` and then reference it in both places?

packages/compass-crud/src/components/insert-json-document.tsx
```diff
@@ -43,16 +67,22 @@ class InsertJsonDocument extends Component<InsertJsonDocumentProps> {
...
-      <div className={styles['editor-container']}>
+      <div
+        className={cx(
+          editorContainerStyles,
+          darkMode ? editorContainerStylesDark : editorContainerStylesLight
+        )}
+      >
         <Editor
-          className={styles.editor}
+          className={editorStyles}
```
Anemy, line 85: Would it work if the background color is something the editor controls? I'm wondering if there are other spots that might benefit if the background color setting happens in the `Editor` component instead of here at the consumer. Fine as is too.

Anemy (approved):
Looking good, nice cleanup on the tooltip. It might have become unused in some of the recent changes. Left a comment on a comment we should probably cleanup and a possible code improvement.

packages/compass-crud/src/components/insert-json-document.tsx
lerouxb: I don't know. The colours are part of editor container which is outside editor. But this is not the only editor container, so if we change something there I'd rather do that in a separate PR so we can change validation and whatever else is doing that as well.

The `editorStyles` variable you commented on only sets the height.

PR #3838, opened 2022-11-29
fix(compass-preferences-model): avoid race condition when installing listeners

addaleax:
The following sequence of events is currently possible:
- A preference value is changed in the main process
- The main process informs the renderer about it
- Consequentially, the renderer starts to refresh its cache
- Something (X) running in the renderer gets the preference value via `getPreferences()` and then listens for changes using `onPreferenceValueChanged()` (as e.g. the React preference utils do)
- The cache refresh operation finishes and updates the cache

X then uses an outdated value of the preference and is not informed about the change in value.

Fix this by keeping track of active cache refresh operations, and if there are any, call the newly installed listener once with the fresh value.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-preferences-model/src/renderer-ipc.ts
+++ b/packages/compass-preferences-model/src/renderer-ipc.ts
@@ -14,7 +14,7 @@ import { createSandboxAccessFromProps } from './setup-preferences';
  * API to communicate with preferences from the electron renderer process.
  */
 export const makePreferencesIpc = (ipcRenderer: HadronIpcRenderer) => {
-  let cachedPreferences = {} as AllPreferences;
+  let cachedPreferences = {} as Readonly<AllPreferences>;
   let inflightCacheRefresh: Promise<AllPreferences> | undefined;
   async function refreshCachedPreferences(): Promise<AllPreferences> {
     inflightCacheRefresh = ipcRenderer.invoke('compass:get-preferences');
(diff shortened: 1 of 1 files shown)


packages/compass-preferences-model/src/renderer-ipc.ts
```diff
@@ -14,13 +14,13 @@ import { createSandboxAccessFromProps } from './setup-preferences';
...
  */
 export const makePreferencesIpc = (ipcRenderer: HadronIpcRenderer) => {
-  const cachedPreferences = {} as AllPreferences;
+  let cachedPreferences = {} as AllPreferences;
+  let inflightCacheRefresh: Promise<AllPreferences> | undefined;
   async function refreshCachedPreferences(): Promise<AllPreferences> {
-    const result: AllPreferences = await ipcRenderer.invoke(
-      'compass:get-preferences'
-    );
-    Object.assign(cachedPreferences, result);
```
addaleax, line 22: (drive-by change, there’s no reason not to just fully re-assign the variable; `Object.assign` would not delete keys that no longer exist, which currently does not happen but might at some point in the future)

Anemy (approved):

PR #3858, opened 2022-12-01
chore(compass-components): use darkMode with LeafyGreenProvider COMPASS-6331

Anemy:
COMPASS-6331

Now that LeafyGreen's provider provides darkMode to its components we can use that instead of providing darkMode ourselves in our own wrappers.


packages/compass-home/src/components/home.tsx
```diff
@@ -399,7 +399,7 @@ function ThemedHome(
   }, [setIsSettingsOpen]);
 
   return (
-    <LeafyGreenProvider>
+    <LeafyGreenProvider darkMode={theme.enabled && theme.theme === Theme.Dark}>
       <ThemeProvider theme={theme}>
```
gribnoysup, line 403: Why are we keeping our own provider? In what cases we should use leafygreen useDarkMode hook and in what cases ours?

gribnoysup:
This probably should take into account places that use nesting our theme provider to either force a specific theme or provide theme in places where the context is not correctly provided otherwise, like ag-grid component:

|main|this patch|
|---|---|
|image|image|

mcasimir:
> Why are we keeping our own provider? In what cases we should use leafygreen useDarkMode hook and in what cases ours?

> This probably should take into account places that use nesting our theme provider to either force a specific theme (like a tooltip content) or provide theme in places where the context is not correctly provided otherwise, like ag-grid component:

I guess this is a reason to keep the `ThemeProvider`?

If it is safe to stack providers like this, we should consider just putting this in the `ThemeProvider` and we could be leaving the outermost `LeafyGreenProvider` as it is for other customizations.

```
<LeafyGreenProvider darkMode={theme.enabled && theme.theme === Theme.Dark}>
```

packages/compass-home/src/components/home.tsx
Anemy: Updated and removed ours, I don't think there's any case currently where we'd need our own. Thanks for raising!

packages/compass-components/src/components/leafygreen.tsx
```diff
@@ -1,25 +1,23 @@
 import React from 'react';
```
mcasimir, line 1: nothing urgent, but is this file still necessary?

mcasimir (approved):
🥳

packages/compass-components/src/components/leafygreen.tsx
Anemy: Not really necessary, no, this could all go in the `index.ts`. I think it might be nice to have this structure to keep that file a bit more readable and have a dedicated place for any future temporary wrappers we might need around LeafyGreen. Happy to change it if you feel strongly about where things should go.

PR #3868, opened 2022-12-05
chore(compass-components): move `StoreConnector` to compass-components, remove `hadron-react-components`

Anemy:
Moves the last component in `hadron-react-components` to `compass-components`.

The last component is `StoreConnector` which connects a reflux store to a react component. This makes it pass in the store's state as props to the wrapped component. We'll be moving away from reflux in time, so this should go away eventually. However, as this component is still used and not planned to be removed anytime soon we're moving it into `compass-components` so we can remove the `hadron-react-components` package.


mcasimir (approved):
🔥

PR #3878, opened 2022-12-08
feat(aggregations): Update aggregations stage components COMPASS-6234

lerouxb:
**TODO:**

* [ ] tests
* [ ] the editor background has to be white, but all the editors everywhere are gray. need an override somehow
* [ ] not sure about the warning and error line colours in dark mode. Same as light for now.
* [ ] the designs had the stage type dropdown inside the editor. I haven't done that yet and I'm not really sure we should.
* [ ] $out & $merge has lots of code I don't know how to test or get to
* [ ] need  to add + buttons between stages
* [ ] maybe a smaller Add Stage button?
* [ ] need to add text above the Add Stage button
* [ ] some weird typescript errors around type inference of connect() I need to solve or ignore somehow

Screenshot 2022-12-08 at 17 21 30


packages/compass-aggregations/src/components/stage-preview.tsx
```diff
@@ -0,0 +1,399 @@
...
+  saveDocuments,
+  stageValue
+}: OutProps) {
+    if (isComplete) {
+      if (hasServerError) {
+        return (<div className={stagePreviewOutStyles} />);
+      }
+
+      // TODO: is it still possible to get here? How do we test this?
+      return (
```
mabaasit, line 129: Yes, for both the stages `$out/$merge`, but only on cloud. You can simple test is by setting atlas in the state.

packages/connection-form/src/components/connect-form-actions.tsx
```diff
@@ -8,18 +8,28 @@ import {
   palette,
   css,
   cx,
+  useDarkMode,
```
mabaasit, line 11: are changes to this file related to the current PR?

lerouxb: no. commit message "drive-by". I saw that line too many times :)

lerouxb: I can extract it, I suppose

mabaasit: was just curious. you can leave it!

packages/compass-aggregations/src/components/atlas-stage-preview.tsx
```diff
@@ -25,7 +25,7 @@ const atlasTextStyles = css({
   maxWidth: '400px',
 });
 
-export const AtlastStagePreivew = ({
```
mabaasit, line 28: 🙈

packages/compass-aggregations/src/components/stage.tsx
```diff
@@ -0,0 +1,162 @@
...
+  isExpanded,
+  hasSyntaxError,
+  hasServerError,
+  isAutoPreviewing
+}: StageProps) {
+  const opacity = isEnabled ? 1 : DEFAULT_OPACITY;
+
+  return (<KeylineCard className={cx(
+      stageStyles,
+      hasSyntaxError && stageWarningStyles,
```
mabaasit, line 135: Currently we are not showing any warning if the stage was just added (no operator or value has been set and as a result syntaxError is null).

lerouxb: I pushed a commit and now it shouldn't show the yellow border if there is no stage operator or value.

lerouxb: (behavior for the banner should be unchanged and it will still show the yellow border if there is some other syntax error. ie. if there is an operator and value)

PR #3879, opened 2022-12-08
chore(compass-components): bump LeafyGreen dependencies for more darkMode support

Anemy:
The `Modal` and `Table` components recently received support for receiving `darkMode` from the `LeafyGreenProvider`.

https://github.com/mongodb/leafygreen-ui/blob/main/packages/modal/CHANGELOG.md#major-changes
https://github.com/mongodb/leafygreen-ui/blob/main/packages/table/CHANGELOG.md#1000


mcasimir (approved):

Anemy:
- [] Before merging going to check that our hack for `SegmentedControl` still works as intended, and I'll click around a bit incase there are any small styling tweaks that might have happened that need addressing.

Anemy:
Looks like the `Code` component language switching option needs to be clicked twice before the language can be changed. Asked about it in slack: https://mongodb.slack.com/archives/CFFA5BS79/p1670528898443499 
Going to hold off on adding a workaround till I hear back.

Anemy:
I'm thinking we wait for another release that fixes the double click on language selection in the code component before updating our LeafyGreen versions.
As the new darkMode isn't out yet it isn't something folks will likely need in the short term. If y'all think we should do a workaround for now though I'm up for it. I tried only updating the modal components and the table as those are the two which gain darkMode and the issue still happens as it's a patch dependency that they rely on that gets updated.

mcasimir:
I would create a ticket to enable the LG darkmode and put it in blocked.

The only thing we should clarify about it is if it makes sense to re-add the hack, since we are technically thinking to release the LG dark mode as an experimental feature already in the settings.

Anemy:
Passing the theme here: https://github.com/mongodb-js/compass/pull/3889
Going to close this, we can revisit when COMPASS-6367 is unblocked

PR #3886, opened 2022-12-12
feat: index tab UI improvements COMPASS-6323, COMPASS-6329

mcasimir:
Screenshot 2022-12-15 at 13 24 10


Anemy (approved):
Changes look good, I also like Claudia's suggestion of removing the toolbar box here, looks a bit nicer without it imo.

packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
```diff
@@ -178,6 +175,14 @@ export const IndexesTable: React.FunctionComponent<IndexesTableProps> = ({
                 )}
               </Cell>
             )}
+            <Row className={removePaddingStyles}>
```
mabaasit, line 178: took me a lot of time to figure out that nesting rows creates a Collapse button. How did you find it? it's not even documented.

but then it does not look like other accordians we have in Compass, where the whole title is clickable.

mcasimir: yeah I've searched threads in the leafygreen channel

mcasimir: > where the whole title is clickable

mmm that's a good point .. should be fine, the problem with the accordion is that the cell content is really tough to maintain aligned, and It was already messed up after the LG update

packages/compass-indexes/src/components/indexes-table/property-field.tsx
```diff
@@ -110,11 +112,11 @@ const PropertyField: React.FunctionComponent<PropertyFieldProps> = ({
         />
       )}
       {extra.status === 'inprogress' && (
-        <Badge variant={BadgeVariant.Blue}>In Progress...</Badge>
+        <Badge variant={BadgeVariant.Blue}>In Progress ...</Badge>
       )}
       {extra.status === 'failed' && (
         <ErrorBadgeWithTooltip
-          tooltip={String(extra.error)}
+          tooltip={extra.error ? String(extra.error) : ''}
```
mcasimir, line 119: drive by, if error is `undefined` this actually would show a tooltip with the `undefined` text in it

mabaasit (approved):

packages/compass-indexes/src/modules/indexes.ts
```diff
@@ -211,6 +229,25 @@ const _convertToModels = (indexes: Document[]): IndexDefinition[] => {
...
+  indexes: IndexDefinition[],
+  inProgressIndexes: IndexDefinition[]
+) {
+  for (const inProgressIndex of inProgressIndexes) {
+    const index = indexes.find((index) => index.name === inProgressIndex.name);
+
+    if (index) {
+      index.extra = index.extra ?? {};
+      index.extra.status = inProgressIndex.extra.status;
+      index.extra.error = inProgressIndex.extra.error;
```
Anemy, line 242: How do you feel about avoiding the mutation of the passed in index objects here? I've found that when things are mutated in methods it can lead to bugs down the line, as the changes to the objects can easily be hidden and unexpected.

packages/compass-indexes/src/components/indexes-table/key-list.tsx
```diff
@@ -0,0 +1,44 @@
...
+      {keys.map(({ field, value }) => (
+        <Badge
+          key={field}
+          data-testid={`${field}-key`}
+          variant={BadgeVariant.LightGray}
+          className={badgeStyles}
+          role="listitem"
+        >
+          {field}
+          <IndexIcon direction={value} />
```
Anemy, line 37: [unrelated nit feel free to ignore] Something I noticed about the index icon direction is that there is no `title` attribute. It might be a slightly nicer ux to provide that so folks can be know the terminology behind `ascending`/`descending`. Raising this since we're doing some other improvements to this view in this pr already.

packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
```diff
@@ -178,6 +175,14 @@ export const IndexesTable: React.FunctionComponent<IndexesTableProps> = ({
...
             )}
+            <Row>
+              <Cell
+                className={cx(nestedRowCellStyles, cellStyles)}
+                colSpan={canDeleteIndex ? 6 : 5}
+              >
+                <KeyList keys={index.fields.serialize()} />
+              </Cell>
+            </Row>
           </Row>
```
Anemy, line 186: Small thing, bringing it up since we're updating some other parts of this table, feel free to do nothing. I found it a bit hard to know exactly which row I was deleting when hovering over the delete button. Should we do something like alternating the color of rows, or having a row hover background color to make it more apparent which row is being highlighted to delete? 
Screen Shot 2022-12-15 at 5 13 09 PM

Anemy (approved):
lgtm, some smaller improvements/thoughts I was thinking about when trying out this functionality. Nice improvements.

packages/compass-indexes/src/modules/indexes.ts
mcasimir: the indexes are cloned in the caller, but I like the extra safety

packages/compass-indexes/src/components/indexes-table/key-list.tsx
mcasimir: i tried to add the title but nothing happens, something with the icon component I suspect

packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
mcasimir: Yeah that would be nice, but we may need some design input for that though

packages/compass-indexes/src/modules/indexes.ts
mcasimir: changed

PR #3897, opened 2022-12-15
chore(compass-schema): remove global css grid styles, update schema styles COMPASS-5579

Anemy:
Part of COMPASS-5579
Removes the `grid.css` and `utilities.css` global styles. Updates the two places using these styles to not use them. Also updates the `field` component in `compass-schema` to typescript as it has caused a few errors in the last weeks which would have been caught in typescript.

| before | after |
| - | - |
| Screen Shot 2022-12-14 at 7 09 21 PM | Screen Shot 2022-12-14 at 7 09 05 PM |


packages/compass-schema/src/components/field.tsx
```diff
@@ -0,0 +1,290 @@
...
+  // Check if the type represents geo coordinates, if privacy settings allow.
+  if (!enableMaps) {
+    return type;
+  }
+  const coords = detectCoordinates(type);
+  if (coords) {
+    return {
+      ...type,
+      name: 'Coordinates',
+      values: coords,
```
Anemy, line 168: Changed this function to not mutate the field types passed.

mcasimir (approved):

PR #3914, opened 2022-12-21
fix: export to Language (Java) has incorrect class name COMPASS-6159

alenakhineika:
## Description

Use a proper `AggregateIterable` class name when exporting to java instead of an incorrect `FindIterable`.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/bson-transpilers/package.json
```diff
@@ -12,6 +12,7 @@
   "main": "index.js",
   "compass:main": "index.js",
   "scripts": {
+    "bootstrap": "npm run compile",
```
addaleax, line 15: I’m wondering if there’s a reason this isn’t part of bootstrapping (maybe because it involves a fairly complex java build step)?

addaleax (approved):

packages/bson-transpilers/package.json
alenakhineika: CI just failed because it was unable to locate a Java Runtime, so I removed bootstrapping.

packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js
```diff
@@ -116,7 +116,12 @@ module.exports = (ANTLRVisitor) => class CodeGenerationVisitor extends ANTLRVisi
...
       .reduce((obj, c) => {
-        obj[c] = this.Imports[c].template(this.requiredImports[c]);
+        if (c === 'driver') {
+          obj[c] = this.Imports[c].template(mode);
+        } else {
+          obj[c] = this.Imports[c].template(this.requiredImports[c]);
+        }
+
         return obj;
       }, {});
```
addaleax, line 126: It might be a good idea to refactor this entire statement to a `for ... of`  loop, just for readability…

packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js
```diff
@@ -116,7 +116,12 @@ module.exports = (ANTLRVisitor) => class CodeGenerationVisitor extends ANTLRVisi
         );
       })
       .reduce((obj, c) => {
-        obj[c] = this.Imports[c].template(this.requiredImports[c]);
```
addaleax, line 119: Would it make sense to pass both `this.requiredImports[c]` and `mode` as arguments always? That way, all the templates have the same function signature, rather than one being special-cased in a way that’s not obvious just when looking at the YAML files

packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js
alenakhineika: Replaced with `for .. in`.

packages/bson-transpilers/codegeneration/CodeGenerationVisitor.js
```diff
@@ -107,18 +107,16 @@ module.exports = (ANTLRVisitor) => class CodeGenerationVisitor extends ANTLRVisi
...
-          this.Imports[code] &&
-          this.Imports[code].template
-        );
-      })
-      .reduce((obj, c) => {
-        obj[c] = this.Imports[c].template(this.requiredImports[c]);
-        return obj;
-      }, {});
+    const imports = {};
+    for (const code in this.requiredImports) {
```
addaleax, line 111: tiny nit, but it’s a good practice to avoid `for ... in` unless it’s intentional to include enumerable keys from an object’s prototype chain (i.e. what the `Object.prototype.hasOwnProperty.call()` call would have tested for as well):

```suggestion
    for (const code of Object.keys(this.requiredImports)) {
```

addaleax (approved):

PR #3920, opened 2022-12-27
fix(compass-schema): fix name prop missing in multi type array field COMPASS-6397

Anemy:
COMPASS-6397 

Fixes a recent bug on `main` (not on stable release) where a document shown in the schema view would fail to render if it had an array field with multiple types in the array.
This PR typescripts a few more of the `compass-schema` components and adds the first few renderer unit tests to the package. This PR also fixes a bug where the sample values shown would not update when switching between a type that is in an array and a non-array type, this is shown in the videos below. 
I'll typescript more of the views and add more tests in later prs when I have the time.

**Before**

Screen Shot 2022-12-27 at 3 22 00 PM

https://user-images.githubusercontent.com/1791149/209717490-239ab2b6-ea9c-4110-85a2-18d97cfa7409.mp4

**After**

https://user-images.githubusercontent.com/1791149/209717485-6523546e-73f6-4aaf-b6c5-a1c7109109e8.mp4


lerouxb:
Can you explain what caused the bug and where/when it crept in?

Anemy:
@lerouxb Thanks for asking! I missed that in the description.

I introduced this bug in recent changes updating parts of the field to use LeafyGreen components. What caused the bug is that we were not accounting for the active type for a field to be undefined. Typescript would have caught the bug. In the changes we stopped using the object reference of the active selected type for a field's schema as a check for if the type is active, and instead started using the type name. Here's the line of the pr: https://github.com/mongodb-js/compass/pull/3897/files#diff-3aff6ae7d9dc37bfcf759bac92cc4641227e59d6113fa16d99e19f5201a8d8e6R101

lerouxb (approved):

PR #3928, opened 2023-01-03
chore: remove deprecated global styles, wrap home in body element COMPASS-5579

Anemy:
COMPASS-5579
We wrap our home element in a `Body` typography element so that height and font-size are standardized in Compass. Previously we were applying these through css, and slightly inconsistently as they would not apply to portal items and other places of Compass not wrapped in the component, like the settings page.


mcasimir (approved):
🥳

PR #3934, opened 2023-01-05
fix(telemetry): initialize before identify and use get-os-info from npm

mcasimir:
- Make sure to initialize the segment client before calling `identify`
- Use `get-os-info` from npm


addaleax (approved):

packages/compass-e2e-tests/tests/import-export-connections.test.ts
```diff
@@ -17,6 +17,9 @@ describe('Connection Import / Export', function () {
   let originalDisableKeychainUsage: string | undefined;
   let telemetry: Telemetry;
 
+  const getTrackedEvents = (): any[] =>
+    telemetry.events().filter((e: any) => e.type === 'track');
```
addaleax, line 21: We _could_ have a regression test using the inverse condition, i.e. make sure that
```
telemetry.events().filter((e: any) => e.type === 'identify').length === 1;
```

I think the existing test in logging.test.ts doesn’t catch this because it’s starting Compass in first-run mode and then immediately enabling telemetry, which doesn’t cause the same issue that this PR is fixing.

That’s not blocking though, if you want you can just merge this and I’ll add it later.

addaleax (approved):

packages/compass-e2e-tests/tests/import-export-connections.test.ts
mcasimir: ah interesting, yeah that's a good idea, I'll follow up with that. Was wondering what makes the other test pass

addaleax: Well yeah, as I said, it runs Compass in first-run mode and then afterwards enables telemetry (by closing the welcome modal without further actions). That would not trigger the same code path as telemetry being enabled from the start.

PR #3935, opened 2023-01-05
feat(aggregations): add stage button between stages COMPASS-6382

mabaasit:
feat(aggregations): add stage button between stages COMPASS-6382

image
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-builder-ui-workspace.jsx
```diff
@@ -7,11 +7,25 @@ import AddStage from '../add-stage';
...
 import { sortableContainer, sortableElement } from 'react-sortable-hoc';
+import { css } from '@mongodb-js/compass-components';
 
 import styles from './pipeline-builder-ui-workspace.module.less';
 
+const stageContainerStyles = css({
+  display: 'flex',
+  flexDirection: 'column',
+  '&.dragging .add-stage-button': {
+    display: 'none'
```
mabaasit, line 18: hide add stage button when dragging the stage.

lerouxb:
We seem to have lost some space between the input keyline card and the add stage button in the empty state somewhere along the way:

Screenshot 2023-01-05 at 15 57 02

Not sure if it was this PR or maybe caused by me in recent PRs.

lerouxb:
The button is only appearing after the first stage. I think the design / plan was to add it before each stage, ie. before the first one as well so you can use the button to add a stage at the start.

Screenshot 2023-01-05 at 15 59 24


Are you planning on removing the old + button that's in the toolbar in a separate PR?

packages/compass-aggregations/src/components/add-stage/add-stage.tsx
```diff
@@ -0,0 +1,61 @@
...
+        <div className={linkContainerStyles}>
+          <Link href='https://www.mongodb.com/docs/manual/reference/operator/aggregation-pipeline/'>Learn more about aggregation pipeline stages</Link>
+        </div>
+      </div>
+    );
+  }
+
+  return (
+    <div className={containerStyles}>
+      <IconButton aria-label='Add stage' data-testid="add-stage" onClick={() => onAddStageClick(index)}>
```
lerouxb, line 49: We should probably add a title="Add stage" attribute here.

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-builder-ui-workspace.jsx
lerouxb: When you start dragging a stage there's a jump and I wonder if it is caused by this. Maybe we need a better way to hide it like visibility: 'hidden' or if for some reason that doesn't work then opacity: 0 and pointer-events: 'none'?

Something that will cause it to still take up space.

mabaasit: not sure what part jumped, when dropping the stage, did stage jump or the other content of the page?
but, filling up the whole space is a better idea. i added that

lerouxb (approved):

PR #3941, opened 2023-01-06
chore(connection-form): add placeholder for host input

Anemy:
| before | after |
| - | - |
| Screen Shot 2023-01-05 at 8 40 26 PM | Screen Shot 2023-01-05 at 8 42 17 PM |
|  | Screen Shot 2023-01-05 at 8 42 23 PM |


mcasimir:
If I recall correctly, not having placeholders was an intentional decision

Anemy:
sgtm!

PR #3943, opened 2023-01-06
test(telemetry): add a test for identify calls after first run

mcasimir:


addaleax (approved):

packages/compass-e2e-tests/tests/logging.test.ts
```diff
@@ -371,6 +371,29 @@ describe('Logging and Telemetry integration', function () {
...
+    let telemetry: Telemetry;
+
+    before(async function () {
+      telemetry = await startTelemetryServer();
+      compass = await beforeTests();
+    });
+
+    after(async function name() {
+      await telemetry.stop();
+      await afterTests(compass);
```
addaleax, line 385: Without having dug into the test failures in a ton of detail, I think this *might* fix it and would probably be necessary in order for this test not to be flaky anyway:

```suggestion
      await afterTests(compass);
      await telemetry.stop();
```

mcasimir: Thanks!!

packages/compass/src/main/telemetry.ts
```diff
@@ -79,7 +81,8 @@ class CompassTelemetry {
     });
   }
 
-  private static async identify() {
+  // Keep this method synchronous to avoid race conditions.
+  private static identify() {
```
addaleax, line 85: @mcasimir I looked into why this was failing, turns out the `fs.readFile` call that `getOsInfo` makes on linux delays the `this.analytics.identify()` call so much that it only happens after the first `.track()` event.

PR #3946, opened 2023-01-06
fix(settings): tweak settings modal text alignment COMPASS-6413 

lerouxb:
before:

image (4)

after:

Screenshot 2023-01-06 at 17 25 56

### DIFF
--- a/packages/compass-settings/src/components/settings/theme.tsx
+++ b/packages/compass-settings/src/components/settings/theme.tsx
@@ -29,6 +29,19 @@ const checkboxStyles = css({
   marginBottom: spacing[3],
 });
 
+const radioBoxStyles = css({
+  div: {
+    textAlign: 'left',
+    padding: spacing[3],
+    justifyContent: 'flex-start',
+  },
+});
+
+const themePreviewStyles = css({
+  marginRight: spacing[2],
+  maxWidth: '128px',
+});
+
 function LightThemePreview() {
   return (
     <svg
@@ -36,6 +49,7 @@ function LightThemePreview() {
       xmlnsXlink="http://www.w3.org/1999/xlink"
       fill="none"
       viewBox="0 0 171 119"
+      className={themePreviewStyles}
     >
       <rect width="171" height="119" fill="#fff" rx="6" />
       <rect width="42" height="119" fill="#F9FBFA" rx="6" />
@@ -58,6 +72,7 @@ function DarkThemePreview() {
       xmlnsXlink="http://www.w3.org/1999/xlink"
       fill="none"
       viewBox="0 0 171 119"
+      className={themePreviewStyles}
     >
       <rect width="171" height="119" fill="#3D4F58" rx="6" />
       <rect width="42" height="119" fill="#1C2D38" rx="6" />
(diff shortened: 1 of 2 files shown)


packages/compass-settings/src/components/settings/theme.tsx
```diff
@@ -29,13 +29,27 @@ const checkboxStyles = css({
...
+  div: {
+    textAlign: 'left',
+    padding: spacing[3],
+    justifyContent: 'flex-start',
+  },
+});
+
+const themePreviewStyles = css({
+  marginRight: spacing[2],
+  maxWidth: '128px',
```
lerouxb, line 42: Kinda arbitrary, but should leave enough space for the text to the right of it in both cases and at all the sizes we support.

packages/compass-settings/src/components/settings/theme.tsx
```diff
@@ -29,13 +29,27 @@ const checkboxStyles = css({
   marginBottom: spacing[3],
 });
 
+const radioBoxStyles = css({
+  div: {
```
lerouxb, line 33: Bit of a hack, but I think so is sticking an image plus text in here :)

lerouxb: We could have more control by only having an image as the content of the radio button. Maybe a task for a follow-up.

packages/compass-settings/src/components/sidebar.tsx
```diff
@@ -54,7 +53,8 @@ const SettingsSideNav: React.FunctionComponent<SidebarProps> = ({
       aria-labelledby="modal-title"
     >
       {items.map((item) => (
-        <Button
+        <button
```
lerouxb, line 56: I don't think the settings navigation should have been leafygreen buttons anyway, because they just don't look like them. Same thing in the main / connetions sidebar. Menu items are just too custom.

packages/compass-settings/src/components/settings/theme.tsx
```diff
@@ -120,10 +135,12 @@ export const ThemeSettings: React.FunctionComponent<ThemeSettingsProps> = ({
           id="theme-selector"
           onChange={handleSelectorChange}
           value={themeValue}
+          size="full"
```
lerouxb, line 138: There's compact, default and full. Full leaves enough space for the image and text. I think we'll have the best control if we just don't have separate text or integrate it into the image.

packages/compass-settings/src/components/settings/theme.tsx
```diff
@@ -29,13 +29,27 @@ const checkboxStyles = css({
...
+  div: {
+    textAlign: 'left',
+    padding: spacing[3],
+    justifyContent: 'flex-start',
+  },
+});
+
+const themePreviewStyles = css({
+  marginRight: spacing[2],
+  maxWidth: '50%',
```
lerouxb, line 42: Kinda arbitrary, but should leave enough space for the text to the right of it in both cases and at all the sizes we support.

packages/compass-settings/src/components/sidebar.tsx
Anemy: Small thing, can we update the commend on line 26 since it's no longer LeafyGreen? https://github.com/mongodb-js/compass/pull/3946/files#diff-e8dc5dba7d636d3aa573305a44ec9b4fafc28e1977a7a16e34e62afddc44ee5aR26

addaleax (approved):

PR #3958, opened 2023-01-10
testing to see if analyzeSchema() could be used for gathering the list of fields for the select fields table

lerouxb:
Just a test implementation for use in Import.gatherFields()

Not crazy about how mongodb-schema goes through great lengths to stream yet schema-analysis in here loads all the docs from the entire sample into memory at once, but I suppose this will do for now.

### DIFF
--- a/packages/compass-schema/src/modules/schema-analysis.spec.js
+++ b/packages/compass-schema/src/modules/schema-analysis.spec.js
@@ -21,7 +21,8 @@ function schemaToPaths(fields, parent = []) {
     if (array) {
       const arrayDoc = array.types.find((f) => f.bsonType === 'Document');
       if (arrayDoc) {
-        paths.push(...schemaToPaths(arrayDoc.fields, [...path, '*']));
+        //paths.push(...schemaToPaths(arrayDoc.fields, [...path, '*']));
+        paths.push(...schemaToPaths(arrayDoc.fields, path));
       }
     }
   }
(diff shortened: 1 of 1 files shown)


packages/compass-schema/src/modules/schema-analysis.spec.js
```diff
@@ -3,6 +3,32 @@ import bson from 'bson';
...
+    if (doc) {
+      paths.push(...schemaToPaths(doc.fields, path));
+    }
+
+    // recurse on array
+    const array = field.types.find((f) => f.bsonType === 'Array');
+    if (array) {
+      const arrayDoc = array.types.find((f) => f.bsonType === 'Document');
+      if (arrayDoc) {
+        paths.push(...schemaToPaths(arrayDoc.fields, [...path, '*']));
```
lerouxb, line 24: I suppose the `*` isn't technically necessary and we can just know that path components inside arrays mean object fields later when we build the json doc or csv row. Undecided about it.

packages/compass-schema/src/modules/schema-analysis.spec.js
```diff
@@ -169,5 +195,46 @@ describe('schema-analyis', function () {
...
+        {}
+      );
+
+      console.dir(schema, { depth: Infinity });
+
+      const paths = schemaToPaths(schema.fields);
+      console.dir(paths, { depth: Infinity });
+
+      /*
+        [
```
lerouxb, line 229: One potential problem seems to be that we lost the field order so we can't loop through these when building the json doc or csv row - we'd have to loop through the document that came from the database or the csv fields in teh order as determined when we write the header row.

PR #3961, opened 2023-01-11
fix(compass-crud): don't show negative count on delete when no document count COMPASS-5996

Anemy:
COMPASS-5996

before

https://user-images.githubusercontent.com/1791149/211903316-92292ea9-77f1-4b3d-8ec1-e371dd02b9e7.mp4

after

https://user-images.githubusercontent.com/1791149/211903339-65b3ba40-446b-49ad-a196-e35138de2bdf.mp4


Anemy:
Looks like a legit failure. Interesting it's only on windows. Will look more when back from vacation, if anyone wants to pick this up before then please feel free.

mabaasit (approved):

PR #3964, opened 2023-01-12
fix(crud): If a date is in the safe range, go with relaxed EJSON rather than canonical COMPASS-5744

lerouxb:
Some reference:

https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#datetime
https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#conversion-table
https://github.com/mongodb/js-bson/blob/35bb2a279343439e1306a0eeb8d65d1abf910929/test/node/specs/bson-corpus/datetime.json#L9-L10

### DIFF
--- a/packages/hadron-document/src/utils.ts
+++ b/packages/hadron-document/src/utils.ts
@@ -1,6 +1,8 @@
 import { EJSON } from 'bson';
 import type { TypeCastMap, TypeCastTypes } from 'hadron-type-checker';
 
+const maxFourYearDate = new Date('9999-12-31T23:59:59.999Z').valueOf();
+
 export function fieldStringLen(value: unknown): number {
   const length = String(value).length;
   return length === 0 ? 1 : length;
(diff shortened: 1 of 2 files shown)


packages/hadron-document/src/utils.ts
```diff
@@ -70,6 +76,12 @@ function makeEJSONIdiomatic(value: EJSON.SerializableTypes): void {
       }
       continue;
     }
+    if (entry.$date && entry.$date.$numberLong) {
+      const number = entry.$date.$numberLong;
+      if (number >= 0 && number <= maxFourYearDate) {
+        entry.$date = new Date(+number);
```
addaleax, line 82: Should this be

```suggestion
        entry.$date = new Date(+number).toISOString();
```

?

addaleax: (also, you may want to add tests for this)

lerouxb: I don't know. It seems to work. I suppose I can just add that to be safe.

lerouxb: Yeah I see this file has no tests yet :(

addaleax: It does: https://github.com/mongodb-js/compass/blob/03211164e94cb45075ead41b2090cacd99a71301/packages/hadron-document/test/document.test.ts#L2317-L2382

lerouxb: Added now.

addaleax (approved):

PR #3972, opened 2023-01-16
feat(collections): when dropping a collection or database, redirect to either the database or databases view COMPASS-6018, COMPASS-6434 

lerouxb:
This should now only redirect if it is the active collection or database being dropped.

If you delete the last collection in the database it will redirect you to the instance's databases view. Otherwise it will redirect to the database's list of collections.


packages/compass-e2e-tests/tests/instance-sidebar.test.ts
```diff
@@ -204,5 +211,12 @@ describe('Instance sidebar', function () {
 
     // wait for it to be gone
     await collectionElement.waitForExist({ reverse: true });
+
+    // the app should have redirected to the the database Collections tab
+    // because we were on the collection Docuemnts tab and the database has
```
addaleax, line 216:
```suggestion
    // because we were on the collection Documents tab and the database has
```

addaleax (approved):

lerouxb:
Parking this for now. Will come back to it next time I'm on backlog work.

PR #3974, opened 2023-01-17
feat(databases): when dropping a database, redirect to the databases view COMPASS-6434

lerouxb:
Also includes e2e test changes to make sure this works.

### DIFF
--- a/packages/compass-app-stores/src/stores/instance-store.js
+++ b/packages/compass-app-stores/src/stores/instance-store.js
@@ -265,6 +265,10 @@ store.onActivated = (appRegistry) => {
     appRegistry.emit('select-namespace', metadata);
   });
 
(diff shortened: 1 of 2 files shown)


packages/compass-app-stores/src/stores/instance-store.js
```diff
@@ -265,6 +265,10 @@ store.onActivated = (appRegistry) => {
     appRegistry.emit('select-namespace', metadata);
   });
 
+  appRegistry.on('database-dropped', async() => {
```
addaleax, line 268: Does this event also fire when another database/collection is open and the dropped database was just a random one from the sidebar?

lerouxb: So yeah you could currently be viewing a different database. Don't know if/how we could detect that.

PR #3986, opened 2023-01-22
Update the display to Projection

vrajesh27:
## Description
Project might be a little misleading when shown in UI when it actually means "Projection".

### Checklist

## Motivation and Context

- [x] Misc

## Types of changes


mcasimir:
Hi @vrajesh27, thanks for the suggestion and PR! We will discuss this further internally and try to understand if using "projection" makes more sense or if we should update our documentation instead. We are going to close this one for now and keep the wording to `project` since it mirrors the query API syntax.

PR #3995, opened 2023-01-24
feat(compass-explain-plan): adds a cancellable loader to explain

himanshusinghs:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-explain-plan/src/modules/explain.js
```diff
@@ -389,6 +396,8 @@ export const fetchExplainPlan = (query) => {
       explain.resultId = resultId();
       explain.error = error;
       dispatch(explainPlanFetched(explain));
```
mabaasit, line 398: In other tabs, we hide the content when there's an error. we can have similar behaviour here as well?

packages/compass-explain-plan/src/modules/explain.js
```diff
@@ -406,6 +415,22 @@ export const changeExplainPlanState = (explainState) => {
...
+export const startExplainPlan = () => {
+  return (dispatch) => {
+    const abortController = new AbortController();
+    return dispatch(explainStateChanged(EXPLAIN_STATES.REQUESTED, abortController));
+  };
+};
+
+export const cancelExplainPlan = () => {
+  return (dispatch, getStore) => {
+    getStore().explain.abortController?.abort();
```
gribnoysup, line 430: Because you are aborting without specifying a `reason` and not handling cancellation as some special case as soon as you merge this branch with main (which updates electron), you'll start getting a pretty garbage message. I would suggest to either explicitly handle abort state as [we do in aggregation plugin](https://github.com/mongodb-js/compass/blob/e4dda02d4c291f949fd862ac0347beb1b12f86d7/packages/compass-aggregations/src/modules/aggregation.ts#L235-L251) or at least explicitly provide an error here to make sure that the message you will get in UI is controlled by the code here, [similar to compass-crud](https://github.com/mongodb-js/compass/blob/c6cf421f51c55178ef005cec77dcc0fc37494b4a/packages/compass-crud/src/stores/crud-store.ts#L1395)

Also as we are keeping the controller in the store, it would be nice to clean it up after it was used

gribnoysup (commented):
We probably don't need to show the "Summary" in case when user canceled the operation?

image

Maybe something that should be true for other errors too

packages/compass-explain-plan/src/components/explain-states/explain-states.jsx
```diff
@@ -136,29 +147,42 @@ class ExplainStates extends Component {
...
-            switchToJSONView={this.props.switchToJSONView}
-            viewType={this.props.explain.viewType}
-          />
-        }
-      >
-        {this.renderZeroState()}
-        {this.renderContent()}
-      </WorkspaceContainer>
-    );
+      isExplainLoading
```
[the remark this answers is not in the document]
mcasimir: That would be nice yeah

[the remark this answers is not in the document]
himanshusinghs: Thanks, I overlooked that aspect. It is now in the new changes

packages/compass-explain-plan/src/components/explain-states/explain-states.jsx
```diff
@@ -136,29 +147,42 @@ class ExplainStates extends Component {
...
+            <CancelLoader
+              data-testid="query-explain-cancel"
+              cancelText="Cancel"
+              onCancel={() => this.props.cancelExplainPlan()}
+              progressText="Running explain"
+            />
+          </div>
+        : (
+        <WorkspaceContainer
+          className='boys-2'
```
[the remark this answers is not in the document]
himanshusinghs: Good catch, definitely not intended 😅

packages/compass-explain-plan/src/modules/explain.js
himanshusinghs: So I looked at the document tab and schema analysis tab as starting points:
- On error both of them display the error banner, content is gone (or not there already)
- On cancel of operation, document tab displays an error banner with content "Operation aborted" but the schema analysis tab simply switches back to zero content screen which I guess is the welcome page of the feature itself. I found the latter more clear and was also something that I guess Rhys was trying to mention in the comment [here](https://jira.mongodb.org/browse/COMPASS-4620?focusedCommentId=4354087&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4354087).

To move forward I have adapted the behaviour that is there in schema analysis tab - switching back to zero content. Let me know how you feel about that?

gribnoysup: FWIW resetting to the initial state becomes confusing when you already have results in the view before re-running, because if it's not an error that we are showing I would expect it to go back to the previous state, not the initial empty one. We can add this as a discussion point for the design sync, would be indeed nice to decide on one approach across views.

packages/compass-explain-plan/src/modules/explain.js
mcasimir: I have a PR that would probably simplify that a bit already, moving the tree-stages.js out of the store, so the store would only need the `explain`, would it make sense to merge this and do the change in a follow-up? I'd be in favor of cleaning up explain.js, seems unnecessarily convoluted for a store that at the moment have only one action.

himanshusinghs: If you meant merging your PR first and then adapting this PR accordingly (if needed) then I am good that.

packages/compass-explain-plan/src/modules/explain.js
himanshusinghs: Design team agreed on having the behaviour same as what we have at the moment in aggregation tab, also documented [here](https://jira.mongodb.org/browse/EXPO-2181?focusedCommentId=5160056&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-5160056). Required changes are applied 👍

packages/compass-explain-plan/src/modules/explain.js
himanshusinghs: User cancelling the explain is now a more explicit action - and restores the previous state (which ever it was).

packages/compass-explain-plan/src/modules/explain.js
```diff
@@ -320,16 +348,25 @@ export const fetchExplainPlan = (query) => {
       return;
     }
 
+    const abortController = new AbortController();
```
gribnoysup, line 351: Before starting a new fetch, you probably want to check whether another one is already running. Otherwise you can put this tab in a weird state by clicking "Explain" button multiple times

packages/compass-e2e-tests/tests/collection-explain-plan-tab.test.ts
```diff
@@ -38,4 +44,57 @@ describe('Collection explain plan tab', function () {
     const stages = await browser.$$(Selectors.ExplainStage);
     expect(stages).to.have.lengthOf(1);
   });
+
+  it('shows a loading state while explain is running', async function () {
+    // Popuplate existing collection with some more data
+    await createNumbersCollection(collectionName, 10_00_000);
```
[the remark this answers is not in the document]
himanshusinghs: Oh that's perfect, thanks! Will update it 👍

packages/compass-explain-plan/src/modules/explain.js
himanshusinghs: Yea you're right. I just noticed that we don't have a disabled state for button for when the operation is running. In that case, I guess we can assume that user would like the result for his last click and hence we just cancel the previous explain.

packages/compass-explain-plan/src/modules/explain.js
```diff
@@ -291,6 +318,7 @@ export function isAggregationExplainOutput(explainOutput) {
 
 /**
  * Fetches the explain plan.
+ * TODO: Declutter this middleware as it is getting difficult to read
```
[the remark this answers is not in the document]
himanshusinghs: Makes sense - also I didn't do justice in explaining what that todo was meant to be.

I thought that it would be much clearer if this middleware would dispatch explicit actions like `explainPlanFetched`, `explainPlanFetchFailed`, `explainPlanParseFailed` (just some examples) instead of always dispatching `explainPlanFetched` with nested properties of explain modified in different cases. It was not clear to understand upfront what state certain properties (being changed) were related to.

packages/compass-explain-plan/src/modules/explain.js
himanshusinghs: changes implemented

gribnoysup (approved):

packages/compass-explain-plan/src/modules/explain.js
```diff
@@ -320,16 +348,28 @@ export const fetchExplainPlan = (query) => {
       return;
     }
 
+    // Cancel previous run of aggregation if there is any
+    explain.abortController?.abort();
+
+    const abortController = new AbortController();
+    const abortSignal = abortController.signal;
+    const oldExplain = { ...explain };
+    dispatch(startExplainPlan(abortController, oldExplain));
```
mabaasit, line 357: You can directly dispatch `explainStateChanged` here or simply 
```js
{
  type: EXPLAIN_STATE_CHANGED,
  explainState,
  abortController,
  oldExplain,
}```

Currently `startExplainPlan` calls `explainStateChanged`, which does the same thing. So you can avoid this and directly dispatch the action. 
Also in `doChangeExplainPlanState`, you can access the `oldExplain` from the state and don't have to pass it along.

himanshusinghs: thanks, it definitely make sense. changes implemented

himanshusinghs: ahh wait just to clarify - I have kept `explainStateChanged` as it was used in components and to respond to events on registry. For them it would still make sense to keep it as it makes it more clear what is happening

PR #4001, opened 2023-01-24
feat(import): Function to analyze a CSV file and autodetect the columns' types COMPASS-6423

lerouxb:
I haven't added the sample rows here like originally planned - I was thinking that it will be better to have that has part of listCSVFields() so it can appear while we run the full analysis. So I made a separate PR to add it there rather: https://github.com/mongodb-js/compass/pull/4003

We might add support for inferring more types soon based on looking for some patterns and we'll gain more interesting test fixtures once we can export complex.json. But I think this is the main structure I had planned.

I changed the result object's type a bit from what I had in the tech design. I updated it there to match.


packages/compass-import-export/src/import/analyze-csv-fields.ts
```diff
@@ -0,0 +1,240 @@
...
+  totalRows: number;
+  fields: Record<string, CSVField>;
+};
+
+function detectFieldType(value: PapaValue): CSVFieldType {
+  // papaparse handles some types for us, so get those out of the way first
+  if (value === null) {
+    return 'null';
+  }
+  if (value instanceof Date) {
```
addaleax, line 75: This probably doesn’t break here, but I also see no reason to do the “real” brand check and use `Object.prototype.toString.call()` instead

packages/compass-import-export/src/utils/csv-header.ts
```diff
@@ -0,0 +1,3 @@
+export function csvHeaderNameToFieldName(name: string) {
+  return name.replace(/\[\d+\]/, '');
```
addaleax, line 2: Not a case I’d like to think about, but I guess the point of this project is to handle the complex stuff well … this should probably replace all of the array indices? :grimacing: 

```suggestion
  return name.replace(/\[\d+\]/g, '');
```

addaleax (commented):
Looks great so far!

packages/compass-import-export/src/import/analyze-csv-fields.ts
```diff
@@ -0,0 +1,240 @@
...
+
+function addRowToResult(
+  result: AnalyzeCSVFieldsResult,
+  headerFields: string[],
+  data: PapaRowData,
+  currentRow: Record<string, string>
+) {
+  for (const field of Object.values(result.fields)) {
+    for (const columnIndex of field.columnIndexes) {
+      const fieldName = headerFields[columnIndex];
```
lerouxb, line 141: ugh I just realised that I used the variable fieldName for two different things in two places which could be confusing so I'll rename them. in this one it is the field as it is in the csv header and in the other it is the field once it has been "normalised" ie. all array indexes removed.

packages/compass-import-export/src/import/analyze-csv-fields.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      const abortController = new AbortController();
+      const progressCallback = sinon.spy();
+      const result = await analyzeCSVFields({
+        input: fs.createReadStream(filepath),
+        delimiter: csvDelimiter,
+        abortSignal: abortController.signal,
+        progressCallback,
+      });
+
+      const resultPath = filepath.replace(/.csv$/, '.analyzed.json');
```
addaleax, line 30:
```suggestion
      const resultPath = filepath.replace(/\.csv$/, '.analyzed.json');
```

packages/compass-import-export/src/import/analyze-csv-fields.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      } catch (err) {
+        // This helps to tell you which file is missing and what the expected
+        // content is which helps when adding a new .csv fixture
+        console.log(resultPath);
+        console.log(JSON.stringify(result, null, 2));
+        throw err;
+      }
+
+      const expectedResult = JSON.parse(text);
+      expect(result, basename.replace(/.csv$/, '.analyzed.json')).to.deep.equal(
```
addaleax, line 43:
```suggestion
      expect(result, basename.replace(/\.csv$/, '.analyzed.json')).to.deep.equal(
```

packages/compass-import-export/src/import/analyze-csv-fields.ts
```diff
@@ -0,0 +1,243 @@
...
+  aborted: boolean;
+  fields: Record<string, CSVField>;
+};
+
+function detectFieldType(value: PapaValue): CSVFieldType {
+  // papaparse handles some types for us, so get those out of the way first
+  if (value === null) {
+    return 'null';
+  }
+  if (Object.prototype.toString.call(value) === '[object Date') {
```
addaleax, line 76:
```suggestion
  if (Object.prototype.toString.call(value) === '[object Date]') {
```

addaleax (approved):

packages/compass-import-export/src/import/analyze-csv-fields.spec.ts
lerouxb: oh wow. some days..

packages/compass-import-export/test/csv/types/double.csv
```diff
@@ -5,9 +5,7 @@ double,notes
 -1.2
 "1.0"
 "1.2"
-1e-15
+1e-15,a very small number close to 0
```
addaleax, line 8: Feel free to ignore, but you might want to add a test for some pathological cases like `1e0` here

addaleax (approved):

packages/compass-import-export/test/csv/types/double.csv
lerouxb: Cool that caught something. Keep them coming ;)

PR #4010, opened 2023-01-28
feat(focus-mode): resizable editor in focus mode COMPASS-6446

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


gribnoysup (commented):
Seems like with the header of the editor not wrapping you can break the layout (allowing flex wrap and setting max width for some elements there seems to help with this, I think editor container would also need to be similarly updated a bit to now overflow):

image

packages/compass-aggregations/src/components/focus-mode/focus-mode-stage-editor.tsx
```diff
@@ -9,9 +9,8 @@ import StageOperatorSelect from '../stage-toolbar/stage-operator-select';
 
 const containerStyles = css({
   height: '100%',
-  display: 'grid',
-  gridTemplateRows: 'min-content 1fr',
-  gridTemplateColumns: '1fr',
+  display: 'flex',
```
gribnoysup, line 12: If you are changing this to flex, then you should be explicit about what children you want to be flexible to avoid unpredictable behavior, in this case you probably want `headerStyles` to have flex set to `none` and `editorStyels` to have flex grow set to 1 (and you probably don't want to have height set to 100% there because this is 100% of the container, which doesn't account for the container)

packages/compass-aggregations/src/components/focus-mode/focus-mode-stage-editor.tsx
```diff
@@ -20,6 +19,13 @@ const headerStyles = css({
   justifyContent: 'flex-start',
   gap: spacing[2],
   marginLeft: spacing[2],
+  marginRight: spacing[2],
+  flexWrap: 'wrap',
+});
+
+const linkStyles = css({
+  whiteSpace: 'nowrap',
+  overflow: 'hidden',
```
gribnoysup, line 28: Any particular reason we are adding overflow hidden here? If you want to keep it, you need to add some padding on top (and bottom so that center alignment still works) too to account for the icon, otherwise it is cut when this link wraps:

image

packages/compass-aggregations/src/components/focus-mode/focus-mode.tsx
```diff
@@ -80,6 +90,97 @@ type FocusModeProps = {
   onCloseModal: () => void;
 };
 
+const INPUT_RESIZER_PROPS = {
```
gribnoysup, line 93: Feel free to keep it as is, but I am wondering why are we keeping this out of the actual render? Jumping back and forth between render method and prop definitions make it a bit awkward to read

gribnoysup (approved):
Overall looks and works good, left some very small comments here and there

packages/compass-aggregations/src/components/focus-mode/focus-mode.tsx
mabaasit: to avoid react. but i can use `useMemo` and have it there.

gribnoysup: Is this a recommended way to use this library or causing performance issues otherwise? I'm assuming that some of those objects being "re-created" every render would be handled by the component itself correctly and will not cause issues and a lot of these are primitive. I would suggest not to over-optimize like that unless you have a more or less real case for this

gribnoysup (approved):

PR #4022, opened 2023-02-01
feat(import): Function to import a CSV file COMPASS-6425

lerouxb:
I had to go back and change analyzeCSVFields() because I treated empty CSV fields (ie. empty strings) as `null` whereas I should have treated them as `undefined`. null maps to the bson type whereas what we really want is for those to be excluded from the document. Unless ignoreEmptyStrings is unchecked and then (for whatever reason) it should insert a blank string. The only way to get to null would be to import the string `'Null'` which is what compass would export a `null` as.

Worth pointing out again that `mongoexport` exports `null` as an empty string. But if we interpret every empty string as an explicit `null` value, then many imports will result in collections full of `null`s.

There are now tons of tests for the ignoreEmptyStrings flag when analysing and importing.

I also did quite a bit of refactoring of moving constants, types and utility functions around.

Maybe I should try and split that out into a separate PR?

--

Next steps (after this PR):

* Fix WritableCollectionStream to actually stop on errors when it should. Maybe also turn it into a read and write stream so we can pipe the errors / status somewhere.

* Port existing import code to use the new function including the existing tests to make sure that we pass existing regression tests. (ie. cover all the previously known edge cases)


packages/compass-import-export/src/import/import-csv.ts
```diff
@@ -0,0 +1,122 @@
...
+    header: true,
+    transformHeader: function (header: string, index: number): string {
+      debug('importCSV:transformHeader', header, index);
+      headerFields.push(header);
+      return header;
+    },
+  });
+
+  return new Promise((resolve, reject) => {
+    pipeline(input, parseStream, docStream, collectionStream, function (err) {
```
lerouxb, line 105: Apparently pipeline can return a promise and take the abort signal in a last-param options hash, but typescript is NOT happy if I try and do that. My best guess is that the type of collectionStream is confusing something.

packages/compass-import-export/src/import/analyze-csv-fields.ts
```diff
@@ -212,11 +154,11 @@ export function analyzeCSVFields({
           initResultFields(result, headerFields);
         }
 
-        addRowToResult(result, headerFields, results.data);
+        addRowToResult(result, headerFields, results.data, ignoreEmptyStrings);
 
         ++result.totalRows;
 
-        progressCallback(result.totalRows);
+        progressCallback && progressCallback(result.totalRows);
```
addaleax, line 161:
```suggestion
        progressCallback?.(result.totalRows);
```

packages/compass-import-export/src/import/import-csv.ts
addaleax: I think you’d want the promise version of `pipeline` from `import('stream').pipeline`, not the one from `import('stream')`: https://nodejs.org/api/stream.html#streams-promises-api (i.e. not https://nodejs.org/api/stream.html#streampipelinesource-transforms-destination-callback)

packages/compass-import-export/src/utils/csv.ts
```diff
@@ -0,0 +1,310 @@
...
+          // inside an object or a blank string at the top level that is the
+          // name of an object.field.
+          parts.push({ type: 'field', name: '' });
+        }
+        snippet = [];
+        continue;
+      }
+    } else {
+      if (char === ']') {
+        parts.push({ type: 'index', index: parseInt(snippet.join(''), 10) });
```
addaleax, line 279: Maybe rather

```suggestion
        parts.push({ type: 'index', index: +snippet.join('') });
```

+ some handling for `NaN` results? `parseInt()` cuts off processing at the first invalid character, if there are any, so that’s probably something we should handle here, and some invalid strings would result in `NaN` with either conversion method

packages/compass-import-export/src/utils/csv.ts
```diff
@@ -0,0 +1,310 @@
...
+  if (type === 'boolean') {
+    if (value === 'true' || value === 'TRUE') {
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  if (type === 'date') {
+    return new Date(value);
```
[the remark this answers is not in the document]
lerouxb: I guess. If we want to support that. I can add code for it.

packages/compass-import-export/src/utils/csv.ts
lerouxb: I'll add some unit tests for this function. Would have to throw and catch somewhere, make that kind of thing cause the header parsing to fail because there's not much we can do to recover from broken header strings like that.

packages/compass-import-export/src/import/import-csv.ts
lerouxb: Hmm still the same error. I'm just gonna go with what I have for now as it works.

addaleax: Btw, this compiles fine for me (it’s just a bit weird/annoying that it doesn’t consider the `signal` option to be optional in the options argument):

```diff
diff --git a/packages/compass-import-export/src/import/import-csv.ts b/packages/compass-import-export/src/import/import-csv.ts
index 63c5fc89cb65..f35d006a1c1a 100644
--- a/packages/compass-import-export/src/import/import-csv.ts
+++ b/packages/compass-import-export/src/import/import-csv.ts
@@ -1,4 +1,4 @@
-import { Transform, pipeline } from 'stream';
+import { Transform, promises as streamPromises } from 'stream';
 import type { Readable } from 'stream';
 import Papa from 'papaparse';
 import toNS from 'mongodb-ns';
@@ -10,6 +10,7 @@ import type { Delimiter, IncludedFields, PathPart } from '../utils/csv';
 import { makeDoc, parseHeaderName } from '../utils/csv';
 import { createDebug } from '../utils/logger';
 
+const { pipeline } = streamPromises;
 const debug = createDebug('import-csv');
 
 type ImportCSVOptions = {
@@ -26,7 +27,7 @@ type ImportCSVOptions = {
 
 type ImportCSVResult = CollectionStreamStats & { aborted?: boolean };
 
-export function importCSV({
+export async function importCSV({
   dataService,
   ns,
   input,
@@ -101,22 +102,17 @@ export function importCSV({
     },
   });
 
-  return new Promise((resolve, reject) => {
-    pipeline(input, parseStream, docStream, collectionStream, function (err) {
-      if (err) {
-        if (err.code === 'ERR_STREAM_PREMATURE_CLOSE') {
-          const result = {
-            ...collectionStream.getStats(),
-            aborted: true,
-          };
-          resolve(result);
-          return;
-        }
-        reject(err);
-        return;
-      }
+  try {
+    await pipeline(input, parseStream, docStream, collectionStream, abortSignal ? { signal: abortSignal } : undefined);
+  } catch (err: any) {
+    if (err.code === 'ERR_STREAM_PREMATURE_CLOSE') {
+      return {
+        ...collectionStream.getStats(),
+        aborted: true,
+      };
+    }
+    throw err;
+  }
 
-      resolve(collectionStream.getStats());
-    });
-  });
+  return collectionStream.getStats();
 }
```

lerouxb: Thanks! It was indeed the optional abortSignal that was confusing it.

packages/compass-import-export/src/import/import-csv.ts
```diff
@@ -0,0 +1,182 @@
...
+  const params = [
+    input,
+    parseStream,
+    docStream,
+    collectionStream,
+    ...(abortSignal ? [{ signal: abortSignal }] : []),
+  ] as const;
+
+  // This is temporary until we change WritableCollectionStream so it can pipe
+  // us its errors as they occur.
```
lerouxb, line 147: ☝️ Will tackle this when I fix the existing stopOnErrors bug in the collection write stream. In a separate PR.

Anemy (approved):
Left one question. Lgtm. Will be cool to start using this with the feature flags.

packages/compass-import-export/src/utils/csv.ts
```diff
@@ -0,0 +1,485 @@
...
+    // this supports the edge case where the field name is a blank string (either
+    // the whole field is a blank string or the last object field was a blank
+    // string)
+    parts.push({ type: 'field', name: '' });
+  }
+
+  return parts;
+}
+
+export function formatHeaderName(path: PathPart[]): string {
```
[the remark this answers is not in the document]
lerouxb: Currently in tests, but I pulled it out (I had a TODO) because export to CSV will need it.

PR #4033, opened 2023-02-03
chore(e2e): reduce e2e aggregation test flakyness, check for list visible

Anemy:
Hopefully reduces the `Collection aggregations tab` e2e test `supports paginating aggregation results` flakyness that we've been seeing:

https://evergreen.mongodb.com/task/10gen_compass_main_macos_arm_test_packaged_app_60x_enterprise_edc0f5aded54da149151f93256e5491cc203bb1a_23_02_02_15_41_54
https://evergreen.mongodb.com/task/10gen_compass_main_macos_arm_test_packaged_app_60x_enterprise_63dc5c971e2d17059312ecbe_23_02_03_01_00_07
https://evergreen.mongodb.com/task/10gen_compass_main_macos_arm_test_packaged_app_60x_enterprise_b3ec79156eca402d13b1499a9d0340f9cf6c3d50_23_02_02_10_23_52

I'm thinking after switching the views it would take a moment before the document json list contents was populated. If this doesn't do the trick we can check for the `DocumentJSONEntry` selector existing instead.

### DIFF
--- a/packages/compass-crud/src/components/document-json-view.tsx
+++ b/packages/compass-crud/src/components/document-json-view.tsx
@@ -81,7 +81,10 @@ class DocumentJsonView extends React.Component<DocumentJsonViewProps> {
    */
   render() {
     return (
-      <ol className={cx(LIST_CLASS, this.props.className)}>
+      <ol
+        className={cx(LIST_CLASS, this.props.className)}
+        data-testid="document-json-list"
+      >
         {this.renderDocuments()}
       </ol>
     );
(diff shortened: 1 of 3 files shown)


gribnoysup:
Doesn't hurt to add this wait, but I think worth noting that on the screenshot of the failure there is a visible server error for the aggregation pipeline, so there are really no documents when this fails. If we want to fix the flake, we probably should figure out why this happens.

Anemy:
Totally forgot to check the screenshots - thanks for catching! I'll reopen if I find a consistent reproduction and fix.

Anemy:
@gribnoysup Ah that screenshot is definitely helpful. Looks like what happens is the first stage is selected, and then while that stage's preview documents are being loaded the 2nd stage's stage selector is opened and scrolled. Then *sometimes* the documents are returned then, and if so when the new documents are rendered the scroll of the selector for the 2nd stage is resets to the top, where the `$bucket` stage is incorrectly selected.

PR #4043, opened 2023-02-07
feat(focus-mode): show guide cue COMPASS-6497

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


alenakhineika (approved):

packages/compass-aggregations/src/components/stage-toolbar/index.tsx
```diff
@@ -97,6 +102,8 @@ const DISABLED_TEXT = 'Stage disabled. Results not passed in the pipeline.';
 const COLLAPSED_TEXT =
   'A sample of the aggregated results from this stage will be shown below.';
 
+const GUIDE_CUE_DELAY = 700;
```
mcasimir, line 105: what is the delay for?

mcasimir (commented):
How did we solved the issue with the cue flying around if the aggregation tab is not visible?

packages/compass-aggregations/src/components/stage-toolbar/index.tsx
mabaasit: I added this delay to give some time for the content to load (Preview documents) and not show Guide Cue immediately.

mabaasit:
> Just a question, how did we solve the issue with the cue flying around if the aggregation tab is not visible?

We don't have that issue in this case. Adding a stage is only possible on this screen and once stage is added we show the Cue.  (by default the aggregation pipeline is empty, so no stage is rendered).

In the example I showed you, the content of the Aggregations tab is rendered even when Documents tab is active and I had added Cue in Aggregations toolbar.

gribnoysup:
I just loaded a saved pipeline and when cue popped up and I switched to another tab, it stayed on the screen it seems:

image

mabaasit:
> I just loaded a saved pipeline and when cue popped up I switched to another tab, it stayed on the screen it seems:

Nice catch. So, currently GuideCue doesn't hide it self as user clicks outside. I'll add an event handler for that. I reached out to Claudia regarding this and want input if we should hide it permanently (or just hide it on outside click and only hide it permanently only when user interacts with it either by clicking on **Got it** button or pressing **Esc**).

packages/compass-aggregations/src/components/stage-toolbar/index.tsx
mcasimir: did we consider other options too? what is the issue without the delay?

gribnoysup:
I don't think clicking outside is the only issue here, if I click while the saved aggregation is loading and there is no stages on the screen yet (but they are loaded afterwards), I get this state:

image

Another thing that caught my eye is that is that when the button is not in the viewport, we still show the popup:

image

I think instead of listening to some event that will tell you the active tab, we should rather use something like IntersectionObserver to detect whether the trigger node is visible or not at all, then it will be more universal. This also might be something that we right away add as a behavior to the guide cue inside compass components, so other places that use this don't need to reimplement this

And this is just an idea, but maybe you can then also keep the localStorage part of the logic inside an exported component, making a unique key some sort of a prop on the component so that similarly every other place that will be using it don't need to worry about the local storage part of it (but maybe this is a bit of a stretch)

mabaasit:
Thanks for the detailed comment. I just saw it (post my [3ddb129](https://github.com/mongodb-js/compass/pull/4043/commits/3ddb12960aa250a0f3232350a0bbad991839ad36) commit). 

Yes, using an intersection observer would solve this issue, but after the chat with design team, i made the fixes. The Cue is now in the proximity of the trigger (focus mode button). It scrolls now with the stage and also is only visible if the stage is in viewport.

gribnoysup:
All those are valid design considerations, from the technical perspective if you will be trying to show the popup when the trigger is not visible, you will always run into corner cases, this is for example the same case where I switched to another tab before popup was rendered, and then switched back to aggregation tab: 

image

mcasimir:
Is there something like `usePortal={false}` that could help?

packages/compass-aggregations/src/components/stage-toolbar/index.tsx
```diff
@@ -132,13 +183,32 @@ export function StageToolbar({
...
+              data-testid="focus-mode-guide-cue"
+              open={isGuideCueIntersecting && isGuideCueVisible}
+              setOpen={() => setGuideCueVisited()}
+              refEl={focusModeButtonRef}
+              numberOfSteps={1}
+              popoverZIndex={2}
+              scrollContainer={containerRef.current}
+              portalContainer={containerRef.current}
+              title="Focus Mode"
+              tooltipAlign="bottom"
```
gribnoysup, line 197: The way our logic for showing the cue (in particular the scroll container options we use so that the cue can scroll with the ref el) combines with the way leafygreen popups lock focus and thus auto scrolls the parent to fit the focused content leads to a pretty nasty behavior where as soon as we render the cue the current scroll position is completely reset:

Kapture 2023-02-09 at 18 59 40

([here's the minimal repro](https://codesandbox.io/s/restless-architecture-8tqrn8?file=/src/App.js) that I made)

I am not really sure that we can do anything about it here, other than either removing all the container props (but this means that the cue will not scroll) or removing the `topAlign` here, at least in that case the chances of focusable elements inside the cue not being on the screen and stealing the scroll position will be minimized?

packages/compass-aggregations/src/components/stage-toolbar/index.tsx
```diff
@@ -104,12 +111,56 @@ export function StageToolbar({
...
+    if (focusModeButtonRef.current) {
+      const observer = new IntersectionObserver(
+        (entries) => {
+          if (entries[0].isIntersecting) {
+            setIsGuideCueIntersecting(true);
+          } else {
+            setIsGuideCueIntersecting(false);
+          }
+        },
+        { threshold: 1 }
```
gribnoysup, line 140: I think we might want this to be a close to zero number instead of 1, otherwise the cue disappears even when the tiniest part of the icon is not on the screen anymore:

Kapture 2023-02-09 at 19 02 57

gribnoysup (approved):

PR #4060, opened 2023-02-13
fix(hardon-document): cancel edit on non-existent field COMPASS-6505

mabaasit:
When cancelling from editing a field in a document that's non-existent, the compass throws an error.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/hadron-document/src/element.ts
+++ b/packages/hadron-document/src/element.ts
@@ -699,7 +699,7 @@ export class Element extends EventEmitter {
    */
   revert(): void {
     if (this.isAdded()) {
-      this.parent!.elements!.remove(this);
+      this.parent?.elements?.remove(this);
       this._bubbleUp(Events.Removed, this, this.parent);
       delete (this as any).parent;
     } else {
--- a/packages/hadron-document/test/element.test.ts
+++ b/packages/hadron-document/test/element.test.ts
@@ -1760,6 +1760,22 @@ describe('Element', function () {
         });
       });
     });
+
+    context('when the element parent is not defined', function () {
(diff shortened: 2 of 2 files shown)


addaleax (approved):
Should this have a test?

addaleax (approved):

packages/hadron-document/src/element.ts
```diff
@@ -699,7 +699,7 @@ export class Element extends EventEmitter {
    */
   revert(): void {
     if (this.isAdded()) {
-      this.parent!.elements!.remove(this);
+      this.parent?.elements?.remove(this);
       this._bubbleUp(Events.Removed, this, this.parent);
```
mcasimir, line 703: is it ok that `this._bubbleUp` is triggered anyway?

packages/hadron-document/src/element.ts
```diff
@@ -699,7 +699,7 @@ export class Element extends EventEmitter {
    */
   revert(): void {
     if (this.isAdded()) {
-      this.parent!.elements!.remove(this);
+      this.parent?.elements?.remove(this);
```
mcasimir, line 702: just a question, in which case do we end up with a document that doesn't exist?

mcasimir (approved):
🔥

packages/hadron-document/src/element.ts
mabaasit: i tried really hard to find the root cause, but was not. I am giving up and will merge this PR.

PR #4064, opened 2023-02-14
fix(import): Stop on errors when stopOnErrors is true COMPASS-6518

lerouxb:
To test this you need an error coming from the database when docs get written and the easiest way is to add validation to a collection: https://www.mongodb.com/docs/manual/core/schema-validation/specify-json-schema/#std-label-schema-validation-json

Then just import a file. Try with the checkbox checked and unchecked. Seems to work for me in that the error shows in a banner and compass does not blow up.

I also rewrote the tests so that it will test with and without errors, with and without FLE2, with and without stopOnErrors. All of them run multiple batches (2 batches + 1 doc) to try and hit edge cases.


packages/compass-import-export/src/utils/collection-stream.ts
```diff
@@ -223,7 +227,22 @@ export class WritableCollectionStream extends Writable {
...
+  _makeStreamError(): CollectionStreamError | undefined {
+    if (this.stopOnErrors && this._errors.length) {
+      const error = this._errors[0];
+      if (error instanceof Error) {
+        return error;
+      }
+      return {
+        name: 'CollectionStreamError',
+        message: 'Something went wrong while writing data to a collection',
+        original: error,
```
addaleax, line 242: Would it make sense to name this `cause` instead of `original` to be consistent with the standard JS error property?

packages/compass-import-export/src/utils/collection-stream.ts
```diff
@@ -223,7 +227,22 @@ export class WritableCollectionStream extends Writable {
...
 
-    return callback();
+    return callback(this._makeStreamError());
+  }
+
+  _makeStreamError(): CollectionStreamError | undefined {
+    if (this.stopOnErrors && this._errors.length) {
+      const error = this._errors[0];
+      if (error instanceof Error) {
+        return error;
```
addaleax, line 237:
```suggestion
      if (Object.prototype.toString.call(error) === '[object Error]') {
        return error as Error;
```

?

addaleax (approved):

packages/compass-import-export/src/utils/collection-stream.ts
lerouxb: Sure, but I do wonder if/how that could ever break.

packages/compass-import-export/src/utils/collection-stream.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+  isFLE: boolean,
+  stopOnErrors: boolean
+) {
+  if (stopOnErrors) {
+    return 1;
+  }
+
+  if (isFLE) {
+    // one attempted batch at the batch size (followed by insertOne() on retry), then subsequent batches are all size 1.
+    return numDocs > BATCH_SIZE ? 1 + numDocs - BATCH_SIZE : 1;
```
lerouxb, line 90: ☝️ this shows how often the progress event can be called for FLE2. And that becomes a call straight to dispatch() here: https://github.com/mongodb-js/compass/blob/6f0050b373c609355b053aa3c643c494b6177eb5/packages/compass-import-export/src/modules/import.ts#L290-L292

PR #4070, opened 2023-02-16
feat(import): importJSON() COMPASS-6424

lerouxb:
This works for json files (ie. an array of docs) and also for json streams / json lines. ie. either one doc per line or just one doc after another regardless of whitespace. Notably this fixes the old ticket where you can't just import one doc (ie. object) formatted over multiple lines like what you would have in insert document.

For importing CSV I think the progress event worked fine because you could see how many docs have been processed and an earlier step already returned the number of docs when we analysed the file. This is enough to accurately show the progress as designed. For importing JSON we don't know how many docs there are in advance. So I'll have to augment the progress callback with how many bytes have been processed. Then the progress can be the number of bytes out of the total. I'll do that in a follow-up PR because i'll add it to both importCSV() and importJSON() so they can remain compatible. ie. you would have the number of docs and number of bytes processed. Total bytes in the file can be read outside the function and for CSV we have the additional advantage that we already have the total number of docs.

Other than that this is pretty straight forward since stream-json and bson's EJSON already does all the heavy lifting.


packages/compass-import-export/src/import/import-json.ts
```diff
@@ -0,0 +1,114 @@
...
+    ...(jsonVariant === 'json'
+      ? [Parser.parser(), StreamArray.streamArray()]
+      : []),
+    docStream,
+    collectionStream,
+    ...(abortSignal ? [{ signal: abortSignal }] : []),
+  ] as const;
+
+  try {
+    await pipeline(...params);
```
lerouxb, line 91: ☝️ typescript is not happy here. The tests pass, though..

packages/compass-import-export/src/import/import-csv.spec.ts
```diff
@@ -634,6 +636,45 @@ describe('importCSV', function () {
...
+          required: ['xxx'],
+        },
+      },
+    });
+
+    const promise = importCSV({
+      dataService,
+      ns,
+      fields,
+      input: Readable.from(lines.join('\n')),
```
addaleax, line 665: Somewhat unrelated random thought: Do we have a test for files with \r\n line endings somewhere?

packages/compass-import-export/src/import/import-json.spec.ts
```diff
@@ -0,0 +1,544 @@
...
+
+        const totalRows = progressCallback.callCount;
+
+        expect(stats).to.deep.equal({
+          nInserted: totalRows,
+          nMatched: 0,
+          nModified: 0,
+          nRemoved: 0,
+          nUpserted: 0,
+          ok: Math.ceil(totalRows / 1000),
```
himanshusinghs, line 103: Just out of curiosity, why this?

addaleax (approved):

himanshusinghs (approved):
Looking great 👍

packages/compass-import-export/src/import/import-csv.spec.ts
lerouxb: I've just assumed that it works because that's up to the parser lib to interpret. But I can add some more tests. Should probably add ones for BOM characters as well and weird non-utf8 charsets just to see.

packages/compass-import-export/src/import/import-json.spec.ts
lerouxb: The previously existing WritableCollectionStream has `ok` has the number of batches processed and the batch size is currently hardcoded as 1000.

lerouxb: (What goes into that object is kinda the responsibility of WritableCollectionStream and its own tests, I just include this here as a sanity check and because I made its output also be the output of importJSON() so it is useful to have something fail if that changes because it would also change this interface.)

packages/compass-import-export/src/import/import-csv.spec.ts
lerouxb: Actually I think I'm gonna merge this PR then add tests for that in a separate one because I probably need those kinds of tests for importCSV, guessFileType and analyzeCSVFields as well. So I'll just group it together.

packages/compass-import-export/src/import/import-json.spec.ts
himanshusinghs: I see. Thanks for explanation :)

PR #4075, opened 2023-02-20
test(import): import tests for windows line endings, non-utf8 encodings, BOM chars COMPASS-6529

lerouxb:


packages/compass-import-export/src/import/import-csv.spec.ts
```diff
@@ -780,6 +781,101 @@ describe('importCSV', function () {
...
+
+    expect(docs).to.have.length(3);
+
+    for (const doc of docs) {
+      expect(Object.keys(doc)).to.deep.equal(['_id', 'value']);
+    }
+  });
+
+  it('does not mind if a file is not valid utf8', async function () {
+    const latin1Buffer = buffer.transcode(
```
lerouxb, line 817: A bit tricky and I don't know if this is the best way to get a file that's in a different encoding, but it seems that import at least doesn't totally fall over in that case, so probably good enough?

packages/compass-import-export/src/import/analyze-csv-fields.spec.ts
```diff
@@ -180,4 +182,40 @@ describe('analyzeCSVFields', function () {
...
+    });
+    expect(Object.keys(result.fields)).to.deep.equal(['_id', 'value']);
+  });
+
+  it('does not mind if a file is not valid utf8', async function () {
+    const latin1Buffer = buffer.transcode(
+      Buffer.from('ê,foo\n1,2'),
+      'utf8',
+      'latin1'
+    );
```
addaleax, line 202: This is equivalent to:

```suggestion
    const latin1Buffer = Buffer.from('ê,foo\n1,2', 'latin1');
```

packages/compass-import-export/src/import/import-csv.spec.ts
addaleax: > but it seems that import at least doesn't totally fall over in that case, so probably good enough?

It will break if the file is in ASCII-incompatible encodings like UTF-16.

I think ideally we’d _want_ to break on invalid UTF-8 as input. I can’t say for sure but I’d expect that 99 % of the time users would rather have no data or only partial data imported rather than broken data.

lerouxb: Can you think of a test for that and code to check for it?

addaleax: In terms of a test, the one here should work just fine. For checking the encoding, this should work:

```js
class Utf8Validator extends Transform {
  decoder = new TextDecoder('utf8', { fatal: true, ignoreBOM: true })

  _transform(chunk: Buffer, enc: unknown, cb: (err: null | Error, chunk?: Buffer) => void) {
    try {
      this.decoder.decode(chunk, { stream: true });
    } catch (err: any) {
      cb(err);
      return;
    }
    cb(null, chunk);
  }
}

Readable.from(Buffer.from('héllo', 'utf8')).pipe(new Utf8Validator()).pipe(process.stdout);
// Readable.from(Buffer.from('héllo', 'latin1')).pipe(new Utf8Validator()).pipe(process.stdout);
```

lerouxb: Thanks! Updated now.

packages/compass-import-export/src/import/analyze-csv-fields.ts
```diff
@@ -155,6 +158,10 @@ export function analyzeCSVFields({
 
         if (!headerFields) {
           headerFields = results.meta.fields ?? [];
+          if (headerFields.length) {
+            // if the file contained a BOM it will be included at the start of the first field
+            headerFields[0] = headerFields[0].replace(/^\ufeff/, '');
+          }
```
addaleax, line 164: If we're using `strip-bom-stream` anyway, I think you could also add that at the top here similar to the Utf8Validator part

addaleax (approved):
looks great!

packages/compass-import-export/src/import/analyze-csv-fields.ts
lerouxb: aah I meant to go back to that and change it then forgot. Will fix, quickly.

PR #4085, opened 2023-02-22
WIP use importCSV and importJSON

lerouxb:
WIP because this is maybe more "exploration" than a proper PR. Will probably redo in bits once I know what to do.

### DIFF
--- a/packages/compass-import-export/src/import/import-csv.ts
+++ b/packages/compass-import-export/src/import/import-csv.ts
@@ -7,11 +7,14 @@ import type { DataService } from 'mongodb-data-service';
 import stripBomStream from 'strip-bom-stream';
 
 import { createCollectionWriteStream } from '../utils/collection-stream';
-import type { CollectionStreamStats } from '../utils/collection-stream';
 import { makeDoc, parseHeaderName } from '../utils/csv';
-import { processParseError, processWriteStreamErrors } from '../utils/import';
+import {
+  processParseError,
+  processWriteStreamErrors,
+  makeImportResult,
+} from '../utils/import';
 import type { Delimiter, IncludedFields, PathPart } from '../utils/csv';
-import type { ErrorJSON } from '../utils/import';
+import type { ErrorJSON, ImportResult } from '../utils/import';
 import { createDebug } from '../utils/logger';
 import { Utf8Validator } from '../utils/utf8-validator';
 import { ByteCounter } from '../utils/byte-counter';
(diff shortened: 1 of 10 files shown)


packages/compass-import-export/src/import/import-csv.ts
```diff
@@ -22,7 +25,7 @@ type ImportCSVOptions = {
   dataService: DataService;
   ns: string;
   input: Readable;
-  output: Writable;
+  output?: Writable;
```
lerouxb, line 28: This is the stream where the log stuff goes. Making it optional for now because the modal doesn't have a place to access it anyway. That makes integration in smaller steps easier.

packages/compass-import-export/src/import/import-csv.ts
```diff
@@ -159,12 +151,12 @@ export async function importCSV({
         errorCallback,
       });
 
-      return {
-        ...collectionStream.getStats(),
-        aborted: true,
-      };
+      return makeImportResult(collectionStream, numProcessed, true);
```
lerouxb, line 154: I didn't think much about what the result of importCSV and importJSON should be. Now it should be the same for both and give us everything we need to update the UI.

packages/compass-import-export/src/import/list-csv-fields.ts
```diff
@@ -32,6 +32,8 @@ export async function listCSVFields({
     preview: [],
   };
 
+  // TODO: deal with BOM and utf-8 validation here too
```
lerouxb, line 35: Missed this in my previous PR where I added tests and validation to only three out of four endpoints by accident. Will make a PR just for that separately.

packages/compass-import-export/src/modules/import.ts
```diff
@@ -22,7 +22,7 @@
  * action or not. Not great, but it has saved me a considerable amount of time vs.
  * larger scale refactoring/frameworks.
  */
-
+import _ from 'lodash';
```
lerouxb, line 25: This whole file is going to be a tricky diff to view because of the feature flag and branch.

packages/compass-import-export/src/modules/import.ts
```diff
@@ -268,142 +309,282 @@ export const startImport = () => {
...
+
+        // for now, call onGuesstimatedDocsTotal() so that the existing progress bar worlks
+        const guessedTotal = Math.max(numDocs, Math.ceil(fileSize / numBytes));
+        dispatch(onGuesstimatedDocsTotal(guessedTotal));
+
+        dispatch(onGuesstimatedProgress(numDocs, guessedTotal));
+
+        dispatch(
+          onProgress({
+            docsWritten: numDocs, // TODO: this is a lie
```
lerouxb, line 352: I don't really see why we even need docsWritten. Progress should just be one thing and I think docsProcessed is fine?

If we DO need both processed and written, then I think we should add a consistent third param to the progress callback in importJSON and importCSV.

lerouxb:
Closing in favour of https://github.com/mongodb-js/compass/compare/use-new-import?expand=1

PR #4086, opened 2023-02-22
feat(compass-schema): Support LeafyGreen darkmode COMPASS-6533

mcasimir:
Most of the changes are in `packages/compass-schema/src/components/compass-schema.tsx` which is collapsed in the diff. 

NOTE: it would probably be better to do a more thorough cleanup if we want to use this plugin in other places, it relies a lot on global styles and selectors.

Changes:
- moves the styles to css-in-js and allows for dark mode
- use a different map tile for dark mode
- styles d3 tips: that is using a library that was discontinued a while ago and that renders tooltips as a direct child of `document.body`, outside of React. To work around this quirk we detect the darkmode looking for an element with the `[data-theme="Dark"]` selector (which is rendered in compass home).

Screenshot 2023-02-22 at 22 35 34
Screenshot 2023-02-22 at 22 48 36
Screenshot 2023-02-22 at 22 49 09


packages/compass-schema/src/components/field.tsx
```diff
@@ -55,6 +55,7 @@ const fieldContainerStyles = css({
 });
 
 const fieldStyles = css({
+  overflow: 'hidden',
```
mcasimir, line 58: prevents the charts to overflow while resizing

packages/compass-schema/src/components/compass-schema.tsx
```diff
@@ -0,0 +1,467 @@
...
+
+const Schema: React.FunctionComponent<{
+  actions: Record<string, any>;
+  store: Record<string, any>;
+  analysisState: AnalysisState;
+  outdated?: boolean;
+  isActiveTab?: boolean;
+  errorMessage?: string;
+  maxTimeMS?: number;
+  schema?: any;
```
mcasimir, line 390: @Anemy looking forward to using the shiny new types from `mongodb-schema`

packages/compass-schema/src/modules/create-d3-tip.js
```diff
@@ -0,0 +1,40 @@
+import { css, palette } from '@mongodb-js/compass-components';
+import d3 from 'd3';
+require('./d3-tip')(d3);
+
+const toolTipStyles = (darkMode) => css`
+  z-index: 5;
+  line-height: 1;
+  padding: 8px;
+  background: ${darkMode ? palette.white : palette.black};
+  color: ${darkMode ? palette.black : palette.white};
```
gribnoysup, line 10: Having styles that are not changing in one place and the ones that are changing separate from each other will make refactoring this in the future easier

packages/compass-schema/src/components/coordinates-minichart/coordinates-minichart.jsx
```diff
@@ -278,5 +283,7 @@ class CoordinatesMinichart extends PureComponent {
   }
 }
 
+const CoordinatesMinichart = withDarkMode(UnthemedCoordinatesMinichart);
```
mabaasit, line 286: When zooming out, the map container shows light color in dark mode.

image

packages/compass-schema/src/components/compass-schema.tsx
```diff
@@ -0,0 +1,456 @@
...
+`;
+
+const schemaStyles = css`
+  width: 100%;
+  padding: 0 ${spacing[3]}px;
+  flex-grow: 1;
+  overflow: auto;
+`;
+
+const minichartStyles = (darkMode: boolean) => css`
```
gribnoysup, line 64: There seems to be only a few styles here that switch by `darkMode`, as we are not refactoring this in any way, would it make more sense to just use injectGlobal and just put dark mode specific stuff behind the data-theme right in css. I'm thinking it might make it a bit clearer that the work here is not finished for the future changes. And as I mentioned in another comment separating stuff that changes based on dark mode from main styles just makes it easier to refactor in the future, it was quite hard to read though all the code here to see what exactly is switched based on the dark more

gribnoysup (approved):
Left a few comments, feel free to ignore if they don't make sense to you

packages/compass-schema/src/components/compass-schema.tsx
mcasimir: it totally makes sense, yeah, it didn't come to mind as an option. It is not making a big difference I'd keep it as it is for now, I'd really like to move them where they are used.

packages/compass-schema/src/modules/create-d3-tip.js
mcasimir: not sure if this is a good or a bad thing I'm doing here 😄 .. ideally I would like to move all the styles where they are used and avoid global classes entirely, but that would require a bit more work

gribnoysup: I'm just suggesting to have something like (a made up example):

```js
const commonStyles = css`
  position: relative;
  padding: 10px;
`

const darkStyles = css`
  color: black;
`

const lightStyles = css`
  color: white;
`

...

render() {
  return <div className={cx(commonStyles, darkMode ? darkStyles : lightStyles)}>
}
```

instead of having whole style switched because reading through them and figuring out what is the dark mode part there is a bit easier that way

mcasimir: Aaahh ok i see now i understand what you mean, that makes sense, how you do that with stuff like `&:after` ? wouldn't those classes be kind-of structurally coupled between each other?

gribnoysup: Not exactly sure what you mean by that, but you totally can separate theming and layout for pseudo elements same way you can for everything else:

```js
const commonStyles = css`
  position: relative;
  padding: 10px;
  &:after {
    position: absolute;
    content: ""
  }
`

const darkStyles = css`
  color: black;
  &:after {
    color: white;
  }
`

... etc
```

and will have a similar effect of clear separation between theming and layout in the code

PR #4107, opened 2023-02-27
feat(import): use new import backend COMPASS-6519, COMPASS-6520, COMPASS-6565, COMPASS-6566

lerouxb:
Use the new importCSV() and importJSON() in place of the old functions. Same with guessFileType() and listCSVFields().

analyzeCSVFields() is left for a future PR because that will require some UI changes to show a loading state.

Replaces https://github.com/mongodb-js/compass/pull/4085/files

I removed the useNewImportBackend feature flag, because I don't think we need it. The functions do exactly what the old ones do, just more. So there's no reason to branch all over the place and use either the old or the new ones. It just creates lots more work.

I left useNewExportBackend for now and we can see when we get to export if that's going to be a different situation requiring a different solution.


packages/compass-import-export/src/components/import-modal.tsx
```diff
@@ -64,7 +65,7 @@ const closeButtonStyles = css({
 type ImportModalProps = {
   isOpen: boolean;
   ns: string;
-  startImport: () => void;
+  startImport: (useNewImportBackend?: boolean) => void;
```
lerouxb, line 68: This is kinda weird, but temporary. Felt like the cleanest, easiest way to get the feature flag would be a react hook in the component. The component will likely need to branch based off the hook soon enough anyway. And by the end it will just be removed.

packages/compass-import-export/src/import/analyze-csv-fields.ts
```diff
@@ -163,8 +162,6 @@ export function analyzeCSVFields({
       delimiter,
       header: true,
       step: function (results: Papa.ParseStepResult<PapaRowData>, parser) {
-        debug('analyzeCSVFields:step', results);
```
lerouxb, line 166: I removed a bunch of this debugging because it is all in a tight loop (once per row or even once per csv cell) and that really messes with performance plus it just floods the console. The performance only really effects development, but it is nice to be able to get a sense of how this would perform "in real life". We can just add back any debugging we need temporarily when we need it.

packages/compass-import-export/src/modules/import.ts
```diff
@@ -252,6 +282,7 @@ export const startImport = () => {
     const fileSize = fileStats?.size || 0;
 
     log.info(
+      // eslint-disable-next-line @mongodb-js/compass/unique-mongodb-log-id
```
lerouxb, line 285: I'm keeping the same log ids, at least for now while the logs are unchanged. So I had to add some linting and other comments to make the checks happy.

packages/compass-import-export/src/utils/csv.ts
```diff
@@ -293,11 +293,15 @@ export function parseValue(
     if (ISO_DATE.test(value)) {
       // iso string
       date = new Date(value);
-    } else {
-      // fall back to assuming it is an int64 value
+    } else if (!isNaN(+value)) {
```
lerouxb, line 296: Bug found with e2e tests: I made the assumption that anything that doesn't match the regex can only be an int or not a date. But the old code just passed whatever we got to `new Date()`, so things like YYYY-MM-DD parsed. Luckily we had some test fixtures that happened to hit that.

packages/compass-import-export/src/utils/csv.ts
```diff
@@ -84,8 +84,8 @@ const FLOAT = /^\s*-?(\d+\.?|\.\d+|\d+\.\d+)([eE][-+]?\d+)?\s*$/;
 // from papaparse: https://github.com/mholt/PapaParse/blob/aa0046865f1b4e817ebba6966d6baf483e0652d7/papaparse.js#L1025
 const ISO_DATE =
   /^((\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z))|(\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d([+-][0-2]\d:[0-5]\d|Z))|(\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d([+-][0-2]\d:[0-5]\d|Z)))$/;
-const TRUTHY_STRINGS = ['true', 'TRUE', 'True'];
-const FALSY_STRINGS = ['false', 'FALSE', 'False'];
+const TRUTHY_STRINGS = ['t', 'true', 'TRUE', 'True'];
```
lerouxb, line 87: While manually testing I noticed this pattern for true/false in some open source datasets and figured it would be nice if we could automatically detect and parse it.

packages/compass-import-export/src/modules/import.ts
```diff
@@ -540,18 +782,14 @@ export const selectImportFileName = (fileName: string) => {
...
       const exists = await checkFileExists(fileName);
       if (!exists) {
         throw new Error(`File ${fileName} not found`);
       }
-      const stats = await getFileStats(fileName);
+      const fileStats = await getFileStats(fileName);
 
-      fileStats = {
-        ...stats,
-        type: mime.lookup(fileName),
```
lerouxb, line 553: I noticed we don't actually use this as we go with whatever detectImportFile returns. The types don't even include it anyway.

packages/compass-import-export/src/components/import-modal.tsx
lerouxb: Actually I'm changing my mind. I'm replacing more and more stuff in another branch and it is becoming a mess. I think I'm going to try to just make a new import store / module next to the old one and then in the components we'll use either the old and the new. Otherwise it is branching sprinkled all over the place.

packages/compass-import-export/src/modules/import.ts
```diff
@@ -625,16 +597,15 @@ export const setDelimiter = (delimiter: CSVDelimiter) => {
       delimiter: delimiter,
     });
 
-    if (previewLoaded) {
-      debug('preview needs updated because delimiter changed', {
+    // TODO: This is certainly a race condition. What if it is busy loading?
```
lerouxb, line 600: I'm thinking it is probably better to just always dispatch loadCSVPreviewDocs() again. Ideally we would abort the previous attempt, but that's probably overkill. The new attempt should most likely finish after the previous one.

Thoughts?

packages/compass-import-export/src/components/import-preview.tsx
```diff
@@ -60,7 +59,7 @@ function ImportPreview({
     type: string;
     checked: boolean;
   }[];
-  values: Document[];
+  values: string[][];
```
lerouxb, line 62: The type was wrong before. Noticed when I wrote code to turn an array of string arrays into an array of docs and it didn't work.

The preview values are an array of up to length 10 where each item in the array is an array of strings which represent the CSV values in that row.

packages/compass-import-export/src/modules/import.ts
Anemy: It should be pretty fast right? I think a good comment is enough in that case. If there's a good chance `loadCSVPreviewDocs()` will be slow lets track the promise and abort/discard it then.

lerouxb: OK I'm making it re-update regardless of previewLoaded and I updated the comment.

Anemy (approved):
lgtm! nice work

PR #4108, opened 2023-02-27
feat(hotkeys): hook for keyboard shortcuts COMPASS-6551

mabaasit:
In this PR i added a hook for handling keyboard shortcuts. The hook exposes:
1. react reference which can be used to scope the shortcut when the element is focused 
2. an OS specific shortcut to display the hot key.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-components/src/hooks/use-hotkeys.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  return {
+    shortcut,
+    ref,
+  };
+};
+
+const mapKeyToShortcut = (key: string) => {
+  if (isMac()) {
+    return key.replace('meta', '⌘').replace('alt', 'option');
```
mcasimir, line 35: is replacing enough here? one issue that comes to mind is that we may forget or use the spacing inconsistently `ctrl+c` vs `ctrl + c`.

packages/compass-components/src/hooks/use-hotkeys.ts
```diff
@@ -0,0 +1,38 @@
...
+  const shortcut = useMemo(() => {
+    if (Array.isArray(key)) {
+      return key.map(mapKeyToShortcut);
+    }
+    return mapKeyToShortcut(key);
+  }, [key]);
+
+  return {
+    shortcut,
+    ref,
```
mcasimir, line 29: where / how is this used?

packages/compass-components/src/hooks/use-hotkeys.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  const shortcut = useMemo(() => {
+    if (Array.isArray(key)) {
+      return key.map(mapKeyToShortcut);
+    }
+    return mapKeyToShortcut(key);
+  }, [key]);
+
+  return {
+    shortcut,
```
mcasimir, line 28: Not a big deal, but can we avoid `useHotkeys` to also render the shortcut string? I feel that like this we would lose some flexibility. For example, I can only use the function that renders the shortcut consistently if I'm also binding it, which may not be the case: code mirror may register hotkeys that we may still want to document.

Could we have this rendering as a separate function or component?

packages/compass-components/src/hooks/use-hotkeys.ts
```diff
@@ -0,0 +1,38 @@
...
+ * -------- | ----------- | -------------
+ * ctrl     | control     | ctrl
+ * shift    | shift       | shift
+ * alt      | option      | alt
+ * meta     | command     | ctrl
+ *
+ */
+export const useHotkeys = (...args: Parameters<typeof useGlobalHotkeys>) => {
+  const [key] = args;
+  const ref = useGlobalHotkeys(...args);
```
mcasimir, line 18: may be good to leave a comment that `meta` is already interpreted as Ctrl or Cmd by `useGlobalHotkeys`, I may be wrong on this, but it could be a bit unexpected, given that the meaning of `metaKey` is (or was) confusing even between browsers

packages/compass-components/src/hooks/use-hotkeys.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+  after(function () {
+    Object.defineProperty(window.navigator, 'userAgent', {
+      value: initialUserAgent,
+      writable: true,
+    });
+  });
+
+  context('on macOS', function () {
+    before(function () {
+      Object.defineProperty(window.navigator, 'userAgent', {
```
mcasimir, line 20: nice trick!

packages/compass-find-in-page/src/components/find-in-page-input.tsx
```diff
@@ -102,34 +103,27 @@ function FindInPageInput({
...
-        evt.preventDefault();
-        const searchValue = findInPageInputRef.current?.value;
-        if (!searchValue || searchValue === '') {
-          return dispatchStopFind();
-        }
-
-        const back = evt.shiftKey;
-        dispatchFind(searchValue, !back, searching);
-      }
+      const back = evt.shiftKey;
```
mcasimir, line 114: this feels a bit odd at this point, I wouldn't mix the approaches, why don't we use `useHotkeys` for the key strokes and do something like this?

``` js
const find = (evt, back) => { evt.preventDefault(), const searchValue = ... };
const findNext = useCallback(() => find(evt, false))
const findPrev = useCallback(() => find(evt, true))

useHotkeys('enter', findNext, { enableOnFormTags: ['INPUT'] });
useHotkeys('shift + enter', findPrev, { enableOnFormTags: ['INPUT'] });
```

packages/compass-shell/src/components/shell-header/shell-header.jsx
```diff
@@ -8,9 +8,10 @@ import {
   keyframes,
   SpinLoader,
   withDarkMode,
+  useHotkeys,
```
mcasimir, line 11: didn't know where to add this comment, but if we have a way to use the rendering without binding the keys, it would be nice to adapt also this file: https://github.com/mongodb-js/compass/blob/c7ee317492da4ebf944f7867a9bd8ca342a3c1ec/packages/compass-shell/src/components/shell-info-modal/keyboard-shortcuts-table.jsx#L31

mcasimir (commented):
🎹

This looks fantastic! it would make it so easy to add shortcuts everywhere 🔥 . I left some comments around the code, except for the spacing in the rendering is mostly just random ideas to improve it (or make it worse :)), let me know what you think!

packages/compass-shell/src/components/shell-header/shell-header.jsx
mabaasit: Yes, that will be great. Totally missed on this one.

packages/compass-components/src/hooks/use-hotkeys.ts
mabaasit: Yes we can avoid doing that. Actually this hook does not render it, it just maps it for rendering. With this approach we are forcing user to render what has been binded. Now that I am adding a component to render the shortcut, I'd still prefer to leave this here and return the shortcut (without mapping) and whenever user wants to render the shortcut, use the `Hotkey` component which will map and render it. Something like this:

```ts
const { shortcut } = useHotkeys('ctrl + a');

return (
	<Hotkey shortcut={shortcut} />
);
```

Alternatively, we can remove it from the hook and then let users render using `Hotkey` component. 

```ts
useHotkeys('ctrl + a');

return (
	<Hotkey shortcut="ctrl + a" />
);
```
Personally i like the former approach.

packages/compass-components/src/hooks/use-hotkeys.ts
mabaasit: this i left intentionally like this and let the user decide how it should look. if they want space, they will add a shortcut with space `ctrl + x`.

packages/compass-components/src/hooks/use-hotkeys.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+import sinon from 'sinon';
+
+const initialUserAgent = window.navigator.userAgent;
+
+describe('use-hotkeys', function () {
+  after(function () {
+    Object.defineProperty(window.navigator, 'userAgent', {
+      value: initialUserAgent,
+      writable: true,
+    });
```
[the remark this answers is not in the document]
mabaasit: as its a readonly prop, ts is not happy about it.

packages/compass-components/src/hooks/use-hotkeys.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+  });
+
+  context('on macOS', function () {
+    before(function () {
+      Object.defineProperty(window.navigator, 'userAgent', {
+        value:
+          'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/92.0.4515.159 Safari/537.36',
+        writable: true,
+      });
```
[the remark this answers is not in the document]
mabaasit: Yes it does and adding `configurable: true` throws with `TypeError: Cannot redefine property: userAgent`

packages/compass-components/src/hooks/use-hotkeys.ts
mcasimir: > the user decides how it should look

I don't see why that would be useful for us. We are the users here and having different users to make shortcuts look differently is what we want to avoid: the main benefit of having this as a shared function is to get consistent results, I believe we should choose a way (with or without spaces), stick to that, and make sure the result is what we expect.

If there is a reason to allow different formats I'd rather have that as an option, so we are sure we don't use it by mistake. Personally I would prefer that we don't even have the option to mess it up though.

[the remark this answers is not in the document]
mabaasit: added [here](https://github.com/mongodb-js/compass/pull/4108/commits/5f00f63af8b075c40aa669209b9d29c4cdf568fc#diff-0ca4f234b23391bba7137f47e9a314b40dfa18d85a79fdeee81a6a71b04d0953R37) in 5f00f63af8b075c40aa669209b9d29c4cdf568fc

addaleax (approved):

packages/compass-components/src/hooks/use-hotkeys.ts
mabaasit: added [here](https://github.com/mongodb-js/compass/pull/4108/commits/5f00f63af8b075c40aa669209b9d29c4cdf568fc#diff-0ca4f234b23391bba7137f47e9a314b40dfa18d85a79fdeee81a6a71b04d0953R35) in [5f00f63](https://github.com/mongodb-js/compass/pull/4108/commits/5f00f63af8b075c40aa669209b9d29c4cdf568fc)

packages/compass-components/src/hooks/use-hotkeys.tsx
```diff
@@ -0,0 +1,59 @@
...
+export const useHotkeys = (
+  key: string,
+  callback: GlobalHotkeysArgs[1],
+  options?: GlobalHotkeysArgs[2],
+  deps?: GlobalHotkeysArgs[3]
+) => {
+  return useGlobalHotkeys(key, callback, options, deps);
+};
+
+export const mapHotkeyToShortcut = (key: string) => {
```
mcasimir, line 36: I'm not sure this is the clearest name for a public function, hotkey, and shortcut are almost synonyms, out of context, and without reading the code I'm not sure I would be able to tell what this does. Is there a name that would better reveal the intent of what this function does?

packages/compass-components/src/hooks/use-hotkeys.tsx
```diff
@@ -0,0 +1,59 @@
...
+  // map the modifier keys to the platform specific keys
+  const shortcut = isMac()
+    ? key.replace(/\bmeta\b/, '⌘').replace(/\balt\b/, 'option')
```
mcasimir, line 39: Something makes me think that `useGlobalHotkeys` is accepting different "cases", like probably `Meta` would also work, given `ArrowUp` does, should we cover that too in this regex?

packages/compass-components/src/hooks/use-hotkeys.tsx
```diff
@@ -0,0 +1,59 @@
...
+    : key.replace(/\bmeta\b/, 'ctrl');
+  // normalize the shortcut and add spaces between + signs
+  return shortcut
+    .replace(/\+/g, ' + ') // Add space on both sides of each '+'
+    .replace(/\s+/g, ' ') // Replace multiple spaces with a single space
+    .replace('ArrowUp', '↑')
+    .replace('ArrowDown', '↓')
+    .trim();
+};
```
mcasimir, line 48: It would be really nice to also make sure that the resulting case is consistent, for example always CamelCase or lowercase

mcasimir (approved):

PR #4110, opened 2023-02-28
feat(compass-sidebar): adds the refresh CTA to sidebar

himanshusinghs:
Closes
--
COMPASS-6431

## Description
Moves the refresh CTA from options menu to an easily accessible location in sidebar.

<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>image
	<td>image
<tr>
	<td>image
	<td>image
</table>

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-e2e-tests/helpers/commands/show-shell.ts
+++ b/packages/compass-e2e-tests/helpers/commands/show-shell.ts
@@ -13,3 +13,12 @@ export async function showShell(browser: CompassBrowser): Promise<void> {
     await browser.clickVisible(Selectors.ShellInput);
   });
 }
+
+export async function executeInShell(
+  browser: CompassBrowser,
+  command: string
+): Promise<void> {
+  await browser.showShell();
+  await browser.setAceValue(Selectors.ShellContent, command);
+  await browser.keys(['Enter']);
+}
--- a/packages/compass-e2e-tests/helpers/commands/show-shell.ts
+++ b/packages/compass-e2e-tests/helpers/commands/show-shell.ts
@@ -13,12 +13,3 @@ export async function showShell(browser: CompassBrowser): Promise<void> {
     await browser.clickVisible(Selectors.ShellInput);
   });
 }
-
-export async function executeInShell(
-  browser: CompassBrowser,
-  command: string
-): Promise<void> {
-  await browser.showShell();
-  await browser.setAceValue(Selectors.ShellContent, command);
-  await browser.keys(['Enter']);
-}
--- a/packages/compass-e2e-tests/tests/instance-sidebar.test.ts
+++ b/packages/compass-e2e-tests/tests/instance-sidebar.test.ts
@@ -188,12 +188,15 @@ describe('Instance sidebar', function () {
   });
 
   it('can refresh the databases', async function () {
-    const db = `db_${Date.now()}`;
+    const db = 'test';
     const coll = `coll_${Date.now()}`;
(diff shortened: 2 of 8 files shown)


packages/compass-e2e-tests/helpers/commands/show-shell.ts
```diff
@@ -13,3 +13,12 @@ export async function showShell(browser: CompassBrowser): Promise<void> {
     await browser.clickVisible(Selectors.ShellInput);
   });
 }
+
+export async function executeInShell(
```
lerouxb, line 17: Our convention is one helper command file per command, so you probably want to add execute-in-shell.ts next to show-shell.ts and add a line for it next to wherever show-shell.ts was added.

packages/compass-e2e-tests/tests/instance-sidebar.test.ts
```diff
@@ -186,4 +186,14 @@ describe('Instance sidebar', function () {
       .$(Selectors.databaseTab('Collections', true))
       .waitForDisplayed();
   });
+
+  it('can refresh the databases', async function () {
+    const db = `db_${Date.now()}`;
+    const coll = `coll_${Date.now()}`;
+    await browser.executeInShell(`use ${db};`);
```
lerouxb, line 193: I probably wouldn't have bothered adding it via the shell since we usually have a client somewhere in tests where we can just run commands against the DB, but this is nice in that it mimicks the common use case 👍

packages/compass-e2e-tests/helpers/commands/show-shell.ts
lerouxb: We started off having all the commands in one file but that quickly grew out of control so Sergey helped me with some amazing typescript: https://github.com/mongodb-js/compass/blob/main/packages/compass-e2e-tests/helpers/compass-browser.ts#L1-L13

packages/compass-e2e-tests/tests/instance-sidebar.test.ts
himanshusinghs: So I started looking for a client to otherwise execute it without my helper and realised that there is already `shellEval` supposedly doing the same stuff as `executeInShell`. I will probably just used that and get rid of my helper.

packages/compass-e2e-tests/tests/instance-sidebar.test.ts
```diff
@@ -186,4 +186,14 @@ describe('Instance sidebar', function () {
       .$(Selectors.databaseTab('Collections', true))
       .waitForDisplayed();
   });
+
+  it('can refresh the databases', async function () {
+    const db = `db_${Date.now()}`;
```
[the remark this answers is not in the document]
himanshusinghs: Thanks - I missed on the cleanup. Will take care of it.

packages/compass-e2e-tests/tests/instance-sidebar.test.ts
lerouxb: Oh yeah you're right. I totally forgot about that. I think it was one of the first few commands we had.

himanshusinghs: perfect - changes are ready for another review 😄

PR #4112, opened 2023-02-28
feat(compass-import-export): open file input before import modal COMPASS-6535

Anemy:
COMPASS-6535

This updates the import flow to open the file input window before we show the import modal. If the users hits cancel on the file input then the modal is not shown.

https://user-images.githubusercontent.com/1791149/221929118-ce7c8d5e-7e0f-4b02-9bde-ea6aa881980c.mp4


packages/compass-e2e-tests/tests/collection-import.test.ts
```diff
@@ -19,12 +19,14 @@ async function importJSONFile(browser: CompassBrowser, jsonPath: string) {
   await insertDocumentOption.waitForDisplayed();
   await browser.clickVisible(Selectors.ImportFileOption);
 
-  // wait for the modal to appear and select the file
+  // Select the file.
+  await browser.selectFile(Selectors.ImportFileInput, jsonPath);
```
lerouxb, line 23: Something I thought about with all these selectFile bits moved around: I wonder if we should have something somewhere changing the file (ie. maybe clicking again and then cancel, then again and re-selecting it) to make sure all that works. We can also add it later or I can add it to the new tests for the new backend.

packages/compass-import-export/package.json
```diff
@@ -128,7 +128,6 @@
     "react-dom": "^17.0.2",
     "react-redux": "^8.0.5",
     "redux": "^4.2.0",
-    "redux-mock-store": "^1.5.4",
```
lerouxb, line 131: 👏

packages/compass-import-export/src/components/import-file-input.tsx
```diff
@@ -0,0 +1,66 @@
...
+      } else if (typeof onCancel === 'function') {
+        onCancel();
+      }
+    },
+    [onCancel, selectImportFileName]
+  );
+
+  const backend = useMemo(() => {
+    // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+    let electron: typeof import('@electron/remote');
```
lerouxb, line 34: We should file a follow-up ticket for the whole project with what we need to do / change to get import and export to work in DE and this kind of thing should probably be mentioned in it.

packages/compass-import-export/src/modules/import.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+
+type DispatchFunctionType = ThunkDispatch<RootImportState, void, AnyAction>;
+
+const mockEmptyState = {
+  importData: {
+    ...INITIAL_STATE,
+  },
+};
+
+describe('import [module]', function () {
```
lerouxb, line 24: Thanks for replacing the old tests. Saves me some effort :)

packages/compass-e2e-tests/tests/collection-import.test.ts
lerouxb: I'm also a little surprised that the [selectFile helper](https://github.com/mongodb-js/compass/blob/6b97a8ff0f178c31cf6c873948a76865b1958005/packages/compass-e2e-tests/helpers/commands/select-file.ts#L1-L31) works with the file dialog already open and the modal still closed. But I'm not gonna think about it too much if it works :)

packages/compass-import-export/src/components/import-file-input.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+function ImportFileInput({
+  autoOpen,
+  onCancel,
+  selectImportFileName,
+  fileName,
+}: ImportFileInputProps) {
+  const handleChooseFile = useCallback(
+    (files: string[]) => {
+      console.log('on change', files);
```
lerouxb, line 22: console.log left in, btw.

packages/compass-import-export/src/components/import-file-input.tsx
lerouxb: Actually thinking about it more, in the browser we'd have to keep a reference to the file object and not just the filename because we wouldn't be able to just read from the file with the fs module. So we would still need work here.

Anemy: Good thinking, I'll remove the try catch here so it explicitly fails in non-electron environments for now.

packages/compass-e2e-tests/tests/collection-import.test.ts
Anemy: Sounds good, let's add a test for that. Didn't include it in this pr, will in the file input ui updates if not sooner.

PR #4131, opened 2023-03-03
chore(import): remove the (manual) file type selector from import (part of COMPASS-6545)

lerouxb:
Screenshot 2023-03-03 at 17 01 11

Screenshot 2023-03-03 at 17 01 00

Had to add a new file broken in a different way so the E2E test that tests parse errors will continue to work. This makes it much harder to break import.

With the new guessFileType() it will reliably pick JSON whenever something begins to parse as JSON. If it doesn't detect JSON, then changing the type to JSON would guarantee to cause a parse error.

This also gets around the awkwardness of json (ie. an array of objects) vs jsonl (ie. one object per line or objects after another) and how the existing code didn't clear the multi line var and gave you no way of manually specifying it.


packages/compass-import-export/src/modules/import.ts
```diff
@@ -554,30 +554,6 @@ export const selectImportFileName = (fileName: string) => {
...
-export const selectImportFileType = (fileType: 'json' | 'csv') => {
-  return (
-    dispatch: ThunkDispatch<RootImportState, void, AnyAction>,
-    getState: () => RootImportState
-  ) => {
-    const { fileName, delimiter } = getState().importData;
-
-    dispatch({
-      type: FILE_TYPE_SELECTED,
-      fileType: fileType,
```
lerouxb, line 571: This never dealt with `fileIsMultilineJSON` and now we never have to.

Anemy (approved):

lerouxb:
There are existing flakes that seem to be unrelated as I get different errors on retry. Will just merge for now.

PR #4132, opened 2023-03-03
chore(compass-import-export): show message when import is in progress COMPASS-6538, COMPASS-6539

Anemy:
COMPASS-6538
When import is in progress and someone clicks to import data, for instance through the app menu, we show a message indicating that only one import can happen at once.
A bit more context, in the future, once we can have multi-toasts, we'll remove this limitation. This message comes into play once we add the in progress toast in COMPASS-6540.

Initially I was thinking we'd need an import context for disabling buttons around Compass. We'll still need that for export, as there are more export buttons, however for import this should suffice.

How do y'all feel about the text here? Any suggestions?

Screenshot 2023-03-03 at 12 07 22 PM


lerouxb (approved):

Anemy:
Looks like the failures are unrelated recent flakes. Merging this, will revert if they were somehow caused by this.

PR #4142, opened 2023-03-09
chore: update mocha to latest

mcasimir:
Just noticed in the Snyk report that we are still using old versions of mocha, this seems to be also one of the biggest misalignments in the version of dependencies that we still have in Compass.

- Bump `mocha` to `^10.2.0`
- Migrate `compass-field-store` and `compass-app-stores` to the new configuration
- Drop `mocha.opts`


Anemy (approved):
Yess. When the diff is `-30,000` 🥳 
:shipit:

PR #4155, opened 2023-03-15
feat(import): analyse CSV fields and auto-select the correct type COMPASS-6536

lerouxb:
Screenshot 2023-03-20 at 16 37 42

Screenshot 2023-03-20 at 16 41 05


packages/compass-import-export/src/components/import-preview.tsx
```diff
@@ -47,49 +59,101 @@ const cellUncheckedStyles = css({
...
+});
+
+const fieldTypeContainerStyles = css({
+  display: 'flex',
+  alignItems: 'center',
+  gap: spacing[1],
+});
+
+const infoIconCSS = css({
+  height: spacing[3],
```
Anemy, line 74: It feels a bit weird to set the height of this icon container element, do we need to? I could see it leading to some cropping or scrolling.

packages/compass-import-export/src/components/import-preview.tsx
```diff
@@ -47,49 +59,101 @@ const cellUncheckedStyles = css({
...
+        <div {...props}>
+          {children}
+          <div className={infoIconCSS}>
+            <Icon glyph="InfoWithCircle"></Icon>
+          </div>
+        </div>
+      )}
+    >
+      <>
+        <p>
```
Anemy, line 116: Would the LeafyGreen `Body` element work here?

packages/compass-import-export/src/components/import-preview.tsx
```diff
@@ -109,29 +173,108 @@ function ImportPreview({
...
-                    <span title={field.path}>{field.path}</span>
-                  </Label>
-                  <SelectFieldType
-                    fieldPath={field.path}
-                    selectedType={field.type}
-                    onChange={(newType: string) =>
-                      setFieldType(field.path, newType)
-                    }
+                <Label
+                  id={`toggle-import-field-label-${field.path}`}
```
Anemy, line 177: It was like this before, looks like this id doesn't match the `aria-labelledby` on the checkbox.  Can we align them in these changes?

packages/compass-import-export/src/components/import-preview.tsx
```diff
@@ -47,49 +59,101 @@ const cellUncheckedStyles = css({
...
+) {
+  return (
+    <TableHeader
+      key={`col-${field.path}`}
+      label={
+        <div
+          className={columnHeaderStyles}
+          data-testid={`preview-field-header-${field.path}`}
+        >
+          {field.type === 'placeholder' ? (
```
Anemy, line 152: Total small thing, feel free to ignore, might be a touch less code to write this instead as
```
{field.type !== 'placeholder' && (
  <>
  ...
```

packages/compass-import-export/src/components/import-preview.tsx
```diff
@@ -109,29 +173,108 @@ function ImportPreview({
...
+  analyzed,
+}: {
+  fields: Field[];
+  values: string[][];
+  onFieldCheckedChanged: (fieldPath: string, checked: boolean) => void;
+  setFieldType: (fieldPath: string, fieldType: string) => void;
+  loaded: boolean;
+  analyzed: boolean;
+}) {
+  if (!loaded) {
```
Anemy, line 231: Will this take long ever? Should we show a `SpinLoader`?

packages/compass-import-export/src/modules/import.ts
```diff
@@ -439,25 +434,144 @@ export const cancelImport = () => {
...
+      analyzeAbortController.abort();
+      dispatch({
+        type: ANALYZE_CANCELLED,
+      });
+    }
+
+    const input = fs.createReadStream(fileName);
+
+    listCSVFields({ input, delimiter })
+      .then((result) => {
```
Anemy, line 550: Would `await` syntax here? Or is there a reason to use the promise .then syntax?

packages/compass-import-export/src/components/import-preview.tsx
```diff
@@ -47,49 +59,101 @@ const cellUncheckedStyles = css({
...
-              >
+    <Tooltip
+      align="top"
+      justify="middle"
+      delay={500}
+      trigger={({ children, ...props }) => (
+        <div {...props}>
+          {children}
+          <div className={infoIconCSS}>
+            <Icon glyph="InfoWithCircle"></Icon>
```
lerouxb, line 110: I'm not sure what the exact right thing is to do here. I want a little icon for launching the tooltip, but should it be on hover? On click?

Anemy (commented):
Looking good, left a couple comments/questions

packages/compass-import-export/src/components/import-preview.tsx
lerouxb, line 152: All this mention of 'placeholder' relates to when there are arrays in the CSV. We won't be exporting stuff like this until export is changed, but import now already supports CSV files like that.

Screenshot 2023-03-17 at 14 41 47

Here's the file that produced that:

```
a,foo[0],foo[1],foo[2],bar[0][0],bar[0][1],z,notes
a,1,2,3,,,z,foo is an array of simple values
a,1,,,,,z,foo is not always of the same length
```

See the columns in the preview still match one to one with the cells in the CSV file, but for arrays the type is for the values inside the array and the whole array gets one type.

packages/compass-import-export/src/components/import-preview.tsx
Anemy: Good question for Diancheng + design (I'm also not sure what the best pattern is).

packages/compass-import-export/src/utils/csv.ts
```diff
@@ -37,9 +37,29 @@ export type CSVParsableFieldType =
   | 'timestamp'
   | 'decimal'
   | 'regex'
-  | 'number'
+  | 'number' // like 'mixed', but for use when everything is an int, long or double.
   | 'mixed';
 
+export const CSVFieldTypeLabels: Record<CSVParsableFieldType, string> = {
```
lerouxb, line 43: We might want to think through this order since it is the order that the options in the field type selects will appear in and right now it is essentially arbitrary. But it is easy enough to change in a PR later once people had a chance to play with it.

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: No amount of flexbox css helped me get the icon aligned with the selectbox. Let me know if you find alternative ways :)

packages/compass-import-export/src/components/import-preview.tsx
Anemy: Probably not for this iteration, but I could see folks sometimes having `placeholder` items in arrays like this. We might want a way to collapse/expand those fields so the table is still scrollable. This also goes back to your comments on the UX of having this table format for negotiating the import.

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: No longer than before. It just looks at the first row of the CSV file to get the columns and the first 10 rows or so to get the preview. Which was already there before. Maybe?

packages/compass-import-export/src/modules/import.ts
lerouxb: I just copied the style from the other thunk actions in the file.

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: We're already in Body so it feels strange to me to do Body as="p", but whatever ;)

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: I mean more like the technical right accessible ideal way to do this since an icon is a span and all that.

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: I'm not sure how much we should worry about it because you can only get into that situation when re-importing CSV we exported. Which we should absolutely support, but we'd also be warning people NOT to use CSV for that sort of thing. If you're going to export it only to re-import back into mongodb, then extended JSON is obviously a much better format.

So in reality CSV should be for when exporting for use with a third party system or when importing from some third party system.

lerouxb: Aah yes that's left over from when this was a lot more complicated.

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: (If yes, then we could just do a follow-up since I'd say this is exactly as before)

packages/compass-import-export/src/components/import-preview.tsx
Anemy: Sounds good 👍 
I wonder if we should start recommending people the json export type when we do the export work.

packages/compass-import-export/src/modules/import.ts
Anemy: My preference would be towards using `await`, I find it makes error handling and readability a bit better. In my current work I'm updating one of the other occurrences in this file.

packages/compass-import-export/src/components/import-preview.tsx
Anemy: If it's an easy add and looks alright I think it makes sense, then we're fully transparent to folks as to whats happening. If it's flashing in then out though I think it's fine as is. Your call, feel free to ignore.

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: We should. It is in the design :)

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: I think it is pretty much guaranteed to take only a split second. Often with our loaders we end up adding some kind of delay before they appear to stop the flashing anyway. So I'm not sure how often that's going to be an improvement. Not that I have much of an opinion.

packages/compass-import-export/src/modules/import.ts
lerouxb: I'll change it, then ;)

packages/compass-import-export/src/components/import-preview.tsx
Anemy: Isn't the `Body` component already `p` elements by default? Why do we need to pass `as="p"`?

packages/compass-import-export/src/components/import-preview.tsx
Anemy: Super small thing, is there any reason why we're using `CSS` at the end of the variable name? The usual convention is `Styles` at the end of the name, like `variableNameStyles`.

Anemy: Yeah it looks like the line height from the table header from LeafyGreen is what makes it weird. Maybe it makes sense to set ```lineHeight: `${spacing[3]}`, ``` instead of the height with a note. all good though, thanks

packages/compass-import-export/src/components/import-preview.tsx
```diff
@@ -47,22 +59,176 @@ const cellUncheckedStyles = css({
...
+                  className={fieldPathHeaderStyles}
+                  htmlFor={`toggle-import-field-checkbox-${field.path}`}
+                >
+                  <span title={field.path}>{field.path}</span>
+                </Label>
+              </div>
+              <div className={fieldTypeContainerStyles}>
+                {analyzed ? (
+                  <>
+                    <SelectFieldType
```
Anemy, line 189: Small thing, feel free to ignore, and something I should have caught in the older import/export LeafyGreen work:
This `SelectFieldType` looks like a simple component wrapper around a `Select`. Do you feel it's nicer to have it all in this file and not a seperate `select-field-type.tsx`?

Anemy (approved):

packages/compass-import-export/src/components/select-field-type.tsx
```diff
@@ -17,24 +14,27 @@ function SelectFieldType({
   onChange,
 }: {
   fieldPath: string;
-  selectedType: string;
+  selectedType: CSVParsableFieldType;
+  summary?: string;
```
Anemy, line 18:
```suggestion
```

packages/compass-e2e-tests/tests/collection-import.test.ts
```diff
@@ -54,6 +54,8 @@ async function selectFieldType(
 
   const fieldTypeSelectSpan = await fieldTypeSelectMenu.$(`span=${fieldType}`);
   await fieldTypeSelectSpan.waitForDisplayed();
+  await fieldTypeSelectSpan.scrollIntoView();
+  await browser.pause(1000);
```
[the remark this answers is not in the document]
Anemy: Any idea what's causing the flakyness? Is it the rendering time of the table? Or the placeholder disappearing?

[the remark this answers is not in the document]
lerouxb: rendering or scrolling of the field type options inside the select. The symptom is that we click on the option and then wait for the popup menu thingy containing the options to disappear, but it never disappears and the select's selected option never updates. So the click did nothing. Then 120 seconds later the test fails and the screenshot shows the select dropdown still open.

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: Filing a follow-up for this.

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: Weirdly lineHeight: `${spacing[3]}px` (px is important because lineHeight allows unitless values) does not do the same. I just added a comment.

PR #4157, opened 2023-03-15
chore(compass-editor): validation autocompleter with json COMPASS-6572

mabaasit:
Validation autocompleter

<table>
<tr>
	<td> Old </td>	<td> New </td>
<tr>
	<td>

https://user-images.githubusercontent.com/1305718/225344193-6a1e6193-c1f4-4528-8519-84c351b9890a.mov

</td>	<td> 

https://user-images.githubusercontent.com/1305718/225344216-84d342a6-ebb6-4240-90f5-fe9ca4f5ed0d.mov

 </td>
</table>

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-editor/src/codemirror/validation-autocompleter.ts
```diff
@@ -0,0 +1,176 @@
...
+  from: number,
+  detail?: string
+) => {
+  return {
+    from,
+    filter: false,
+    options: completions
+      .filter((completion) => {
+        return completion.value.toLowerCase().startsWith(text.toLowerCase());
+      })
```
gribnoysup, line 36: Unless you are doing this for some specific reason, a recommended approach from codemirror docs is to just return all results with `validFor` instead of doing filtering yourself, [here's an example](https://github.com/mongodb-js/compass/pull/4163/files#diff-09b51c2c02afdec3ad7c224bb84ec90e1878d0724ca903f904c9fa84cfdde309R47-R64)

packages/compass-editor/src/codemirror/validation-autocompleter.ts
```diff
@@ -0,0 +1,176 @@
...
+      .map((completion) => ({
+        ...completion,
```
gribnoysup, line 38: There are props here that are not valid for codemirror, I'd suggest to do mapping more explicit. This also doesn't handle snippets, [here's an example](https://github.com/mongodb-js/compass/pull/4163/files#diff-09b51c2c02afdec3ad7c224bb84ec90e1878d0724ca903f904c9fa84cfdde309R13-R32)

packages/compass-editor/src/codemirror/validation-autocompleter.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    const ancestors = getAncestryOfToken(token, document);
+
+    if (isCompletingRequired(ancestors) || isCompletingProperties(ancestors)) {
+      const ancestor = getNestingAncestor(ancestors);
+      return createCompletions(
+        filterFieldsByAncestor(fieldCompletions, ancestor),
+        textBefore,
+        context.pos - textBefore.length,
+        'Field'
```
gribnoysup, line 157: We use lowercase ones everywhere else for detail

packages/compass-editor/src/codemirror/validation-autocompleter.ts
```diff
@@ -0,0 +1,176 @@
...
+      const value = x.value.replace(`${ancestor}.`, '');
+      return {
+        ...x,
+        value,
+      };
+    })
+    .filter((x) => x.value);
+};
+
+export const createValidationAutocompleter = (
```
gribnoysup, line 111: Seems like this is not doing any autocomplete that was happening before, in particular for the case where you are using [query validation](https://www.mongodb.com/docs/manual/core/schema-validation/specify-query-expression-rules/#std-label-schema-validation-query-expression) and not jsonSchema. Here's current behavior:

image

gribnoysup (commented):
Left a few comments, also doesn't look like this is covering some basic cases we have right now. In this other autocomplete PR I figured out how to fix testing setup for autocompleter, so maybe we can merge it first and then you can add some tests here?

packages/compass-editor/src/codemirror/validation-autocompleter.ts
mabaasit: Thanks. I'll review your PR and then I can reuse this functionality. Also the tests.

packages/compass-editor/src/codemirror/validation-autocompleter.ts
mabaasit: From the docs, the root validation object either contains `$jsonSchema` or the query operators and not the document fields, but somehow the current auto-completer suggests that. 

When the user is inside the query operator validation, I'll reuse the queryAutoCompleter to handle that efficiently.

gribnoysup: Gotcha, wasn't obvious if this was removed this on purpose or not

gribnoysup (approved):

PR #4169, opened 2023-03-20
feat(compass-import-export): show import progress in toast, make import background COMPASS-6540 COMPASS-6555

Anemy:
COMPASS-6540 COMPASS-6555
This pr updates how we show import progress in Compass. Previously we would show a progress bar in a modal that would block other activities in the ui while an import is in progress. Now, when a user clicks to import, we hide the modal and show a toast that indicates the progress.

When errors are encountered in importing we show the errors in the toast. If the user didn't set to `stopOnErrors` we create and import error log file, and let the user open it at the end. The log file is located in the user data folder in a folder named `ImportErrorLogs`.  The file name is `import-FILE_NAME.log` where FILE_NAME is the name of the file being imported.

This pr includes e2e tests for the import flow, let me know if something else would be good to add a test for.

Questions for reviewers:
- How do you feel about the error log file location and name?
- Are there additional tests we should add here?

___
*Successful import*

https://user-images.githubusercontent.com/1791149/226411791-0d68cb41-c3a2-41b6-9f13-e4d32bf42025.mp4

___
*Aborted import*

https://user-images.githubusercontent.com/1791149/226411832-7e8dbe11-80c0-4a30-bbaf-10208eb5acf5.mp4

___
*Import with error and stop on errors*

https://user-images.githubusercontent.com/1791149/226411892-243282e1-de98-4071-8084-4274b9bf2666.mp4

___
*Import with error and no stop on errors (error log file)*

https://user-images.githubusercontent.com/1791149/226411971-ed63894a-68ff-4a33-83a0-a12f095113da.mp4


packages/compass-e2e-tests/tests/collection-import.test.ts
```diff
@@ -757,4 +771,179 @@ describe('Collection import', function () {
...
+    // Select the file.
+    await browser.selectFile(Selectors.ImportFileInput, csvPath);
+
+    // Wait for the modal to appear.
+    const importModal = await browser.$(Selectors.ImportModal);
+    await importModal.waitForDisplayed();
+
+    // Confirm import.
+    await browser.clickVisible(Selectors.ImportConfirm);
+    g;
```
Anemy, line 922:
```suggestion
```

packages/compass-import-export/src/modules/import.ts
```diff
@@ -348,63 +362,107 @@ export const startImport = () => {
...
+      // Show the first two errors and a message that more errors exists.
+      const statusMessage = `${errors
+        .slice(0, 2)
+        .map((error) => error.message)
+        .join('\n')}${
+        errors.length > 2
+          ? '\nMore errors occurred, open the error log to view.'
+          : ''
+      }`;
+      openToast(importToastId, {
```
gribnoysup, line 422: Should all of these toasts inside this action use the same id?

packages/compass-e2e-tests/tests/collection-import.test.ts
```diff
@@ -26,16 +27,20 @@ async function importJSONFile(browser: CompassBrowser, jsonPath: string) {
...
+  await importModal.waitForDisplayed({ reverse: true });
 
-  // wait for the modal to go away
-  await importModal.waitForDisplayed({ reverse: false });
+  // Wait for the done toast to appear and close it.
+  const toast = Selectors.ImportSucceededToast;
+  await browser.$(toast).waitForDisplayed();
+  await browser.$(Selectors.closeToastButton(toast)).waitForDisplayed();
+  await browser.waitForAnimations(toast);
+  await browser.clickVisible(Selectors.closeToastButton(toast));
```
gribnoysup, line 41: `clickVisible` does `waitForAnimations` internally, I am a bit surprised that you need to wait first and then click 🤔

packages/compass-e2e-tests/helpers/selectors.ts
```diff
@@ -550,16 +550,22 @@ export const incompleteSchemaForClonedDocMsg =
...
-  '[data-testid="import-modal"] [data-testid="done-button"]';
-export const ImportErrorBox = '[data-testid="import-error-box"]';
+export const ImportInProgressToast = '[data-testid="import-toast-in-progress"]';
+export const ImportSucceededToast =
+  '[data-testid="import-toast-result-completed"]';
+export const ImportCompletedWithErrorsToast =
+  '[data-testid="import-toast-result-errors-completed"]';
+export const ImportAbortedToast = '[data-testid="import-toast-result-aborted"]';
+export const ImportFailedToast = '[data-testid="import-toast-result-failed"]';
+export const ImportToastAbort =
```
gribnoysup, line 567: Do you need these specific testids if you are correctly checking for text inside the toast in tests already?

packages/compass-e2e-tests/tests/collection-import.test.ts
```diff
@@ -26,16 +27,20 @@ async function importJSONFile(browser: CompassBrowser, jsonPath: string) {
...
-  const doneButton = await browser.$(Selectors.ImportDone);
-  await doneButton.waitForDisplayed();
-  await browser.clickVisible(Selectors.ImportDone);
+  // Wait for the modal to go away.
+  await importModal.waitForDisplayed({ reverse: true });
 
-  // wait for the modal to go away
-  await importModal.waitForDisplayed({ reverse: false });
+  // Wait for the done toast to appear and close it.
+  const toast = Selectors.ImportSucceededToast;
```
gribnoysup, line 37: Feel free to leave this, but in other tests if we need to refer to the same element like that we do 

```suggestion
  const toastElement = await browser.$(Selectors.ImportSucceededToast);
```

instead of copying a string in another variable

packages/compass-components/src/hooks/use-toast.tsx
```diff
@@ -128,19 +137,24 @@ export const ToastArea: React.FunctionComponent = ({ children }) => {
...
-        ))}
+        {toasts.map(
+          ([
+            id,
+            { dataTestId, dismissible, title, body, variant, progress },
+          ]) => (
+            <Toast
+              className={toastStyles}
+              key={id}
+              data-testid={dataTestId ?? `toast-${id}`}
```
gribnoysup, line 148: Super nit, but I would prefer not to have this to be honest. Passed id is as good testid as any to target the element if needed and we should be checking expected content in toasts, not attributes

gribnoysup (commented):
Seems like now that import is non blocking, we need to account for things like disconnecting in the middle of it:

image

packages/compass-import-export/src/modules/import.ts
```diff
@@ -280,17 +261,37 @@ export const startImport = () => {
     const abortSignal = abortController.signal;
     dispatch(onStarted(abortController));
 
-    let promise;
+    openToast(importToastId, {
+      dataTestId: 'import-toast-in-progress',
+      title: `Importing ${fileType} file...`,
```
gribnoysup, line 266: Should we pass a filename here or something to keep some context about what's getting imported?

gribnoysup, line 266:
```suggestion
      title: `Importing ${fileType} file…`,
```

packages/compass-import-export/src/hooks/use-toast-action.tsx
```diff
@@ -0,0 +1,35 @@
...
+  actionHandler: () => void;
+  actionText: string;
+}) {
+  return (
+    <>
+      {statusMessage}
+      <Link
+        data-testid={`toast-action-${actionText}`}
+        onClick={actionHandler}
+        href="#"
```
gribnoysup, line 28: I know that we don't do any a11y anymore, but can we please render this as button instead of a link? It's very low effort to do

packages/compass-import-export/src/modules/import.ts
```diff
@@ -300,23 +301,36 @@ export const startImport = () => {
...
           errors: errors.slice(), // make sure it is not the same variable
         })
       );
+      if (!importDone) {
+        // Update the toast with the new progress.
+        openToast(importToastId, {
+          dataTestId: 'import-toast-in-progress',
+          title: `Importing ${fileType} file...`,
+          body: useToastAction({
+            statusMessage: `${docsWritten}/${guessedTotal}`,
```
gribnoysup, line 322: We had a `title` in the previous implementation that would show you what these numbers mean actually as without context they are kinda confusing. Would it make sense to have something similar here? Maybe just use this text as a message instead of just numbers?

packages/compass-import-export/src/hooks/use-toast-action.tsx
```diff
@@ -0,0 +1,35 @@
...
+        className={toastActionStyles}
+      >
+        {actionText}
+      </Link>
```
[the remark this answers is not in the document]
Anemy: Started floating it to the right 👍 let me know how that looks.

packages/compass-import-export/src/hooks/use-toast-action.tsx
Anemy: Good catch, updated

packages/compass-import-export/src/modules/import.ts
Anemy: Nice suggestion, I was using the type from the designs, but I like the sound of the filename instead of just the file type. I'll double check with product before updating.

packages/compass-components/src/hooks/use-toast.tsx
Anemy: Removed 👍

packages/compass-e2e-tests/helpers/selectors.ts
Anemy: Removed the extra selectors for variants 👍 definitely cleaner now, nice.

packages/compass-e2e-tests/tests/collection-import.test.ts
Anemy: Ah yup. This was not needed.

packages/compass-import-export/src/modules/import.ts
```diff
@@ -245,6 +214,18 @@ export const startImport = () => {
       transform,
     } = importData;
 
+    // Create the error log output file.
+    const userDataPath = getUserDataFolderPath();
+    if (!userDataPath) {
+      throw new Error('cannot access user data for error log generation');
+    }
```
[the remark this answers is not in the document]
Anemy: Good question, I could see both sides having advantages:
- If we don't break early here a user wouldn't be able to get a full list of import errors that occur. That could be potentially frustrating if errors do occur, a user wouldn't be able to discern what went wrong and what data made it in.
- On the other hand if a user can't import because of a file issue, they wouldn't be able to run into that possible situation, however they're probably more frustrated as they can't import to begin with.

I just updated it to still proceed and add the error as the first error so it's always shown in the modal.

packages/compass-import-export/src/modules/import.ts
```diff
@@ -300,23 +301,36 @@ export const startImport = () => {
...
           docsProcessed,
           errors: errors.slice(), // make sure it is not the same variable
         })
       );
+      if (!importDone) {
+        // Update the toast with the new progress.
+        openToast(importToastId, {
+          dataTestId: 'import-toast-in-progress',
+          title: `Importing ${fileType} file...`,
+          body: useToastAction({
```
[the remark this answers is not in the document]
Anemy: Moved the toast management into react components to make it more structured. 👍

packages/compass-import-export/src/modules/import.ts
Anemy: Updated this to
```
statusMessage={`${docsWritten}/${docsProcessed} document${
  docsWritten !== 1 ? 's' : ''
} written.`}
```
How does that look? Do you think having a `title` on the numbers will still benefit folks?

packages/compass-import-export/src/modules/import.ts
```diff
@@ -300,23 +301,36 @@ export const startImport = () => {
...
       dispatch(
         onProgress({
           docsWritten,
           docsProcessed,
           errors: errors.slice(), // make sure it is not the same variable
         })
       );
+      if (!importDone) {
+        // Update the toast with the new progress.
+        openToast(importToastId, {
```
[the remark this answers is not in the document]
Anemy: I found that having one `openToast` made the use-toast implementation a bit cleaner. An update toast would likely be duplicating the code in `openToast`, and then we'd have to have some sort of way of handling when the toast doesn't exist in the logic. For the updates here we'd likely need to pass most of the props anyway. Do you have a strong preference towards adding an `updateToast`?

`openToast` in `use-toast.tsx`: https://github.com/mongodb-js/compass/blob/19fdffe62918cf6471ec4c70e7d56c95f9057a51/packages/compass-components/src/hooks/use-toast.tsx#L35

packages/compass-import-export/src/modules/import.ts
Anemy: Updated 👍 
Screenshot 2023-03-27 at 11 00 20 AM

Anemy:
Looks like the abort in progress test is a bit flaky. I'll fix that before merging, feel free to review in the meantime.

lerouxb:
Regarding abort tests: I was thinking about it yesterday, actually because I'd like a way to properly test skipping CSV field detection as well. All I can think of would be to generate a giant temporary file locally (ie. probably not part of fixtures) and import that. Then you know it will take several minutes to either analyze (in my case) or import (in yours) so there's enough time for the tests to click either skip or stop while it is running. The exact contents of the file doesn't matter. Just do 100 columns, 1 million or whatever rows, random numbers. Or something like that.

I don't mind doing that as a follow-up - we can do both in one test. ie. skip, then click import, then abort it.

Anemy:
@lerouxb Added a test that imports a csv with a million documents. We skip the analyze csv. How does that look? We can followup with a test that has the csv fields analysis not skipped if we want.

Anemy:
Looks like it imports the 1 million faster than it can click abort sometimes lol. I'll try making it 10m.

gribnoysup (commented):
The app still throws when disconnecting during import and the tooltip is still on the screen on the connection screen:

image

```
Uncaught (in promise) TypeError: Cannot read properties of null (reading 'databases')
    at store.refreshNamespaceStats (instance-store.js:188:23)
    at EventEmitter.eval (instance-store.js:342:11)
    at EventEmitter.emit (index.js:202:33)
    at AppRegistry.emit (app-registry.ts:270:26)
    at reducer (global-app-registry.ts:60:13)
    at combination (redux.js:571:29)
    at dispatch (redux.js:299:22)
    at eval (index.js:24:16)
    at dispatch (redux.js:702:28)
    at eval (import.ts:390:5)
```

Based on the stack trace not everything is cleaned up fully and something still emits refresh event in the import-export plugin even after disconnecting causing instance-store to throw

packages/compass-import-export/src/components/import-toast.tsx
```diff
@@ -0,0 +1,255 @@
...
+  }, [errors, errorLogFilePath, cancelImport, status]);
+
+  useEffect(() => {
+    return () => {
+      // Abort the import operation when it's in progress and the import is going away.
+      abortIfInProgress.current?.();
+    };
+  }, []);
+
+  // We return null as we're using this hook as a standalone component.
```
gribnoysup, line 237: I am not sure I follow why is this a component in the first place? All the effects here seem to be just parts of the async action flow that actually does the import, so feel like this belongs in the store, not here

packages/compass-import-export/src/modules/import.ts
```diff
@@ -252,17 +253,27 @@ export const startImport = () => {
...
+    const errors: ErrorJSON[] = [];
+
+    let errorLogFilePath;
+    try {
+      errorLogFilePath = await getErrorLogPath(fileName);
+    } catch (err: any) {
+      errors.push(err as Error);
+    }
+    const errorLogWriteStream = errorLogFilePath
+      ? fs.createWriteStream(errorLogFilePath)
```
gribnoysup, line 274: Are we sure opening a write stream will always work? Should this be guarded by try catch or do we want to hard fail if for some reason we can't write logs to the file?

packages/compass-import-export/src/modules/import.ts
```diff
@@ -276,45 +287,51 @@ export const startImport = () => {
...
+        // If we are importing a massive file with many errors we don't
+        // want to run out of memory.
+        errors.push(err);
+      }
     };
 
+    let importDone = false;
+
+    // Note: This progress callback can be called
+    // after the import has completed/aborted.
```
gribnoysup, line 320: Not sure I follow why would we want that 🤔 throttle returns both flush and cancel methods on the returned throttled function, seems like in case of abort we would want to cancel any progress updates, or am I missing something?

packages/compass-import-export/src/components/import-modal.tsx
```diff
@@ -203,6 +153,10 @@ function ImportModal({
     React
   );
 
+  const handleImportBtnClicked = useCallback(() => {
+    startImport();
+  }, [startImport]);
```
gribnoysup, line 158: Feel free to leave this, but as I pointed out in another PR for this modal, this effectively does nothing but adds a bunch of additional function calls without actually bringing any memoization benefits as there is only one dependency in this callback that is directly used inside

packages/compass-import-export/src/modules/import.ts
Anemy: Ah thanks, should be a few lines up in the try catch. Updated.

packages/compass-import-export/src/components/import-toast.tsx
Anemy: Thanks for bringing this up. Moved the handling of these toast functions to the import flow like they were before. The rendering of the toasts is still in this `components/import-toast.tsx` file, however it's not directly called from the import async flow. Does that make sense to you? Is there a better place?

packages/compass-import-export/src/modules/import.ts
Anemy: I might have written this before I flushed the throttled function 😅 . I'll check again and make sure it's happening as expected and remove this extra flag then.

packages/compass-import-export/src/modules/import.ts
```diff
@@ -276,44 +297,62 @@ export const startImport = () => {
...
-      const guessedTotal = Math.max(
-        docsProcessed,
-        Math.ceil(fileSize / averageSize)
-      );
-      dispatch(onGuesstimatedDocsTotal(guessedTotal));
+      if (importDone) {
+        // When on error occurs in importing it will end the import.
+        // Sometimes this is while the stream is waiting on an insert response
+        // from the MongoDB server. In that case we ignore the progress
+        // so that the errors aren't overridden.
```
gribnoysup, line 347: Just for me to understand better: does it mean that even when you cancel, we still continue to write documents? If you aborted the write, shouldn't any in-progress request fail with an error instead of finishing and emitting progress? 🤔 If import function can emit onProgress after import is done, this sounds like the flow I would expect to be handled by the import code, not by the store slice

packages/compass-import-export/src/modules/import.ts
```diff
@@ -353,63 +392,98 @@ export const startImport = () => {
...
+      });
+      debug('Error while importing:', err.stack);
+
+      showFailedToast(err);
+
+      return dispatch(onFailed(err));
+    } finally {
+      importDone = true;
+      errorLogWriteStream?.close();
+      progressCallback.flush();
```
gribnoysup, line 422: I don't think you want flush here for the case when the error happened though, it's convenient to handle both in finally, but stopping import (whether by aborting or error being thrown in the process) should stop reporting progress too, so why would you want to flush these especially after doing all the error handling that affects the state?

packages/compass-import-export/src/modules/import.ts
```diff
@@ -353,63 +392,98 @@ export const startImport = () => {
...
+      ignoreBlanks,
+      stopOnErrors,
+      hasExcluded: exclude.length > 0,
+      hasTransformed: transform.length > 0,
+    };
+
+    // Don't emit when the data service is disconnected.
+    if (dataService === state.dataService.dataService) {
+      dispatch(globalAppRegistryEmit('import-finished', payload));
+    }
```
gribnoysup, line 486: I'm not sure this will have an effect that expect it to have (there is still an error thrown when you disconnect during import and tooltip stays on the screen) for two reasons:

1. `state` is a reference to the state created at the very beginning of the import as is this `dataService` variable (at the very top of the action you have `dataService = state.dataService.dataService`). So even when if your state changes and new `dataService` is set in state, you are referencing stale state, this condition will always be true inside this action. This is why `redux-thunk` docs even have a section on [how to correctly access state in actions](https://redux.js.org/usage/writing-logic-thunks#accessing-state):

> Since the state is synchronously updated as soon as the reducers process an action, you can call getState after a dispatch to get the updated state.

2. even if the state was correctly accessed here, there is no code in this store that would actually clean up data service from the store when data service is disconnected. I see that you are emitting a `cancelImport` action when data service disconnects, but this event will not do anything about dataService being in the store. This is one of the reasons redux documentation suggests actions to be modelled around [application events that occur](https://redux.js.org/style-guide/#model-actions-as-events-not-setters) and not specific "multi-purpose" setters / getters. If you do it like that, it's much easier to reason about what is the application flow and how different parts of the store should react to this. In this case `dataServiceDisconnected` can handle cancelling any ongoing operations and reducer can react to this action and clean up the state, which would ensure that you don't have disconnected dataService there

packages/compass-import-export/src/modules/import.ts
gribnoysup: Ah, gotcha, missed the flush at the end, left a comment there

packages/compass-import-export/src/components/import-toast.tsx
gribnoysup: I think you found the best way to split it, nice!

packages/compass-import-export/src/modules/import.ts
Anemy: What was happening is the insert request is in progress when the import is aborted, when that request succeeds it is emitting on progress as more documents were written. Thinking about your suggestion though it makes sense that the store shouldn't have to manage that. I'll update it so that this can't happen and we can remove this long comment.

packages/compass-import-export/src/modules/import.ts
Anemy: Yes - thanks!
Updated the `data-service-disconnected` event handler to emit a redux event that is handled by two places:
- The data service reducer is now cleaned up/reset when the disconnected event happens
- The import reducer now aborts and resets the abortControllers on the disconnected event.

Updated this check to check for the the active `dataService` from `getState` matching the dataService that was initially used. No more error on disconnect while an import is in progress.

packages/compass-import-export/src/modules/import.ts
Anemy: Makes sense, updated to flush only on success, and updated both `importCSV` and `importJSON` to not report progress when they've been aborted.
A bit unrelated, also removed some unused state related to the progress which wasn't being used anymore.

packages/compass-import-export/src/utils/get-user-data-file-path.ts
```diff
@@ -0,0 +1,9 @@
+import path from 'path';
+import { getStoragePaths } from '@mongodb-js/compass-utils';
+const { appName, basepath } = getStoragePaths() || {};
+
+export function getUserDataFolderPath() {
+  if (appName === undefined || basepath === undefined) return;
```
lerouxb, line 6: This should maybe assert or just throw an error in this case because you now have a mixed return type.

lerouxb (approved):

packages/compass-import-export/src/modules/import.ts
```diff
@@ -169,64 +167,53 @@ export const INITIAL_STATE: State = {
...
+  type: FILE_SELECT_ERROR,
+  error,
 });
 
-export const onFailed = (error: Error) => ({ type: FAILED, error });
+async function getErrorLogPath(fileName: string) {
+  // Create the error log output file.
+  const userDataPath = getUserDataFolderPath();
+  if (!userDataPath) {
+    throw new Error('cannot access user data for error log generation');
```
lerouxb, line 205: I'd move this check and this error into getUserDataFolderPath().

packages/compass-import-export/src/utils/get-user-data-file-path.ts
Anemy: We throw now 👍

Anemy:
A lot of the feedback was from Sergey, however as he's on vacation (no ping) there won't be a chance to get a final check from him on this. When he's back I'll follow up and see if there's any more to do here. Thanks for the good thorough review.

Once CI is green I'll merge.

PR #4173, opened 2023-03-21
feat(compass-import-export): add export json COMPASS-6427

Anemy:
COMPASS-6427

This adds the backend for exporting json. Not yet used, only tested with units tests. It will be integrated in further tickets in [the epic](https://jira.mongodb.org/browse/COMPASS-5576).

Question: Should we zip the fixtures we use here like we do for some of the end to end test fixtures?


packages/compass-import-export/src/export/gather-fields.ts
```diff
@@ -0,0 +1,169 @@
...
+    throw err;
+  }
+}
+
+type GatherFieldsOptions = {
+  dataService: DataService;
+  ns: string;
+  abortSignal: AbortSignal;
+  filter?: Document;
+  progressCallback: (index: number) => void;
```
lerouxb, line 67: You can probably make this optional like we we now do with other progressCallbacks. If anything it makes testing easier when you really don't care about progress.

packages/compass-import-export/src/export/gather-fields.ts
```diff
@@ -0,0 +1,169 @@
...
+      }
+
+      current = current[fieldName] as Projection;
+    }
+  }
+
+  return projection;
+}
+
+// TODO(COMPASS-6426): We will fill out the rest of this function and
```
lerouxb, line 148: I'm actually now wondering if progress here would be overkill given that a) it is a sample and b) it is largely blocking: The only waits are around getting the sample and then running mongoSchema. mongodb-schema's parseSchema streams internally, but we're giving it an array of docs so it is Readable.from() with all the data in one go. And it has no progress mechanism yet anyway.

packages/compass-import-export/src/export/export-json.ts
```diff
@@ -0,0 +1,129 @@
...
+const debug = createDebug('export-json');
+
+type ExportJSONOptions = {
+  dataService: DataService;
+  ns: string;
+  output: Writable;
+  abortSignal: AbortSignal;
+  query?: ExportQuery;
+  aggregation?: ExportAggregation;
+  progressCallback: (index: number) => void;
```
lerouxb, line 24: I'd make progressCallback and abortSignal optional to keep in line with where import ended up.

packages/compass-import-export/src/export/export-json.ts
```diff
@@ -0,0 +1,129 @@
...
+  if (variant === 'relaxed') {
+    return {
+      relaxed: true,
+    };
+  }
+  return variant === 'canonical'
+    ? {
+        relaxed: false, // canonical
+      }
+    : undefined; // default
```
lerouxb, line 45: Does this end up making to relaxed true/false or a third thing? Just wondering if default used to mean "compass style" ie. we could be losing some custom processing. Like a "semi-relaxed".

packages/compass-import-export/src/export/export-json.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+export async function exportJSON({
+  dataService,
+  ns,
+  output,
+  abortSignal,
+  query = { filter: {} },
```
lerouxb, line 53: We have to be able to export an aggregation as well, I think. So I'm wondering if we shouldn't just pass in something like a cursor. ie. build the whole query and run it on the outside. 🤔

packages/compass-import-export/src/export/gather-fields.ts
```diff
@@ -0,0 +1,169 @@
...
+
+async function analyzeSchema({
+  dataService,
+  ns,
+  abortSignal,
+  filter,
+  sampleSize,
+}: AnalyzeSchemaOptions) {
+  try {
+    // TODO(COMPASS-6426): Should we use the other aspects of the query in this sample. (project/sort/limit/skip)
```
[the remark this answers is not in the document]
lerouxb: And if there's already a projection for a find query we also skip the select fields step and just use the projection.

So project can probably be ignored. The others we can/should probably include?

Maybe you should just replace the filter parameter with the same query parameter that exportJSON takes 🤔

packages/compass-import-export/src/export/export-json.ts
lerouxb: I remember Anna mentioning that we do something for dates or longs or something to make the json less verbose when it doesn't need to be, but I can't find it now.

packages/compass-import-export/src/export/export-json.spec.ts
```diff
@@ -0,0 +1,455 @@
...
+        ns: testNS,
+        output: temp.createWriteStream(),
+        variant: 'default',
+        abortSignal: abortController.signal,
+        progressCallback: () => {},
+      });
+
+      throw new Error('did not expect export to succeed');
+    } catch (err: any) {
+      expect(err.message).to.equal('example error cannot fetch docs');
```
lerouxb, line 362: if you look at some of my tests you can:

```typescript
    await expect(promise).to.be.rejectedWith(
      Error, // class of error
      'Parser cannot parse input: expected a value' // message
    );
```

That way you can make sure the code throws whereas your tests now would pass if the code somehow succeeds. a bit neater than the try/catch as well.

So it would be something like:

```typescript
await expect(exportJSON({
        dataService,
        ns: testNS,
        output: temp.createWriteStream(),
        variant: 'default',
        abortSignal: abortController.signal,
        progressCallback: () => {},
      }).to.reject(Error, 'example error cannot fetch docs');
```

It would also give you better test failure output. ie. it would print what it _did_ get.

packages/compass-import-export/src/export/gather-fields.ts
Anemy: Sounds good, if there's no progress to track let's remove it. We can do that in the COMPASS-6426 ticket work.

packages/compass-import-export/src/export/export-json.ts
Anemy: Good thinking. Updated the `exportJSON` function to take an `input` which is a cursor, then created an `exportJSONFromQuery` and `exportJSONFromAggregation`. Makes it a bit more general and makes these types nicer. I'm thinking when we integrate these we can revisit and see if there's a nicer interface to provide here 👍

packages/compass-import-export/src/utils/collection-stream.ts
```diff
@@ -296,23 +299,42 @@ export const createCollectionWriteStream = function (
...
+  limit?: number;
+  skip?: number;
+  projection?: Document;
+};
+
+type ReadableCollectionStreamOptions = {
+  dataService: DataService;
+  ns: string;
+  query: ExportQuery;
+  aggregation?: ExportAggregation;
```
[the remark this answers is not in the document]
Anemy: 👍 updated and answered in https://github.com/mongodb-js/compass/pull/4173#discussion_r1149613367

packages/compass-import-export/src/export/export-json.spec.ts
Anemy: Ah this is good. Thanks for mentioning. Updated tests.

lerouxb (approved):
Looks good. I think my only remaining concern is around the "compass" ejson export variant we apparently currently have, but we can get back to it in a future PR.

PR #4182, opened 2023-03-23
fix(import): Optimise CSV field type detection

lerouxb:
I was investigating if we can easily speed up analyzeCSVFields() because for very large files it can take quite long. I realised you can only speed it up by around 2x so we'll probably make it sample one way or another anyway, but then I figured I might as well open the PR regardless.

I added a script for analyzing a csv file from the commandline which is handy for benchmarking and debugging. I figure we might as well keep it here. Or do we have a better place for that sort of thing?

Beyond this point there are rapidly diminishing returns. Big files are big and javascript is only so fast. You'd probably need WebAssembly and/or worker threads. I commented out the detection code and made it return just string and even that isn't _that_ much faster. If you comment out the looping through the cells in the row to be analyzed entirely it only speeds up by maybe a further 2 or 3 times.

Another nice thing about this PR is that it brings the stream pipeline pattern in line with what we do in importCSV and importJSON. I wrote those after this one and I think it is better than this event based approach.


packages/compass-import-export/src/import/analyze-csv-fields.spec.ts
```diff
@@ -88,6 +88,7 @@ describe('analyzeCSVFields', function () {
         'objectId',
         'regex',
         'timestamp',
+        'symbol',
```
lerouxb, line 91: I ran into the same thing when I tried replacing the CSV parser yesterday: I have no idea how this passed before because we can't detect symbols as a type. The file is basically even empty and it should have been skipped all along.

packages/compass-import-export/src/import/analyze-csv-fields.spec.ts
```diff
@@ -198,7 +199,7 @@ describe('analyzeCSVFields', function () {
     });
 
     // only looked at the first row because we aborted before even starting
-    expect(result.totalRows).to.equal(1);
+    expect(result.totalRows).to.equal(0);
```
lerouxb, line 202: We now abort the pipeline using node stream promises's pipeline machinery so the timing it subtly different. It is kinda arbitrary whether we looked at 0, 1 or 2 rows before it kicked in.

packages/compass-import-export/src/import/analyze-csv-fields.ts
```diff
@@ -141,77 +137,81 @@ function pickFieldType(field: CSVField): CSVParsableFieldType {
...
-          }
-
-          initResultFields(result, headerFields);
-        }
-
-        addRowToResult(result, headerFields, results.data, ignoreEmptyStrings);
-
-        ++result.totalRows;
+        initResultFields(result, headerFields);
+        fields = Object.values(result.fields);
```
lerouxb, line 176: This caches the result of Object.values() outside the loop. The kind of micro-optimisation I'd normally not bother with, but it does begin to matter if you're looking at millions of rows.

packages/compass-import-export/src/import/analyze-csv-fields.ts
```diff
@@ -141,77 +137,81 @@ function pickFieldType(field: CSVField): CSVParsableFieldType {
...
-    input = input.pipe(validator).pipe(byteCounter).pipe(stripBomStream());
-
-    Papa.parse(input, {
-      delimiter,
-      header: true,
-      step: function (results: Papa.ParseStepResult<PapaRowData>, parser) {
-        if (abortSignal?.aborted && !aborted) {
-          aborted = true;
-          result.aborted = true;
-          parser.abort();
```
lerouxb, line 178: Oh and like listCSVFields() this also didn't properly destroy the input stream whereas `await pipeline()` wit the signal passed in does.

Anemy (approved):
lgtm!
I'm thinking we might want to make the `abortSignal` option required and not optional to clean it up a bit. I don't think there are cases where we aren't using it. Unrelated so doesn't have to happen here, but I'm curious on your thoughts.

PR #4190, opened 2023-03-24
WIP: skippable analyzeCSVFields()

lerouxb:
https://user-images.githubusercontent.com/69737/227579422-27e847a3-f17a-474b-bb84-097a0ce57848.mov

### DIFF
--- a/packages/compass-import-export/src/components/import-modal.tsx
+++ b/packages/compass-import-export/src/components/import-modal.tsx
@@ -9,6 +9,7 @@ import {
   css,
   spacing,
   FormFieldContainer,
+  Body,
 } from '@mongodb-js/compass-components';
 import { useTrackOnChange } from '@mongodb-js/compass-logging';
 
@@ -31,6 +32,7 @@ import formatNumber from '../utils/format-number';
 import {
   startImport,
   cancelImport,
+  skipCSVAnalyze,
   selectImportFileName,
   setDelimiter,
   setStopOnErrors,
@@ -61,11 +63,18 @@ const closeButtonStyles = css({
   marginRight: spacing[2],
 });
 
+const progressStyles = css({
+  display: 'flex',
+  flexDirection: 'column',
+  alignItems: 'center',
+});
+
 type ImportModalProps = {
   isOpen: boolean;
   ns: string;
   startImport: () => void;
   cancelImport: () => void;
+  skipCSVAnalyze: () => void;
   closeImport: () => void;
   errors: Error[];
   status: ProcessStatus;
@@ -92,6 +101,9 @@ type ImportModalProps = {
   guesstimatedDocsTotal: number;
   guesstimatedDocsProcessed: number;
 
+  analyzeBytesProcessed: number;
+  analyzeBytesTotal: number;
+
   /**
    * See `<ImportPreview />`
    */
@@ -114,6 +126,8 @@ function ImportModal({
   cancelImport,
   closeImport,
 
+  skipCSVAnalyze,
+
   errors,
   status,
 
@@ -132,6 +146,9 @@ function ImportModal({
   docsWritten,
   guesstimatedDocsTotal,
   guesstimatedDocsProcessed,
+
+  analyzeBytesProcessed,
+  analyzeBytesTotal,
 
   fields,
   values,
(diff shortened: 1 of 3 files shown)


packages/compass-import-export/src/modules/import.ts
```diff
@@ -443,31 +448,76 @@ export const cancelImport = () => {
...
     });
 
     const input = fs.createReadStream(fileName);
 
+    const progressCallback = _.throttle(function ({
+      bytesProcessed,
+    }: {
+      bytesProcessed: number;
+    }) {
+      console.log({ bytesProcessed, fileSize });
```
Anemy, line 506:
```suggestion
```

packages/compass-import-export/src/components/import-modal.tsx
```diff
@@ -61,11 +65,34 @@ const closeButtonStyles = css({
   marginRight: spacing[2],
 });
 
+const progressStyles = css({
+  display: 'flex',
+  flexDirection: 'column',
+  alignItems: 'center',
+  backgroundColor: palette.gray.light3, // TODO: dark mode
```
gribnoysup, line 72: Is it passable in dark mode without this? If not, maybe we can come up with some color here before merging?

packages/compass-import-export/src/components/import-modal.tsx
```diff
@@ -154,6 +189,10 @@ function ImportModal({
     startImport();
   }, [startImport]);
 
+  const handleSkipCSVAnalyze = useCallback(() => {
+    skipCSVAnalyze();
+  }, [skipCSVAnalyze]);
```
gribnoysup, line 194: Just a note that when written like that, useCallback adds no memoization value here, but an additional function call, which is sort of unnecessary. I'm guessing it just follows the pattern in the component, so maybe something to keep in mind for the future

packages/compass-import-export/src/components/import-preview.tsx
```diff
@@ -322,12 +309,7 @@ function ImportPreview({
         data={values}
         columns={gapOrFields.map((field) => {
           if (typeof field !== 'string' && 'path' in field) {
-            return FieldHeader(
-              field,
-              analyzed,
-              onFieldCheckedChanged,
-              setFieldType
-            );
+            return FieldHeader(field, onFieldCheckedChanged, setFieldType);
```
gribnoysup, line 312: Just wondering, are we working around some weird leafygreen limitation here? I got really confused by this "component" when going through code

packages/compass-import-export/src/components/import-modal.tsx
```diff
@@ -216,18 +255,40 @@ function ImportModal({
...
+        {fileType === 'csv' && !csvAnalyzed && (
+          <FormFieldContainer className={progressStyles}>
+            <Body weight="medium">Detecting field types</Body>
+            {analyzeBytesTotal && (
+              <div className={loaderStyles}>
+                <SpinLoader />
+                <Body>
+                  {Math.round(
+                    (analyzeBytesProcessed / analyzeBytesTotal) * 100
+                  )}
```
gribnoysup, line 279: Maybe also something to keep in mind for the future as I'm guessing you are just following a pattern here. The only place where these numbers are used is this placeholder, so following [redux design guidelines](https://redux.js.org/style-guide/#connect-more-components-to-read-data-from-the-store), you might want to rather have a separate "analyze in-progress" component, which will be the only one that subscribes to these values via connect, not the whole modal

gribnoysup (approved):

packages/compass-import-export/src/components/import-modal.tsx
lerouxb: aaargh I forgot about dark mode

packages/compass-import-export/src/components/import-modal.tsx
lerouxb: yeah I just copied the stuff around it. But good point.

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: I don't actually know what that is about, to be honest. I found it like that, I think. The diff / blame is a bit incriminating, but I think I just moved it. I can just merge it in?

packages/compass-import-export/src/components/import-modal.tsx
lerouxb: Screenshot 2023-03-27 at 12 17 41

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: Removed it.

packages/compass-import-export/src/components/import-modal.tsx
lerouxb: Cleaned that up now.

PR #4213, opened 2023-03-30
fix(compass-components): display width on default file input variant

Anemy:
Looks like passing a `leftGlyph` to LeafyGreen's Button made the width apply over our passed width. Also refactored to not use enums, and to follow lower case variant convention.
| before | after |
| - | - |
| Screenshot 2023-03-30 at 11 06 20 AM | Screenshot 2023-03-30 at 11 27 45 AM |

### DIFF
--- a/packages/compass-components/src/components/file-input.tsx
+++ b/packages/compass-components/src/components/file-input.tsx
@@ -59,12 +59,10 @@ const buttonSmallStyles = css({
   },
 });
 
-const buttonHorizontalStyles = css({
-  width: '100%',
-});
-
-const buttonVerticalStyles = css({
-  width: '100%',
+const buttonDefaultStyles = css({
+  // We !important here to override the LeafyGreen button width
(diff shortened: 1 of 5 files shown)


lerouxb (approved):

packages/compass-components/src/components/file-input.tsx
```diff
@@ -320,11 +314,9 @@ function FileInput({
         <Button
           id={id}
           data-testid="file-input-button"
-          className={cx({
-            [buttonSmallStyles]: variant === Variant.Small,
-            [buttonHorizontalStyles]: variant === Variant.Horizontal,
-            [buttonVerticalStyles]: variant === Variant.Vertical,
-          })}
+          className={
+            variant === 'small' ? buttonSmallStyles : buttonDefaultStyles
```
lerouxb, line 318: Aah much nicer 👍

PR #4219, opened 2023-03-31
feat(export): exportCSV() COMPASS-6428 COMPASS-6521 COMPASS-6637

lerouxb:
Looks much bigger than it is because it adds lots of test fixtures.

Factored out parts of exportJSON to be able to use it here.

I also updated import a bit so we can import what we now export.

Refactored some of the files that were just called things like "utils.ts". Which does a little part of  [COMPASS-6673](https://jira.mongodb.org/browse/COMPASS-6673)


packages/compass-import-export/src/export/export-csv.ts
```diff
@@ -0,0 +1,457 @@
...
+      return value.toISOString();
+    }
+
+    if (['number', 'boolean'].includes(typeof value)) {
+      return formatValue(value.toString() as string);
+    }
+
+    // Arrays and plain objects that somehow made it here plus unforeseen things
+    // that don't have a _bsontype.
+    // NOTE: Not escaping quotes for objects and arrays. This results in
```
lerouxb, line 174: As mentioned elsewhere I'm having second thoughts about this...

packages/compass-import-export/src/export/export-utils.spec.ts
```diff
@@ -0,0 +1,156 @@
+/* eslint-disable mocha/max-top-level-suites */
```
Anemy, line 1: Is this a useful lint to have? I wonder if we should just disable it if it's not helping us prevent anything.

packages/compass-import-export/src/export/export-csv.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+import { importCSV } from '../import/import-csv';
+import { importJSON } from '../import/import-json';
+
+import allTypesDoc from '../../test/docs/all-bson-types';
+
+const { expect } = chai;
+chai.use(sinonChai);
+chai.use(chaiAsPromised);
+
+describe('exportCSV', function () {
```
Anemy, line 31: In `exportJSON` we have a test for exporting an empty collection, would that be useful to have here as well?

packages/compass-import-export/src/import/import-csv.spec.ts
```diff
@@ -210,14 +192,29 @@ describe('importCSV', function () {
       const ns = 'db.col';
 
       const totalRows = analyzeResult.totalRows;
-      const fields = _.mapValues(
-        analyzeResult.fields,
+      const fields = _.mapValues(analyzeResult.fields, (field, name) => {
```
Anemy, line 195: Small thing, I like to reduce lodash usage where possible, since we're updating this line can we remove the lodash dependency here? We could do something like 
```const fields = Object.entries(analyzeResult.fields).map(([key, value]) => { ... })```

File: packages/compass-import-export/src/csv/csv-utils.ts [hunk 5]
```diff
@@ -11,116 +11,212 @@ import {
...
 const FALSY_STRINGS = ['f', 'false', 'FALSE', 'False'];
 const NULL_STRINGS = ['Null', 'NULL', 'null'];
 
-export function detectFieldType(
+const parenthesis = {
+  '{': '}',
+  '[': ']',
+};
+
+function isEJSON(value: string) {
```
Anemy, line 181: Should we check if the value can be parsed as valid json? (`JSON.parse(value)` doesn't throw).  Or is that too much compute?

packages/compass-import-export/src/import/import-types.ts
```diff
@@ -0,0 +1,37 @@
...
+  op?: any;
+  errorInfo?: Document;
+  /*
+  e.index = index;
+  e.code = code;
+  e.op = op;
+  e.errInfo = errInfo;
+  // https://www.mongodb.com/docs/manual/reference/method/BulkWriteResult/#mongodb-data-BulkWriteResult.writeErrors
+  e.name = index && op ? 'WriteError' : 'WriteConcernError';
+*/
```
Anemy, line 30:
```suggestion
```
Leftover comment? Looks like this was already there, just noticing because of the file move.

packages/compass-import-export/test/docs/all-bson-types.js
```diff
@@ -10,47 +10,51 @@ import {
   MaxKey,
   MinKey,
   ObjectId,
-  Symbol as BSONSymbol,
+  //Symbol as BSONSymbol,
```
Anemy, line 13:
```suggestion
```
Might be nicer to remove the deprecated or leave a comment as to why it's commented. (Same with where `Symbol/BSONSymbol` is used below)

packages/compass-import-export/test/docs/all-bson-types.js
```diff
@@ -10,47 +10,51 @@ import {
...
 } from 'bson';
 
 export default [
   {
+    _id: new ObjectId('642d766b7300158b1f22e972'),
     double: new Double(1.2), // Double, 1, double
     string: 'Hello, world!', // String, 2, string
     object: { key: 'value' }, // Object, 3, object
     array: [1, 2, 3], // Array, 4, array
+    // TODO: this is not being formatted in a way that gets seen by the parser. comes out as blank/null somehow
```
Anemy, line 25: Is this referring to the binData?

Anemy (commented):
Nice work, I'm liking the refactoring. Great tests around all types. Left some comments/questions/thoughts.

packages/compass-import-export/src/csv/csv-types.ts
```diff
@@ -0,0 +1,102 @@
...
+  double: 'Double',
+  boolean: 'Boolean',
+  date: 'Date',
+  string: 'String',
+  null: 'Null',
+  undefined: 'Undefined',
+  objectId: 'ObjectId',
+  binData: 'Binary',
+  uuid: 'UUID',
+  md5: 'MD5',
```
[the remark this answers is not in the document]
lerouxb: We can, but I think that's an implementation detail? What gives me pause is that we used to just call them `UUID` before.

packages/compass-import-export/src/export/export-csv.ts
```diff
@@ -0,0 +1,308 @@
...
+      this.progressCallback?.(this.docsWritten, 'WRITE');
+      cb(null, line);
+    } catch (err: any) {
+      cb(err as Error);
+    }
+  }
+}
+
+// You probably want to use exportCSVFromAggregation() or exportCSVFromQuery() rather
+export async function exportCSV({
```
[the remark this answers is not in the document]
lerouxb: Makes sense. I was on the fence around that. It also doesn't have to be exported.

packages/compass-import-export/src/export/export-csv.ts
```diff
@@ -0,0 +1,308 @@
...
+  ) {
+    //debug('ColumnStream', { chunk });
+    this.columnRecorder.addToColumns(chunk);
+    this.progressCallback?.(this.docsProcessed, 'DOWNLOAD');
+    this.docsProcessed++;
+    cb(null, `${EJSON.stringify(chunk, { relaxed: true })}\n`);
+  }
+
+  getColumns() {
+    // TODO: In case of very large arrays or objects we should probably have a
```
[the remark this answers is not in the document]
lerouxb: I'll remove the comment, actually. We should probably rather speak about it at design sync or whatever.

packages/compass-import-export/src/export/export-utils.spec.ts
lerouxb: We probably should remove it. That just leads to an extra level of nesting for no good reason.

packages/compass-import-export/src/import/import-csv.spec.ts
lerouxb: The file already uses _.isPlainObject() :)

packages/compass-import-export/src/csv/csv-types.ts
```diff
@@ -0,0 +1,102 @@
...
+  int: 'Int32',
+  long: 'Long',
+  double: 'Double',
+  boolean: 'Boolean',
+  date: 'Date',
+  string: 'String',
+  null: 'Null',
+  undefined: 'Undefined',
+  objectId: 'ObjectId',
+  binData: 'Binary',
```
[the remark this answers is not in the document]
lerouxb: I don't know. We never had them before, so that's probably a design sync topic.

packages/compass-import-export/src/csv/csv-types.ts
```diff
@@ -0,0 +1,102 @@
...
+
+// the subset of bson types that we can parse
+export type CSVParsableFieldType =
+  | CSVDetectableFieldType
+  | 'binData'
+  | 'md5'
+  | 'timestamp'
+  | 'decimal'
+  | 'number' // like 'mixed', but for use when everything is an int, long or double.
+  | 'mixed';
```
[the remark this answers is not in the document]
lerouxb: There isn't a serialisation for javascript, javascriptWithScopy, Symbol, DBRef, etc. They just become EJSON and are covered by the ejson detected and parsed type.

packages/compass-import-export/test/docs/all-bson-types.js
lerouxb: Sorry old comment. The thing I was going on about in #dev-tools. Fixed now.

packages/compass-import-export/src/csv/csv-types.ts
```diff
@@ -0,0 +1,102 @@
...
+  | 'double'
+  | 'boolean'
+  | 'date'
+  | 'string'
+  | 'objectId'
+  | 'uuid'
+  | 'regex'
+  | 'minKey'
+  | 'maxKey'
+  | 'ejson' // not a real type, but the fallback for otherwise unserializable values
```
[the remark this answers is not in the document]
lerouxb: I've expanded the comment. It is also explained everywhere else ejson as a type is referenced.

packages/compass-import-export/src/csv/csv-utils.ts
lerouxb: Actually if there's a lot of stuff like that in the file it will still be really slow because each field would it make past those checks. And this is just meant to be heuristics - if it fails during import the parse error will be caught and logged and the user can still select something else manually.

Plus the column would have to look like EJSON for each row encountered for it to detect that which would also help weed out false positives.

So I think this is probably the right compromise for now. We can always make it stricter in a follow-up if it is ever a problem.

lerouxb: Another compromise might be to add an option strict: true/false. Then when analyzing we can set it to false and the second detection that gets done at import time if the column is specified as Mixed can pass strict because then the stakes are a little bit higher if it gets it wrong.

packages/compass-import-export/src/export/export-utils.spec.ts
lerouxb: (but I think we can file a ticket and do that as a follow-up)

Anemy (approved):
Looks like there's a failing test, at least on windows. Once those are green lgtm !

lerouxb:
test-packaged-app-latest is failing due to a change in CSFLE which is a known issue. I'll just merge with that red for now.

I also haven't done any stress testing of exportCSV() with extremely large files yet, but we'll just have to do that when we integrate. It is going to be much easier to do once we can just export stuff from the UI.

PR #4227, opened 2023-04-03
feat(stage-editor): adds support for wizard stages in stage-editor store: COMPASS-6676

himanshusinghs:
## Description
This PR adds subtle changes in our `stage-editor` store to support creating wizard stages that will later be used to render stage wizard forms.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


himanshusinghs:
A few more places will need updating to complete the change, I will convert this back to ready when I am done with changes.

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
```diff
@@ -453,23 +507,49 @@ export const addStage = (
...
+    } = getState();
+    const addAfter = after ?? stages.length;
+    const wizardsUntilAfter = stages
+      .slice(0, addAfter + 1)
+      .filter(({ type }) => type === 'wizard');
+    const addAfterIdxInPipeline = addAfter - wizardsUntilAfter.length;
+
+    const stage = pipelineBuilder.addStage(addAfterIdxInPipeline);
     track('Aggregation Edited', {
       num_stages: getState().pipelineBuilder.stageEditor.stages.length,
```
mabaasit, line 523: The `num_stages` needs to filter out the wizards now. We can create a helper and reuse it across

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
```diff
@@ -38,6 +39,7 @@ export const enum StageEditorActionTypes {
   StageAdded = 'compass-aggregations/pipeline-builder/stage-editor/StageAdded',
   StageRemoved = 'compass-aggregations/pipeline-builder/stage-editor/StageRemoved',
   StageMoved = 'compass-aggregations/pipeline-builder/stage-editor/StageMoved',
+  WizardStageAdded = 'compass-aggregations/pipeline-builder/stage-wizard/WizardStageAdded',
```
mabaasit, line 42: nit: WizardAdded instead?

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
```diff
@@ -764,6 +909,43 @@ const reducer: Reducer<StageEditorState> = (
...
 
+  if (
+    isAction<WizardStageAddAction>(
+      action,
+      StageWizardActionTypes.WizardStageAdded
+    )
+  ) {
+    const { after, usecaseId, formValues } = action;
+    const stages = [...state.stages];
+    stages.splice(after + 1, 0, {
```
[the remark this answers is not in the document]
himanshusinghs: Yup and thanks. I recently updated it.

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
himanshusinghs: Took care of all the tracking that needs updated.

packages/compass-aggregations/src/components/stage-toolbar/index.tsx
```diff
@@ -158,8 +162,15 @@ type StageToolbarOwnProps = Pick<
...
-    const stage = state.pipelineBuilder.stageEditor.stages[ownProps.index];
+    const {
+      pipelineBuilder: {
+        stageEditor: { stages },
+      },
+    } = state;
+    const stage = stages[ownProps.index];
+    assertReduxStage(stage);
     return {
+      idxInPipeline: storeIndexToPipelineIndex(stages, ownProps.index),
```
mabaasit, line 173: Wouldn't this always be same as `ownProps.index`? We are rendering all the stage components (preview, editor and toolbar) inside a stage which is rendered only for ReduxStage.

packages/compass-aggregations/src/components/focus-mode/focus-mode-modal-header.tsx
```diff
@@ -296,12 +314,20 @@ export default connect(
...
+    assertReduxStage(stage);
+
     return {
       stageIndex,
       isEnabled: !stage?.disabled,
-      stages: stages.map((stage) => {
-        return stage.stageOperator;
-      }),
+      stages: stages.reduce<Stage[]>((accumulator, stage, idxInStore) => {
+        if (stage.type === 'stage') {
```
mabaasit, line 323: Can we add test cases to this? Add wizards in the store and then validate the expected behaviour of focus mode (navigation)?

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
```diff
@@ -224,8 +261,10 @@ export const loadPreviewForStagesFrom = (
   return (dispatch, getState) => {
     getState()
       .pipelineBuilder.stageEditor.stages.slice(from)
```
mabaasit, line 263: we can use `pipelineFromStore` helper here as well to avoid `if` check in `forEach`

packages/compass-aggregations/src/components/focus-mode/focus-mode-modal-header.tsx
himanshusinghs: Yes I am on it.

packages/compass-aggregations/src/components/stage-toolbar/index.tsx
himanshusinghs: It won't remain the same as soon as we have Wizards in the state. With wizards the index will shift but because we want to render the stage titles with their actual position, we calculate this.

I did a recent change, where we don't calculate but simply takes it from stage.

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
himanshusinghs: We will have to keep it because our action dispatcher works on indexes of store but pipelineFromStore will filter out the wizards and thus the indexes will be skewed.

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
```diff
@@ -453,28 +542,57 @@ export const addStage = (
...
+
+    pipelineBuilder.cancelPreviewForStage(idxInPipeline);
+    const stage = pipelineBuilder.removeStage(idxInPipeline);
     track('Aggregation Edited', {
-      num_stages,
+      num_stages: pipelineFromStore(stages).length,
       stage_action: 'stage_deleted',
       stage_name: stage.operator,
-      stage_index: at + 1,
+      stage_index: idxInPipeline + 1,
```
himanshusinghs, line 595: I have a feeling that this is incorrect and was incorrect in old implementation as well. Could you help me confirm this @mabaasit ?

The `idxInPipeline` is already the correct index of item to be removed, we don't need to do `+1` to it.

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
```diff
@@ -453,28 +542,57 @@ export const addStage = (
...
+      // that while calculating the corresponding afterIdx in pipeline
+      includeIndex: true,
+    });
+
+    const stage = pipelineBuilder.addStage(addAfterIdxInPipeline);
     track('Aggregation Edited', {
-      num_stages: getState().pipelineBuilder.stageEditor.stages.length,
+      num_stages: pipelineFromStore(stages).length,
       stage_action: 'stage_added',
       stage_index: stage.id + 1,
```
himanshusinghs, line 561: I don't this is correct as well. The `stage.id` is an incremental number which does not necessarily corresponds to the stage index. We probably should just use, `addAfterIdxInPipeline  + 1` here

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
```diff
@@ -490,41 +608,141 @@ export const moveStage = (
...
+      stageAtFromIdx.idxInPipeline === toIdxInPipeline;
+
+    if (pipelineWasNotModified) {
+      dispatch({ type: StageEditorActionTypes.StageMoved, from, to });
+    } else {
+      track('Aggregation Edited', {
+        num_stages: pipelineFromStore(stages).length,
+        stage_action: 'stage_reordered',
+        stage_name: stageAtFromIdx.stageOperator,
+        stage_index: stageAtFromIdx.idxInPipeline + 1,
```
himanshusinghs, line 641: Same concern here regarding the index

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
mabaasit: Yes, this seems to be wrong. We should fix this

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
mabaasit: It seems to be correct in the old and new implementation. We are adding +1 in every event i think.

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
himanshusinghs: Disregarding this like discussed

packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
```diff
@@ -453,28 +542,57 @@ export const addStage = (
...
+      },
+    } = getState();
+    const stageInStore = stages[at];
+    if (!stageInStore || stageInStore.type !== 'stage') {
+      return;
+    }
+
+    const { idxInPipeline } = stageInStore;
+
+    pipelineBuilder.cancelPreviewForStage(idxInPipeline);
```
mabaasit, line 589: 🔥

mabaasit (approved):

PR #4237, opened 2023-04-11
fix(compass-import-export): remove re-count when not available COMPASS-5179, COMPASS-6649

Anemy:
COMPASS-5179, COMPASS-6649

When there's a count from running a query we display that count as the documents to export. This number was only used for ui purposes, so I think removing the count operations when its not available alright. It's not available when the count hasn't been run successfully, which happens on complex queries that timeout, also with views, and time series collections. We already handle the no count case nicely from the exporting aggregations work.

We're also in the process of updating export, which will also involve removing this count, so these changes help there as well.

### DIFF
--- a/packages/compass-import-export/src/modules/export.ts
+++ b/packages/compass-import-export/src/modules/export.ts
@@ -406,10 +406,7 @@ export const openExport =
     const isAggregation = !!aggregation;
     track('Export Opened', { type: isAggregation ? 'aggregation' : 'query' });
 
-    let count = null;
-    if (maybeCount !== undefined) {
-      count = maybeCount;
-    }
+    const count = maybeCount ?? null;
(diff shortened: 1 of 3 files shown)


packages/compass-import-export/src/modules/export.ts
```diff
@@ -398,27 +398,17 @@ export const openExport =
...
-      count =
-        maybeCount ??
-        (!isAggregation
-          ? await fetchDocumentCount(dataService!, namespace, query)
-          : null);
-    } catch (e: unknown) {
-      dispatch(onError(e as Error));
+    if (maybeCount !== undefined) {
+      count = maybeCount;
     }
```
addaleax, line 412: At this point the code here basically does `const count = maybeCount ?? null;`, right?

addaleax (approved):

PR #4241, opened 2023-04-12
chore(compass-e2e-tests): Tests for aggregation stage wizard (COMPASS-6657)

himanshusinghs:
## Description
This PR adds e2e tests for Aggregation stage wizard. Herein, we cover the UI interaction for the wizard card component, aggregation side panel but we don't dive much into testing individual use case forms because we decided to leave that as part of testing the use case forms themselves.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/sort/sort.tsx
```diff
@@ -117,7 +117,7 @@ export const SortForm = ({
   const comboboxStyles = useMemo(() => {
     return {
       width: `calc(${String(
-        Math.max(...fields.map((label) => label.length))
+        Math.max(...fields.map((label) => label.length), 10)
```
himanshusinghs, line 120: This is to make sure that we have a minimum desirable width for the select fields when all the field names are 1/2 letters only.

packages/compass-e2e-tests/helpers/commands/open-aggregation-side-panel.ts
```diff
@@ -0,0 +1,12 @@
+import type { CompassBrowser } from '../compass-browser';
+import * as Selectors from '../selectors';
+
+export async function openAggregationSidePanel(
```
mabaasit, line 4: should we make this toggle and then we unify open and close methods?

himanshusinghs: I had them separate because it is convenient to trigger an individual state explicitly. If what you're suggesting is `toggleSidePanel` with an extra arg (`state: 'open' | 'close'`) then that works too.

packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/sort/sort.tsx
```diff
@@ -133,35 +133,42 @@ export const SortForm = ({
...
-              );
-            })}
-          </Select>
+          <div data-testid={`sort-form-${index}-direction`}>
+            {/* @ts-expect-error leafygreen unresonably expects a labelledby here */}
+            <Select
+              className={sortDirectionStyles}
+              allowDeselect={false}
+              aria-label="Select direction"
+              // usePortal={false}
```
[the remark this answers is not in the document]
himanshusinghs: Ahh no that was intended but not as a comment. Just to explain - when Options are not rendered with `usePortal: false` they ends up outside of the component tree (portals). Our test helper `browser.selectOption` that we use, was built to work with options that are rendered with `usePortal={false}`. Now I could have created another version of it (`browser.selectOption`) that works with Portal but I decided to cut short some work here and instead simply render the option without the portal. Please let me know if this is not desirable and in that case I will keep that code as it is and instead create the helper to work with portal variants as well.

packages/compass-e2e-tests/helpers/commands/open-aggregation-side-panel.ts
himanshusinghs: I just did that. Could you please check again?

mabaasit (approved):

PR #4246, opened 2023-04-13
feat(import): Import progress COMPASS-6721

lerouxb:
https://user-images.githubusercontent.com/69737/231783601-01c09ed3-e0e6-4b64-b521-de3b2508a53f.mov

### DIFF
--- a/packages/compass-import-export/src/csv/csv-utils.ts
+++ b/packages/compass-import-export/src/csv/csv-utils.ts
@@ -192,11 +192,20 @@ const parenthesis = {
 };
 
 function isEJSON(value: string) {
-  return (
+  if (
     value.length &&
     ['{', '['].includes(value[0]) &&
     value[value.length - 1] === parenthesis[value[0] as '{' | '[']
-  );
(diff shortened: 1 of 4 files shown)


packages/compass-import-export/src/csv/csv-utils.ts
```diff
@@ -192,11 +192,20 @@ const parenthesis = {
...
 function isEJSON(value: string) {
-  return (
+  if (
     value.length &&
     ['{', '['].includes(value[0]) &&
     value[value.length - 1] === parenthesis[value[0] as '{' | '[']
-  );
+  ) {
+    try {
+      JSON.parse(value);
```
lerouxb, line 201: drive-by. @Anemy originally suggested this and we discussed it before and I was on the fence about it. But I just noticed that one of my fixtures happens to be full of false positives so I tried this with it and even though it is a fairly large file it didn't noticeably slow it down, so now I'm convinced.

Anemy (approved):
🔥

PR #4257, opened 2023-04-17
chore(e2e): disable CSFLE tests on latest-alpha

lerouxb:
The CSLFE e2e tests have been failing for more than a week which makes it impossible to get a green CI run.

Re-enabling covered by https://jira.mongodb.org/browse/COMPASS-6600

### DIFF
--- a/packages/compass-e2e-tests/tests/in-use-encryption.test.ts
+++ b/packages/compass-e2e-tests/tests/in-use-encryption.test.ts
@@ -148,7 +148,7 @@ describe('CSFLE / QE', function () {
       }
 
       // temporarily disable for latest-alpha because the CSFLE fails until updated
-      if (!serverSatisfies('< 6.x.x', true)) {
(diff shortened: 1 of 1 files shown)


packages/compass-e2e-tests/tests/in-use-encryption.test.ts
```diff
@@ -146,6 +146,11 @@ describe('CSFLE / QE', function () {
       if (!serverSatisfies('>= 6.0.0', true)) {
         return this.skip();
       }
+
+      // temporarily disable for latest-alpha because the CSFLE fails until updated
+      if (!serverSatisfies('< 6.x.x', true)) {
```
addaleax, line 151: This would currently skip the tests entirely, I think you want something like

```suggestion
      if (!serverSatisfies('<= 6.x.x', true)) {
```

addaleax (approved):

PR #4262, opened 2023-04-17
feat(compass-import-export): add export modal COMPASS-6577

Anemy:
COMPASS-6577

#### Description
This adds a new export modal and store/reducer to `compass-import-export`. This has some key differences from the current export: 
- We now support projections in queries and exporting a full collection
- We've improved the user experience around selecting which fields to export.
- We've added more information banners describing how the export process works and how CSV exports are not guaranteed to keep typing.

#### Selecting fields to export
When a user has clicked `Export query results` and then uses the `Select fields to export` option in the export modal we show a view with a table of fields to export. These fields are from a sample of 50 documents. While fetching sample documents and performing schema analysis are in progress we show a loader. Once loaded, we start with all of the fields deselected. When a user chooses a field with nested fields in it we hide the nested fields. This is because the resulting projection would already include them. (cc @lerouxb thanks for the ux help here).

#### Screenshots
**New export options on the crud toolbar (previously was one export)**
Screenshot 2023-04-17 at 3 33 15 PM

**Export full collection**
Screenshot 2023-04-17 at 3 33 44 PM

**Export CSV banner**
Screenshot 2023-04-17 at 3 33 49 PM

**Export query results (no projection)**
Screenshot 2023-04-17 at 3 35 00 PM

**Export select fields**
Screenshot 2023-04-17 at 3 35 13 PM

**Export with generated projection from selecting fields**
Screenshot 2023-04-17 at 3 35 41 PM

**Export with projection in query**
Screenshot 2023-04-17 at 3 36 11 PM

**Export aggregation**
Screenshot 2023-04-17 at 3 39 33 PM

### Note to reviewers
- [ ] Before merging there are a few TODOs in test files that will be added. I figured as this contains a good amount of code changes it would be good to open it today while I work on the tests tomorrow.


packages/compass-import-export/src/components/export-select-fields.tsx
```diff
@@ -0,0 +1,373 @@
...
+const retryButtonContainerStyles = css({
+  margin: spacing[2],
+  textAlign: 'center',
+});
+
+const addNewFieldRowStyles = css({
+  marginBottom: spacing[5],
+});
+
+const placeholdersCount = 6;
```
alenakhineika, line 81: What does this placeholder do?

packages/compass-import-export/src/modules/export.ts
```diff
@@ -0,0 +1,537 @@
...
+    }
+  };
+};
+
+export const runExport = (): ExportThunkAction<
+  Promise<void>,
+  RunExportAction
+> => {
+  return async (dispatch) => {
+    // TODO(COMPASS-6580): Run export.
```
alenakhineika, line 284: So at this point export does not happen yet?

lerouxb: Aah I just came looking for this :)

packages/compass-import-export/src/modules/export.ts
```diff
@@ -0,0 +1,537 @@
...
+};
+
+// Fields can only be prefixed with one '$'. Otherwise $
+// is not allowed, so this should be a safe separator.
+const FieldSplittingSymbol = '$$$$';
+export function getIdForSchemaPath(schemaPath: SchemaPath) {
+  return schemaPath.join(FieldSplittingSymbol);
+}
+export function getLabelForFieldId(fieldId: string) {
+  return fieldId.split(FieldSplittingSymbol).join('.');
```
lerouxb, line 28: Why not just store the field paths as arrays? Then you never have to worry about special strings. If you need that as a unique string you can use as a cache key or something you can just JSON.stringify() the array.

packages/compass-import-export/src/modules/export.ts
Anemy: Not yet no, this is feature flagged. It will be done in https://jira.mongodb.org/browse/COMPASS-6580

packages/compass-import-export/src/modules/export.ts
Anemy: JSON.stringify works so much better thanks. updated.
I was opting to keep them as maps using the string key because we do a check using the keys to see if the parent of the field is checked here https://github.com/mongodb-js/compass/pull/4262/files#diff-ecb1d137cdb3159f737b523e73fb7fcc1cd0bdd8561e81b23f9378771c7d26bfR222 it would be slower if we have to iterate over an array every time. We could use a tree structure to be more efficient but I think that would introduce more complexity.

packages/compass-import-export/src/components/export-select-fields.tsx
Anemy: They are used to show the skeleton loaders on the page when the sample documents are being loaded. I updated the variable name to be `loadingPlaceholderCount` to be a bit more descriptive. 

Screenshot 2023-04-18 at 8 22 50 AM

packages/compass-import-export/src/modules/export.ts
lerouxb: I think if you want to convert both ways JSON.stringify and JSON.parse makes sense. That's effectively the definition of serialization and what JSON is made for ;) If you only wanted something to use as a key where you don't need to be able to work backwards then a hash is probably more analogous. 

But I think JSON.stringify is probably right here.

packages/compass-import-export/src/components/export-code-view.tsx
```diff
@@ -0,0 +1,74 @@
...
+  selectedFieldOption,
+}: ExportCodeViewProps) {
+  const code = useMemo(() => {
+    if (selectedFieldOption === 'select-fields') {
+      return newGetQueryAsShellJSString({
+        query: {
+          ...(query ?? {
+            filter: {},
+          }),
+          projection: createProjectionFromSchemaFields(
```
Anemy, line 35: We should set `_id: false` when it's not selected to be sure it's not automatically included.

Anemy: done

packages/compass-import-export/src/components/export-field-options.tsx
```diff
@@ -0,0 +1,80 @@
...
+          checked={fieldsToExportOption === 'select-fields'}
+        >
+          Select fields in table
+        </RadioBox>
+      </RadioBoxGroup>
+      {showProjectInfoMessage && (
+        <Banner
+          className={messageBannerStyles}
+          dismissible
+          onClose={() => setShowProjectInfoMessage(false)}
```
lerouxb, line 70: Why do we bother making some of these dismissable? 😄

Anemy: Those were in the designs and I think I the pattern is information about what you can do vs information on what's going to happen. Things you can do with some additional info are dismissible, while banners with information about what will happen are not dismissible.

packages/compass-import-export/src/modules/export.ts
```diff
@@ -0,0 +1,531 @@
...
+          selected: false,
+        };
+      }
+
+      dispatch({
+        type: ExportActionTypes.FetchFieldsToExportSuccess,
+        fieldsToExport: fields,
+      });
+    } catch (err: any) {
+      if (fieldsToExportAbortController.signal.aborted) {
```
lerouxb, line 257: How would you get here given that aborted wouldn't throw and you'd land in the gatherFieldsResult.aborted check above?

I also don't think that aborted is an error here. Right now you can't really abort it, so I wouldn't worry about it and in future aborting would just mean "skip", ie. it still succeeds and the fact that it aborted is really just metadata. ie. you could use it to decide if you should display a banner.

I'd also only try {} catch {} around the `await gatherFieldsFromQuery()` line. Errors outside of that are programmer errors and we want those to propagate, not "runtime" errors like network or database ones.

lerouxb (approved):

packages/compass-import-export/src/components/export-field-options.tsx
lerouxb: I just see it as added complexity for zero gain. It doesn't even remember that you closed it so it will just reappear next time. But it is here now, so might as well keep it.

packages/compass-import-export/src/modules/export.ts
Anemy: Makes sense, thanks for catching. Removed that check in the catch and moved the other parts out of the try catch. 👍

PR #4268, opened 2023-04-18
feat(stage-wizard): dynamic stage wizard fields

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/utils/get-schema.ts
```diff
@@ -0,0 +1,42 @@
...
+      if (isBson) {
+        keys.push(key);
+      } else {
+        const nestedKeys = Array.isArray(value)
+          ? getArrayKeys(value)
+          : getObjectKeys(value);
+        nestedKeys.forEach((nestedKey) => {
+          keys.push(`${key}.${nestedKey}`);
+        });
+      }
```
himanshusinghs, line 33:
```suggestion
      if (!isBson) {
        const nestedKeys = Array.isArray(value)
          ? getArrayKeys(value)
          : getObjectKeys(value);
        nestedKeys.forEach((nestedKey) => {
          keys.push(`${key}.${nestedKey}`);
        });
      }
```
Since you already push the same key on Line - 19, we probably can avoid inserting duplicate keys this way? then we won't have to do the `filter` operation in the `getArrayKeys` method.

packages/compass-aggregations/src/components/stage-wizard/index.tsx
```diff
@@ -145,15 +160,37 @@ type WizardOwnProps = {
...
+        stageEditor: { stages },
+      },
+    } = state;
+
+    const wizard = stages[ownProps.index] as Wizard;
+
+    const previousStage = stages
+      .slice(0, ownProps.index)
+      .reverse()
+      .find((x) => x.type === 'stage' && !x.disabled) as StoreStage | undefined;
```
himanshusinghs, line 176:
```suggestion
      .find((x): x is StoreStage => x.type === 'stage' && !x.disabled);
```
This will take care of both.

himanshusinghs (commented):
Yayy! We have dynamic fields now 😄 🚀 
Left two small suggestions for your review.

packages/compass-aggregations/src/utils/get-schema.ts
```diff
@@ -0,0 +1,40 @@
+import type { Document } from 'mongodb';
+
+const getArrayKeys = (records: Document[]) => {
+  return records
+    .map((item) => getObjectKeys(item))
+    .flat()
+    .filter((x, i, a) => a.indexOf(x) === i)
```
himanshusinghs, line 7:
```suggestion
```
Now we don't need this I guess, right?

mabaasit: we do need it here. here we do flat on the list of keys and then it can have duplicates. so we are filtering them here.

himanshusinghs: ahh that's right

himanshusinghs (approved):

PR #4273, opened 2023-04-19
feat(import): Surface type warnings COMPASS-6543

lerouxb:
During CSV import at the field type selection step, if the user selects a field type that we know is incompatible with the detected fields, it now adds a warning similar to the one for Mixed or Number.

Screenshot 2023-04-20 at 12 08 30
Screenshot 2023-04-20 at 12 04 35
Screenshot 2023-04-20 at 12 04 09

While testing this I also realised that we don't really support the deprecated type `Undefined` and we can't actually turn things into that type, so I removed it and changed the types around that a bit to not include it in detectable and parsable types.


packages/compass-import-export/src/csv/csv-utils.spec.ts
```diff
@@ -877,3 +886,260 @@ describe('formatCSVHeaderName', function () {
...
+  selectedType: CSVParsableFieldType,
+  types: (CSVParsableFieldType | 'undefined')[],
+  expected: (CSVParsableFieldType | 'undefined')[]
+) {
+  types.sort();
+  expected.sort();
+  try {
+    expect(types, selectedType).to.deep.equal(expected);
+  } catch (err) {
+    console.dir(types);
```
lerouxb, line 900: This console.dir() leaves you with an array that's easy to copy/paste into code when we change the types. The diff shown by the error is a bit of a pain to work with alone in that context.

packages/compass-import-export/src/components/import-preview.tsx
```diff
@@ -200,6 +255,58 @@ function MixedWarning({
...
+      delay={500}
+      trigger={({ children, ...props }) => (
+        <div {...props}>
+          {children}
+          <WarningIcon />
+        </div>
+      )}
+    >
+      <>
+        <Body as="p">This field has these detected types:</Body>
```
Anemy, line 289: Isn't `Body` by default a `p` element? We might not need we `as` property here (and below).

Anemy (approved):
lgtm! Left one small non blocker suggestion.

packages/compass-import-export/src/components/import-preview.tsx
lerouxb: I removed it now. I had some issue where it added extra css you can't remove at some point, but can't recreate it now so was probably just confused.

PR #4278, opened 2023-04-20
fix(import): If listCSVFields() or analyzeCSVFields() fails it will display the error in the modal COMPASS-6737

lerouxb:
Without this the error would be invisible (only in the console) and analyze stuck at (probably) 0%.
Screenshot 2023-04-20 at 11 52 09


packages/compass-e2e-tests/helpers/insert-data.ts
```diff
@@ -66,6 +66,7 @@ export async function createDummyCollections(): Promise<void> {
   promises.push(createBlankCollection(db, 'extended-json-file'));
   promises.push(createBlankCollection(db, 'csv-file'));
   promises.push(createBlankCollection(db, 'bom-csv-file'));
+  promises.push(createBlankCollection(db, 'latin1'));
```
Anemy, line 69: Maybe not for this pr, but we're running these on every `beforeEach`. That will recreate these collections even when they aren't being used. I'm thinking we could refactor it to either run in `before` if each collection is used individually, or we could have it so a scoped `beforeEach` calls certain collections to be made so we don't have to do all of them. Would that make sense for you?

lerouxb: Makes sense. We export some more specific helpers to create some collections already. I was just trying to avoid having to also make a another connection to the db inside the tests and also making sure that collections happen in known databases that we know to drop before tests run so we can guarantee that things get cleaned up.

btw, I don't think it takes much time to create an empty collection once you already have a connection open. But I don't mind changing this code either.

packages/compass-e2e-tests/tests/collection-import.test.ts
```diff
@@ -1001,7 +1032,12 @@ describe('Collection import', function () {
       const toastElement = await browser.$(Selectors.ImportToast);
       // Check it displays that the import was aborted.
       const toastText = await toastElement.getText();
-      expect(toastText).to.include('Import aborted.');
+      try {
+        expect(toastText).to.include('Import aborted.');
+      } catch (err) {
+        console.log(toastText);
```
lerouxb, line 1038: The expect failure is not showing the full error message and there's a flake here. I'm trying to see what's going on.

packages/compass-e2e-tests/tests/collection-import.test.ts
```diff
@@ -990,6 +990,32 @@ describe('Collection import', function () {
...
+
+      // Select the file.
+      await browser.selectFile(Selectors.ImportFileInput, csvPath);
+
+      // Wait for the modal to appear.
+      const importModal = await browser.$(Selectors.ImportModal);
+      await importModal.waitForDisplayed();
+
+      // Confirm import.
+      await browser.clickVisible(Selectors.ImportConfirm);
```
lerouxb, line 1018: The diff doesn't make it obvious, but much of this test's logic was in the beforeEach(). I don't particularly like that style because it is test code and if it fails it is the beforeEach() that fails, not  the test. Also in this particular case it meant that there was no screenshot showing the failure (once ever the Import confirm button couldn't be clicked because it was disabled) because the screenshot is in an afterEach and the test never ran, so no screenshot.

As a general rule I try not to use the pattern of executing the test in hooks and only doing assertions in the test. Makes all the test code that can fail contained inside the test while only leaving genuine common code that has nothing to do with this specific test case in hooks. Easier to reason about as well and to isolate the effects from other tests in the file.

lerouxb:
There are some flakes downloading node.js. I'm merging because I'd really like to get a beta out today.

PR #4279, opened 2023-04-20
feat(stage-wizard): lookup use-case COMPASS-6659

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/lookup/lookup.tsx
+++ b/packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/lookup/lookup.tsx
@@ -1,0 +1,180 @@
+import {
+  Body,
+  spacing,
+  css,
+  TextInput,
+  Combobox,
+  ComboboxWithCustomOption,
+  ComboboxOption,
+} from '@mongodb-js/compass-components';
+import React, { useEffect, useMemo, useState } from 'react';
+import { connect } from 'react-redux';
+import type { RootState } from '../../../../modules';
+import { fetchCollectionFields } from '../../../../modules/collections-fields';
+
+type LookupFormState = {
+  from: string;
+  localField: string;
+  foreignField: string;
+  as: string;
+};
+
+const containerStyles = css({
+  display: 'flex',
+  flexDirection: 'row',
+  alignItems: 'center',
+  gap: spacing[2],
+  flexWrap: 'wrap',
+});
+
+const formElementWithText = css({
+  display: 'flex',
+  gap: spacing[2],
+  alignItems: 'center',
+});
+
+const comboboxStyles = (fields: string[]) => {
+  return {
+    width: `calc(${String(
+      Math.max(...fields.map((label) => label.length), 10)
+    )}ch)`,
+  };
+};
+
+export const LookupForm = ({
+  fields,
+  collectionsFields,
+  onSelectCollection,
+  onChange,
+}: {
+  fields: string[];
+  collectionsFields: Record<string, { isLoading: boolean; fields: string[] }>;
+  onSelectCollection: (collection: string) => void;
+  onChange: (value: string, error: Error | null) => void;
+}) => {
+  const [formData, setFormData] = useState<LookupFormState>({
+    as: '',
+    from: '',
+    foreignField: '',
+    localField: '',
+  });
+
+  useEffect(() => {
+    if (formData.from) {
+      console.log(formData.from);
+      onSelectCollection(formData.from);
+    }
+  }, [formData.from, onSelectCollection]);
+
+  useEffect(() => {
+    onChange(JSON.stringify(formData), null);
+  }, [formData, onChange]);
+
+  const onSelectOption = (
+    option: keyof LookupFormState,
+    value: string | null
+  ) => {
+    if (!value) {
+      return;
+    }
+
+    const newData: LookupFormState = { ...formData, [option]: value };
+
+    // We set the "as" form field if:
+    // 1. The value was not set initially.
+    // 2. The value was set initially by us programmatically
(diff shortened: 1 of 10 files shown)


packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/lookup/lookup.tsx
```diff
@@ -0,0 +1,191 @@
...
+};
+
+const containerStyles = css({
+  display: 'flex',
+  flexDirection: 'column',
+  gap: spacing[2],
+});
+
+const titleStyles = css({
+  minWidth: `${'Join documents in'.length}ch`,
```
Anemy, line 30: I'm thinking if we to store this `Join documents in` string as a variable and reference it in both places we can prevent styling issues down the line if the text changes and someone misses this style.

packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/lookup/lookup.tsx
```diff
@@ -0,0 +1,191 @@
...
+    from: '',
+    foreignField: '',
+    localField: '',
+  });
+
+  useEffect(() => {
+    if (formData.from) {
+      onSelectCollection(formData.from);
+    }
+  }, [formData.from, onSelectCollection]);
```
Anemy, line 66: This `useEffect` and the one below it look to me to be a bit circuitous. Is there a reason we don't call the `onSelectCollection` directly when it's changed in `onSelectOption`?

packages/compass-aggregations/src/modules/collections-fields.ts
```diff
@@ -0,0 +1,204 @@
...
+          ? await dataService.find(
+              namespaceToQuery,
+              {},
+              {
+                sort: { $natural: -1 },
+                limit: 1,
+                maxTimeMS: FETCH_SCHEMA_MAX_TIME_MS,
+              }
+            )
+          : await dataService.sample(
```
Anemy, line 174: Something I'm not sure on is `$sample` performance with time series collections and views. We want to avoid a collection scan if possible (collection scans happen when running count operations on those). Does it perform well?

packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/lookup/lookup.tsx
mabaasit: nice catch. added in d71dfbc0e2de21067a67d1526925bbd6896dafa3

packages/compass-aggregations/src/modules/collections-fields.ts
mabaasit: We did add timeout there, but it its definitely not the best way to fetch a single document to figure out the schema. 

With collections, we are using find and it just works fine.

packages/compass-aggregations/src/modules/collections-fields.spec.ts
```diff
@@ -0,0 +1,284 @@
...
+            },
+          ]) as any
+        );
+
+        const spy = sinon.spy(
+          (ns: string, stage: Document, options: Document) => {
+            expect(ns).to.equal('test.listings');
+            expect(stage).to.deep.equal({ size: 1 });
+            expect(options).to.deep.equal({ maxTimeMS: 5000 });
+            return Promise.resolve([{ _id: 12, name: 'test' }]);
```
himanshusinghs, line 220: Question: Is this a common pattern? For this I probably would have done my expects on the arguments of the calls to the spy function?
Nothing to do here as I am just looking for new patterns to look into :D

packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/lookup/lookup.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+      // Call onChange with the new value and an error if any of the fields is empty
+      const anyEmptyValue = Object.values(newData).some((x) => !x);
+      onChange(
+        JSON.stringify(newData),
+        anyEmptyValue ? new Error('Enter all the fields') : null
+      );
+
+      // Fetch the fields of the selected collection if it was changed
+      if (option === 'from' && newData.from) {
```
himanshusinghs, line 90: Would this function get triggered if the user selects the same `from` option (the one that is already selected)? If yes, then we probably should apply the suggestion:

```suggestion
      if (option === 'from' && newData.from !== formData.from) {
```

himanshusinghs (approved):
Looks great 👍

packages/compass-aggregations/src/modules/collections-fields.spec.ts
mabaasit: Good point. Umm, well I added these args in the spy as ts was complaining when I tried to access the arguments of spy `spy.getCalls()[0].args[0]`. And when I added the args, then unused args error. So I ended I doing this as I am expecting only one call to be made.

Normally I'd do this outside of the spy.

packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/lookup/lookup.tsx
mabaasit: added

packages/compass-aggregations/src/modules/collections-fields.ts
Anemy: Ah yup all good.

PR #4288, opened 2023-04-24
feat(compass-import-export): show an add more fields button when more fields to show COMPASS-6762

Anemy:
COMPASS-6762

Sometimes folks can have polymorphic data with generated field names. This would cause us to render possibly thousands of fields and would brick the ui. This PR adds a `Show 100 more fields` button to the export select fields ui when there are a lot of fields and more to show.

https://user-images.githubusercontent.com/1791149/234038348-defb85f0-671e-4380-903b-c47427e1f723.mp4


lerouxb (approved):

Anemy:
Merging with known failing e2e flake (json validation)

PR #4291, opened 2023-04-24
chore(compass-e2e-tests): fix cancel import test flakyness

Anemy:


packages/compass-e2e-tests/tests/collection-import.test.ts
```diff
@@ -979,8 +979,7 @@ describe('Collection import', function () {
       );
       const documentsText = await messageElement.getText();
       expect(documentsText).to.not.equal('1 – 20 of 16116');
-      const result = await getFirstListDocument(browser);
-      expect(result._id).to.exist;
```
Anemy, line 983: This wasn't what was flaking, but it could if the test aborts instantly, so we should remove it.

packages/compass-import-export/src/import/import-json.ts
```diff
@@ -75,8 +75,9 @@ export async function importJSON({
         const doc = EJSON.deserialize(chunk.value);
         if (process.env.COMPASS_E2E_TEST_IMPORT_ABORT_TIMEOUT === 'true') {
           setTimeout(() => {
+            // Give the test more than enough time to click the abort before we continue.
             callback(null, doc);
-          }, 2000);
+          }, 10000);
```
Anemy, line 80: I think the flakyness was just caused by 2 seconds not being enough time for the test to click the abort button.

PR #4294, opened 2023-04-24
fix(export): fix exported CSV field order for large sparse arrays COMPASS-6761

lerouxb:
The most obvious way this would break is if you have docs like this:

```
{ foo: ['a']}
{ bar: 1}
{ foo: ['b, 'c'] }
```

The CSV fields should be

```
foo[0]
foo[1]
bar
```

in that order and not `foo[0]`, `bar`, `foo[1]`.

This is because there would be three columns in the CSV file and then if you re-import the file also in the import preview table, but only two of them have field type selects: one for `foo` and one for `bar`. `foo` is a two-element array where we detect and the user sets the type once for the array. If the fields are out of order, then things get confusing and broken pretty quickly.

This should give you an idea of what happens if you export the above to CSV and then import it again:

Screenshot 2023-04-24 at 21 37 44

And here's the exported file:

[test.array-order-test.csv](https://github.com/mongodb-js/compass/files/11314512/test.array-order-test.csv)


lerouxb:
Hmm. We could probably join together all the array value fields and that will result in a lot less horizontal scrolling. I'll file a ticket.

addaleax (approved):

PR #4298, opened 2023-04-25
fix(export): support collation in export COMPASS-6728

lerouxb:
We now also print the collation in the export code view:
Screenshot 2023-04-25 at 11 51 44

And it is passed along with the query to dataService.findCursor().


packages/compass-import-export/src/export/export-csv.ts
```diff
@@ -260,7 +260,7 @@ export async function exportCSVFromAggregation({
   dataService: DataService;
   aggregation: ExportAggregation;
 }) {
-  debug('exportJSONFromAggregation()', { ns: toNS(ns) });
```
lerouxb, line 263: oof.

PR #4303, opened 2023-04-26
chore(import): Make the whole header the tooltip target COMPASS-6631

lerouxb:
tooltip-target


Anemy (approved):
One small code style comment, not a blocker

PR #4309, opened 2023-04-27
feat(import): Combine array fields into one in the import CSV preview COMPASS-6766

lerouxb:
This significantly speeds up the preview when importing documents containing large arrays and it cuts down on most of the horizontal scrolling.

In this example the amenities field would have been 50 or so different columns in the table:

Screenshot 2023-04-28 at 11 21 18

It isn't obvious in this screenshot, bit every row doesn't have the same number of amenities values. The CSV file has enough columns for the maximum number, though. So most cells in the file are empty.

In this example reviews is an array of objects and we're specifying reviews[]._id's type. It joins together the 350-ish review id columns into one and the user specify's the field type once. Same thing for ever other field inside the review:

Screenshot 2023-04-28 at 11 21 44

Note that every row doesn't have the same amount of items for that column, because not every entry has the same number of reviews. But the CSV file has columns for the maximum number. So again most cells in the file are empty.

There's also a new e2e test fixture that demonstrates the problem of having arrays that have hundreds of items, each one an object. In that case it results in a very dramatic improvement.


packages/compass-import-export/src/components/import-preview.tsx
```diff
@@ -322,10 +318,6 @@ function FieldTypeHeading({
   onFieldCheckedChanged: (fieldPath: string, checked: boolean) => void;
   setFieldType: (fieldPath: string, fieldType: string) => void;
 }) {
-  if (field.type === 'placeholder') {
-    return null;
-  }
-
   const children = (
```
Anemy, line 321: Should we cap the length of strings shown? Is it a big performance hit if there are some huge arrays or strings?

lerouxb: The arrays become strings and the table already turns them into ellipses. Seems fine? We could always already have giant description fields in there.

lerouxb: I'm capping the string length now. Will push the changes..

Anemy (approved):

PR #4310, opened 2023-04-27
chore(compass-import-export, compass-e2e-tests): add export e2e tests COMPASS-6731

Anemy:
First part of COMPASS-6731

This PR adds e2e tests for the new export flow. It tests what we already had in e2e tests for export, but this time with the new export flow that is currently feature flagged. In the next pr I'll add tests for the things listed in the `TODO` at the bottom of the test file, we'll be increasing the test coverage for different paths here. Doing separate prs to make it easier to review, and so some tests could be reviewed while others are being worked on.

One or two drive by code cleanups included, can move them into separate pr if desired. Also one fix for export (feature flagged) that is included here, I'll leave a comment on that one. (Thanks to e2e tests for catching.)

### DIFF
--- a/packages/compass-import-export/src/modules/export.ts
+++ b/packages/compass-import-export/src/modules/export.ts
@@ -65,7 +65,7 @@ type ExportOptions = {
   aggregation?: ExportAggregation;
   fieldsToExport: FieldsToExport;
 
-  selectedFieldOption: undefined | FieldsToExportOption;
+  selectedFieldOption: FieldsToExportOption;
 };
 
 export type ExportStatus =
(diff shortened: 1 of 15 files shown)


packages/compass/src/index.d.ts
```diff
@@ -32,9 +32,6 @@ declare module 'process' {
         HADRON_METRICS_SEGMENT_API_KEY?: string;
         HADRON_METRICS_SEGMENT_HOST?: string;
         HADRON_AUTO_UPDATE_ENDPOINT: string;
-
-        // Legacy feature flags. Add new feature flags to the preferences model directly!
-        COMPASS_LG_DARKMODE?: 'true' | 'false';
```
Anemy, line 37: Doesn't look like this is being used anywhere anymore.

packages/compass-import-export/src/modules/export.ts
```diff
@@ -362,16 +362,25 @@ export const runExport = ({
       dataService: { dataService },
     } = getState();
 
+    let fieldCount = 0;
+
     const query =
-      selectedFieldOption === 'select-fields'
+      exportFullCollection || aggregation
```
Anemy, line 368: Here's the fix mentioned in the description. We were actually using the query when exporting a full collection. Thanks to adding tests we catch these things :)

lerouxb (approved):

PR #4332, opened 2023-05-04
feat(compass-import-export): add export json format variants COMPASS-6611

Anemy:
COMPASS-6611

This PR adds json format options to the export flow. Bumps `@leafygreen-ui/radio-group` for a darkmode fix and the `description` prop.
More information about extended json: https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/

| | |
| - | - |
| Screenshot 2023-05-04 at 12 48 53 PM | Screenshot 2023-05-04 at 12 49 11 PM |


lerouxb:
It would be nice if, for the examples, you could choose a number > 2^53 and then for relaxed ejson show the loss of precision in the number. ie:

```
> Math.pow(2, 53)
9007199254740992
> Math.pow(2, 53) + 1
9007199254740992
> Math.pow(2, 53) + 2
9007199254740994
```

Or just show how `new Long("123456789123456784")` becomes 123456789123456780.

packages/compass-import-export/src/components/export-json-format-options.tsx
```diff
@@ -0,0 +1,70 @@
...
+        className={radioGroupStyles}
+        data-testid="export-json-format-options"
+        onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+          setJSONFormatVariant(event.target.value as ExportJSONFormat)
+        }
+      >
+        <Radio
+          value="default"
+          checked={jsonFormat === 'default'}
+          description='Example:  { "fortyTwo": 42, "oneHalf": 0.5, "bignumber": { "$numberLong": "5000000000" } }'
```
lerouxb, line 39: I think it would be a nice subtle hint (and bit if detail) if you could use a number > 2^^32. Then for default and canonical it would be correct and relaxed it would be wrong. ie. if you use 123456789123456784, then relaxed would turn that into 123456789123456780.

packages/compass-import-export/src/components/export-json-format-options.tsx
```diff
@@ -0,0 +1,70 @@
...
+        >
+          Default Extended JSON
+        </Radio>
+        <Radio
+          value="relaxed"
```
lerouxb, line 44: Would be nice if we had a warning next to relaxed that large integvers (> 2^^53) will change if you use relaxed.

lerouxb (approved):

packages/compass-import-export/src/components/export-json-format-options.tsx
Anemy: Good call, added it to the description and now we show a warning banner when relaxed is chosen.

PR #4336, opened 2023-05-05
feat(compass-import-export): remove export feature flag, use new export, remove old code COMPASS-6582

Anemy:
COMPASS-6582

Removes the export feature flag, so we now always use the new export flow. Removes old code. Updates the export aggregation end to end test.


lerouxb (approved):
NICE! So much stuff that got removed. 🎉

PR #4340, opened 2023-05-08
fix: allow updates on a sharded collection COMPASS-6058

alenakhineika:
## Description

The `findAndModify` command is not allowed on sharded collections and throws the `"Query for sharded findAndModify must contain the shard key"` error. We check if the command returned the `ShardKeyNotFound` error name, we try using updateOne/replaceOne instead of findOneAndUpdate/findOneAndReplace to perform the update operation.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1584,44 +1570,104 @@ type ErrorOrResult =
 export async function findAndModifyWithFLEFallback(
   ds: DataService,
   ns: string,
-  doFindAndModify: (
-    ds: DataService,
-    ns: string,
-    opts: { returnDocument: 'before' | 'after'; promoteValues: false }
-  ) => Promise<ErrorOrResult>
+  query: Partial<QueryState>,
+  object: unknown,
```
addaleax, line 1574: These types don't seem quite right. `query` is not related to `queryState`, and `object` is typed as `unknown` here but we always cast it using `as Document` below

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1584,44 +1570,104 @@ type ErrorOrResult =
...
+          options
+        );
       } else {
-        [d] = docs;
+        document = await ds.findOneAndReplace(
+          ns,
+          query,
+          object as Document,
+          options
+        );
```
addaleax, line 1649: There's quite a bit of duplication now, this code block shows up 4 times in the same function essentially... I'd either store the arguments as an array and then simplify these calls to `await ds.findOneAndReplace(...args);`, or dynamically select the method (`const method = modificationType === 'update' ? 'findOneAndUpdate' : 'findOneAndReplace'; await ds[method](ns, query, object, options);`)

packages/data-service/src/data-service.ts
```diff
@@ -1488,6 +1520,28 @@ class DataServiceImpl extends WithLogContext implements DataService {
     return (await coll.findOneAndUpdate(filter, update, options)).value;
   }
 
+  @op(mongoLogId(1_001_000_183))
```
addaleax, line 1523: Did the script suggest this number? It should always be suggesting one that's higher than all other used numbers

alenakhineika: It was, I think the problem is that I merged the main a couple of times and for this actually, it was fine, but for replaceOne 1_001_000_184 was errored as a duplicate and I keep changing it.

addaleax (approved):

PR #4358, opened 2023-05-09
fix(import): fix guessFileType()  when JSON fails and CSV lines are huge

lerouxb:
This fixes the bug Julia found.

If you select a CSV file that has super long lines, then guessFileType() can stall because the JSON detection fails almost immediately, but then the CSV detection stops processing when a buffer fills up.

The fix is to just destroy the split JSON/CSV-specific streams as they resolve rather than to depend on the base stream they are split from to be destroyed first. That way when one "branch" resolves the other one can continue processing.

Also added it as a test case. (Just the header and first row of the file.)


packages/compass-import-export/test/csv/many-columns.analyzed.json
```diff
@@ -0,0 +1,1312 @@
+{
```
Anemy, line 1: Is it worth having these text fixtures zipped like we do in `compass-e2e-tests`? https://github.com/mongodb-js/compass/blob/main/packages/compass-e2e-tests/package.json#L22
Looks like the `test` folder is getting close to 1mb already, might be more now, not sure what a threshold might be.

lerouxb: I think its fine and we can file a ticket and deal with it if it ever becomes a problem. These files don't change much so I don't see it affecting day to day git operations.

Anemy (approved):

PR #4361, opened 2023-05-10
fix: wrap schema analysis telemetry computation in try/catch HELP-45308

addaleax:
See HELP-45308 / COMPASS-6809.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-schema/src/stores/store.js
+++ b/packages/compass-schema/src/stores/store.js
@@ -290,15 +290,31 @@ const configureStore = (options = {}) => {
     },
 
     _trackSchemaAnalyzed(analysisTime) {
-      const { schema } = this.state;
-      const trackEvent = {
-        with_filter: Object.entries(this.query.filter).length > 0,
-        schema_width: schema ? schema.fields.length : 0,
-        schema_depth: schema ? this.calculateSchemaDepth(schema) : 0,
-        geo_data: schema ? this.schemaContainsGeoData(schema) : false,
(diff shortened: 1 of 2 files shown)


packages/compass-schema/src/stores/store.js
```diff
@@ -290,15 +290,31 @@ const configureStore = (options = {}) => {
...
+          analysis_time_ms: analysisTime,
+        };
+        track('Schema Analyzed', trackEvent);
+      } catch (err) {
+        // TODO(COMPASS-6809): We know that calculateSchemaDepth() can fail
+        // with a recursion/max call stack size exceeded error, but don't quite
+        // know why. (MongoDB servers reject BSON documents with more than 200
+        // levels of nesting, so deeply nested BSON documents do not seem like
+        // a likely cause.)
+        log.error(
```
Anemy, line 309: How do you feel about still tracking a telemetry event, but with limited properties? I'd like to avoid having a chance to lose a lot of these events from a future regression in one of the schema calculation methods. 
We do some bit similar when connection telemetry fails:
https://github.com/mongodb-js/compass/blob/adff1d022d2aa490d9173a42ef76a74db9e36dd2/packages/compass-logging/src/logger.ts#L101

Anemy (approved):
Nice improvement on the logging!

PR #4373, opened 2023-05-12
refactor: fallback to modifyOneMethod with query COMPASS-6058

alenakhineika:
## Description

- Fallback to modifyOneMethod for both `ShardKeyNotFound` and `FLE returnDocument: "after"` errors.
- Use the whole query in modifyOneMethod.
- Remove extra find before modifyOneMethod.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-crud/src/stores/crud-store.ts
+++ b/packages/compass-crud/src/stores/crud-store.ts
@@ -39,6 +39,7 @@ import type AppRegistry from 'hadron-app-registry';
 import { BaseRefluxStore } from './base-reflux-store';
 export type BSONObject = TypeCastMap['Object'];
 export type BSONArray = TypeCastMap['Array'];
+import type { ObjectId } from 'bson';
 type Mutable<T> = { -readonly [P in keyof T]: T[P] };
 
 export type CrudActions = {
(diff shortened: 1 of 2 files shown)


packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1575,67 +1576,45 @@ export async function findAndModifyWithFLEFallback(
...
 
-    if (document) {
-      try {
-        await ds[modifyOneMethod](ns, { _id: document._id }, object);
-      } catch (e) {
-        return [e, undefined] as ErrorOrResult;
-      }
-    }
-  } else if (+(error?.code ?? 0) === 63714_02) {
-    // 6371402 is "'findAndModify with encryption only supports new: false'"
```
addaleax, line 1616: I think this comment is missing now?

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1575,67 +1576,45 @@ export async function findAndModifyWithFLEFallback(
...
-      returnDocument: 'before',
-      promoteValues: false,
-    };
     try {
-      document = await ds[findOneAndModifyMethod](ns, query, object, options);
-    } catch (e) {
-      // Return the current findOneAndModify error here
-      // since we already know the exact error from the previous attempt
-      // and want to know what went wrong with the second attempt with the fallback options,
-      // e.g. return the `Found indexed encrypted fields but could not find __safeContent__` error.
```
addaleax, line 1627: Ditto here

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1575,67 +1576,45 @@ export async function findAndModifyWithFLEFallback(
...
-      return [e, undefined] as ErrorOrResult;
-    }
-  }
-
-  try {
-    if (document) {
-      const docs = await ds.find(ns, { _id: document._id }, options);
+      const docs = await ds.find(
+        ns,
+        { _id: query._id as ObjectId },
```
addaleax, line 1611: Did something require this type change? `_id` is not necessarily an `ObjectId` and the types should account for that

addaleax (approved):

packages/compass-crud/src/stores/crud-store.ts
alenakhineika: TS complins then:

Type 'unknown' is not assignable to type 'Condition<ObjectId> | undefined'.ts(2322)
mongodb.d.ts(6272, 5): The expected type comes from property '_id' which is declared here on type 'Filter<Document>'

addaleax: Mh yeah ... I think `as any` would be more appropriate here because this could really be any BSON value. Not blocking though

addaleax (approved):

PR #4375, opened 2023-05-12
fix(import): Detect line breaks, pass it on to papaparse COMPASS-6819

lerouxb:
I'm not exactly sure what's going on in papaparse's [guessLineEndings()](https://github.com/mholt/PapaParse/blob/b5365550968b46c5ec4a9fa37ea26365c92c9a15/papaparse.js#L1358-L1382), but @graboskyc ran into a case where a CSV file with windows style line endings was dropping the last field in the CSV file when imported.

I investigated and realised the \r from the \r\n line ending was still in the header. Upon further investigation I eventually realised that papaparse detected unix style line endings. I added a bunch of unit tests that would replace our test fixtures' unix style line endings with windows style ones so that we can test both cases everywhere. I also made a simplified version of his file as a new test fixture. I quickly realised it almost always detected the windows ones as unix ones, but I can't figure out why not all files trigger it.

guessLineEndings() in papaparse is a bit complicated, but also seems to have been written before they added streams support and my hunch is that streams has something to do with it getting confused.

The reason the import ended up dropping the field is because this interacted with [another papaparse bug](https://github.com/mholt/PapaParse/issues/974) that we're working around. For windows style line endings it includes the \r in the header and we're working around it by removing it. So _our_ header row had the \r removed, but the actual header and data rows not, so then when it built objects out of the rows (the chunk var inside the transform stream) the final field's property name still had the \r which meant it couldn't be found when we built the docs and that's why it went with undefined.

The fix I came up with is to just detect the line endings ourselves when we detect the file type and CSV delimiter and then pass that to papaparse so that it will skip its newline autodetection.

### DIFF
--- a/packages/compass-import-export/src/import/import-csv.spec.ts
+++ b/packages/compass-import-export/src/import/import-csv.spec.ts
@@ -31,7 +31,7 @@ const { expect } = chai;
 chai.use(sinonChai);
 chai.use(chaiAsPromised);
 
-describe.only('importCSV', function () {
+describe('importCSV', function () {
   let dataService: DataService;
 
   beforeEach(async function () {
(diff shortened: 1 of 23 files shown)


packages/compass-import-export/src/import/import-csv.ts
```diff
@@ -107,6 +110,14 @@ export async function importCSV({
         });
       }
 
+      // Part of the fix for papaparse's header fields quirk mentioned above.
+      // The final field's property name inside chunk still contains the
+      // trailing \r and so does each row's value.
+      //if (fixupFrom !== fixupTo) {
```
lerouxb, line 116: With this commented I'm expecting more tests to fail, yet most pass? 🤯

packages/compass-import-export/src/import/import-csv.spec.ts
```diff
@@ -57,114 +58,131 @@ describe('importCSV', function () {
...
-        input: fs.createReadStream(filepath),
-        delimiter: csvDelimiter,
-        ignoreEmptyStrings: true,
-      });
-
-      const abortController = new AbortController();
-      const progressCallback = sinon.spy();
-
-      const ns = 'db.col';
+    for (const lineEnding of ['unix', 'windows']) {
```
lerouxb, line 61: Our fixtures all have unix line endings. Here I loop through them twice and the second time we also replace all the line endings with windows ones and make sure it detects the correct line ending and that it inserts the exact same docs in both cases.

packages/compass-import-export/src/import/import-csv.spec.ts
```diff
@@ -173,104 +191,121 @@ describe('importCSV', function () {
...
-      });
-      assert(typeResult.type === 'csv');
-      const csvDelimiter = typeResult.csvDelimiter;
-
-      const analyzeResult = await analyzeCSVFields({
-        input: fs.createReadStream(filepath),
-        delimiter: csvDelimiter,
-        ignoreEmptyStrings: true,
-      });
+    for (const lineEnding of ['unix', 'windows']) {
```
lerouxb, line 194: And the same thing for these per-type tests.

gribnoysup (approved):

PR #4386, opened 2023-05-16
feat(compass-preferences-model, connection-form, compass-settings): toggle state of "Edit connection string" in connection form is now judged also based on the new global preference

himanshusinghs:
## Description
In this PR, we added a new global preference setting called `protectConnectionStringsForNewConnections` ~`connectionStringEditableByDefault`~. This preference is used by connection form to determine wether the "Edit connection string" toggle will be rendered by default in an on or off state for new connections. The default toggle state for existing connections remain unaffected.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-preferences-model/src/preferences.ts
```diff
@@ -558,6 +559,22 @@ const modelPreferencesProps: Required<{
...
+  protectConnectionStringsForNewConnections: {
+    type: 'boolean',
+    required: false,
+    default: false,
+    ui: true,
+    cli: true,
+    global: true,
+    description: {
+      short:
+        'Sets the default state of "Edit connection string" toggle for new connections',
```
addaleax, line 575: It sets the inverse of it, right? 

```suggestion
        'If true, "Edit connection string" is disabled for new connections by default',
```

addaleax (approved):

packages/compass-preferences-model/src/preferences.ts
himanshusinghs: That is right, thanks for the copy :) I will accept it here, as it is.

PR #4394, opened 2023-05-17
chore(import): Increase importCSV() debugging to debug CI flake

lerouxb:


packages/compass-import-export/src/import/import-csv.spec.ts
```diff
@@ -255,9 +255,6 @@ describe('importCSV', function () {
           ignoreEmptyStrings: true,
         });
 
-        const errorLog = await fs.promises.readFile(output.path, 'utf8');
```
lerouxb, line 258: I'm suspicious of this line since the error we're seeing is ENOENT after import seems to have completed. Although with no stack trace for some reason.

packages/compass-import-export/src/import/import-csv.ts
```diff
@@ -165,13 +165,17 @@ export async function importCSV({
     await pipeline(...params);
   } catch (err: any) {
     if (err.code === 'ABORT_ERR') {
+      debug('importCSV:aborting');
```
lerouxb, line 168: importCSV:aborting and importCSV:completing will log as soon as we abort or complete, before we write the import errors to file. importCSV:aborted and importCSV:completed should log with the result _after_ writing errors to file.

Hopefully that should be enough to show where things got to and where it went wrong the next time it flakes.

packages/compass-import-export/src/import/import-csv.spec.ts
```diff
@@ -242,6 +244,8 @@ describe('importCSV', function () {
 
         const output = temp.createWriteStream();
 
+        debug('output path', output.path);
```
lerouxb, line 247: So we can confirm that the path we see in the ENOENT error is actually the temp output path.

gribnoysup (approved):

PR #4401, opened 2023-05-18
chore(import-export, schema, field-store): update mongodb-schema, use schema paths helper COMPASS-6720

Anemy:
COMPASS-6720 COMPASS-6809

This updates our [`mongodb-schema`](https://github.com/mongodb-js/mongodb-schema) dependency in Compass.

Updating to the new version involves a few changes:
- Schema field paths now use string arrays instead of dot separated strings. This will make it easier for us to handle cases like dots in field names (previously we would join and split on dots).
- Better schema type definitions are now exported from `mongodb-schema`. This involved some changes in `compass-schema` and `compass-import-export` as we were supporting our own types.
- Removed the `getSchemaPaths` function from `compass-import-export` and updated to use the one from the `mongodb-schema` package,.
- Removed the unused `dimensionality` and `nestedFields` properties and calculations in `compass-field-store`. (could be separated to another pr if preferred). 
- In `compass-field-store` renamed the `aceFields` flattened map used for editor field autocompletion to `autocompleteFields` to be more generic as we are using this in code-mirror and no longer have ace. (could be separated to another pr if preferred).
- Updated schema property variable names, `has_duplicates`, `average_length`, and `total_count`, are now `camelCase`.
- When integrating changes in `compass-schema` there were updates needed in the telemetry methods, `schemaContainsGeoData` and `calculateSchemaDepth`. I ended up refactoring both and hopefully simplifying `calculateSchemaDepth` a bit, it looks like it might fix a recent error we had. Moved them off the store. Updated tests to be easier to maintain. We could consider moving the `calculateSchemaDepth` into `mongodb-schema`, but that's not included in these changes. I'm thinking this should fix COMPASS-6809 cc @addaleax


lerouxb (approved):
😵

PR #4405, opened 2023-05-19
chore(connection-form): rename connect-form components to connection-form

Anemy:
I think it was named `connect-form` when the package was also named that and we didn't update this file when we renamed the package.
I always search for `connection-form` when looking for this file and miss it lol, maybe other folks do the same.


mcasimir (approved):

packages/connection-form/src/components/connection-form.spec.tsx
```diff
@@ -8,8 +8,8 @@ import {
 } from '@testing-library/react';
 import { expect } from 'chai';
 
-import ConnectForm from './connect-form';
-import type { ConnectFormProps } from './connect-form';
+import ConnectForm from './connection-form';
```
mcasimir, line 11: Should we rename also ConnectForm to ConnectionForm?

mcasimir (approved):
❤️

PR #4419, opened 2023-05-22
feat(data-service): share OIDC state with embedded mongosh COMPASS-6802

addaleax:
This commit is sufficient to make OIDC connectivity work when using the connection string to set OIDC as the auth mechanism.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


configs/webpack-config-compass/src/index.ts
```diff
@@ -67,6 +67,11 @@ const sharedResolveOptions = (
       // Additionally `ampersand-sync` brings into the bundle a number of other dependencies
       // that are outdated and having known vulnerabilities.
       'ampersand-sync': false,
+      // `jose` provides a browser export that uses webcrypto APIs and returns
+      // webcrypto objects to represent keys, but openid-client requires
+      // KeyObject instances from the Node.js crypto API (https://tinyurl.com/2rrtu2hy).
```
Anemy, line 72: Why do we have this as a tinyurl?

packages/data-service/src/connect-mongo-client.spec.ts
```diff
@@ -143,7 +155,7 @@ describe('connectMongoClient', function () {
 
     it('should run the ping command with the specified ReadPreference', async function () {
       const commands: CommandStartedEvent[] = [];
-      const [metadataClient, crudClient] = await connectMongoClient(
+      const [metadataClient, crudClient, , state] = await connectMongoClient(
```
Anemy, line 158: Does the extra comma mean we ignore the `tunnel` item? I don't think I've seen this syntax used before, cool.

packages/compass-e2e-tests/tests/oidc.test.ts
```diff
@@ -0,0 +1,168 @@
...
+  let i = 0;
+  let tmpdir: string;
+  let server: ChildProcess;
+  let serverExit: Promise<unknown>;
+  let connectionString: string;
+
+  before(async function () {
+    // TODO(MONGOSH-1306): Get rid of all the setup code to download mongod here... :(
+    if (process.platform !== 'linux') {
+      return this.skip();
```
Anemy, line 38: Copying some of the comment over from https://github.com/mongodb-js/mongosh/pull/1458/files#diff-00a4109f7ba4dab51e247b01caab1dc7ceff5e3f32b8e5a03cab89eedd14d8d8R33
```suggestion
      // OIDC is only supported on Linux in the 7.0+ enterprise server.
      return this.skip();
```

packages/data-service/src/data-service.ts
```diff
@@ -1381,6 +1389,11 @@ class DataServiceImpl extends WithLogContext implements DataService {
         this._tunnel
           ?.close()
           .catch((err) => debug('failed to close tunnel', err)),
+        this._state
+          ?.destroy()
+          .catch((err) =>
+            debug('failed to destryo DevtoolsConnectionState', err)
```
Anemy, line 1395: Sorry for typo corrections - I now have this vscode extension that points them all out to me and I can't unsee 😅 
```suggestion
            debug('failed to destroy DevtoolsConnectionState', err)
```

packages/compass-e2e-tests/tests/oidc.test.ts
```diff
@@ -0,0 +1,168 @@
...
+          'authenticationMechanisms=SCRAM-SHA-256,MONGODB-OIDC',
+          // enableTestCommands allows using http:// issuers such as http://localhost
+          '--setParameter',
+          'enableTestCommands=true',
+          '--setParameter',
+          `oidcIdentityProviders=${JSON.stringify([serverOidcConfig])}`,
+          '--dbpath',
+          path.join(tmpdir, 'db'),
+          '--port',
+          '0',
```
Anemy, line 93: TIL port=0 makes it find a random (unused?) port. nice

packages/compass-e2e-tests/tests/oidc.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      process.execPath
+    } ${path.resolve(__dirname, '..', 'fixtures', 'curl.js')}`;
+    compass = await beforeTests();
+    browser = compass.browser;
+  });
+
+  after(async function () {
+    delete process.env.COMPASS_TEST_OIDC_BROWSER_DUMMY;
+    await afterTests(compass, this.currentTest);
+    server.kill();
```
Anemy, line 136:
```suggestion
    server?.kill();
```

Anemy (approved):
Looking good, left a couple questions/comments, I don't think any are blockers.
Looks like check is failing with the missing `jose` dependency we resolve in the webpack.

packages/compass-e2e-tests/tests/oidc.test.ts
addaleax: Yeah, it’s a nice feature of the OS and it’s good that the server also allows specifying this. Ideally we’ll want to use this for our mongodb-runner/mlaunch replacement so that we’ll just never have to worry about port conflicts.

packages/data-service/src/connect-mongo-client.spec.ts
addaleax: Yep, it just skips the item :+1:

configs/webpack-config-compass/src/index.ts
addaleax: It’s literally just because it’s a long URL that’s wider than most comments here. :slightly_smiling_face:  I don’t really care, happy to put the full link here as well :)

packages/data-service/src/connect-mongo-client.ts
```diff
@@ -49,6 +53,14 @@ export default async function connectMongoClientCompass(
     autoEncryption: connectionOptions.fleOptions?.autoEncryption,
   };
 
+  // TODO(COMPASS-6849): Add a way to properly override openBrowser
+  if (process.env.COMPASS_TEST_OIDC_BROWSER_DUMMY) {
```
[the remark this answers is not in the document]
addaleax: :woman_shrugging: 

For local testing, if I do a `throw` inside this block, it works as expected (i.e. doesn’t work), so it does pick up the environment variable there.

If there’s something odd happening in CI, I’m not sure, but it seems to pass and it’s only very temporary anyway, so I think I’m just okay with leaving this as it is? Worst case we’re opening browser Windows on the CI hosts, but I think those would be cleaned up when our Xvfb server shuts down anyway

PR #4429, opened 2023-05-25
feat(connection-form): add OIDC connection form options COMPASS-6799

Anemy:
COMPASS-6799

Here's how they look:
Screenshot 2023-05-30 at 1 30 31 PM
Screenshot 2023-05-30 at 1 20 45 PM

I'm thinking we should add e2e tests for connecting with the various options we provide (at least `redirectURI` and device auth). Not included in this pr (only unit tests around setting these options and e2e connection form tests).

Resources:

- PR where we added OIDC auth to mongosh: https://github.com/mongodb-js/mongosh/pull/1458
- [tech design](https://docs.google.com/document/d/1ztbmJwPOHxlSkbqIA0KW7749wegAGdDfUNcSF-x_tU8/edit#heading=h.vqmmf538fdxs)
- OIDC auth drivers spec https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#mongodb-oidc


packages/compass-preferences-model/src/preferences.ts
```diff
@@ -255,6 +256,23 @@ const featureFlagsProps: Required<{
...
+   */
+  enableOidc: {
+    type: 'boolean',
+    required: false,
+    default: undefined,
+    ui: true,
+    cli: true,
+    global: true,
+    description: {
+      short: 'OIDC Authentication',
```
addaleax, line 272:
```suggestion
      short: 'Enable OIDC Authentication',
```

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
```diff
@@ -0,0 +1,167 @@
...
+  Accordion,
+  Checkbox,
+  Description,
+  FormFieldContainer,
+  Label,
+  TextInput,
+} from '@mongodb-js/compass-components';
+import type ConnectionStringUrl from 'mongodb-connection-string-url';
+import type { ConnectionOptions } from 'mongodb-data-service';
+import type { AuthFlowType } from '@mongodb-js/oidc-plugin';
```
addaleax, line 12: Do you want to infer this from `OIDCOptions` below? That would remove the dependency on `@mongodb-js/oidc-plugin` here

```suggestion
type ExtractArrayEntryType<T> = T extends (infer U)[] ? U : never;
type AuthFlowType = ExtractArrayEntryType<OIDCOptions['allowedFlows']>;
```

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
```diff
@@ -0,0 +1,167 @@
...
+        value,
+      });
+    },
+    [updateConnectionFormField]
+  );
+
+  const authMechanismProperties =
+    parseAuthMechanismProperties(connectionStringUrl);
+  const allowedHosts = authMechanismProperties.get('ALLOWED_HOSTS');
+  const hasEnabledUntrustedEndpoints = allowedHosts === '*';
```
addaleax, line 52: Wasn’t quite sure whether to comment this here or in `connection-string-helpers.ts`, but `ALLOWED_HOSTS` is not an option that can be set in the URL, it needs to be set in the driver `authMechanismProperties` struct.

It might be easier to extends `ConnectionOptions['oidc']` with a boolean option here, e.g. `oidc?: OIDCOptions & { enableUntrustedEndpoints?: boolean }`; or similar.

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
```diff
@@ -0,0 +1,167 @@
...
+          data-testid="connection-oidc-username-input"
+          onChange={({
+            target: { value },
+          }: React.ChangeEvent<HTMLInputElement>) => {
+            updateConnectionFormField({
+              type: 'update-username',
+              username: value,
+            });
+          }}
+          label="Principal"
```
addaleax, line 71: Do we want to call this username? I know the specs and docs use the language a bit interchangeably, but at the end of the day, I think “username” would give users a more intuitive idea about what this field means.

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
```diff
@@ -0,0 +1,167 @@
...
+              data-testid="connection-oidc-auth-code-flow-redirect-uri-input"
+              onChange={({
+                target: { value },
+              }: React.ChangeEvent<HTMLInputElement>) => {
+                handleFieldChanged('redirectURI', value);
+              }}
+              optional
+              label="Auth Code Flow Redirect URI"
+              value={connectionOptions?.oidc?.redirectURI || ''}
+            />
```
addaleax, line 91: Do we want to add a short description here similar to the one below? E.g. “This value needs to match the configuration of the Identity Provider used by the server.”

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
```diff
@@ -0,0 +1,167 @@
...
+          </FormFieldContainer>
+
+          <FormFieldContainer>
+            <Checkbox
+              onChange={({
+                target: { checked },
+              }: React.ChangeEvent<HTMLInputElement>) => {
+                if (checked) {
+                  handleFieldChanged('allowedFlows', ['device-auth']);
```
addaleax, line 139: Doesn’t this override `allowedFlows = ['auth-code']` to `allowedFlows = ['device-auth']` when checked, and then `allowedFlows = ['device-auth']` to `allowedFlows = []` when unchecked?

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
```diff
@@ -0,0 +1,167 @@
...
+                handleFieldChanged(
+                  'allowedFlows',
+                  newAllowedFlows.length > 0 ? newAllowedFlows : undefined
+                );
+              }}
+              data-testid="oidc-enable-device-auth-flow-input"
+              id="oidc-enable-device-auth-flow-input"
+              label={
+                <Label htmlFor="oidc-enable-device-auth-flow-input">
+                  Enable device authentication flow
```
addaleax, line 155: I’ve usually seen this capitalized because it’s a proper noun, not sure if that looks weird:

```suggestion
                  Enable Device Authentication Flow
```

For this one we’ll definitely want a description, e.g. “Less secure authentication flow that can be used as a fallback when browser-based authentication is unavailable”.

This is probably extremely rare for Compass users because we’re already in a GUI environment to begin with, so I’d maybe consider putting this behind a setting similar to what we do for the Kerberos password field (but that doesn’t have to be part of this PR).

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
Anemy: This is some cool typing! Thanks, nice suggestion.

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
Anemy: I flip flopped a bit on this when writing it, sounds good.

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
Anemy: I've updated it to use a `enableUntrustedEndpoints` on the `OIDCOptions` which is then resolved before connecting in `data-service`. Was that similar to what you had in mind? Haven't added a test yet.

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
Anemy: Clarifying, from my understanding we'll need to supply a `notifyDeviceFlow` handler and show a ui for the device `userCode` in this work? https://github.com/mongodb-js/mongosh/pull/1458/files#diff-d998d2f6015503912abc1073f3bdb46da4c12651a7ba3d2e6170711edb19b45aR808
Since [`allowedFlows` is defaulted to include `auth-code` and `device-auth`](https://github.com/mongodb-js/oidc-plugin/blob/7bbbd1246833871491b75266193cd5a87bf845d5/src/plugin.ts#L266), shouldn't this be setting allowedFlows to only be `['device-auth']` to avoid `auth-code`?

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
Anemy: Nice catch, the standard in other tabs is capitalized. Sounds good, lets visit that in that ticket.

packages/compass-connections/src/stores/connections-store.ts
```diff
@@ -398,11 +425,31 @@ export function useConnections({
...
+
       const newConnectionDataService = await newConnectionAttempt.connect(
-        adjustConnectionOptionsBeforeConnect(
-          connectionInfo.connectionOptions,
-          appName
-        )
+        adjustConnectionOptionsBeforeConnect({
+          connectionOptions: connectionInfo.connectionOptions,
+          defaultAppName: appName,
+          notifyDeviceFlow,
```
addaleax, line 451: I think you’ll need to replace `getMongoClientConnectionOptions()` in DataServiceImpl with 

```
  return _.omit(this._mongoClientConnectionOptions, 'options.oidc.notifyDeviceFlow');
```

so that the shell doesn’t choke on the function property

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
```diff
@@ -0,0 +1,169 @@
...
+              id="oidc-allow-untrusted-endpoint-input"
+              label={
+                <>
+                  <Label htmlFor="oidc-allow-untrusted-endpoint-input">
+                    Enable Untrusted Target Endpoint
+                  </Label>
+                  <Description>
+                    Allow connecting when the target endpoint is not in the list
+                    of trusted endpoints &#40;this sets the driver&apos;s
+                    authMechanismProperties&apos; ALLOWED_HOSTS list to *&#41;
```
addaleax, line 113: I’m not sure I have a specific good suggestion here, but this sounds a bit ambiguous – what the user should know when they check this box is that they promise that this is a _trusted_ endpoint, rather than relying on MongoDB’s preconfigured list of known trusted endpoints. But “Enable Untrusted Target Endpoint” makes it sound a bit like it would put Compass into a more “secure” mode where it handles connections to potentially untrustworthy endpoints well too, if you know what I mean? Maybe something like this:

```suggestion
                    Consider Target Endpoint Trusted
                  </Label>
                  <Description>
                    Allow connecting when the target endpoint is not in the list
                    of endpoints that are considered trusted by default.
                    Only use this option when connecting to servers that you trust.
```

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
addaleax: > Was that similar to what you had in mind?

Yes :)

> Would it work/be cleaner nicer to add the `authMechanismProperties` to the `ConnectionOptions` and then when we connect we pull it in?

Yeah, I think on the one hand this is probably a bit closer to the spirit of `ConnectionOptions`, on the other hand, we really don’t want to add more usage of `authMechanismProperties` outside of the connection string. If we do add something to `ConnectionOptions`, it would probably be more about including all possible `DevtoolsConnectOptions` in it, so that we can handle e.g. `ConnectionFleOptions` there as well.

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
addaleax: > I added one, can remove if it's not needed here.

No, it looks great! This is exactly what we want :+1:

> Since [`allowedFlows` is defaulted to include `auth-code` and `device-auth`](https://github.com/mongodb-js/oidc-plugin/blob/7bbbd1246833871491b75266193cd5a87bf845d5/src/plugin.ts#L266), shouldn't this be setting allowedFlows to only be `['device-auth']` to avoid `auth-code`?

I think we’ll want to default to `['auth-code']` in Compass like we do in mongosh, and I’d lean towards this button making it so that the value is `['auth-code', 'device-auth']` rather than `['device-auth']` only. We can probably re-evaluate that choice if it turns out to be an issue, but in practice it really shouldn’t be.

The default in the OIDC plugin is mostly that way because we made the decision to default to Auth-Code-Flow-only in mongosh fairly late (and it’s not currently a real security problem because we only connect to trusted endpoints anyway). Maybe also relevant: https://github.com/mongodb/docs-node/pull/698/files#r1210137439

packages/compass-connections/src/components/connecting/connecting.tsx
```diff
@@ -71,8 +87,28 @@ function Connecting({
...
+              <Link href={oidcDeviceAuthVerificationUrl} target="_blank">
+                {oidcDeviceAuthVerificationUrl}
+              </Link>
+              <Body className={textContentStyles}>
+                Enter the following code on that page:
+              </Body>
+              <InlineCode baseFontSize={16}>
+                {oidcDeviceAuthUserCode}
+              </InlineCode>
+            </div>
```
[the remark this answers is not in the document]
Anemy: Screenshot 2023-05-29 at 7 38 02 PM
Screenshot drive by folks to see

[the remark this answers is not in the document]
addaleax: Yeah, I think that would be great :+1:

packages/compass-connections/src/stores/connections-store.ts
```diff
@@ -67,6 +80,12 @@ type Action =
       connectionAttempt: ConnectionAttempt;
       connectingStatusText: string;
     }
+  | {
+      type: 'oidc-attempt-connect-notify-device-auth';
+      // TODO: Do these make sense elsewhere?
```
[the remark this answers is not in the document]
Anemy: Ah yup, it's a leftover I missed - I'll remove shortly, thanks for noticing.

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
Anemy: Sounds good, let's keep it as is with the `enableUntrustedEndpoints` and we can looks at `DevtoolsConnectOptions` down the line.

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
Anemy: Updated to default 👍 
I'm thinking we'll want an e2e test to cover device auth in Compass, separate pr?

packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-oidc.tsx
Anemy: Good thinking - thanks for updating!

PR #4430, opened 2023-05-26
chore: update electron

svc-devtoolsbot:
- Update electron


Anemy:
Going to merge this once tests pass as the browsers lists have updated and are failing on main with the existing electron

PR #4450, opened 2023-06-01
chore(e2e): wait for animations on the option before clicking to select it

lerouxb:
Try and fix the flake we've been seeing in CI where it selects minPoolSize rather than maxPoolSize.

### DIFF
--- a/packages/compass-e2e-tests/helpers/commands/set-connect-form-state.ts
+++ b/packages/compass-e2e-tests/helpers/commands/set-connect-form-state.ts
@@ -1,7 +1,28 @@
+import _ from 'lodash';
 import { expect } from 'chai';
 import type { CompassBrowser } from '../compass-browser';
 import * as Selectors from '../selectors';
 import type { ConnectFormState } from '../connect-form-state';
+
+async function waitForElementAnimations(browser: CompassBrowser, element: any) {
+  let previousResult = {
+    ...(await element.getLocation()),
+    ...(await element.getSize()),
+  };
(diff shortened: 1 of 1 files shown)


packages/compass-e2e-tests/helpers/commands/set-connect-form-state.ts
```diff
@@ -1,8 +1,29 @@
+import _ from 'lodash';
 import { expect } from 'chai';
 import type { CompassBrowser } from '../compass-browser';
 import * as Selectors from '../selectors';
 import type { ConnectFormState } from '../connect-form-state';
 
+async function waitForElementAnimations(browser: CompassBrowser, element: any) {
```
lerouxb, line 7: I can't figure out what the type of element should be. I thought it would be Element<'async'> to mirror the type of Browser, but then it complains (with a warning, not error) that it is WebdriverIO.Element which is unsafe. But I can't figure out how to even import that to try it.

packages/compass-e2e-tests/helpers/commands/set-connect-form-state.ts
```diff
@@ -1,8 +1,29 @@
...
+  let previousResult = {
+    ...(await element.getLocation()),
+    ...(await element.getSize()),
+  };
+  await browser.waitUntil(async function () {
```
lerouxb, line 12: actually I just realised that waitUntil also exists on element, but then we use beowser.pause which does not anyway, so whatever.

PR #4452, opened 2023-06-01
fix(compass-import-export): add file type filters when exporting COMPASS-6890

Anemy:
Previously we would only add the filetype to the file name. Now we will always set it to the filetype unless the user changes the filter to `All Files`.

image


lerouxb (approved):

addaleax (approved):

Anemy:
Merging, failure looks like unrelated crud flake

PR #4454, opened 2023-06-01
fix: pass through AbortSignal for connection cancellation COMPASS-6871

addaleax:
### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/data-service/src/connect-mongo-client.ts
+++ b/packages/data-service/src/connect-mongo-client.ts
@@ -57,6 +57,7 @@ export function prepareOIDCOptions(
 export async function connectMongoClientCompass(
   connectionOptions: Readonly<ConnectionOptions>,
   setupListeners: (client: MongoClient) => void,
+  signal?: AbortSignal,
   logger?: UnboundDataServiceImplLogger
 ): Promise<
   [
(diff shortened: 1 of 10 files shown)


packages/data-service/src/connect-mongo-client.ts
```diff
@@ -73,6 +74,9 @@ export async function connectMongoClientCompass(
   );
 
   const oidcOptions = prepareOIDCOptions(connectionOptions);
+  // @ts-expect-error Will go away on @types/node update
+  // with proper `AbortSignal` typings
+  oidcOptions.oidc.signal = signal;
```
Anemy, line 79: Small thing, fine as is here too, how do you feel about moving this line into `prepareOIDCOptions`. (Then we can update the unit tests for that function).

Anemy (approved):
lgtm! Nice test add. One small suggestion, not a blocker.

packages/data-service/src/connect-mongo-client.ts
addaleax: @anemy Is something like 9c9c929b0323c28f580908219c1fe8c383ebb301 what you had in mind here?

PR #4455, opened 2023-06-01
chore(compsas): compass window.webContents will not add utm parameters for all *.mongodb.com links

himanshusinghs:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass/src/main/window-manager.ts
+++ b/packages/compass/src/main/window-manager.ts
@@ -25,6 +25,24 @@ import {
 
 const debug = createDebug('mongodb-compass:electron:window-manager');
 
+const urlShouldHaveUtmParams = ({ hostname }: URL) => {
+  const EXCLUDED_MONGODB_HOSTS = [
+    'compass-maps.mongodb.com',
+    'evergreen.mongodb.com',
+    'downloads.mongodb.com',
+    'cloud.mongodb.com',
+  ];
+
+  try {
+    return (
+      /^(.*\.)?mongodb\.com$/.test(hostname) &&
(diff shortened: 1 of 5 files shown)


packages/compass/src/main/window-manager.ts
```diff
@@ -25,6 +25,31 @@ import {
...
+export const urlWithUtmParams = (urlString: string): string => {
+  try {
+    const url = new URL(urlString);
+    const urlShouldHaveUtmParams =
+      /^(.*\.)?mongodb\.com$/.test(url.hostname) &&
+      !EXCLUDED_MONGODB_HOSTS.includes(url.hostname);
+
+    if (urlShouldHaveUtmParams) {
+      url.searchParams.set('utm_source', 'compass');
+      url.searchParams.set('utm_medium', 'product');
```
mabaasit, line 45: Here we are overriding the exisiting `utm` params in url. If that's intended, looks good.

himanshusinghs: Yes, that is intended. We came to this conclusion after a small thought work [here](https://github.com/mongodb-js/compass/pull/4446#issuecomment-1569949401). The appropriate condition to not enforce these utm values would have been the case when the code will be shared across other tools but fortunately, with your last feedback, this part of the logic is very compass specific.

himanshusinghs: But we can always revisit this piece if the situation changes in the future.

gribnoysup (approved):

PR #4456, opened 2023-06-01
chore: run snyk test manually on each package

mcasimir:
Snyk introduced proper support for npm workspaces only recently, but apparently is still not perfect. Some of the dependencies not hoisted are not detected properly. The workaround is to run snyk test manually for each workspace


scripts/snyk-test.js
```diff
@@ -0,0 +1,89 @@
...
+    );
+
+    const res = JSON.parse(await fs.readFile(tmpPath));
+    console.info(`testing ${cwd} done.`);
+    return res;
+  } catch (err) {
+    console.error(`testing ${cwd} failed. ${err.message}`);
+  } finally {
+    try {
+      await fs.rm(tmpPath);
```
addaleax, line 42:
```suggestion
      await fs.rm(tmpPath, { recursive: true });
```

scripts/snyk-test.js
```diff
@@ -0,0 +1,89 @@
...
+const os = require('os');
+const { glob } = require('glob');
+
+async function spawn(cmd, args, opts) {
+  return new Promise((resolve, reject) => {
+    const proc = childProcess.spawn(cmd, args, opts);
+    proc.on('error', (err) => reject(err));
+    proc.on('close', (code) => resolve(code));
+  });
+}
```
addaleax, line 13:
```suggestion
const { promisify } = require('util');
const spawn = promisify(childProcess.execFile);
```

(from the Node.js docs:)

> If this method is invoked as its [util.promisify()](https://nodejs.org/api/util.html#utilpromisifyoriginal)ed version, it returns a Promise for an Object with stdout and stderr properties. The returned ChildProcess instance is attached to the Promise as a child property. In case of an error (including any error resulting in an exit code other than 0), a rejected promise is returned, with the same error object given in the callback, but with two additional properties stdout and stderr.

> ```js
> const util = require('node:util');
> const execFile = util.promisify(require('node:child_process').execFile);
> async function getVersion() {
>   const { stdout } = await execFile('node', ['--version']);
>   console.log(stdout);
> }
> getVersion();
> ```

scripts/snyk-test.js
mcasimir: does it have an effect if is always a file?

addaleax: ah no you’re right, I didn’t notice that this only a single file – sorry!

PR #4464, opened 2023-06-02
feat(compass-settings): add OIDC settings, add enable device auth setting COMPASS-6849

Anemy:
COMPASS-6849

In COMPASS-6801 we'll add a setting for persisting OIDC state across sessions so this settings list won't look so empty :)

Here's how it looks:

https://github.com/mongodb-js/compass/assets/1791149/a1335a04-0ca8-4376-87b5-5d2d3cd59fc2


addaleax (approved):

Anemy:
Chatting with Anna, going to do a follow up pr that adds `browser` to these settings.

PR #4469, opened 2023-06-04
feat(compass-settings): add OIDC browser setting COMPASS-6849

Anemy:
COMPASS-6849

The e2e tests already use a custom browser, this pr adds one more that ensures the browser command from the settings is used.

Screenshot 2023-06-04 at 12 00 08 PM

### DIFF
--- a/packages/compass-e2e-tests/tests/oidc.test.ts
+++ b/packages/compass-e2e-tests/tests/oidc.test.ts
@@ -34,6 +34,17 @@ const DEFAULT_AUTH_INFO = {
   authenticatedUserRoles: [{ role: 'dev/testgroup', db: 'admin' }],
 };
 
+const host = '127.0.0.1';
+
+function getTestBrowserCommand() {
+  return `${process.execPath} ${path.resolve(
+    __dirname,
+    '..',
+    'fixtures',
+    'curl.js'
+  )}`;
+}
+
 describe('OIDC integration', function () {
   let compass: Compass;
   let browser: CompassBrowser;
@@ -44,6 +55,7 @@ describe('OIDC integration', function () {
   let oidcMockProvider: OIDCMockProvider;
 
   let i = 0;
+  let port: number;
   let tmpdir: string;
   let server: ChildProcess;
   let serverExit: Promise<unknown>;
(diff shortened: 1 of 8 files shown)


packages/compass-e2e-tests/tests/oidc.test.ts
```diff
@@ -141,18 +153,19 @@ describe('OIDC integration', function () {
         })(),
       ]);
 
-      connectionString = `mongodb://127.0.0.1:${port}/?authMechanism=MONGODB-OIDC`;
+      connectionString = `mongodb://${host}:${port}/?authMechanism=MONGODB-OIDC`;
     }
-
-    process.env.COMPASS_TEST_OIDC_BROWSER_DUMMY = `${
-      process.execPath
-    } ${path.resolve(__dirname, '..', 'fixtures', 'curl.js')}`;
```
addaleax, line 149: Do you also want to remove the corresponding piece from the data-service package?

packages/compass-e2e-tests/tests/oidc.test.ts
```diff
@@ -169,58 +182,108 @@ describe('OIDC integration', function () {
...
+  describe('with test browser', function () {
+    before(function () {
+      process.env.COMPASS_TEST_OIDC_BROWSER_DUMMY = getTestBrowserCommand();
+      `${
+        process.execPath
+      } ${path.resolve(__dirname, '..', 'fixtures', 'curl.js')}`;
+    });
+
+    after(function () {
+      delete process.env.COMPASS_TEST_OIDC_BROWSER_DUMMY;
```
addaleax, line 194: Fwiw, I imagined that this env var/the corresponding bit in the data-service package would go away with this setting, and that we’d just always use the setting instead … does that seem reasonable to you?

addaleax (approved):

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -532,6 +533,21 @@ const modelPreferencesProps: Required<{
...
+  browserCommandForOIDCAuth: {
+    type: 'string',
+    required: false,
+    default: undefined,
+    ui: true,
+    cli: true,
+    global: true,
+    description: {
+      short: 'Browser command to use for OIDC Authentication',
+      long: 'Change the browser command that is run to start the browser for authenticating with the OIDC identity provider. Leave this empty for default behavior.',
```
addaleax, line 548: Actually, this should probably mention that this string will be passed to a _shell_, i.e. special characters have the meaning that they have in the shell, e.g.suggestion
      long: 'Specify a shell command that is run to start the browser for authenticating with the OIDC identity provider. Leave this empty for default browser.',
```

**Comment by addaleax** [comment]
> Should we let folks set this to false?

I’d lean towards “no” for now, we can always revisit this if users ask for it, but just like with device auth flow in general, we’re really expecting to almost always be in a situation in Compass in which we can open a browser window. (Users could also use a command that always fails here, like e.g. the literal `false` on UNIX; that’s a bit slower because we’d still open the local HTTP server and everything, but it would have the same effect.)

File: packages/compass-e2e-tests/tests/oidc.test.ts (hunk 2 above)
Comment (Anemy) [reply packages/compass-e2e-tests/tests/oidc.test.ts]: Missed that TODO, makes sense, updated. Thanks for noticing!

**Review by addaleax — APPROVED** [review]

File: packages/compass-e2e-tests/tests/oidc.test.ts [hunk 4]
```diff
@@ -141,18 +153,23 @@ describe('OIDC integration', function () {
         })(),
       ]);
 
-      connectionString = `mongodb://127.0.0.1:${port}/?authMechanism=MONGODB-OIDC`;
+      connectionString = `mongodb://${host}:${port}/?authMechanism=MONGODB-OIDC`;
+      await browser.setFeature(
+        'browserCommandForOIDCAuth',
+        getTestBrowserShellCommand()
+      );
```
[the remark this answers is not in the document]
Anemy: 🤦 ty, morning coding xD

PR #4473, opened 2023-06-05
chore(telemetry): add telemetry for windows events

mcasimir:
Poc to track window events:

- **`'Window Open'`**: A new window is created
  - `number_of_windows_open`: number of window open at the time the event is fired

- **`'Window Closed'`**: A window is closed
  - `number_of_windows_open`: number of window open at the time the event is fired

- **`'Window Focused'`**: A window receive focus
  - `number_of_windows_open`: number of window open at the time the event is fired

- **`'Window Blurred'`**: A window lose focus
  - `session_duration_secs`: seconds passed since the last focus on the same window
  - `focus_to_other_compass_window`:
  - `number_of_windows_open`: number of window open at the time the event is fired

- **`'Application Blurred'`**: The application lose focus (user switch to other app)
  - `session_duration_secs`: seconds passed since the last focus on any window
  - `number_of_windows_open`: number of window open at the time the event is fired


packages/compass/src/main/window-manager.ts
```diff
@@ -259,6 +260,66 @@ async function onAppReady() {
   }
 }
 
+function trackWindowEvents(electronApp: App) {
+  const windowFocusedAt = new Map<number, number>();
```
addaleax, line 264:
```suggestion
  // Map of [Window id] -> [opening timestamp in milliseconds]
  const windowFocusedAt = new Map<number, number>();
```

addaleax (approved):

PR #4479, opened 2023-06-06
fix: select combobox option onBlur COMPASS-6511

alenakhineika:
## Description

Select the combobox value not only when clicking on the combobox option from a dropdown, but also when clicking outside the combobox area. Otherwise, the combobox input may contain a visible search value, but the value is not actually being selected, what causes the `You must select a field name and type` error on the indexes modal submit. This PR also replaces the current combobox with ComboboxWithCustomOption component, that is basically the same combobox, so we keep the logic of adding a custom value to the combobox options list in a single place.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-components/src/components/combobox-with-custom-option.tsx
```diff
@@ -44,6 +44,16 @@ export const ComboboxWithCustomOption = <
...
       multiselect={multiselect}
       onFilter={setSearch}
+      onBlur={() => {
+        // Select an option on blur event to fix missing values on modal submit COMPASS-6511
+        if (!onChange) return;
+        if (search) {
+          if (!userOptions.find((x) => x.value === search)) {
+            setCustomOptions([{ value: search } as K]);
+          }
+          (onChange as onChangeType<false>)(search);
```
addaleax, line 54: What happens if `multiselect` is `true`?

alenakhineika: Added a comment that we skip the onBlur event for multiselect, because in this case, a user should explicitly check the value in the list.

packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/field-combobox.tsx
```diff
@@ -81,7 +81,6 @@ export const FieldCombobox = ({
       overflow="scroll-x"
       {...props}
       options={fields}
-      value={value}
```
alenakhineika, line 84: Looks like setting this value does not change the behavior of the combobox. Let me know if I miss something here.

mabaasit (approved):

PR #4484, opened 2023-06-06
feat: optionally store OIDC tokens as secrets COMPASS-6801

addaleax:
Just a very early WIP right now with most of the work still left to do:

TBD:
- [x] Add a Compass setting to disable this
- [x] Include https://github.com/mongodb-js/devtools-connect/pull/27
- [x] Add e2e tests

### DIFF
--- a/packages/compass-connections/src/stores/connections-store.ts
+++ b/packages/compass-connections/src/stores/connections-store.ts
@@ -30,6 +30,14 @@ type ConnectFn = typeof connect;
 
 export type { ConnectFn };
 
+type RecursivePartial<T> = {
+  [P in keyof T]?: T[P] extends (infer U)[]
+    ? RecursivePartial<U>[]
+    : T[P] extends object | undefined
+    ? RecursivePartial<T[P]>
+    : T[P];
+};
+
 export function createNewConnectionInfo(): ConnectionInfo {
   return {
     id: uuidv4(),
@@ -61,6 +69,7 @@ type State = {
   connections: ConnectionInfo[];
   oidcDeviceAuthVerificationUrl: string | null;
   oidcDeviceAuthUserCode: string | null;
+  connectionMergeInfos: Record<string, RecursivePartial<ConnectionInfo>>;
 };
 
 export function defaultConnectionsState(): State {
@@ -73,6 +82,7 @@ export function defaultConnectionsState(): State {
     connectionErrorMessage: null,
     oidcDeviceAuthVerificationUrl: null,
     oidcDeviceAuthUserCode: null,
+    connectionMergeInfos: {},
   };
 }
 
@@ -113,6 +123,11 @@ type Action =
       type: 'set-connections-and-select';
       connections: ConnectionInfo[];
       activeConnectionInfo: ConnectionInfo;
+    }
+  | {
+      type: 'add-connection-merge-info';
+      id: string;
+      mergeConnectionInfo: RecursivePartial<ConnectionInfo>;
     };
 
 export function connectionsReducer(state: State, action: Action): State {
@@ -177,6 +192,17 @@ export function connectionsReducer(state: State, action: Action): State {
         activeConnectionId: action.activeConnectionInfo.id,
         activeConnectionInfo: action.activeConnectionInfo,
         connectionErrorMessage: null,
+      };
+    case 'add-connection-merge-info':
+      return {
+        ...state,
+        connectionMergeInfos: {
+          ...state.connectionMergeInfos,
+          [action.id]: merge(
+            cloneDeep(state.connectionMergeInfos[action.id]),
+            action.mergeConnectionInfo
+          ),
+        },
       };
     default:
       return state;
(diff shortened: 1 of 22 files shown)


packages/data-service/src/data-service.ts
```diff
@@ -713,6 +714,13 @@ export interface DataService {
    * Register reauthentication handlers with this DataService instance.
    */
   addReauthenticationHandler(handler: ReauthenticationHandler): void;
+
+  /**
+   * Return the current state of ConnectionOptions secrets, which may have changed
+   * since connecting (e.g. OIDC tokens). The `connectionInfoSecretsChanged` event
+   * is being emitted when this value changes.
+   */
+  getUpdatedSecrets(): Promise<Partial<ConnectionOptions>>;
```
mcasimir, line 723: Just a random thought, somehow I would rather have something less generic so the domain and intent stands out more, both here and in the `connectionMergeInfos`. Like  `getUpdatedOIDCTokens`. In know is less future-proof but we could easily refactor it later if we would have more secrets.

packages/compass-connections/src/stores/connections-store.ts
```diff
@@ -190,11 +218,28 @@ async function loadConnections(
 ) {
   try {
     const loadedConnections = await connectionStorage.loadAll();
+    const toBeReSaved: ConnectionInfo[] = [];
```
mcasimir, line 221: Not blocking. There is something odd to me with the fact that the frontend doesn't directly use the tokens, and could operate without its knowledge .. but here it seems to be the thing that orchestrates storing it. 

Maybe a bit of a stretch and unnecessary, but did we consider having a separate store for tokens and/or push this logic out of the frontend?

mcasimir (commented):
I left a couple of non-blocking comments, probably someone with more info should do the actual review

packages/compass-connections/src/stores/connections-store.ts
addaleax: > Not blocking. There is something odd to me with the fact that the frontend doesn't directly use the tokens, and could operate without this knowledge .. but here it seems to be the thing that orchestrates storing it.

Sooo … that’s part of the idea behind what you commented on above – this part of the code _should_ operate with as little knowledge about what’s happening here under the hood.

> Maybe a bit of a stretch and unnecessary, but did we consider having a separate store for tokens and/or push this logic out of the frontend?

I think the part that I’m having a hard time understanding here is where the boundaries of “frontend” would be, or where a better place for this logic would be.

It would be great to refactor this code to have a layer between `useConnections()`, which should really only be a React store, and `ConnectionStorage`, which should really only provide primitives for loading/saving connection info objects.

Right now, that layer doesn’t really exist, though, and I’m not sure how much work adding it would be here. If we think it’s worth it, we could make this part of this project.

packages/compass-connections/src/stores/connections-store.ts
```diff
@@ -59,6 +69,7 @@ type State = {
   connections: ConnectionInfo[];
   oidcDeviceAuthVerificationUrl: string | null;
   oidcDeviceAuthUserCode: string | null;
+  connectionMergeInfos: Record<string, RecursivePartial<ConnectionInfo>>;
```
Anemy, line 72: Would it help to leave a comment here? The name `connectionMergeInfos`, at least for me, doesn't say much about what it is or used for. Can't think of a better name myself.
Maybe:
```suggestion
  // Additional connection information that is merged with the connection info
  // when connecting. This is useful for instances like OIDC sessions where we
  // have a setting on the system for storing credentials.
  // When the setting is on this `connectionMergeInfos` would have the session
  // credential information and merge it before connecting.
  connectionMergeInfos: Record<string, RecursivePartial<ConnectionInfo>>;
```

Anemy (approved):
lgtm, one documentation suggestion. Nice work

packages/compass-connections/src/stores/connections-store.ts
addaleax: Yeah, that makes perfect sense to me :+1:

packages/compass-connections/src/stores/connections-store.ts
mcasimir: Yeah, no worries, feel free to ignore it. I'd love the idea of `useConnections` but let's keep it for later. 

I was thinking more about moving it out of the redux cycle, and avoid the generic merge:

```
ds = new Dataservice(storeOidcToken = () => {
   ConnectionStorage.updateOidcToken(connectionId, token)
});
```

But is very dependent on my read of all the moving parts, i see `ConnectionStorage` as a repository for domain objects, which means that it is correct for it to have understanding of domain concepts like oidc, and not just a crud thing.

> should operate with as little knowledge about what’s happening here under the hood.

We are generalising what’s happening here under the hood so it doesn't feel like it, but we are still doing it .. and there is nothing wrong with it, especially if is necessary or the best practical approach.

addaleax: > I was thinking more about moving it out of the redux cycle, and avoid the generic merge:

To be clear, I like that approach and I would prefer it, but unfortunately, the compass-connections package assumes that for the lifetime of the Compass process, _it_ is the thing that maintains the state of the connection storage (because it essentially keeps another, in-memory copy of that). We _should_ get rid of that, and move the actual connection "collection management" logic out of it – it’s just not quite straightforward to do.

I’ve opened https://jira.mongodb.org/browse/COMPASS-6909 as a tech debt ticket to track that work, and also listed another example of where this is already problematic (namely, that after importing connections, the import logic has to call the hook’s expose `reloadConnections` method which is being passed through the actual frontend).

mcasimir: Awesome, thanks! We should try to prioritise that soon

PR #4490, opened 2023-06-07
fix(compass-shell): let enableShell setting control runtime start/stop

addaleax:
Make it so that the `enableShell` toggle does not only control whether the shell is displayed, but actually create or tear down the shell runtime instance in according to it.

This obviously does not fix any issues on its own, but it does unblock development on e.g. https://github.com/mongodb-js/compass/pull/4484, and seems like a good idea in general, because it also gives users a workaround around not just the current Electron bug, but also future problems where having the shell process running in the background leads to undesirable behavior in the main Compass app.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-shell/src/stores/store.js
+++ b/packages/compass-shell/src/stores/store.js
@@ -1,9 +1,10 @@
 import { createStore } from 'redux';
 import reducer from '../modules';
-import { setupRuntime } from '../modules/runtime';
+import { changeEnableShell, setupRuntime } from '../modules/runtime';
 import { globalAppRegistryActivated } from '@mongodb-js/mongodb-redux-common/app-registry';
 import { setupLoggerAndTelemetry } from '@mongosh/logging';
 import createLoggerAndTelemetry from '@mongodb-js/compass-logging';
+import preferences from 'compass-preferences-model';
 
 const { log, debug, track } = createLoggerAndTelemetry('COMPASS-SHELL');
 
@@ -22,6 +23,12 @@ export default class CompassShellStore {
     appRegistry.on('data-service-connected', this.onDataServiceConnected);
 
     appRegistry.on('data-service-disconnected', this.onDataServiceDisconnected);
+
+    preferences.onPreferenceValueChanged(
+      'enableShell',
+      this.onEnableShellChanged
+    );
+    this.onEnableShellChanged(preferences.getPreferences().enableShell);
 
     setupLoggerAndTelemetry(
       appRegistry,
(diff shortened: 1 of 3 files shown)


packages/compass-shell/src/stores/store.spec.js
```diff
@@ -97,7 +104,7 @@ describe('CompassShellStore [Store]', function () {
       appRegistry.emit('data-service-connected', null, fakeDataService);
       const runtimeState2 = getRuntimeState();
 
-      expect(runtimeState1).to.equal(runtimeState2);
+      expect(runtimeState1).to.deep.equal(runtimeState2);
```
Anemy, line 107: ❤️

packages/compass-shell/src/stores/store.spec.js
```diff
@@ -75,7 +82,7 @@ describe('CompassShellStore [Store]', function () {
       // become available
       await runtimeState.runtime.evaluate('help');
 
-      expect(eventRecieved).to.equal(true);
+      expect(eventReceived).to.equal(true);
```
Anemy, line 85: I still spell this wrong to this day 🤦

packages/compass-shell/src/stores/store.js
```diff
@@ -47,21 +54,17 @@ export default class CompassShellStore {
...
   onDataServiceDisconnected = () => {
-    const {
-      runtime: { runtime },
-    } = this.reduxStore.getState();
-
-    if (runtime) {
-      runtime.terminate();
-    }
-
     this.reduxStore.dispatch(setupRuntime(null, null, null));
```
Anemy, line 68: Do we ever unset the `dataService` on the store?
How do you feel about adding a new redux action for this, that way the reducer then responds to the disconnect action by terminating the runtime if it exists.

Here's how we do something similar in `import-export`:
Calling the action:
https://github.com/mongodb-js/compass/blob/8a7cf9d3bb953cc4b8bbbf8c417575fd4b33eb34/packages/compass-import-export/src/stores/export-store.ts#L37
Here's one place where it's handled:
https://github.com/mongodb-js/compass/blob/8a7cf9d3bb953cc4b8bbbf8c417575fd4b33eb34/packages/compass-import-export/src/modules/export.ts#L624
Then it's unset on the store with the same action here:
https://github.com/mongodb-js/compass/blob/8a7cf9d3bb953cc4b8bbbf8c417575fd4b33eb34/packages/compass-import-export/src/modules/compass/data-service.ts#L49

`dataService` is on this same reducer so we could also unset it when we take down the runtime.

gribnoysup (approved):

packages/compass-shell/src/stores/store.js
addaleax: Maybe I’m missing something, but this here is the exact line that creates the redux action to unset `dataService` on the store and have that call `state.runtime.terminate()` … we could also send another action instead of `setupRuntime(null, null, null)`, I don’t know what the best practices for redux are, but it doesn’t seem to make a practical difference?

Anemy: Yup no difference, fine as is

PR #4498, opened 2023-06-09
fix: fix SNYK-JS-FOLLOWREDIRECTS-2396346 and SNYK-JS-FOLLOWREDIRECTS-2332181 COMPASS-6880, COMPASS-6879

mcasimir:


packages/compass/src/main/telemetry.ts
```diff
@@ -121,6 +121,10 @@ class CompassTelemetry {
     }
   }
 
+  private static _flushTelemetryAndIgnoreFailure() {
+    return CompassTelemetry.analytics?.flush().catch(() => Promise.resolve());
```
mcasimir, line 125: flush returns a promise now, did i replicate the intended behaviour correctly?

lerouxb: I _think_ so but for some reason I'm never 100% sure about how .catch() behaves and have to look it up again every time.

lerouxb (approved):

packages/compass/src/main/telemetry.ts
addaleax: I think you did, yes. But just for consistency with the rest of this file:

```suggestion
    return this.analytics?.flush().catch(() => Promise.resolve());
```

addaleax (approved):

PR #4507, opened 2023-06-09
fix(compass-crud): show 1 as the page number even if in a collection, no entries are present

prashant-th18:
## Description
If we have a collection with non-zero documents, then if we delete all of them, the page number after deleting the last document is shown as '1'; however, it is changed to '0' on refreshing the collection.

In the code, where the 'start', which is responsible for showing the page number, is assigned, I have added a check to see if the length of the documents present is 0, so 0 will be assigned.

### Checklist

## Motivation and Context
- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes
- [ ] *Backport Needed*


packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -864,7 +864,8 @@ class CrudStoreImpl
         ? DOCUMENTS_STATUS_ERROR
         : DOCUMENTS_STATUS_FETCHED_PAGINATION,
       docs: documents.map((doc: BSONObject) => new HadronDocument(doc)),
-      start: skip + 1,
+      // making sure we don't set start to 1 if length is 0
+      start: length == 0 ? 0 : skip + 1,
```
addaleax, line 868: Optional suggestion, but I think we generally prefer `===` and `length` should always be a number anyway?

```suggestion
      start: length === 0 ? 0 : skip + 1,
```

addaleax (approved):

packages/compass-crud/src/stores/crud-store.ts
prashant-th18: Yes.

PR #4509, opened 2023-06-09
chore: Electron 23 COMPASS-6897

lerouxb:


.evergreen/tasks.yml
```diff
@@ -25,7 +25,7 @@ tasks:
       - func: bootstrap
       - func: test-electron
         vars:
-          debug: 'hadron*,mongo*'
+          debug: 'hadron*,mongo*,*COMPASS-CONNECT-UI*'
```
lerouxb, line 28: Trying to see what's going on here https://github.com/mongodb-js/compass/blob/c18ea4d531f14aea20b161ec1c563ea340eed80f/packages/compass-connections/src/modules/telemetry.ts#L51

configs/mocha-config-compass/register/dns-config-register.js
```diff
@@ -0,0 +1,2 @@
+const dns = require('dns');
+dns.setDefaultResultOrder('ipv4first');
```
addaleax, line 2: This should come with some sort of explanation and a TODO comment about removing this as soon as we can

PR #4521, opened 2023-06-13
chore: update node to match the electron version COMPASS-6915

mcasimir:


lerouxb (approved):

addaleax (approved):

scripts/testserver.js
```diff
@@ -12,6 +13,8 @@ const yargsParser = require('yargs-parser');
 const { rimraf } = require('rimraf');
 const glob = require('glob');
 
+dns.setDefaultResultOrder('ipv4first');
```
addaleax, line 16: `// TODO(COMPASS-6516)` everywhere?

lerouxb: Yes, but right now I'm wondering if this is more of an investigation PR :) At least windows is green with these added.

PR #4524, opened 2023-06-14
feat(guide-cue): component implementation COMPASS-6334

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/components/aggregation-side-panel/index.tsx
```diff
@@ -108,23 +103,12 @@ export const AggregationSidePanel = ({
...
         drag_and_drop: false,
       });
-      markCueVisited();
     },
-    [onSelectUseCase, markCueVisited]
+    [onSelectUseCase]
   );
 
-  // Hide guide cue when use-case card is dragged from the list
-  useDndMonitor({
```
mabaasit, line 117: this is now handled by dismissing the GuideCue if user does not engage with it

packages/compass-components/src/components/guide-cue/guide-cue-groups.ts
```diff
@@ -0,0 +1,11 @@
+// When adding data to groups, make it readonly to make
+// GroupName type strict.
+// export const GROUPS = [] as const; // of type {id: string, steps: number}
+export const GROUPS: Array<{ id: string; steps: number }> = [];
+
+export const GROUP_TO_STEPS = Object.fromEntries(
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  Object.entries(GROUPS).map(([_, val]) => [val.id, val.steps])
+);
```
gribnoysup, line 9: Super nit: instead of `Object.fromEntries`, a `Map` seems more appropriate here

packages/compass-components/src/components/guide-cue/guide-cue.tsx
```diff
@@ -0,0 +1,200 @@
...
+  | 'open'
+  | 'onPrimaryButtonClick'
+  | 'setOpen'
+> &
+  GroupAndStep & {
+    cueId: string;
+    trigger: ({
+      ref,
+    }: {
+      ref: React.MutableRefObject<T | null>;
```
gribnoysup, line 50: This should rather be `LegacyRef<T>` or at least `Ref<T>` to match React html elements ref types better (otherwise you'll struggle to pass it directly to the element)

packages/compass-components/src/components/guide-cue/guide-cue.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+      guideCueService.onCueIntersectionChange(
+        entry.isIntersecting,
+        cueData.cueId,
+        cueData.groupId
+      );
+
+      if (!entry.isIntersecting && isOpen) {
+        setIsOpen(false);
+      }
```
gribnoysup, line 86: You can remove this isOpen check from there I think (unless I'm missing something), this creates a dependency in this callback that can be avoided, meaning that your observer in the effect below is also not recreated everytime isOpen changes

gribnoysup (commented):
Haven't done a thorough look as this is a draft, but a few small things caught my eye. At a glance this is already looking pretty awesome 👏

packages/compass-aggregations/src/components/aggregation-side-panel/index.tsx
```diff
@@ -150,42 +134,38 @@ export const AggregationSidePanel = ({
...
+          <div className={contentStyles} data-testid="side-panel-content">
+            {filteredUseCases.map((useCase, index) => (
+              <div
+                key={index}
+                ref={(r) => {
+                  if (index === 0) {
+                    ref.current = r;
+                  }
+                }}
+              >
```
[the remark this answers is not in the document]
mabaasit: today, i woke up to this thought in the morning and was planning to make this change. thanks for suggesting it :)

packages/compass-components/src/components/guide-cue/guide-cue.tsx
mabaasit: nice catch. will fix it

packages/compass-components/src/components/guide-cue/guide-cue-groups.ts
```diff
@@ -0,0 +1,9 @@
+// When adding data to groups, make it readonly to make
+// GroupName type strict.
+// const GROUPS = [] as const; // of type {id: string, steps: number}
+const GROUPS: Array<{ id: string; steps: number }> = [];
+
+export const GROUP_STEPS_MAP = new Map<GroupName, number>();
+GROUPS.forEach(({ id, steps }) => GROUP_STEPS_MAP.set(id, steps));
```
gribnoysup, line 7: Just fyi: another way to do it, Map takes entries as an argument in constructor:

```suggestion
export const GROUP_STEPS_MAP = new Map<GroupName, number>(
  GROUPS.map(({ id, steps }) => [id, steps])
);
```

packages/compass-components/src/components/guide-cue/guide-cue-service.ts
```diff
@@ -0,0 +1,267 @@
...
+  private _activeCue: Cue | null = null;
+
+  constructor(private readonly _storage: GuideCueStorage) {
+    super();
+  }
+
+  addCue(cue: Omit<Cue, 'isVisited'>) {
+    const cueIndex = this.getCueIndex(cue.cueId, cue.groupId);
+    if (cueIndex !== -1) {
+      console.warn(`The Cue ${cue.cueId} is already `);
```
gribnoysup, line 63: is already added?

packages/compass-components/src/components/guide-cue/guide-cue.tsx
```diff
@@ -0,0 +1,222 @@
...
+    const node = refEl.current;
+    const observer = new IntersectionObserver(observerCallback, {
+      threshold: 0.5,
+    });
+    observer.observe(node);
+    return () => {
+      observer.unobserve(node);
+      observer.disconnect();
+    };
+  }, [refEl, observerCallback]);
```
gribnoysup, line 109:
```suggestion
  }, [observerCallback]);
```

(also if observerCallback is only used inside this effect, you can totally just move the definition there)

packages/compass-components/src/components/guide-cue/guide-cue.tsx
```diff
@@ -0,0 +1,222 @@
...
+    if (!groupId) {
+      return { cueId, step: 1 };
+    }
+    return { cueId, groupId, step };
+  }, [cueId, groupId, step]);
+
+  const observerCallback = useCallback(
+    (entries: IntersectionObserverEntry[]) => {
+      const [entry] = entries;
+      intersectionRef.current = entry.isIntersecting;
```
gribnoysup, line 81: I might be missing something here, but the way this intersectionRef is used in other hooks, you probably want it to be part of the component state, not just a ref? You use it for rendering, but nothing guarantees that the render will be triggered if this value is just a ref

packages/compass-components/src/components/guide-cue/guide-cue.tsx
```diff
@@ -0,0 +1,222 @@
...
+      return;
+    }
+    guideCueService.addCue({
+      ...cueData,
+      isIntersecting: intersectionRef.current,
+    });
+    return () => {
+      guideCueService.removeCue(cueData.cueId, cueData.groupId);
+    };
+  }, [refEl, cueData, intersectionRef]);
```
gribnoysup, line 136:
```suggestion
  }, [cueData]);
```

packages/compass-components/src/components/guide-cue/guide-cue.tsx
```diff
@@ -0,0 +1,222 @@
...
+    <>
+      {readyToRender && (
+        <LGGuideCue
+          {...restOfCueProps}
+          open={isOpen}
+          numberOfSteps={guideCueService.getCountOfSteps(cueData.groupId)}
+          onDismiss={onNextGroup}
+          onPrimaryButtonClick={onNext}
+          setOpen={() => {
+            // noop
```
gribnoysup, line 206: This is a noop because we don't want to give leafygreen any control over component rendering or something? Maybe worth adding a more detailed note here 🤔

gribnoysup (approved):
Overall looks very good, I have one question about `intersectionRef` usage that I think is worth clarifying

packages/compass-components/src/components/guide-cue/guide-cue-groups.ts
mabaasit: oh thanks :)

packages/compass-components/src/components/guide-cue/guide-cue.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+// omit the props we are handling
+export type GuideCueProps<T> = Omit<
+  LGGuideCueProps,
+  | 'currentStep'
+  | 'refEl'
+  | 'numberOfSteps'
+  | 'onDismiss'
+  | 'open'
+  | 'onPrimaryButtonClick'
```
gribnoysup, line 50: I was just trying to add a GuideCue for the signal popover and I think we might want these to be available props? In our case the pimary button says something like "See them in action" and should behave as if the insight was clicked

packages/compass-components/src/components/guide-cue/guide-cue.tsx
```diff
@@ -0,0 +1,222 @@
...
+          }}
+          currentStep={cueData.step || 1}
+          refEl={refEl}
+          data-cueid={getDataCueId(cueData)}
+          tooltipClassName={cx(
+            // Avoid flicker when the component (trigger) becomes invisible
+            !intersectionRef.current && hiddenPopoverStyles
+          )}
+        >
+          {children}
```
gribnoysup, line 216: Extremely small thing, just wondering what you'll think about that. So when adding the GuideCue for the signal popover (and it's actually the same in other places), one thing I found inconvenient is that the component definition looks somewhat like that:


```
cue title

...

trigger body

...

cue description
```

Whereas it seems to me that it would be easier to read if it would be something like that instead:

```
cue title
cue description

...

trigger body
```

I know that this is based on the leafygreen interface here, but maybe we can map their `children` prop to something like `description` prop in our component (and remove `children` prop from the interface)? What do you think? 🤔

packages/compass-components/src/components/guide-cue/guide-cue.tsx
mabaasit: good point. I moved it in there.

packages/compass-components/src/components/guide-cue/guide-cue.tsx
mabaasit: I changed it to be a component state in [8485e90](https://github.com/mongodb-js/compass/pull/4524/commits/8485e90021c490303e0e3e18779464015779c8aa). I had added it prior to adding dismissible behaviour on GuideCue

packages/compass-components/src/components/guide-cue/guide-cue.tsx
mabaasit: Yes. but now looking closely at it, i added it back here ([893a43f](https://github.com/mongodb-js/compass/pull/4524/commits/893a43fcac0a40fc3926ca80c3b8d866ed7c2d91)) and cleaned up onNext and onNextGroup.

packages/compass-components/src/components/guide-cue/guide-cue.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+export const GuideCue = <T extends HTMLElement>({
+  children,
+  trigger,
+  cueId,
+  groupId,
+  step,
+  ...restOfCueProps
+}: React.PropsWithChildren<GuideCueProps<T>>) => {
+  const [isOpen, setIsOpen] = useState(false);
```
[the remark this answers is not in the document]
mabaasit: I am a bit worried about that as it introduces a possibility of having multiple guide-cues open at the same time.

GuideCueService ensures that we only have one GC open at a single moment. If we introduce this flexibility in the component, we will have to hide (if any) active GC and then show this. And if that active GC is part of a group, that may mean to interrupt the group.

packages/compass-components/src/components/guide-cue/guide-cue.tsx
mabaasit: If it makes our workflow easier, yes totally! I don't have anything against it.

packages/compass-components/src/components/guide-cue/guide-cue.tsx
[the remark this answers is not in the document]
gribnoysup: Awesome! Let's do that then!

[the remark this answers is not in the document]
mabaasit: added in [e663fbc](https://github.com/mongodb-js/compass/pull/4524/commits/e663fbcd509f17267bb545622b2a1ef315b2cf7e)

gribnoysup (approved):
Amazing stuff!

PR #4528, opened 2023-06-14
fix(stage-wizard): button label

mabaasit:
### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


himanshusinghs (approved):

himanshusinghs (commented):
Ahh just noticed that it might fail the tests for `(aggregation side panel)` in `packages/compass-aggregations/src/components/aggregation-side-panel/index.spec.tsx`

packages/compass-aggregations/src/modules/side-panel.ts
```diff
@@ -1,5 +1,9 @@
 import type { AnyAction } from 'redux';
 import { isAction } from '../utils/is-action';
+import { type PipelineBuilderThunkAction } from '.';
```
himanshusinghs, line 3: Just a question, does this work the same way as:
```
import type { PipelineBuilderThunkAction } from '.'
```

himanshusinghs (approved):

packages/compass-aggregations/src/modules/side-panel.ts
mabaasit: In this case, yes. But with `import type {} from ''`, you can only import types and with `import {type A, B} from ''`, you can import both.

PR #4549, opened 2023-06-17
chore: automate beta and ga creation from CI COMPASS-6270

mcasimir:
- Switch the branching model to 2 branches: `beta-releases` and `ga-releases`.
- Creates a new beta at 5 AM UTC on Monday merging `main` to `beta-releases`, can be scheduled manually. Uses the latest know release in Jira to bump the version. 
- Adds a workflow to promote the latest beta to a GA (merge a beta branch). Uses the `fixVersion` of a release ticket passed as input to bump the version.

Covers quickfixes and releasing GAs from alternate branches through workflow inputs.


packages/notary-service-client/README.md
```diff
@@ -4,52 +4,6 @@
 
 ## Example
 
-## CLI
```
mcasimir, line 7: The notary service is going to be decommissioned in the upcoming months, didn't see a good reason to keep this here

.github/workflows/start-ga.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Install Dependencies
+        run: |
+          npm -v
+          npm ci
+
+      - name: Start Release
+        env:
+          JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
+        run: |
+          node scripts/release.js ga \
```
Anemy, line 46: Would it work to use typescript/ts-node to run this file? Any reason we don't want to? Then we could have the file typescript.

mcasimir: Sounds simple enough to try, i'm not sure `scripts` is setup for that, i'll check

mcasimir: yeah, scripts needs to be restructured and setup for that .. i'm not particularly keen in doing it in this PR, i'd rather do it in a separate PR, and then rebase

lerouxb (approved):
👍

PR #4595, opened 2023-06-28
fix: Make sure form fields' dropdown menus are positioned above buttons in other form fields when the window loses focus

lerouxb:
It fixes this problem:

eek

Open that selectbox's dropdown then open devtools so the main window area loses focus and then buttons from other fields end up sitting on top of the open dropdown.

### DIFF
--- a/packages/compass-components/src/components/form-field-container.tsx
+++ b/packages/compass-components/src/components/form-field-container.tsx
@@ -4,6 +4,10 @@ import React from 'react';
 
 const formFieldContainerStyles = css({
   margin: `${spacing[4]}px 0`,
+  // This is to prevent buttons from other form fields being positioned on top
+  // of a select dropdown menu in this form field when the window loses focus.
+  position: 'relative',
+  zIndex: 1,
 });
 
 function FormFieldContainer({
--- a/packages/compass-e2e-tests/helpers/commands/add-collection.ts
+++ b/packages/compass-e2e-tests/helpers/commands/add-collection.ts
@@ -1,3 +1,5 @@
+import _ from 'lodash';
+import type { Element } from 'webdriverio';
 import type { CompassBrowser } from '../compass-browser';
 import * as Selectors from '../selectors';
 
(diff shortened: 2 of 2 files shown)


packages/compass-components/src/components/form-field-container.tsx
```diff
@@ -4,6 +4,10 @@ import React from 'react';
 
 const formFieldContainerStyles = css({
   margin: `${spacing[4]}px 0`,
+  // This is to prevent buttons from other form fields being positioned on top
+  // of a select dropdown menu in this form field when the window loses focus.
```
Anemy, line 8: Is/was this a LeafyGreen issue? Might be worth seeing if the new versions (if we aren't on newest) behave the same. Looks like an e2e test is failing. Changes are fine by me. lgtm once green

lerouxb: I don't know.

packages/compass-e2e-tests/helpers/commands/add-collection.ts
```diff
@@ -70,15 +95,24 @@ export async function addCollection(
...
       await span.waitForDisplayed();
+      await waitForAnimations(browser, span as Element<'async'>);
+      await browser.screenshot(`custom-collation-${key}.png`);
       await span.click();
+      await menu.waitForDisplayed({ reverse: true });
+
+      const button = await browser.$(
+        Selectors.createCollectionCustomCollationFieldButton(key)
+      );
+      console.log({ [key]: await button.getText() });
```
Anemy, line 115:
```suggestion
```

packages/compass-e2e-tests/helpers/commands/add-collection.ts
```diff
@@ -31,6 +33,29 @@ export type AddCollectionOptions = {
...
+
+    const result = {
+      ...(await element.getLocation()),
+      ...(await element.getSize()),
+    };
+    const stopped = _.isEqual(result, previousResult);
+    previousResult = result;
+    return stopped;
+  });
+}
```
Anemy, line 57: Do we want to move this into the file https://github.com/mongodb-js/compass/blob/93307133a16a7320bd205cd04424424bbbadaaa1/packages/compass-e2e-tests/helpers/commands/wait-for-animations.ts
Maybe we can reduce the duplication there too.

PR #4600, opened 2023-06-29
chore(storage-mixin): remove from preferences storage COMPASS-6946

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-preferences-model/src/preferences.ts
```diff
@@ -119,10 +104,11 @@ type PostProcessFunction<T> = (
...
 type PreferenceDefinition<K extends keyof AllPreferences> = {
-  /** The type of the preference value, in Ampersand naming */
-  type: AmpersandType<AllPreferences[K]>;
+  /** The type of the preference value */
+  type: PreferenceValue<AllPreferences[K]>;
   /** An optional default value for the preference */
   default?: AllPreferences[K];
+  // todo: do we need it
   /** Whether the preference is required in the Ampersand model */
   required: boolean;
```
mabaasit, line 113: Its safe to remove this property now?

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -807,28 +793,18 @@ export class Preferences {
...
+    const defaultPreferences = Object.fromEntries(
+      Object.entries(defaultUserPreferencesProps)
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        .filter(([_key, value]) => value.default !== undefined)
+        .map(([key, value]) => [key, value.default])
+    ) as UserPreferences;
+
+    this._preferencesStorage = isSandbox
+      ? new SandboxPreferences(defaultPreferences)
+      : new StoragePreferences(defaultPreferences, basepath);
```
mabaasit, line 805: I kept the implementation of these two separate.

packages/compass-preferences-model/src/storage.ts
```diff
@@ -0,0 +1,97 @@
...
+    return join(this.getFolderPath(), this.file);
+  }
+
+  async setup() {
+    // Ensure folder exists
+    await fs.mkdir(this.getFolderPath(), { recursive: true });
+
+    try {
+      this.preferences = JSON.parse(
+        (await fs.readFile(this.getFilePath())).toString()
```
addaleax, line 67:
```suggestion
        await fs.readFile(this.getFilePath(), 'utf8')
```

packages/compass-preferences-model/src/preferences.spec.ts
```diff
@@ -56,57 +64,35 @@ describe('Preferences class', function () {
...
-    const preferences = new Preferences(tmpdir);
+    const preferences = await setupPreferences(tmpdir);
     const calls: any[] = [];
     preferences.onPreferencesChanged((prefs) => calls.push(prefs));
     await preferences.savePreferences({ enableMaps: true });
     expect(calls).to.deep.equal([{ enableMaps: true }]);
   });
 
-  it('notifies callers of preferences changes after fetchPreferences', async function () {
+  it.skip('handles concurrent modifications to different preferences', async function () {
```
addaleax, line 83: Just leaving a comment here about the .skip()

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -119,10 +104,11 @@ type PostProcessFunction<T> = (
 ) => T;
 
 type PreferenceDefinition<K extends keyof AllPreferences> = {
-  /** The type of the preference value, in Ampersand naming */
-  type: AmpersandType<AllPreferences[K]>;
+  /** The type of the preference value */
+  type: PreferenceValue<AllPreferences[K]>;
```
addaleax, line 108: Did Ampersand do validation of the type? I think it did, that's missing now, right?

packages/compass-preferences-model/src/preferences.ts
addaleax: It's similar to `type`, `default`, and `values`, right? They were all used by Ampersand for validation, that's a featur that's going away here

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -278,7 +264,7 @@ const allFeatureFlagsProps: Required<{
   ...featureFlagsProps,
 };
 
-const modelPreferencesProps: Required<{
+const defaultUserPreferencesProps: Required<{
```
addaleax, line 267: I'm not quite sure I follow the name here ... :) These are the ones that are stored, right, not defaults?

packages/compass-preferences-model/package.json
```diff
@@ -47,13 +47,12 @@
...
     "@mongodb-js/compass-utils": "^0.3.1",
     "ampersand-collection-filterable": "^0.3.0",
-    "ampersand-model": "^8.0.1",
     "ampersand-rest-collection": "^6.0.0",
     "ampersand-state": "5.0.3",
     "bson": "^5.2.0",
+    "joi": "^17.9.2",
     "js-yaml": "^4.1.0",
     "lodash": "^4.17.21",
-    "storage-mixin": "^5.1.5",
```
addaleax, line 56: I never realized that this was just not namespaced in any way, not even with the name referencing Ampersand or something like that :)

packages/compass-preferences-model/src/global-config.ts
```diff
@@ -91,11 +91,11 @@ async function loadGlobalPreferences(
 const cliProps = Object.entries(allPreferencesProps).filter(
   ([, definition]) => definition.cli
 );
-function getCliPropNamesByType(type: AmpersandType<any>): string[] {
+function getCliPropNamesByType(type: unknown): string[] {
```
addaleax, line 94: This type could be `Types` from Joi, I think?

```suggestion
function getCliPropNamesByType(type: Types): string[] {
```

packages/compass-preferences-model/src/global-config.ts
```diff
@@ -170,30 +170,15 @@ function validatePreferences(
...
-      !(allPreferencesProps[key].values as unknown[])?.includes(value)
-    ) {
-      error(
-        `Value for option "${key}" is not allowed: expected one of [${String(
-          allPreferencesProps[key].values?.join(', ')
-        )}], received ${String(value)}`
-      );
+
+    const validationResults =
+      allPreferencesProps[key].validator.validate(value);
```
addaleax, line 175: :blue_heart:

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -227,6 +189,7 @@ function featureFlagToPreferenceDefinition(
       featureFlag.stage === 'released'
         ? () => ({ value: true, state: 'hardcoded' })
         : undefined,
+    validator: Joi.boolean().default(false),
```
addaleax, line 192: For my own understanding: `.default()` implies `.optional()`, right?

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -579,51 +499,48 @@ const modelPreferencesProps: Required<{
       long: 'Force connection string properties to take specific values',
     },
     customPostProcess: parseRecord,
+    validator: Joi.array()
+      .items(Joi.array().items(Joi.string(), Joi.string()))
```
addaleax, line 503: I think if you want tuples, this is how you’d do it:

```suggestion
      .items(Joi.array().length(2).items(Joi.string()))
```

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -917,30 +788,30 @@ export class Preferences {
...
-      userPreferencesModel.save.bind(userPreferencesModel)
-    );
+    keys.forEach((key) => {
+      const { error } = storedUserPreferencesProps[key].validator.validate(
+        attributes[key]
+      );
+      if (error) {
+        throw error;
+      }
+    });
```
addaleax, line 811: for-of should work fine, if you want:

```suggestion
    for (const key of keys) {
      const { error } = storedUserPreferencesProps[key].validator.validate(
        attributes[key]
      );
      if (error) {
        throw error;
      }
    }
```

packages/compass-preferences-model/src/preferences.ts
```diff
@@ -917,30 +788,30 @@ export class Preferences {
     attributes: Partial<UserPreferences> = {}
   ): Promise<AllPreferences> {
     const keys = Object.keys(attributes) as (keyof UserPreferences)[];
-    const originalPreferences = await this.fetchPreferences();
+    const originalPreferences = this.getPreferences();
```
addaleax, line 791: Just to mention it explicitly and make sure it’s intentional, this means that if a user has two Compass instances running and changes preferences first in one and then the other, now that second write always undoes the first one, rather than merging the updates

packages/compass-preferences-model/src/storage.ts
```diff
@@ -0,0 +1,93 @@
+import { promises as fs } from 'fs';
+import { join } from 'path';
+import type { UserPreferences } from './preferences';
+
+const ENCODING_UTF8 = 'utf8';
```
addaleax, line 5: This can just be inlined, no?

packages/compass-preferences-model/src/storage.ts
```diff
@@ -0,0 +1,93 @@
...
+    return join(this.getFolderPath(), this.file);
+  }
+
+  async setup() {
+    // Ensure folder exists
+    await fs.mkdir(this.getFolderPath(), { recursive: true });
+
+    try {
+      this.preferences = JSON.parse(
+        (await fs.readFile(this.getFilePath(), ENCODING_UTF8)).toString()
```
addaleax, line 61:
```suggestion
        (await fs.readFile(this.getFilePath(), ENCODING_UTF8))
```

or

```suggestion
        (await fs.readFile(this.getFilePath(), 'utf8'))
```

(Typescript should be aware of the different encodings and know that calling `readFile` with `'utf8'` will return a string)

packages/compass-preferences-model/src/preferences.ts
mabaasit: Yes. it does.

packages/compass-preferences-model/src/storage.ts
mabaasit: Yeah, it can be! Changing that to inline.

packages/compass-preferences-model/src/preferences.ts
mabaasit: Oh! This seems like a breaking change i am introducing. I think I need to align this and avoid any unwanted behaviour.

mabaasit: While going through the changes, I stumbled on [this page](https://www.electronjs.org/docs/latest/tutorial/process-model#the-main-process) and realized that electron has a single main process and all the browser windows share it. So any change made in any Compass window should reflect on other and should not be out of sync. Plus I also tested it and works just fine. Let me know what you think about it.

lerouxb: Two instances of the app or two windows in the same app?

mabaasit: multiple windows within the same app.

addaleax: @mabaasit Right, that’s how it works on macOS or when you use the “New Connection” button. On other OSes, starting Compass multiple times through the regular mechanism for doing so will create actual different Compass main processes.

Again, no particularly strong feelings here and this is still a change in behavior that we can make, we should just be intentional about it.

mabaasit: added [2e12f02](https://github.com/mongodb-js/compass/pull/4600/commits/2e12f020dfece0322a29a6d91be9eed47e30ec25) to address this. 

However, when we have two different compass instances and user updates any preferences it will not be reflected in other one unless some user action triggers fetching of preferences (which is the current behaviour).

addaleax (approved):

PR #4602, opened 2023-06-29
fix(compass-components): case sensitive index key badges COMPASS-6510

Anemy:
COMPASS-6510

As react-testing-library only tests the dom, not the rendered items, I'm not sure what a good regression test would be besides adding onto our end to end tests. Will add one if folks feel it's worth it.

| before | after |
| - | - |
| Screenshot 2023-06-29 at 10 30 55 AM | Screenshot 2023-06-29 at 10 30 44 AM |

### DIFF
--- a/packages/compass-components/src/components/index-keys-badge.tsx
+++ b/packages/compass-components/src/components/index-keys-badge.tsx
@@ -11,6 +11,7 @@ const keyListStyles = css({
 });
 
 const badgeStyles = css({
+  // Override LeafyGreen's uppercase styles as we want to keep the case sensitivity of the key.
   textTransform: 'none',
   gap: spacing[1],
 });
--- a/packages/compass-components/src/components/index-keys-badge.tsx
+++ b/packages/compass-components/src/components/index-keys-badge.tsx
@@ -11,6 +11,7 @@ const keyListStyles = css({
 });
 
 const badgeStyles = css({
+  textTransform: 'none',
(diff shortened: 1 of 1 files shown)


addaleax (approved):

packages/compass-components/src/components/index-keys-badge.tsx
```diff
@@ -11,6 +11,8 @@ const keyListStyles = css({
 });
 
 const badgeStyles = css({
+  // Override LeafyGreen's uppercase styles as we want to keep the case sensitivity of the key.
+  textTransform: 'none',
```
addaleax, line 15: Is there a ticket for a proper prop that we can pass to the LG component? I think it might be good to link that here (and create one if it doesn’t currently exist)

lerouxb (approved):

packages/compass-components/src/components/index-keys-badge.tsx
Anemy: Asked in designs https://mongodb.slack.com/archives/GDTJXPHD0/p1688140011504569?thread_ts=1688049533.185989&cid=GDTJXPHD0
Good q, I have a feeling we probably shouldn't be using the LeafyGreen `Badge` here, and rather do something similar in our own code. Going to merge this so we get it into the next beta/ga, will follow up and add a todo with a ticket when there's a response. (Or will implement our own kind of Badge here).

PR #4619, opened 2023-07-05
feat(compass-query-bar): pass schema and sample docs to ai, basic auth COMPASS-6979

Anemy:
COMPASS-6979
Adds the schema, sample documents, collection name, and basic auth to the ai query request.

Uses the `getSimplifiedSchema` helper in `mongodb-schema` we added in https://github.com/mongodb-js/mongodb-schema/pull/198 (hence the `mongodb-schema` version bump).

We send 4 documents currently. If, when stringified, that's over the current character threshold (arbitrary 5000) we try to only send one, if that's over then we error.

https://github.com/mongodb-js/compass/assets/1791149/8e50fd27-fa7b-4699-82f7-f23a15fe514a


packages/compass-query-bar/src/modules/ai-query-request.ts
```diff
@@ -13,25 +15,69 @@ function getAIQueryEndpoint(): string {
...
+  if (msgBody.length > MAX_REQUEST_SIZE) {
+    // When the message body is over the max size, we try
+    // to see if with fewer sample documents we can still perform the request.
+    // If that fails we throw an error indicating this collection's
+    // documents are too large to send to the ai.
+    msgBody = JSON.stringify({
+      userPrompt,
+      collectionName,
+      schema,
+      sampleDocuments: sampleDocuments?.slice(0, MIN_SAMPLE_DOCUMENTS),
```
addaleax, line 62: Do we want to iteratively reduce the # of sampleDocuments instead of always going from the full list to a single item? Or do we think that it probably doesn’t matter?

Anemy: I think it probably doesn't matter, at least for now. As we start to learn more about how we can best prompt the ai we can update it if we feel it would help.

alenakhineika (approved):
LGTM ✨

PR #4629, opened 2023-07-10
chore: use @mongosh-js/eslint-config-devtools@0.9.7, ignore eqeqeq and noconsole errors

lerouxb:


packages/compass-components/src/components/combobox/Combobox.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable filename-rules/match */
```
lerouxb, line 1: There's something up with linting in compass-components, btw. It complains about this and the leafygreen imports in vscode, but clearly not in `npm run check`. But also doesn't complain about the type imports and a bunch of things below - only in vscode and (possibly) if you run eslint directly in the package. Seems to fail in such a way that it passes.

addaleax (approved):

packages/compass-components/src/components/combobox/Combobox.tsx
```diff
@@ -189,7 +192,7 @@ export function Combobox<M extends boolean>({
    */
   const isMultiselect = useCallback(
     <T extends string>(val?: Array<T> | T | null): val is Array<T> => {
-      if (multiselect && (typeof val == 'string' || typeof val == 'number')) {
+      if (multiselect && (typeof val === 'string' || typeof val === 'number')) {
```
lerouxb, line 195: I changed comparisons to strings like this to `===` but left comparisons to `null` in place and just ignored it. Hard to tell at scale what people meant with each one.

packages/compass-components/src/components/combobox/Combobox.tsx
```diff
@@ -245,9 +248,7 @@ export function Combobox<M extends boolean>({
           }
         }
         setSelection(newSelection as SelectValueType<M>);
-        (onChange as onChangeType<true>)?.(
-          newSelection as SelectValueType<true>
-        );
+        (onChange as onChangeType<true>)?.(newSelection);
```
lerouxb, line 251: I ran --fix on this file to get the type imports thing to go away and it might have gone too far and removed some type assertions that may or may not have been needed. We'll see. I've been bitten by this before.

packages/compass-aggregations/src/utils/get-directory.js
```diff
@@ -4,6 +4,7 @@ let electronApp;
 try {
   electronApp = require('@electron/remote').app;
 } catch (e) {
+  // eslint-disable-next-line no-console
```
lerouxb, line 7: For most files where there are only one or two console lines I just disabled the line. For files with a bunch (often tests) I disabled it for the whole file. For packages that exist basically to console.log (scripts, build packages) I just disabled the rule in its eslint rc file.

packages/compass-components/src/components/combobox/ComboboxOption.tsx
```diff
@@ -335,6 +337,7 @@ const InternalComboboxOption = React.forwardRef<
 InternalComboboxOption.displayName = 'ComboboxOption';
 
 export { InternalComboboxOption };
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
lerouxb, line 340: Again - no idea why nothing caught this before.

PR #4647, opened 2023-07-17
chore(storage-mixin): connections from storage COMPASS-6948 	COMPASS-6949

mabaasit:
In this PR, we are removing the amperand and storage-mixin from connections and are loading/saving connections directly to the filesystem. 

It also removes the `connection-model` package and a result obsolete migrations from compass (which ran in the renderer process).

We also dropped support for legacy connections (connections created using compass < v1.29). Users having legacy connections will see a toast in Compass directing them to migrate those connections.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/connection-storage/src/connection-storage.spec.ts
```diff
@@ -285,55 +189,29 @@ describe('ConnectionStorage', function () {
       expect(error.message).to.be.equal('id must be a uuid');
     });
 
-    it('does not save fle secrets if fleOptions.storeCredentials is false', async function () {
```
mabaasit, line 288: Moved these fle tests to `connection-secrets.spec.ts` as they tested `extractSecrets` and were not relevant here

packages/compass-connections/src/components/connections.tsx
```diff
@@ -122,6 +172,22 @@ function Connections({
     []
   );
 
+  // Handle Legacy Connections toast.
+  const [hasSeenLegacyToast, setHasSeenLegacyToast] = useState(false);
+  useEffect(() => {
+    void connectionStorage
```
addaleax, line 178:
```suggestion
    if (hasSeenLegacyToast) return;
    void connectionStorage
```

scripts/import-test-connections.js
```diff
@@ -282,7 +276,7 @@ async function main() {
       );
       await fs.writeFile(
         path.join(destDir, `${connectionInfoWithId.id}.json`),
-        JSON.stringify(modelWithoutSecrets)
+        JSON.stringify(connectionInfoWithoutSecrets)
```
addaleax, line 279: Shouldn’t the saved JSON objects have a single top-level `connectionInfo` key, not be the ConnectionInfo object itself?

packages/connection-storage/src/connection-storage.ts
```diff
@@ -1,67 +1,123 @@
...
+    await fs.mkdir(this.getFolderPath(), { recursive: true });
 
     try {
-      await fetchConnectionModels();
+      const connections = (await this.getConnections())
+        // Ignore legacy connections and make sure connection has a connection string.
+        .filter(
+          (x: { connectionInfo?: ConnectionInfo }) =>
+            x.connectionInfo &&
+            x.connectionInfo.connectionOptions.connectionString
```
addaleax, line 93:
```suggestion
            x.connectionInfo?.connectionOptions?.connectionString
```

packages/connection-storage/src/connection-storage.ts
```diff
@@ -1,67 +1,123 @@
...
+        )
+        .map((x) => x.connectionInfo) as ConnectionInfo[];
+
+      if (process.env.COMPASS_E2E_DISABLE_KEYCHAIN_USAGE === 'true') {
+        return connections.map((connection) =>
+          this.mapStoredConnectionToConnectionInfo(connection)
+        );
+      }
+
+      const secrets = await this.getKeytarCredentials();
```
addaleax, line 103: Probably not a big deal, but does it make sense to perform `this.getKeytarCredentials()` and the `this.getConnections()` call above in parallel?

mabaasit: that's a good idea.

scripts/import-test-connections.js
mabaasit: Yes, added in [3d6710c](https://github.com/mongodb-js/compass/pull/4647/commits/3d6710cade3deab3d9602b6ce07c32507f66dba3)

.evergreen/functions.yml
```diff
@@ -169,6 +169,13 @@ functions:
           set -e
           eval $(.evergreen/print-compass-env.sh)
 
+          # With removal of storage-mixin and top level import of 
+          # keytar, we need to rebuild the keytar.
```
addaleax, line 173: I’m not sure that the connection to storage-mixin would be obvious here to future readers

```suggestion
          # Pre-built binaries for keytar do not support older RHEL versions.
          # Since we have started importing keytar from top-level scripts in packages,
          # re-built keytar before the bootstrap process so that it is always usable.
```

That being said, if you want to go back to the previous approach and dynamically load keytar at runtime, I don’t think that that’s a bad choice either.

addaleax (approved):
Looks great, thank you for your patience! :shipit:

.evergreen/functions.yml
mabaasit: I was thinking about that approach as well and was unsure. As we will be removing `keytar` now, I think we can leave it as is and probably remove at that point.

PR #4654, opened 2023-07-18
fix: strip unknown preferences when loading COMPASS-7026

lerouxb:


packages/compass-preferences-model/src/storage.spec.ts
```diff
@@ -41,4 +42,28 @@ describe('storage', function () {
...
+    });
+
+    it('will strip unknown prefs', async function () {
+      const prefsStorage = new StoragePreferences(
+        { showedNetworkOptIn: true, foo: true } as unknown as UserPreferences,
+        tmpDir
+      );
+      await prefsStorage.setup();
+
+      // roundabout way to make it load because setup doesn't fill prefsStorage.preferences if it writes the file
```
lerouxb, line 62: This test is a bit hacky and reliant on knowledge of internals. I think this whole class could benefit from lots more unit tests that would hopefully hilight that sort of thing.

packages/compass-preferences-model/src/storage.ts
```diff
@@ -71,7 +73,19 @@ export class StoragePreferences extends BasePreferencesStorage {
   }
 
   private async readPreferences(): Promise<UserPreferences> {
-    return JSON.parse(await fs.readFile(this.getFilePath(), 'utf8'));
+    const preferences = JSON.parse(
+      await fs.readFile(this.getFilePath(), 'utf8')
+    );
+
+    // Exclude old and renamed preferences so we don't try and save those back.
```
lerouxb, line 80: I think the problem here is that we're passing off anything that will JSON.parse() without throwing as valid UserPreferences where it could be `[]` or `{}` or partial UserPreferences or a superset or..

I now strip unknown preferences (think preferences we removed or renamed since last time the user's preferences were saved to disk), but we could have also _added_ preferences since then.

(To not even mention edge cases where someone could have edited the file manually.)

But I tried to keep this fix PR small.

himanshusinghs (approved):

gribnoysup (approved):

PR #4669, opened 2023-07-25
chore: fixes inaccurate detection of fake mongodb deployments - COMPASS-7045

himanshusinghs:
## Description
A follow up of https://github.com/mongodb-js/mongodb-build-info/pull/22, which brings updated logic for fake deployment detection to Compass.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/data-service/src/instance-detail-helper.ts
+++ b/packages/data-service/src/instance-detail-helper.ts
@@ -15,7 +15,6 @@ import toNS from 'mongodb-ns';
 import type {
   AtlasVersionInfo,
   BuildInfo,
-  CmdLineOpts,
   CollectionInfo,
   CollectionInfoNameOnly,
   ConnectionStatusWithPrivileges,
@@ -108,12 +107,12 @@ export type InstanceDetails = {
 };
 
 export async function getInstance(
-  client: MongoClient
+  client: MongoClient,
+  uri: string
 ): Promise<Omit<InstanceDetails, 'csfleMode'>> {
   const adminDb = client.db('admin');
   const [
     connectionStatus,
-    getCmdLineOptsResult,
     hostInfoResult,
     buildInfoResult,
     getParameterResult,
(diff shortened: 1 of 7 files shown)


packages/data-service/src/instance-detail-helper.ts
```diff
@@ -108,12 +107,12 @@ export type InstanceDetails = {
 };
 
 export async function getInstance(
-  client: MongoClient
+  client: MongoClient,
+  uri: string
```
himanshusinghs, line 111: Is there a better way for me to get the connection uri here? MongoClient is in scope but does not provide an accessor to fetch the connection string.

packages/data-service/test/helpers.ts
```diff
@@ -92,5 +95,19 @@ export function createMongoClientMock({
...
+  const connectionString = hosts.reduce<string>(
+    (connectionString, { host, port }, index) => {
+      const hostWithPort = port ? `${host}:${port}` : host;
+      if (index === 0) {
+        return `${connectionString}://${hostWithPort}`;
+      }
+      return `${connectionString},${hostWithPort}`;
+    },
+    'mongodb'
+  );
```
addaleax, line 107: Do you just want to use the `ConnectionString` class from the `mongodb-connection-string-url` package here? Here, you'd technically have to worry about things like escaping characters in `host` and `port` as well (e.g. for IPv6 hosts)

packages/data-service/src/instance-detail-helper.ts
addaleax: I think `MongoClient`s only store the parsed connection string options, not the original string, unfortunately 😐

addaleax (approved):

packages/data-service/test/helpers.ts
himanshusinghs: yea that is right. I will make the adjustments.

PR #4671, opened 2023-07-25
chore(query-bar, atlas-service): update ai query prompt creation parameters COMPASS-6990

Anemy:
COMPASS-6990

We are now passing the `databaseName` and not passing the `sampleDocuments`.

### DIFF
--- a/packages/atlas-service/src/main.spec.ts
+++ b/packages/atlas-service/src/main.spec.ts
@@ -109,6 +109,7 @@ describe('AtlasServiceMain', function () {
         userPrompt: 'test',
         signal: new AbortController().signal,
         collectionName: 'jam',
+        databaseName: 'peanut',
         schema: { _id: { types: [{ bsonType: 'ObjectId' }] } },
         sampleDocuments: [{ _id: 1234 }],
       });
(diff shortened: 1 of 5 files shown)


packages/compass-query-bar/src/stores/ai-query-reducer.ts
```diff
@@ -189,7 +192,8 @@ export const runAIQuery = (
         );
       }
 
-      const query = jsonResponse?.content?.query;
+      const query = EJSON.deserialize(jsonResponse?.content?.query);
```
mcasimir, line 195: how it worked without this before?

packages/atlas-service/src/main.spec.ts
```diff
@@ -147,7 +149,8 @@ describe('AtlasServiceMain', function () {
       try {
         await AtlasService.getQueryFromUserPrompt({
           userPrompt: 'test',
-          collectionName: 'test.test',
+          collectionName: 'test',
+          databaseName: 'peanut',
           sampleDocuments: [{ test: '4'.repeat(60000) }],
```
alenakhineika, line 154: Should we remove sampleDocuments as we were recently discussing?

Anemy: The caller of this function doesn't pass the `sampleDocuments` currently, and we have a regression test to make sure they aren't passed: https://github.com/mongodb-js/compass/pull/4671/files#diff-4cd0f8f72aa57907ac7a7876d0f2c423e04fe5e78d19e043e88aeba548ec1ebdR72
So I think it's fine to leave this part here as we might add it back in for some testing later on. For the same reason we left the sample documents code in the backend.
If you feel strongly about removing them from this function's interface I'm happy to.

alenakhineika: Since we have tests for this, I think it is fine to keep it here.

alenakhineika (approved):

PR #4674, opened 2023-07-26
chore(pipeline-storage): inject basepath as dependency

mabaasit:
In this PR, we aligned PipelineStorage with other storage implementations by injecting basepath as a dependency.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/utils/pipeline-storage.spec.js
```diff
@@ -1,35 +1,28 @@
...
 
-const pipelineStorage = new PipelineStorage();
-
 describe('PipelineStorage', function () {
   let tmpDir;
+  let pipelineStorage;
   beforeEach(function () {
-    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'SavedPipelines'));
+    tmpDir = fs.mkdtempSync(os.tmpdir());
     fs.mkdirSync(path.join(tmpDir, 'SavedPipelines'));
```
mcasimir, line 20: I would expect ensuring the path exists being done by the storages

packages/compass-aggregations/src/stores/store.ts
```diff
@@ -156,7 +159,7 @@ const configureStore = (options: ConfigureStoreOptions) => {
     options.dataProvider.dataProvider as DataService,
     initialPipelineSource
   );
-  const pipelineStorage = new PipelineStorage();
+  const pipelineStorage = new PipelineStorage(basepath);
```
mcasimir, line 162: why is this a better alternative than having it inside the storages? if the reason is testing you can still have a default storage path and have the constructor accept a param to override it.

in this way you are giving a "passing knowledge" to `store.ts` of where the files are saved, which store.ts can't use directly.

packages/compass-aggregations/src/utils/pipeline-storage.spec.js
mabaasit: I was thinking of doing all the directory setup as the app starts in the main process somewhere [here](https://github.com/mongodb-js/compass/blob/main/packages/compass/src/main/application.ts#L60). That way we will ensure we have all the necessary directories setup when we try to read from them.

packages/compass-aggregations/src/stores/store.ts
mabaasit: Yes, I did it for that reason (testability). I do see your point about sharing knowledge in `store.ts`, so I don't have a strong preference here.

packages/compass-aggregations/src/utils/pipeline-storage.spec.js
mcasimir: nice idea! but similar to the other comment, it seems that storages have the info on what's the folder: https://github.com/mongodb-js/compass/pull/4674/files/e7050dd7b6e504b4840c150e6cde7b52d0a4d1b1#diff-cda704877ab6cd60ae17d533f713bd373be3a3458105773b5a06e290f25e997fR68

So the storage would still be the "information expert" here. By requiring `_init` to run `mkdir` we would need to give the `_init` code a piece of knowledge that storages already have, and introduce a stronger coupling: where we want to save files and how (on a filesystem vs somewhere else). 

We could consider adding a `.setup()` method instead and call that in the init code .. that way we would have a fail-fast type of experience where if is not possible to write on a directory .. compass would crash??

Or, to get rid of the coupling entirely .. just use a lazy initialization, and have mkdir being called once if it wasn't before. I would prefer this way if that's what is happening right now, so we don't incur in the risk of having new issues that we didn't have before.

packages/compass-aggregations/src/stores/store.ts
mcasimir: Yeah, btw having a way to change the base directory in the constructor is a great idea, we should do it! If you are ok with it, let's also keep the default in the storages so we make it both testable and we won't leak details in the client code.

mabaasit:
I am closing this PR as we agreed to:
1. Have optional `options` param across storage classes where we will have `basepath`
2. Ensure storage folder (PipelineStorage, FavoriteQuery ...) exist in the storage class (this avoids weird setup in tests as well)
3. Refactor common code to access files

PR #4690, opened 2023-08-01
chore(storage): handle max items in storage COMPASS-6982

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-query-bar/src/utils/query-storage.spec.ts
```diff
@@ -96,10 +106,52 @@ describe('QueryStorage', function () {
...
 
     await queryHistoryStorage.delete(queries[1]._id);
     expect(await queryHistoryStorage.loadAll()).to.have.lengthOf(0);
   });
+
+  context(
+    `limits saved recent queries to ${maxAllowedRecentQueries}`,
+    function () {
+      it('truncates recents queries to max allowed queries', async function () {
+        const queries = createNumberOfQueries(tmpDir, maxAllowedRecentQueries);
```
kmruiz, line 122: What do you think on doing 

```suggestion
        const queries = createNumberOfQueries(tmpDir, maxAllowedRecentQueries + 1);
```

If we use the maximum here, if I understand well the behaviour, we won't test the use case where we surpass the limit.

packages/connection-storage/src/connection-storage.spec.ts
```diff
@@ -230,6 +232,55 @@ describe('ConnectionStorage', function () {
...
+        connectionInfos.forEach((connectionInfo) =>
+          writeFakeConnection(tmpDir, { connectionInfo })
+        );
+
+        return connectionInfos;
+      };
+
+      it('truncates recents to max allowed connections', async function () {
+        const connectionInfos = createNumberOfConnections(
+          maxAllowedConnections
```
kmruiz, line 252: I would suggest to do the same things as earlier in the maximum number of queries:

```suggestion
          maxAllowedConnections + 1
```

To force surpassing the limit. What do you think?

packages/connection-storage/src/connection-storage.spec.ts
```diff
@@ -230,6 +232,55 @@ describe('ConnectionStorage', function () {
...
+
+        expect(
+          deleteSpy.calledOnceWithExactly(
+            connectionInfos[connectionInfos.length - 1].id
+          )
+        ).to.be.true;
+      });
+
+      it('does not remove recent if recent connections are less then max allowed connections', async function () {
+        createNumberOfConnections(8);
```
kmruiz, line 271: I'm being picky here, but I would use maxAllowedConnections - 1, instead of 8, just in case we change the variable we don't break the test.

Unless we want the test to fail if the maximum number of connections allowed is less than 8.

```suggestion
        createNumberOfConnections(maxAllowedConnections - 1);
```

kmruiz (commented):
Just added a few comments. LGTM, but as I'm not yet really experienced in Compass I'll let someone else approve the PR.

Great work!

packages/compass-query-bar/src/utils/query-storage.spec.ts
mabaasit: Here I create the max queries and then after this I create another one using `queryHistoryStorage.saveQuery`. This method handles the max connections by removing any exceeding connection. If you check the assertions after this line, it asserts that max connections is same as `maxAllowedRecentQueries`

packages/connection-storage/src/connection-storage.spec.ts
mabaasit: Isn't that the point of tests - any change in code should break it? Let's say if we change `maxAllowedConnections` to 5, it should break the test and highlight that I should update tests as well.

packages/connection-storage/src/connection-storage.ts
```diff
@@ -143,39 +144,41 @@ export class ConnectionStorage {
...
+            JSON.stringify({ secrets }, null, 2)
+          )
+          .catch((e) => {
+            log.error(
+              mongoLogId(1_001_000_202),
+              'Connection Storage',
+              'Failed to save secrets to keychain',
+              { message: (e as Error).message }
+            );
+          });
```
addaleax, line 179: [optional, I know this was already here before] This can just be a regular try/catch, right?

packages/connection-storage/src/connection-storage.spec.ts
kmruiz: I do agree, I don't have a strong opinion on this specific scenario. I was being picky as it will fail under some changes, not all of them. For example, if you change maxAllowedConnections to something greater than 10, the test will still work. It will only fail if it's lower than 8.

But I don't think this is going to change much so I would keep it like that. The test is awesome, fast and gives good feedback.

mabaasit: well that's a fair point. let me change it :)

packages/connection-storage/src/connection-storage.ts
mabaasit: added here [`a682345` (#4690)](https://github.com/mongodb-js/compass/pull/4690/commits/a68234595840fb4e7bfc1452027b4dcc3953e1b8)

packages/connection-storage/src/connection-storage.spec.ts
mabaasit: added here [`a682345` (#4690)](https://github.com/mongodb-js/compass/pull/4690/commits/a68234595840fb4e7bfc1452027b4dcc3953e1b8)

kmruiz (approved):

PR #4695, opened 2023-08-02
feat(connections): use modal to highlight legacy connections COMPASS-7072

mabaasit:
In this PR, we added a modal to highlight **favorite** legacy connections of a user and removed toast. We also added an option to hide the modal permanently for a user to avoid modal showing up as user reloads or re-opens Compass.

image

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass/src/main/application.ts
```diff
@@ -97,7 +99,7 @@ class CompassApplication {
     this.setupLifecycleListeners();
     this.setupApplicationMenu();
     this.setupWindowManager();
-    this.trackApplicationLaunched(globalPreferences);
+    void this.trackApplicationLaunched(globalPreferences);
```
mcasimir, line 102: This goes from having a a function that can never throw to one that potentially can throw and the rejection would be unhandled, are we sure?

packages/compass-components/src/components/modals/modal-header.tsx
```diff
@@ -57,32 +46,62 @@ const warningIconStyles = css({
   },
 });
 
-const warningIconStylesDark = css({
-  background: `${palette.red.dark2}`,
-});
+const variantLightStyles = {
```
mcasimir, line 49: Some thoughts: what are we trying to do here and for which purpose? Are the current components and styles not enough for this niche use-case? Are we sure we shouldn't negotiate a compromise with design?

packages/compass-components/src/components/modals/modal-header.tsx
```diff
@@ -57,32 +46,62 @@ const warningIconStyles = css({
...
+  [ModalVariant.Danger]: css({
+    background: `${palette.red.light3}`,
```
mcasimir, line 51: Does this change affect the style of any existing modals?

mcasimir (commented):
This was super fast! Left few comments, mostly the unhandled promise bothers me a little, it would be better if we don't have situations where the app can crash or misfunction because of telemetry. 

Everything else looks good although i'm not sure if i would go as far as adding a new component variant for something that we hope will be seen by very few people, not blocking though, just a thought.

packages/compass-components/src/components/modals/modal-header.tsx
mabaasit: currently in modal header we only support two variants (default - with no icon and danger - with red filled warning icon) and the design came up with a warn variant (which uses yellow) as fill color. 

I did try to use danger variant to avoid these changes, but Ben suggested to use warn variant.

mabaasit: I did have another chat with Ben and he agreed to drop this. I'll make the changes.

mabaasit: removed!

packages/compass/src/main/application.ts
mcasimir: More of a nit, feel free to keep it this way or change it and merge: now this is not going to cause problems anymore, but if it was already in this way I would have never spotted this same issue. 

I wonder if it would be better to just `.catch()` explicitly here instead.

mcasimir: mm actually yeah on a second though .. i don't know .. i think that we should not have this `void` here. We may add a line of code throwing inside `trackApplicationLaunched` and never notice.

mabaasit: [via chat] Maurizio helped me realize that `.track` function takes a callback. I used that to determine the track event props.

mcasimir (approved):

PR #4736, opened 2023-08-16
feat(compass-aggregations): generate aggregations with ai COMPASS-7105

Anemy:
COMPASS-7105

Note: This is using the new backend from https://github.com/10gen/compass-mongodb-com/pull/89/files

This adds the ai input to the aggregation pipeline header. User inputs text, gets an aggregation out. This currently replaces whatever aggregation content they currently have in the aggregation view.

A lot of the actions & reducer code behind fetching the ai response is similar between this pipeline-ai module and the query bar query-ai module. We could move them into a shared package in Compass, I think it would be a bit extra to do that now, and might make things a bit too coupled down the line if we want different experiences in each. Something to keep an eye on.

We currently sample documents every time a generation request is made. We could in future share the preview documents and or cache these documents with the query bar. I opted to keep it simple for now by not implementing that yet.

https://github.com/mongodb-js/compass/assets/1791149/db455f19-4100-470a-afe7-3d2b9f234eda

<details>
  
  Screenshot 2023-08-16 at 1 30 20 PM
  Screenshot 2023-08-16 at 1 31 15 PM
  Screenshot 2023-08-16 at 1 37 17 PM
  Screenshot 2023-08-16 at 2 44 31 PM
</details>


packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-ai.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+      expect(clearTextButton).to.be.visible;
+      clearTextButton.click();
+
+      expect(store.getState().pipelineBuilder.aiPipeline.aiPromptText).to.equal(
+        ''
+      );
+    });
+  });
+
+  describe('Query AI Feedback', function () {
```
Anemy, line 78: Will update this to bring in the changes we did to the ai query feedback tests in https://github.com/mongodb-js/compass/pull/4735/commits/d4c12ec432512c9c1302c5fc2e8421ec2a6f8562#diff-d21ac5941571e624a797890f07d97e64bec5852597750e3794ed3716b3c7d471R84

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-ai.ts
```diff
@@ -0,0 +1,422 @@
...
+
+    let pipeline;
+    try {
+      if (!jsonResponse?.content?.aggregation?.pipeline) {
+        throw new Error(
+          'No pipeline returned. Please try again with a different prompt.'
+        );
+      }
+
+      // TODO(COMPASS-7105): Update the format the backend/ai model
```
Anemy, line 228: Linked the wrong Compass ticket here, should be [COMPASS-7100](https://jira.mongodb.org/projects/COMPASS/issues/COMPASS-7100)
Going to likely update this before merging based on when https://github.com/10gen/compass-mongodb-com/pull/89 is merged in.

packages/atlas-service/src/main.ts
```diff
@@ -579,6 +586,70 @@ export class AtlasService {
     return res.json() as Promise<IntrospectInfo>;
   }
 
+  static async getAggregationFromUserInput({
```
Anemy, line 589: I'll add some tests for this function.

packages/compass-query-bar/src/utils/query.ts
```diff
@@ -51,6 +51,35 @@ export function doesQueryHaveExtraOptionsSet(fields?: QueryFormFields) {
   return false;
 }
 
+export function parseQueryStringsToFormFields(query?: {
```
mcasimir, line 54: `parseQueryAttributesToFormFields` or `parseQueryPropertiesToFormFields`? .. maybe is just me but when i read  parseQueryStrings i can only think that we are parsing a URL

packages/atlas-service/src/main.ts
```diff
@@ -604,6 +611,70 @@ export class AtlasService {
...
+      // If that fails we throw an error indicating this collection's
+      // documents are too large to send to the ai.
+      msgBody = JSON.stringify({
+        userInput,
+        collectionName,
+        databaseName,
+        schema,
+        sampleDocuments: sampleDocuments?.slice(0, AI_MIN_SAMPLE_DOCUMENTS),
+      });
+      // Why this is not happening on the backend?
```
mcasimir, line 650: seems totally fine to keep that in the backend

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-ai.ts
```diff
@@ -0,0 +1,408 @@
...
+    );
+
+    dispatch({
+      type: AIPipelineActionTypes.AIPipelineSucceeded,
+    });
+
+    pipelineBuilder.reset(pipelineText);
+
+    dispatch({
+      type: AIPipelineActionTypes.LoadNewPipeline,
```
mcasimir, line 259: `LoadGeneratedPipeline`?

packages/compass-aggregations/src/modules/pipeline-builder/pipeline-ai.ts
```diff
@@ -0,0 +1,408 @@
...
+      pipelineText: pipelineBuilder.source,
+      pipeline: pipelineBuilder.pipeline,
+      syntaxErrors: pipelineBuilder.syntaxError,
+    });
+
+    dispatch(updatePipelinePreview());
+  };
+};
+
+type CancelAIPipelineAction = {
```
mcasimir, line 270: more silly naming suggestions: `CancelAIPipelineGenerationAction`?

mcasimir (commented):
Epic!! Nothing stood out to me so i only left 1000 naming comments, before we merge it is better if someone else also take a look.

packages/atlas-service/src/main.ts
```diff
@@ -604,6 +611,70 @@ export class AtlasService {
...
+      // When the message body is over the max size, we try
+      // to see if with fewer sample documents we can still perform the request.
+      // If that fails we throw an error indicating this collection's
+      // documents are too large to send to the ai.
+      msgBody = JSON.stringify({
+        userInput,
+        collectionName,
+        databaseName,
+        schema,
+        sampleDocuments: sampleDocuments?.slice(0, AI_MIN_SAMPLE_DOCUMENTS),
```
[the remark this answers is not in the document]
Anemy: It's called `AI_MIN_SAMPLE_DOCUMENTS` as this is the length that we slice them to only when the the prompt size is too large with the normal amount of sample documents. So this is the minimum amount we try to send if the normal amount is too much. Happy to name it anything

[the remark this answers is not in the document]
mcasimir: Ah, got it! Something like `AI_FALLBACK_SAMPLE_DOCUMENTS_SIZE`?

packages/atlas-service/src/main.ts
gribnoysup: The comment got removed, but I am still not sure why this is a client validation and not a backend one

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-ai.spec.tsx
```diff
@@ -0,0 +1,146 @@
...
```
gribnoysup, line 85: Pipeline AI Feedback?

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/index.tsx
```diff
@@ -112,39 +117,59 @@ export const PipelineHeader: React.FunctionComponent<PipelineHeaderProps> = ({
...
-          showExplainButton={showExplainButton}
+      {enableAIExperience && (
+        <PipelineAI
+          onClose={() => {
+            onHideAIInputClick?.();
+          }}
+          show={isAIInputVisible}
         />
-      </div>
+      )}
```
gribnoysup, line 160: Is this by design that this is added above the options and not below? Why is it different from query bar?

image

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-ai.tsx
```diff
@@ -0,0 +1,58 @@
...
+};
+
+type PipelineAIProps = Omit<
+  React.ComponentProps<typeof GenerativeAIInput>,
+  'onSubmitFeedback'
+>;
+
+function PipelineAI(props: PipelineAIProps) {
+  return (
+    <GenerativeAIInput
```
gribnoysup, line 34: This component adds margin that should not be added making it misaligned with all the other content in the toolbar:

image

Layout positioning should be handled by parents, not children, especially for reusable components.

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-actions.tsx
```diff
@@ -75,8 +82,13 @@ export const PipelineActions: React.FunctionComponent<PipelineActionsProps> = ({
   onCollectionScanInsightActionButtonClick,
 }) => {
   const showInsights = usePreference('showInsights', React);
+  const enableAIExperience = usePreference('enableAIExperience', React);
+
   return (
     <div className={containerStyles}>
+      {enableAIExperience && showAIEntry && (
+        <AIExperienceEntry onClick={onShowAIInputClick} />
+      )}
```
gribnoysup, line 91: We are already rendering it inside the stages component, why not keep it all in one place and just push it to the right when there are stages in the pipeline?

packages/atlas-service/src/main.spec.ts
```diff
@@ -13,6 +13,20 @@ function getListenerCount(emitter: EventEmitter) {
...
+  aiEndpoint: string;
+}[] = [
+  {
+    functionName: 'getQueryFromUserInput',
+    aiEndpoint: 'mql-query',
+  },
+  {
+    functionName: 'getAggregationFromUserInput',
+    aiEndpoint: 'mql-aggregation',
+  },
```
gribnoysup, line 27: If they are literally the same to the point that the tests are only switching the endpoint name mock, I wounder if it would make sense to implement them as such and not test the same thing copied over twice

gribnoysup (approved):
Overall seems to work fine, left a few comments for the implementation.

I am also wondering what is our release strategy for this? As this is not backwards compatible with the currently deployed backend, if we merge this today, this auto creates a beta on Monday early morning and when it releases, latest beta will be broken until we also deploy the backend, right? Do we want to hold off on merging this until beta is kicked off next week or are we okay with the risks?

packages/atlas-service/src/main.ts
alenakhineika: We do have this validation on the backend.

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-ai.spec.tsx
alenakhineika: In the compass tracking plan, it is called "AI Response Rated" and it is one kind of event for both query and aggregation that has the `editor_view_type: find | stage | text ` prop to distinguish what we leave feedback for. The plan also uses different values for feedback. It is actually called `rating` with `good | bad | dismissed` values. Should we update this format in both places, here and in the query bar?

```
AI Response Rated

editor_view_type: find | stage | text 
query_shape: <stages & operators used>
rating: good | bad | dismissed
num_tokens: <length of repsonse in number of tokens>
```

alenakhineika:
> I am also wondering what is our release strategy for this? As this is not backwards compatible with the currently deployed backend

@gribnoysup the backend already has the aggregations endpoint. Is this PR uses some different froat from what the endpoint supports?

gribnoysup:
Backend change is merged and autodeployed to staging, but not production as far as I can see, nobody promoted it yet. This is also a breaking change, so if we deploy the backend to production right now for example, the AI feature in the current beta release will stop working because this API responses are not compatible. And vice versa, if we merge this and beta is autopublished on Monday without us releasing the backend in time, AI feature will not work. So that's why I'm asking what's the release strategy for this if we are doing a breaking change without changing the API version in the url and whether or not we should consider waiting until after the kick off just in case (or are okay with timings, risks of doing this right before the kick off)

mcasimir:
> > I am also wondering what is our release strategy for this? As this is not backwards compatible with the currently deployed backend
> 
> @gribnoysup the backend already has the aggregations endpoint. Does this PR use some different format from what the endpoint supports?

Yeah, the change in the service is not released as far as i know, when we merge this one we need to do a new beta, and as soon as the beta is ready we need to promote the service, not the most graceful upgrade, after we are live we should treat breaking changes appropriately.

alenakhineika:
> > > I am also wondering what is our release strategy for this? As this is not backwards compatible with the currently deployed backend
> > 
> > 
> > @gribnoysup the backend already has the aggregations endpoint. Does this PR use some different format from what the endpoint supports?
> 
> Yeah, the change in the service is not released as far as i know, when we merge this one we need to do a new beta, and as soon as the beta is ready we need to promote the service, not the most graceful upgrade, after we are live we should treat breaking changes appropriately.

I think this is where @addaleax suggestion to include the compass version in the request would be useful. Then if we do a breaking change we could still support the previous compass versions.

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-ai.spec.tsx
Anemy: Yes, let's do that in a separate pr.

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-actions.tsx
Anemy: I found getting the styles to work nicely without doing it this way ended up being a lot of work. There might be a clean way to do it that's simple but I hadn't found it so I opted for this way.

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/index.tsx
Anemy: Nope! Thanks for catching. Pushed up an update:
Screenshot 2023-08-18 at 11 24 39 AM

Anemy:
Going to merge this in once tests go green. As Maurizio mentioned, once this is in main we'll want a new beta and to promote/update the live service accordingly.

gribnoysup:
Adding metadata to the request payload sounds good, but we also have `version` in the endpoint for versioning as specified in the tech design

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-header/pipeline-actions.tsx
gribnoysup: Feel free to keep it, sure, just FYI, it's two lines of css as far as I can see, maybe will come up helpful next time you need to style something similarly:

```diff
 const containerStyles = css({
+  flex: 1, // so that the pipeline takes all available space in the toolbar
   display: 'flex',
   gap: spacing[2],
   alignItems: 'center',
 });

+const buttonContainerStyles = css({
+  marginLeft: 'auto', // wrapper for your button so that it's pushed to the right in the layout
+});
```

mcasimir:
> Right, didn’t mean to imply that versioning the API endpoint shouldn’t be the primary way of managing API version differences

To make sure that we have 3 ways to version this API and that we use all of them at once, we should also switch to the atlas convention: `Accept: application/vnd.atlas.<yyyy-mm-dd>+json` (https://www.mongodb.com/docs/atlas/api/versioned-api-overview/)

Anemy:
Merging, I think the versioning and api design changes would be better in a pr separate from this one. Thanks for all the good comments!

alenakhineika:
I will open a PR for the backend to support v1 and v2 endpoints before Monday.

PR #4737, opened 2023-08-16
chore(compass-query-bar): update response handling of ai query response COMPASS-7100

Anemy:
COMPASS-7100

Related backend pr: https://github.com/10gen/compass-mongodb-com/pull/89
This also means we will show possibly invalid queries in the query bar.
If the backend updates are merged before the aggregations pr we'll need to update that too.


Anemy:
Pulled these changes into https://github.com/mongodb-js/compass/pull/4736
Backend is already merged

PR #4741, opened 2023-08-17
chore(compass-components): update ai input styles COMPASS-7039

Anemy:
Part of COMPASS-7039

Changes:
- Move the robot button onto the left side of the input.
- Disable generate button when no input from user.
- Update the loader not replace the robot icon.
- Update the feature flag text to include `Do not use this feature with sensitive data.`.

Screenshot 2023-08-17 at 1 29 00 PM


packages/compass-components/src/components/generative-ai/generative-ai-input.tsx
```diff
@@ -110,14 +114,10 @@ const buttonResetStyles = css({
...
-  buttonResetStyles,
-  {
-    padding: spacing[1],
-    display: 'inline-flex',
-  },
-  focusRing
-);
+const closeAIButtonStyles = css(buttonResetStyles, focusRing, {
+  padding: spacing[1],
+  position: 'absolute',
```
gribnoysup, line 119: If you're changing controls to absolute positioning, you probably want to adjust the input size too, otherwise text can go below buttons:

image

gribnoysup (approved):
There is a text overflow issue with the new buttons, otherwise looks good

packages/compass-components/src/components/generative-ai/generative-ai-input.tsx
Anemy: Good catch, updated to have a restricted width:
Screenshot 2023-08-18 at 5 05 57 PM

PR #4756, opened 2023-08-23
fix(connections): add support for downgrade to a version with storage-mixin COMPASS-7174

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/connection-storage/src/connection-storage.ts
+++ b/packages/connection-storage/src/connection-storage.ts
@@ -31,6 +31,16 @@ import type {
 
 const { log, mongoLogId } = createLoggerAndTelemetry('CONNECTION-STORAGE');
 
+type ConnectionLegacyProps = {
+  _id: string;
+  isFavorite?: boolean;
+  name: string;
+};
+
+type ConnectionWithLegacyProps = {
+  connectionInfo: ConnectionInfo;
+} & ConnectionLegacyProps;
+
 export class ConnectionStorage {
   private static calledOnce: boolean;
   private static path: string;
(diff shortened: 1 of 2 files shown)


packages/connection-storage/src/connection-storage.ts
```diff
@@ -102,7 +112,7 @@ export class ConnectionStorage {
     }
   }
 
-  private static async getConnections(): Promise<any[]> {
+  private static async getConnections(): Promise<ConnectionWithLegacyProps[]> {
```
mcasimir, line 115: Why do we need those props to get to this layer? here they shouldn't be needed. Can we just make sure that we save them and we don't read them back?

mcasimir: Ah wait that's what we are doing already?

mcasimir: Nope .. I tried, but i'm a still confused :), why do we read them back? And how is `isFavorite` and `name` related to this change?

If we want to be aware of those, shouldn't it be to remove them from what we read so it never reaches other layers?

And on save we can just re-add them back.

mcasimir (approved):
Super quick fix! 🔥🔥🔥 I left a comment, i would not keep around those legacy props, but is not blocking 👏

packages/connection-storage/src/connection-storage.ts
mabaasit: this method (`getConnections`), just reads all the data from the storage including legacy connections. i just added types to support that as its the expected output from this method. and then in `loadAll` method, we only consider connections with `connectionInfo` and other props (`isFavorite` and `name`) are used in `getLegacyConnections` which's used in connection migration modal (where we show list of legacy connections).

mcasimir: Ahh got it! so `isFavorite` and `name` is not related to this specific `_id` thing! Makes sense, thanks!

PR #4759, opened 2023-08-24
feat: navigate aggregations generated from the query bar ai COMPASS-7106

alenakhineika:
## Description

In case AI returns aggregation on the "Documents" tab, we navigate users to the "Aggregations" tab and create a pipeline. We also show a guide cue so users are aware of what just happened.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/modules/pipeline-builder/pipeline-ai.ts
```diff
@@ -34,18 +37,23 @@ export const initialState: AIPipelineState = {
...
   AIPipelineFailed = 'compass-aggregations/pipeline-builder/pipeline-ai/AIPipelineFailed',
   AIPipelineSucceeded = 'compass-aggregations/pipeline-builder/pipeline-ai/AIPipelineSucceeded',
+  AIPipelineCreatedFromQuery = 'compass-aggregations/pipeline-builder/pipeline-ai/AIPipelineCreatedFromQuery',
   CancelAIPipelineGeneration = 'compass-aggregations/pipeline-builder/pipeline-ai/CancelAIPipelineGeneration',
+  HideGuideCue = 'compass-aggregations/pipeline-builder/pipeline-ai/HideGuideCue',
   ShowInput = 'compass-aggregations/pipeline-builder/pipeline-ai/ShowInput',
   HideInput = 'compass-aggregations/pipeline-builder/pipeline-ai/HideInput',
   ChangeAIPromptText = 'compass-aggregations/pipeline-builder/pipeline-ai/ChangeAIPromptText',
-  LoadGeneratedPipeline = 'compass-aggregations/LoadGeneratedPipeline',
+  LoadAIPipeline = 'compass-aggregations/LoadAIPipeline',
```
gribnoysup, line 56: While makes it definitely clearer, we currently have 3 people working on the code in the same area, I'd maybe stay away from renaming for now just to limit the amount of possible git conflicts between PRs

packages/compass-aggregations/src/stores/store.ts
```diff
@@ -252,6 +253,10 @@ const configureStore = (options: ConfigureStoreOptions) => {
     localAppRegistry.on('fields-changed', (fields) => {
       store.dispatch(fieldsChanged(fields.autocompleteFields));
     });
+
+    localAppRegistry.on('open-aggregation-tab', (data) => {
+      store.dispatch(createPipelineFromQuery(data));
+    });
```
gribnoysup, line 259: So I'd suggest either to make the event name more scoped to the use-case you are introducing or event handler should be handling this as a more generic event that can be used to just programmatically open aggregation subtab inside collection tab, optionally indicating somehow that it's coming from query tab AI feature. Currently you are mixing two approaches: the event name is very generic indicating that it can be used to just open aggregation tab, but action that is being dispatched is purely focused on handling this as if it's always tab being opened in relation to AI endpoint returning query instead of aggregation

packages/compass-query-bar/src/stores/ai-query-reducer.ts
```diff
@@ -225,11 +237,11 @@ export const runAIQuery = (
       return;
     }
 
+    let query: Record<string, unknown>;
     let generatedFields: QueryFormFields;
     try {
-      generatedFields = parseQueryAttributesToFormFields(
-        jsonResponse.content.query
-      );
+      query = jsonResponse?.content?.query as Record<string, unknown>;
```
gribnoysup, line 243: This type assertion is incorrect, we validate the response before returning it from `getQuery` method, so what you get here will always match expected backend schema (or request will fail already and you will not end up in this part of the action)

packages/compass-query-bar/src/stores/ai-query-reducer.ts
```diff
@@ -243,6 +255,34 @@ export const runAIQuery = (
...
+    const aggregation = Object.entries(query ?? {}).find(([key, value]) => {
+      return (
+        key === 'aggregation' &&
+        typeof value === 'object' &&
+        value !== null &&
+        typeof (value as any).pipeline === 'string'
+      );
+    });
+
+    if (aggregation) {
```
gribnoysup, line 269: Should we check that there is no query returned with aggergation / move this `if (aggregation)` block under the check that comes afterwards that determines that query came back empty? When testing backend locally, I saw sometimes that returned response included both query and aggergation sometimes, something like `{query: { find: '{ _id: 1 }', aggregation: { pipeline: '[{ $match: { _id: 1 } }]' } }`, which is a valid query and we probably don't need to switch in this case?

packages/compass-components/src/components/generative-ai/generative-ai-input.tsx
```diff
@@ -150,6 +151,10 @@ type GenerativeAIInputProps = {
   isFetching?: boolean;
   placeholder?: string;
   show: boolean;
+  showGuideCue?: boolean;
+  guideCueTitle?: string;
+  guideCueDescription?: string;
+  onHideGuideCue?: () => void;
```
gribnoysup, line 157: The way this is currently implemented, this state doesn't really do anything, you are mixing controlled and uncontrolled components here. If you want component to be controlled, the state of the guide cue should always be read and changed from the outside, not from internal state, if you want it to be uncontrolled, you don't need `onHideGuideCue` and this `showGuideCue` should rather be called something like `showGuideCueOnMount` or `initialShowGuideCue` to indicate that it's not fully controlling the state of the cue component

packages/compass-query-bar/src/stores/ai-query-reducer.ts
```diff
@@ -243,6 +255,34 @@ export const runAIQuery = (
...
+    // The query endpoint may return the aggregation property in addition to filter, project, etc..
+    // It happens when the AI model couldn't generate a query and tried to fulfill a task with the aggregation.
+    const aggregation = Object.entries(query ?? {}).find(([key, value]) => {
+      return (
+        key === 'aggregation' &&
+        typeof value === 'object' &&
+        value !== null &&
+        typeof (value as any).pipeline === 'string'
+      );
+    });
```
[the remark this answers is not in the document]
alenakhineika: You are too fast with commenting :) I already changed this locally after rebasing from the main, but haven't pushed yet.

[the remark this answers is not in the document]
gribnoysup: Oh, it's a draft, I didn't notice, sorry! Disregard everything that doesn't make sense with the changes you are still doing then 🙇

packages/compass-components/src/components/generative-ai/index.ts
```diff
@@ -1,4 +1,5 @@
 export { GenerativeAIInput } from './generative-ai-input';
+// export { GenerativeAIGuideCue } from './generative-ai-guide-cue';
```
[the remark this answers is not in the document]
alenakhineika: Yep.

packages/compass-query-bar/src/stores/ai-query-reducer.ts
alenakhineika: Yep gone after rebasing from main and refactoring.

packages/compass-query-bar/src/stores/ai-query-reducer.ts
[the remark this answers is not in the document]
alenakhineika: Pushed.

gribnoysup (commented):
Overall already looks good, but I think the ai-input / ai-guide-cue implementation can be simplified and dependency on aggregation store can be completely removed

packages/compass-components/src/components/generative-ai/ai-guide-cue.tsx
```diff
@@ -0,0 +1,78 @@
...
+      setOpen={setOpen}
+      refEl={refEl}
+      numberOfSteps={1}
+      currentStep={1}
+      onPrimaryButtonClick={() => {
+        onHideGuideCue?.();
+      }}
+      onDismiss={() => {
+        onHideGuideCue?.();
+      }}
```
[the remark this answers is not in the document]
alenakhineika: So I kept onHideGuideCue and noped setOpen. Let me know how this look now 🙈

packages/compass-components/src/components/generative-ai/generative-ai-input.tsx
```diff
@@ -150,6 +151,8 @@ type GenerativeAIInputProps = {
   isFetching?: boolean;
   placeholder?: string;
   show: boolean;
+  isAggregationGeneratedFromQuery?: boolean;
+  onHideGuideCue?: () => void;
```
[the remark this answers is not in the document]
alenakhineika: Refactored as we discussed. Please let me know if missed something 🙏

packages/compass-components/src/components/generative-ai/generative-ai-input.tsx
```diff
@@ -246,7 +252,21 @@ function GenerativeAIInput({
...
+              open={isAggregationGeneratedFromQuery}
+              setOpen={() => {
+                /* noop because we control guide cue visibility through onResetIsAggregationGeneratedFromQuery */
+              }}
+              refEl={guideCueRef}
+              onResetIsAggregationGeneratedFromQuery={
+                onResetIsAggregationGeneratedFromQuery
+              }
+              title="Aggregation generated"
+              description="Your query requires stages from MongoDB's aggregation framework. Continue to work on it in our Agaredation Pipeline Builder"
```
Anemy, line 265: There's a typo we should fix here, how do you feel about this?
```suggestion
              description="The generated query requires use of MongoDB's aggregation framework. Open and work on it in the Aggregation Pipeline Builder"
```
or with the typo fixed:
```suggestion
              description="Your query requires stages from MongoDB's aggregation framework. Continue to work on it in our Aggregation Pipeline Builder"
```

packages/compass-query-bar/src/stores/ai-query-reducer.ts
```diff
@@ -246,6 +258,25 @@ export const runAIQuery = (
...
+      // The query endpoint may return the aggregation property in addition to filter, project, etc..
+      // It happens when the AI model couldn't generate a query and tried to fulfill a task with the aggregation.
+      if (query.aggregation) {
+        localAppRegistry?.emit('generate-aggregation-from-query', {
+          userInput,
+          aggregation: query.aggregation,
+        });
+        const msg =
+          'Query requires stages from aggregation framework therefore an aggregation was generated.';
+        trackAndLogFailed({
```
Anemy, line 270: Have we checked with Julia that an `AI Response Failed` event is expected here? Since it's making a usable pipeline it might not be intuitive as a failure.

Anemy (commented):
Looking good, two small comments

packages/compass-components/src/components/generative-ai/generative-ai-input.tsx
alenakhineika: I think "Open and work on it in the Aggregation Pipeline Builder" is a bit confusing because this message shows up as a guide cue when we already navigated to the aggregation. But I will fix the typo, thx! did not notice it 👀

packages/compass-query-bar/src/stores/ai-query-reducer.ts
alenakhineika: Good point. I will be on vacation next week. Can you chat about this with Julia and either adjust this PR or open the follow-up PR if this will be merged?

Anemy: Sure thing, I'll ping her now. Not a blocker.

packages/compass-components/src/components/generative-ai/generative-ai-input.tsx
Anemy: Ah gotcha, I was thinking this was before the transition. Sounds good.

gribnoysup:
@Anemy while Alena is on vacation, I'm making sure we merge this, I did a bit of code clean up (no behavior changes) on top of last changes, do you mind giving this another look and then I'll merge if this looks good to you?

packages/compass-query-bar/src/stores/ai-query-reducer.ts
Anemy: Got lgtm here: https://mongodb.slack.com/archives/C05AH835WQN/p1692996996219599?thread_ts=1692738027.149129&cid=C05AH835WQN

packages/compass-components/src/components/generative-ai/generative-ai-input.tsx
```diff
@@ -150,6 +151,8 @@ type GenerativeAIInputProps = {
   isFetching?: boolean;
   placeholder?: string;
   show: boolean;
+  isAggregationGeneratedFromQuery?: boolean;
```
Anemy, line 154: A nit, fine as is, if we have the value in the store not as possibly undefined I prefer to keep the component interfaces the same.
```suggestion
  isAggregationGeneratedFromQuery: boolean;
```
That way if something changes in the store we're for sure to catch it here. In tests we can then default it to `false`.

Anemy (approved):
lgtm!

PR #4761, opened 2023-08-24
chore(compass-aggregations): update cursor to grabbing to show drag and drop functionality

Anemy:
Updates the cursor on the stage wizard stages to grab and grabbing to show the drag and drop functionality.
Need to check with design if this is what we want to do, I think this came up as part of chatting to users. The wizard stages are also buttons which, on click, add the wizard at the bottom of the pipeline.

Screenshot 2023-08-24 at 11 37 01 AM


gribnoysup (approved):

Anemy:
Checked with Claudia, good to merge.

PR #4779, opened 2023-08-31
fix(query-editor): attempt at fixing paste in the query bar after auto-{} COMPASS-7149

lerouxb:
I looked at https://github.com/mongodb-js/compass/pull/4707/files and can't understand why we added our own onpaste handling. It seems to just put what was there back there? Maybe good if @kmruiz and @gribnoysup has a look - I'm probably missing some use case.

### DIFF
--- a/packages/compass-query-bar/src/components/option-editor.tsx
+++ b/packages/compass-query-bar/src/components/option-editor.tsx
@@ -151,22 +151,11 @@ const OptionEditor: React.FunctionComponent<OptionEditorProps> = ({
 
   const onFocus = () => {
     if (hasAutofix && editorRef.current) {
-      if (editorRef.current.editorContents === '') {
-        rafraf(() => {
-          editorRef.current?.applySnippet('\\{${}}');
-        });
-      }
-    }
-  };
-
-  const onPaste = (event: React.ClipboardEvent<HTMLDivElement>) => {
-    if (hasAutofix && editorRef.current) {
-      const editorContents = editorRef.current.editorContents;
-      const snippet = lenientlyFixQuery(editorContents || '');
-      if (snippet) {
-        editorRef.current.applySnippet(snippet);
-        event.preventDefault();
-      }
+      rafraf(() => {
+        if (editorRef.current?.editorContents === '') {
+          editorRef.current.applySnippet('\\{${}}');
+        }
+      });
     }
   };
 
@@ -189,7 +178,6 @@ const OptionEditor: React.FunctionComponent<OptionEditorProps> = ({
         commands={commands}
         data-testid={dataTestId}
         onFocus={onFocus}
-        onPaste={onPaste}
         onBlur={onBlur}
       />
       {showInsights && insights && (


packages/compass-query-bar/src/components/option-editor.tsx
```diff
@@ -151,22 +151,11 @@ const OptionEditor: React.FunctionComponent<OptionEditorProps> = ({
...
-  const onPaste = (event: React.ClipboardEvent<HTMLDivElement>) => {
-    if (hasAutofix && editorRef.current) {
-      const editorContents = editorRef.current.editorContents;
-      const snippet = lenientlyFixQuery(editorContents || '');
-      if (snippet) {
-        editorRef.current.applySnippet(snippet);
-        event.preventDefault();
-      }
+      rafraf(() => {
+        if (editorRef.current?.editorContents === '') {
```
lerouxb, line 155: In addition to my question about whether we should be doing onpaste, I'm wondering if it isn't better to check that the editor is still empty by the time rafraf fires. In case something delayed like a paste occurred.

packages/compass-query-bar/src/components/option-editor.tsx
```diff
@@ -151,22 +151,11 @@ const OptionEditor: React.FunctionComponent<OptionEditorProps> = ({
...
-          editorRef.current?.applySnippet('\\{${}}');
-        });
-      }
-    }
-  };
-
-  const onPaste = (event: React.ClipboardEvent<HTMLDivElement>) => {
-    if (hasAutofix && editorRef.current) {
-      const editorContents = editorRef.current.editorContents;
-      const snippet = lenientlyFixQuery(editorContents || '');
```
Anemy, line 165: If we don't need the `lenientlyFixQuery` anymore we can remove it. Echoing Le Roux, let's wait and see if Kevin and or Sergey have extra info.

gribnoysup: Paste was added to handle the case where you focus on the editor, we auto-insert the brackets, and it messes with your pasted content if it's a valid. Which in this branch leads to your input being not valid on paste:

image

packages/compass-query-bar/src/components/option-editor.tsx
gribnoysup: Yeah, if I had to guess that's probably the root cause of the bug (although I can't reproduce it the same way ticket describes)

gribnoysup (commented):
The `onPaste` behavior is a requirement from product / design for COMPASS-6530, now that we mess with user input while user is interacting with the query bar, we also need to fix it for some corner cases, this is why it's there for. I still think that this is a bad thing to do in general and we will be getting bug reports for this forever, but this seems like what we want to do anyway, so we can't remove this `onPaste` handler

packages/compass-query-bar/src/components/option-editor.tsx
kmruiz: Yeah I can't reproduce it either, but this sounds a reasonable issue.

packages/compass-query-bar/src/components/option-editor.tsx
kmruiz: Exactly, and that was one of the use cases that we are expected to handle. If after the focus, the customer pastes a valid query, we would get duplicated keys (as @gribnoysup shown in the picture).

lenientlyFixQuery just takes into consideration this and generates the proper snippet.

PR #4783, opened 2023-09-01
chore: set unset_function_vars: true in anticipation of the coming change COMPASS-7139

lerouxb:


packages/compass-aggregations/src/modules/pipeline-builder/stage-editor.ts
```diff
@@ -352,6 +352,7 @@ export const runStage = (
       };
       // We are not handling cancelling, just supporting `aggregatePipeline` interface
       const { signal } = new AbortController();
+      console.log('aggregatePipeline via runStage', new Error().stack);
```
addaleax, line 355: Accidentally committed? :)

Btw, just in case you find this helpful: https://developer.mozilla.org/en-US/docs/Web/API/console/trace

lerouxb: yeah

PR #4798, opened 2023-09-05
chore(compass-indexes): restructure indexes store COMPASS-7160

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-indexes/src/stores/store.js
```diff
@@ -96,8 +95,7 @@ const configureStore = (options = {}) => {
   }
 
   if (options.isReadonly) {
-    const isReadonlyView = options.isReadonly;
-    store.dispatch(readonlyViewChanged(isReadonlyView));
```
mabaasit, line 99: We are using `isReadonlyView` within this plugin to identify a view and `isReadonly` was not set anywhere and always defaulted to its `false` value. So I removed it in this commit.

packages/compass-indexes/src/modules/index.ts
```diff
@@ -6,49 +6,33 @@ import { RESET_FORM } from './reset-form';
...
-import sortOrder, {
-  INITIAL_STATE as SORT_ORDER_INITIAL_STATE,
-} from './sort-order';
-import sortColumn, {
-  INITIAL_STATE as SORT_COLUMN_INITIAL_STATE,
-} from './sort-column';
+import regularIndexes, {
+  INITIAL_STATE as REGULAR_INDEXES_INITIAL_STATE,
+} from './regular-indexes';
 import error, { INITIAL_STATE as ERROR_INITIAL_STATE } from './error';
```
mabaasit, line 18: Keeping `error` here as its shared with `create-index` and `drop-index` stores. I did not want to do their own error implementation given the time limitations.

packages/compass-indexes/src/stores/store.js
gribnoysup: It's not always `false`, it's `true` if you connect to ADL for example, we can't just remove it

gribnoysup: Ah, I see what you mean, the state is always `false`, not the option passed to the `configureStore`. 

Something is super weird going on here. I see that before https://github.com/mongodb-js/compass/pull/3609 this was actually mapping to hadron readonly mode, the UI was updated to read the value from preferences, but the store, in particular `fetchIndexes` method wasn't, I see that you replaced it with reading `isReadonlyView`, but this doesn't seem what the meaning of this value was

packages/compass-indexes/src/modules/index.ts
gribnoysup: Those are separate plugins with their own store, so you can safely move / remove this slice in this particular one

mabaasit: just cleaned it up in [e9a34b4](https://github.com/mongodb-js/compass/pull/4798/commits/e9a34b489c28324390d41502bff6aefabbb69e12)

gribnoysup (approved):

packages/compass-indexes/src/modules/index.ts
gribnoysup: Not sure why we need to identical files, but as we are not doing any clean up that's fine 👍

packages/compass-indexes/src/stores/store.js
gribnoysup: So it was a bug that was accidentally fixed by not checking for preferences there?

mabaasit: Yes it was.

packages/compass-indexes/src/utils/parse-error.ts
```diff
@@ -0,0 +1,21 @@
+import { MongoError } from 'mongodb';
+import type { AnyError } from 'mongodb';
+
+export type IndexesError = AnyError | string;
+/**
+ * Data Service attaches string message property for some errors, but not all
+ * that can happen during index creation/dropping.
+ * Check for data service custom error, then node driver errmsg.
+ */
```
mabaasit, line 9:
```suggestion
```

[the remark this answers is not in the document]
mabaasit: removed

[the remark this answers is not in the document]
gribnoysup: I meant more like the whole function, not only comment because it's literally what data service methods are doing when re-throwing errors coming from driver:

https://github.com/mongodb-js/compass/blob/6273cb62f7bb18907dcc8cdef59ce8ec83dc9d7a/packages/data-service/src/data-service.ts#L742-L749

(it's fine anyway though, just wanted to clarify what I meant)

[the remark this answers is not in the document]
mabaasit: thanks for sharing. yes, i'll remove it

PR #4803, opened 2023-09-05
chore(compass-settings): remove ai feature flag, add preference COMPASS-7147

Anemy:
COMPASS-7147

This PR is a draft as we'll first want the `hello` endpoint to be setup, as it will dictate whether or not we show this preference or not. We also don't want to surface this to users until we are ready to bring it into public beta, so this may be in draft for a week or more.

This PR removes the `enableAIExperience` feature flag, and adds a preference `enableAI` which can be set using the command line and also derives it's value from `networkTraffic`.

- [ ] Run the text of the setting by Julia.

<details>
  
Screenshot 2023-09-05 at 2 45 13 PM

Screenshot 2023-09-05 at 2 37 01 PM

</details>

### DIFF
--- a/packages/compass-aggregations/src/components/pipeline-toolbar/index.tsx
+++ b/packages/compass-aggregations/src/components/pipeline-toolbar/index.tsx
@@ -74,7 +74,7 @@ export const PipelineToolbar: React.FunctionComponent<PipelineToolbarProps> = ({
   pipelineOutputOption,
 }) => {
   const darkMode = useDarkMode();
-  const enableAIExperience = usePreference('enableAIExperience', React);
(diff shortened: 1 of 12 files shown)


mcasimir:
> We also don't want to surface this to users until we are ready to bring it into public beta, so this may be in draft for a week or more.

Can we rather do this also for the public beta and keep the FF till we are actually ready to release? I can quickly add the endpoint and enable it for the 100% of users in the backend. I don't really see a reason to diverge from our usual best practices here.

Anemy:
Closing as we'll do this work as part of the hello endpoint (and also keep the feature flag).

PR #4812, opened 2023-09-06
chore(query-bar): use the returned aggregation from ai query when it exists

Anemy:
This pr makes it so that we now prefer to use an aggregation returned from the query ai endpoint instead of the query fields, in the case where the aggregation exists.
Previously we would default to using the query fields if any of them existed. Through testing various prompt accuracies I was finding the model really likes returning query properties like sort and project even when they're included in the pipeline. This would make the move to aggregations view feature not happen, which is what motivated this pr.

For an example of a prompt that would generate a good valid aggregation, but the ai also puts projections and sorts into the query try:
```
whats the count of each bed count, sorted by lowest bed count first
```
On the `airbnb_sample` dataset.

### DIFF
--- a/packages/compass-query-bar/src/stores/ai-query-reducer.ts
+++ b/packages/compass-query-bar/src/stores/ai-query-reducer.ts
@@ -256,27 +256,30 @@ export const runAIQuery = (
       return;
     }
 
+    // The query endpoint may return the aggregation property in addition to filter, project, etc..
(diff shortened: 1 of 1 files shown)


Anemy:
Going to close this. I do still think this is something we might want, however, we might be able to make the model less likely to generate things in addition to aggregations, and we can handle it on the server side.

PR #4826, opened 2023-09-08
chore: Move all Row & Cell handling into IndexesTable COMPASS-7199

lerouxb:


packages/compass-indexes/src/components/indexes-table/indexes-table.tsx
```diff
@@ -118,21 +137,92 @@ export function IndexesTable<Shape>({
...
+          {({ datum: info }) => {
+            console.log(info.fields);
+            return (
+              <Row
+                key={info.key}
+                data-testid={info['data-testid']}
+                className={rowStyles}
+              >
+                {info.fields.map((field) => {
+                  <Cell
```
lerouxb, line 184: These render blank

packages/compass-indexes/src/components/regular-indexes-table/regular-indexes-table.tsx
```diff
@@ -50,96 +37,73 @@ export const RegularIndexesTable: React.FunctionComponent<
...
-                <IndexKeysBadge keys={index.fields} />
-              </Cell>
-            </Row>
-          </Row>
-        );
-      }}
-    </IndexesTable>
+      data={data}
+      onSortTable={(column: string, direction: SortDirection) =>
+        onSortTable(column as SortColumn, direction)
```
lerouxb, line 105: Trying to think of a way to do this without having to assert here. IndexesTable doesn't know about the specific column types and it is just data at that point.

gribnoysup (approved):

mabaasit (approved):

PR #4831, opened 2023-09-11
feat: add search indexes table COMPASS-7163 [WIP] 

lerouxb:
* [ ] tests for the table
* [ ] tests for the list view where we display one or the other
* [ ] status badge thing
* [ ] empty state
* [ ] spinner somewhere while loading


lerouxb:
lol. rebase gone wrong.

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -1,5 +1,20 @@
...
+import { createLoggerAndTelemetry } from '@mongodb-js/compass-logging';
 import { isAction } from './../utils/is-action';
+import type { IndexesThunkAction } from '.';
+
+import type { SearchIndex } from 'mongodb-data-service';
+
+const { debug } = createLoggerAndTelemetry('COMPASS-INDEXES');
+
+export type SearchSortColumn = keyof typeof sortColumnToProps;
+export type SortDirection = 'asc' | 'desc';
```
lerouxb, line 11: We now have multiple of these. Trying to think of where to put the SortDirection type and if I should bother. It is just 'asc' | 'desc' everywhere anyway.

packages/compass-indexes/src/components/indexes/indexes.tsx
```diff
@@ -87,7 +102,7 @@ export const Indexes: React.FunctionComponent<IndexesProps> = ({
         isWritable={isWritable}
         isReadonlyView={isReadonlyView}
         readOnly={readOnly}
-        errorMessage={error}
+        errorMessage={regularError || searchError}
```
lerouxb, line 105: There are now two error vars: one for regular indexes and one for search indexes. But the toolbar shows the error. I'm passing both as one var which is not ideal. What if indexes broke and then you browsed to search indexes and that broke as well? Maybe a bit of an extreme edge case, but really messy nonetheless.

packages/compass-indexes/src/components/indexes/indexes.tsx
```diff
@@ -123,23 +144,28 @@ const mapState = ({
...
   isWritable,
   isReadonlyView,
   description,
-  error,
+  regularError: regularIndexes.error,
+  searchError: searchIndexes.error,
   localAppRegistry: (appRegistry as any).localAppRegistry,
-  isRefreshing,
+  isRefreshing:
+    regularIndexes.isRefreshing || searchIndexes.status === 'REFRESHING',
```
lerouxb, line 159: We now have two variables (one for regular indexes and one for search indexes) that mean "refreshing". Which is just weird. But then the toolbar needs to know whether you're already refreshing. Will refreshing the regular indexes disable refreshing the search indexes? The design is a bit weird IMHO.

lerouxb: In fact if searchIndexes.status === 'PENDING' then that probably also counts as loading..

packages/compass-indexes/src/components/indexes/indexes.tsx
lerouxb: We disable the create index button if there's an error. Which means if one of them errors you can't create either of them. That will probably byte us eventually.

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -18,39 +33,203 @@ export enum SearchIndexesStatuses {
...
+      isReadonlyView,
+      dataService,
+      namespace,
+      searchIndexes: { sortColumn, sortOrder, status },
+    } = getState();
+
+    if (isReadonlyView) {
+      // TODO: should we just leave the status at pending?
+      return;
+    }
```
mabaasit, line 169: I think we can just return here without any status change (or maybe change the status to NOT_AVAILABLE). With `isReadonlyView` we do not show the tables, so there's no way to get here.

packages/compass-indexes/src/components/indexes/indexes.tsx
```diff
@@ -3,23 +3,29 @@ import { css, spacing } from '@mongodb-js/compass-components';
 import { connect } from 'react-redux';
 import type AppRegistry from 'hadron-app-registry';
 import { withPreferences } from 'compass-preferences-model';
+import type { SearchIndex } from 'mongodb-data-service';
 
 import {
-  sortIndexes,
+  sortRegularIndexes,
```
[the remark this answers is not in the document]
mabaasit: now i am thinking about it, maybe we should have set the active tab in redux and then have one sort function that's exposed to UI. This then checks for active tab and dispatches the correct sort (something like [this](https://github.com/mongodb-js/compass/blob/main/packages/compass-aggregations/src/modules/pipeline-builder/builder-helpers.ts#L8)).

packages/compass-indexes/src/components/indexes/indexes.tsx
mabaasit: Currently in indexes we do not have initial loading, but we can introduce that for search indexes (status === PENDING). 

These all props are tab based - regular/search and having them that way will help us avoid `regularIndexes.isRefreshing || searchIndexes.status === 'REFRESHING'` and show correct state per tab. Similarly for error. 

We can also connect these components (toolbar, search indexes table and regular indexes table) individually to redux and that can clean it up here a lot.

packages/compass-indexes/src/modules/search-indexes.ts
mabaasit: Maybe where we have `RootState` defined?

packages/compass-indexes/src/stores/store.ts
```diff
@@ -114,6 +129,8 @@ const configureStore = (options: ConfigureStoreOptions) => {
   }
 
   void store.dispatch(fetchIndexes());
+  void store.dispatch(fetchSearchIndexes());
```
mabaasit, line 132: This i think we should move to UI. And based on the active tab, dispatch the correct initial action. And this action maybe will also introduce `isLoading` on regular indexes slice and that way we can align both search/regular indexes (or maybe add status there as well and remove these flag?)

packages/compass-indexes/src/components/indexes/indexes.tsx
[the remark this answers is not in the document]
lerouxb: I think it is OK for now with the components doing connect() separately. We can also refactor more in a follow-up if we still feel it is necessary.

packages/compass-indexes/src/components/indexes-toolbar/indexes-toolbar.tsx
```diff
@@ -266,3 +276,28 @@ export const CreateIndexButton: React.FunctionComponent<
...
+
+const mapState = ({
+  isWritable,
+  isReadonlyView,
+  description,
+  serverVersion,
+  appRegistry,
+  searchIndexes,
+}: RootState) => ({
+  searchIndexes: searchIndexes.indexes,
```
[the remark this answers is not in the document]
lerouxb: yup

mabaasit (approved):

PR #4832, opened 2023-09-11
tests(search indexes): setup e2e search indexes test COMPASS-7171

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-e2e-tests/helpers/commands/create-index.ts
```diff
@@ -34,7 +34,24 @@ export async function createIndex(
   const indexName = extraOptions?.indexName ?? `${fieldName}_${indexType}`;
 
   // Open the modal
-  await browser.clickVisible(Selectors.CreateIndexButton);
+  const isWithCreateDropdown = await browser
+    .$(Selectors.CreateIndexDropdownButton)
+    .isExisting();
+  if (!isWithCreateDropdown) {
+    console.log('Creating index using normal create button');
```
lerouxb, line 41: Don't forget about your console.log lines :)

packages/compass-e2e-tests/tests/search-indexes.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    name: 'Atlas Dedicated Cluster',
+    connectionString: process.env.E2E_TESTS_ATLAS_CS_WITH_SEARCH,
+  },
+  // todo: atlas local dev
+];
+
+const DB_NAME = 'e2e_indexes_test';
+const COLL_NAME = 'numbers';
+
+describe.only('Search Indexes', function () {
```
lerouxb, line 40: And your .only

lerouxb: Oh sorry this is a draft. Carry on, then :)

packages/compass-e2e-tests/helpers/commands/create-index.ts
```diff
@@ -34,7 +34,22 @@ export async function createIndex(
   const indexName = extraOptions?.indexName ?? `${fieldName}_${indexType}`;
 
   // Open the modal
-  await browser.clickVisible(Selectors.CreateIndexButton);
+  const isWithCreateDropdown = await browser
+    .$(Selectors.CreateIndexDropdownButton)
+    .isExisting();
+  if (!isWithCreateDropdown) {
```
lerouxb, line 40: This is very nitpicky, but as a style thing can we please `if (true) {} else {}` rather than `if (!true) {} else {}`. It find this way around so confusing 😆

packages/compass-e2e-tests/helpers/commands/create-search-index.ts
```diff
@@ -0,0 +1,39 @@
...
+import * as Selectors from '../selectors';
+
+export async function createSearchIndex(
+  browser: CompassBrowser,
+  indexName: string,
+  indexDefinition: string,
+  screenshotName?: string
+): Promise<void> {
+  await browser.waitUntil(async () => {
+    await browser.clickVisible(Selectors.CreateIndexDropdownButton);
```
lerouxb, line 11: You probably want to move this outside of the wait. And then you can also just:

`await browser.$(Selectors.createIndexDropdownAction('search-indexes')).waitForDisplayed();`

which is builtin and I think a bit more idiomatic.

packages/compass-e2e-tests/tests/search-indexes.test.ts
```diff
@@ -0,0 +1,200 @@
...
+        {
+          const searchIndexes = await dbInstance
+            .collection(COLL_NAME)
+            .listSearchIndexes()
+            .toArray();
+
+          if (searchIndexes.length > 0) {
+            await Promise.all(
+              searchIndexes.map(({ name }) =>
+                dbInstance.collection(COLL_NAME).dropSearchIndex(name)
```
lerouxb, line 159: Won't this run other CI tasks' search indexes? Maybe we should rather create random/unique search index names and then in an afterEach clean them up. We then also don't have to wait for them to be deleted because it will happen after the tests end and we won't care about the name being available because it is random and unique.

packages/compass-e2e-tests/tests/search-indexes.test.ts
```diff
@@ -0,0 +1,200 @@
...
+            );
+            // Wait for a minute to ensure the indexes have been deleted.
+            await wait(60000);
+          }
+        }
+      });
+      afterEach(async function () {
+        await afterEachTestRun(this.currentTest);
+      });
+      it('allows users to create a regular indexes', async function () {
```
lerouxb, line 170: This test should probably be in the regular indexes test file?

packages/compass-e2e-tests/tests/search-indexes.test.ts
```diff
@@ -0,0 +1,200 @@
...
+        await browser.createSearchIndex(INDEX_NAME, INDEX_DEFINITION);
+        await browser.waitForAnimations(Selectors.SearchIndexList);
+
+        // View it
+        await browser
+          .$(Selectors.searchIndexRow(INDEX_NAME))
+          .waitForDisplayed();
+
+        // Drop it
+        await browser.dropSearchIndex(INDEX_NAME);
```
lerouxb, line 193: Maybe a good idea to wait and see if the index eventually goes away from the list too. Or is that a TODO for once we poll?

packages/compass-indexes/src/components/indexes-toolbar/indexes-toolbar.tsx
```diff
@@ -164,8 +164,12 @@ export const IndexesToolbar: React.FunctionComponent<IndexesToolbarProps> = ({
                 className={alignSelfEndStyles}
                 label="Viewing"
                 defaultValue="regular-indexes"
+                data-testid="indexes-segment-controls"
               >
-                <SegmentedControlOption value="regular-indexes">
+                <SegmentedControlOption
+                  data-testid="regular-indexes"
```
lerouxb, line 170: These test ideas are bound to clash eventually when we lookup a test id at a page level. ie. quite possibly with the list or something. maybe view-regular-indexes-option and view-search-indexes-option or something like that?

packages/compass-e2e-tests/tests/search-indexes.test.ts
mabaasit: i added an expectation to check if the status changes to `DELETING` in [2aab6d3](https://github.com/mongodb-js/compass/pull/4832/commits/2aab6d34c88449de8b95e2f3dca652abe7cb81c1)

packages/compass-e2e-tests/tests/search-indexes.test.ts
```diff
@@ -0,0 +1,188 @@
...
+        await browser.waitUntil(async function () {
+          const text = await browser.$(rowSelector).getText();
+          return text.indexOf('PENDING') > -1;
+        });
+
+        // Drop it
+        await browser.dropSearchIndex(indexName);
+        await browser.waitUntil(async function () {
+          const text = await browser.$(rowSelector).getText();
+          return text.indexOf('DELETING') > -1;
```
lerouxb, line 180: I think this will be flaky because waitUntil polls and deleting could be there only for a split second. Plus we'll poll to update the list as well so the row could be gone before we ever see deleting. If they ever speed things up it could be gone by the time we first refresh the list too.

So I think safest here would be to either only do this once we poll and then wait for the row to be gone entirely OR you'd have to check for either the status to become DELETING OR the row to disappear. Maybe a Promise.race()?

lerouxb: In more general terms: If you're going to wait for something in an e2e test you have to wait for a stable state that won't change until the user intervenes. Waiting for an intermediate state that will change by itself will cause flakyness down the line. If you're lucky this works until we add polling and then hopefully it breaks reliably. Otherwise it becomes the kind of thing that breaks rarely and is hard to debug.

lerouxb (approved):

PR #4845, opened 2023-09-14
feat(compass-preferences-model): add preference for enabling ai features COMPASS-7147

Anemy:
COMPASS-7147
This new setting `enableAIFeatures` can be set in the ui, cli, and global config file. Any ideas for better names?
With this we then allow a user to use the ai feature when: 
- This setting is set true.
- They receive a successful hello feature enabled response.
- We have enabled the feature flag (from `development` to `preview`). 

Here's how it looks:
Screenshot 2023-09-14 at 12 23 23 AM

### DIFF
--- a/packages/atlas-service/src/main.ts
+++ b/packages/atlas-service/src/main.ts
@@ -93,7 +93,10 @@ export async function throwIfNotOk(
 }
 
 function throwIfAINotEnabled(atlasService: typeof AtlasService) {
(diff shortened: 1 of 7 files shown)


mcasimir:
Q: enable AI Features is visible also if is not rolled out for you?

Anemy:
@mcasimir Updated to only show it when users have the feature rolled out to them.

mcasimir (approved):

PR #4854, opened 2023-09-15
feat(search-indexes): new modal to update search indexes COMPASS-7166

kmruiz:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-indexes/src/components/search-indexes-modals/base-search-index-modal.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+    const indexDefinitionDoc = parseShellBSON(indexDefinition);
+    submitIndex(indexName, indexDefinitionDoc);
+  }, [submitIndex, parsingError, indexName, indexDefinition]);
+
+  useTrackOnChange(
+    'COMPASS-SEARCH-INDEXES-UI',
+    (track) => {
+      if (isModalOpen) {
+        track('Screen', { name: 'create_search_index_modal' });
```
mabaasit, line 112: This will always track as `create` and not update. I don't think we have existing event for update (regular) index, but probably need to introduce one. 

(Or you can completely remove it in this PR and then we can handle this in [COMPASS-7170](https://jira.mongodb.org/browse/COMPASS-7170))

packages/compass-indexes/src/components/search-indexes-modals/base-search-index-modal.tsx
```diff
@@ -0,0 +1,191 @@
...
+    [isModalOpen],
+    undefined,
+    React
+  );
+
+  return (
+    <Modal
+      open={isModalOpen}
+      setOpen={closeModal}
+      data-testid="create-search-index-modal"
```
mabaasit, line 124:
```suggestion
      data-testid="search-index-modal"
```

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -88,14 +93,45 @@ type CreateSearchIndexCancelledAction = {
   type: ActionTypes.CreateSearchIndexCancelled;
 };
 
+type OpenUpdateSearchIndexModalAction = {
+  type: ActionTypes.OpenUpdateSearchIndexModal;
+  indexName: string;
+  indexDefinition: string;
```
mabaasit, line 99: If we just store the `indexName` in the state and then get the definition when the modal is open (when connecting component with redux)?

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -88,14 +93,45 @@ type CreateSearchIndexCancelledAction = {
...
   isModalOpen: boolean;
   isBusy: boolean;
 };
 
+type UpdateSearchIndexState = {
+  isModalOpen: boolean;
+  isBusy: boolean;
+  indexName: string;
+  indexDefinition: string;
+};
```
mabaasit, line 129:
```suggestion
type UpdateSearchIndexState = {
  isModalOpen: false;
} | {
  isModalOpen: true;
  isBusy: boolean;
  indexName: string;
};
```

This way we do not have to set the indexName as default. We can just set it as `updateIndex: { isModalOpen: false }`

packages/compass-indexes/src/components/search-indexes-modals/base-search-index-modal.tsx
kmruiz: I'll remove it entirely then, so we can implement it properly in COMPASS-7170.

packages/compass-indexes/src/modules/search-indexes.ts
kmruiz: If we do this, the spread operator won't work in the reducer, as it can not infer the current type (isModalOpen can change on runtime), so we will need to specify every field one by one.

packages/compass-indexes/src/modules/search-indexes.ts
kmruiz: I've though about it, no strong opinion TBH. I did it this way because in the component that triggers the action [IndexActions](./packages/compass-indexes/src/components/search-indexes-table/search-index-actions.tsx), we already have the latestDefinition and the indexName. It was straightforward to pass them up and we get all the information from the same place so it's pretty obvious where the data comes from.

packages/compass-indexes/src/modules/search-indexes.ts
mabaasit: don't have a strong opinion here, but if we leave it as it, then we do not need to have the default value for the indexName. technically it can be a null or a string.

packages/compass-indexes/src/modules/search-indexes.ts
mabaasit: this came up during the Technical Design and we agreed to have it such - just have the index identifier in `update` object and use that to find the corresponding index definition.

kmruiz:
I can't 'approve' because I started the PR, but it's ok for me, so feel free to merge it @mabaasit

lerouxb (approved):

PR #4860, opened 2023-09-18
feat(search-indexes): add telemetry COMPASS-7170

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-indexes/src/components/search-indexes-modals/create-search-index-modal.tsx
+++ b/packages/compass-indexes/src/components/search-indexes-modals/create-search-index-modal.tsx
@@ -103,6 +103,9 @@ export const CreateSearchIndexModal: React.FunctionComponent<
     (track) => {
       if (isModalOpen) {
         track('Screen', { name: 'create_search_index_modal' });
+        track('Index Create Opened', {
+          atlas_search: true,
+        });
       }
     },
     [isModalOpen],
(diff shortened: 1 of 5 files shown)


packages/compass-indexes/src/components/search-indexes-modals/create-search-index-modal.tsx
```diff
@@ -103,6 +103,9 @@ export const CreateSearchIndexModal: React.FunctionComponent<
     (track) => {
       if (isModalOpen) {
         track('Screen', { name: 'create_search_index_modal' });
+        track('Index Create Opened', {
```
Anemy, line 106: Should we ask if we can use the `Screen` event with `create_search_index_modal` instead? It seems like we're duplicating the call here when we could filter for the same event in our telemetry. Fine as is too, this is a nit on reducing duplication, feel free to merge as is imo.

packages/compass-indexes/src/stores/create-index.js
```diff
@@ -47,7 +47,9 @@ const configureStore = (options = {}) => {
     });
 
     localAppRegistry.on('open-create-index-modal', () => {
-      track('Index Create Opened');
+      track('Index Create Opened', {
```
Anemy, line 50: Should we have this in the `create-index-modal` instead for consistency with how we do with search indexes then? Similar comment with the `Screen` event duplication as commented on above then.

mabaasit: I aligned the tracking and will follow up with product if we can remove the duplicate events. If agreed, i'll clean it in a follow up :)

PR #4865, opened 2023-09-19
feat(data-service): setup local atlas detection COMPASS-7213

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/data-service/src/instance-detail-helper.ts
+++ b/packages/data-service/src/instance-detail-helper.ts
@@ -118,6 +118,7 @@ export async function getInstance(
     hostInfoResult,
     buildInfoResult,
     getParameterResult,
+    atlasVersionResult,
     isLocalAtlas,
   ] = await Promise.all([
     runCommand(adminDb, { connectionStatus: 1, showPrivileges: true }).catch(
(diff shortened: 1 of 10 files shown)


packages/data-service/src/instance-detail-helper.ts
```diff
@@ -133,12 +135,19 @@ export async function getInstance(
...
-    runCommand(adminDb, { atlasVersion: 1 }).catch(() => {
-      return { atlasVersion: '', gitVersion: '' };
-    }),
+    checkIsLocalAtlas(
+      async (db: string, collection: string, query: Document) => {
+        const res = await runCommand(client.db(db), {
+          count: collection,
+          query,
+        });
+        return res.n;
```
addaleax, line 144: Similar to my [comment in the mongosh PR for this](https://github.com/mongodb-js/mongosh/pull/1629#discussion_r1309091281), I don’t think we’ll want to be using the `count` command and instead use the driver wrapper, if possible?

packages/data-service/src/instance-detail-helper.ts
```diff
@@ -133,12 +135,19 @@ export async function getInstance(
       getParameter: 1,
       featureCompatibilityVersion: 1,
     }).catch(() => null),
-
-    runCommand(adminDb, { atlasVersion: 1 }).catch(() => {
-      return { atlasVersion: '', gitVersion: '' };
-    }),
```
addaleax, line 139: Would we not run `atlasVersion: 1` anymore? This seems to always have been one of the better Atlas identification checks (although it only works for shared-tier clusters)

mabaasit (commented):
can't approve my own pr, but looks good! thanks for the resolution :)

PR #4883, opened 2023-09-20
feat(search-indexes): polling COMPASS-7237

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -505,29 +501,52 @@ export const fetchSearchIndexes = (): IndexesThunkAction<
       return;
     }
 
-    if (!dataService || !dataService.isConnected()) {
-      debug('warning: trying to load indexes but dataService is disconnected');
+    const notFetchableStatuses: SearchIndexesStatus[] = [
```
kmruiz, line 504: Probably this could be extracted out, into a global constant, so it's not recreated on every call. Not too expensive and the runtime could optimise it, but the scoping seems more clear.

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -505,29 +501,52 @@ export const fetchSearchIndexes = (): IndexesThunkAction<
...
+      'NOT_AVAILABLE',
+      'FETCHING',
+      'POLLING',
+      'REFRESHING',
+    ];
+    // If we are currently doing fetching indexes, we will
+    // wait for that
+    if (notFetchableStatuses.indexOf(status) !== -1) {
```
kmruiz, line 512: Picky, but we can use notFetchableStatuses.includes(status).

packages/compass-indexes/src/modules/search-indexes.ts
mabaasit: added in [13d92db](https://github.com/mongodb-js/compass/pull/4883/commits/13d92db4ddd4b56db150eea6594cb8febdc1182c)

packages/compass-indexes/src/modules/search-indexes.ts
mabaasit: added in [13d92db](https://github.com/mongodb-js/compass/pull/4883/commits/13d92db4ddd4b56db150eea6594cb8febdc1182c)

kmruiz (approved):

packages/compass-indexes/src/components/search-indexes-table/search-indexes-table.tsx
```diff
@@ -163,7 +168,15 @@ export const SearchIndexesTable: React.FunctionComponent<
   onEditIndex,
   status,
   onDropIndex,
+  onPollIndexes,
 }) => {
+  useEffect(() => {
+    const id = setInterval(onPollIndexes, POLLING_INTERVAL);
```
lerouxb, line 174: I'm wondering if we shouldn't rather use a setTimeout that we only start once the polling finishes. Just wondering what would happen if search indexes gets slower than 5 seconds yet every 5 seconds we kick off another poll..

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -510,24 +515,41 @@ export const fetchSearchIndexes = (): IndexesThunkAction<
...
 
-export const refreshSearchIndexes = (): IndexesThunkAction<void> => {
+export const refreshSearchIndexes = (): IndexesThunkAction<Promise<void>> => {
+  return async (dispatch, getState) => {
+    const { indexes } = getState().searchIndexes;
+
+    // If we have a list of indexes already, then we are refreshing
+    // or else we are fetching them.
+    const newStatus: SearchIndexesStatus =
+      indexes.length > 0 ? 'REFRESHING' : 'FETCHING';
```
lerouxb, line 545: Wondering about the common case of having no search indexes yet. Then it will be FETCHING every time and what we probably want is REFRESHING? Wondering if we should base this off the previous status rather.

packages/compass-indexes/src/components/search-indexes-table/search-indexes-table.tsx
mabaasit: if it gets slower, then the current one will be still running and the status is `POLLING` and then when fetching we check and skip another fetch

packages/compass-indexes/src/modules/search-indexes.ts
mabaasit: I added another status to keep track of that, `NOT_READY` and in that state we have not fetched anything yet. once we do that, we do not go back to that state again.

lerouxb: What is FETCHING, then? Can't it just be NOT_READY?

lerouxb: Seems like you can only transition from NOT_READY to FETCHING or from READY to REFRESHING. then from FETCHING or REFRESHING to READY. or READY to POLLING and back to READY.

mabaasit: Yes, from NOT_READY to FETCHING once. And FETCHING is like the initial loading state where we probably want to show centered spinner. 

And to READY state, we can get from all the states that fetch data (FETCHING, REFRESHING or POLLING).

lerouxb (approved):

PR #4887, opened 2023-09-20
fix(search-indexes): clean up the modal state and do not update on no change

mabaasit:
### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-indexes/src/modules/search-indexes.ts
+++ b/packages/compass-indexes/src/modules/search-indexes.ts
@@ -1,4 +1,5 @@
 import type { AnyAction } from 'redux';
+import { isEqual } from 'lodash';
 import { createLoggerAndTelemetry } from '@mongodb-js/compass-logging';
 import { isAction } from './../utils/is-action';
 import {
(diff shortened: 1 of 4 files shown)


packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -450,7 +451,20 @@ export const updateIndex = (
...
+      namespace,
+      dataService,
+      searchIndexes: { indexes },
+    } = getState();
+
+    const currentIndexDefinition = indexes.find(
+      (x) => x.name === indexName
+    )?.latestDefinition;
+    if (isEqual(currentIndexDefinition, indexDefinition)) {
+      dispatch(closeUpdateModal());
```
lerouxb, line 464: I can't immediately see where we close the update modal in the normal case?

lerouxb: (Was trying to run this locally and manually test it but now bootstrap is broken for me. unrelated to this PR, though..)

mabaasit: do we want to show the error message to the user in this case? I am not sure how to handle this case and the best for me is to dismiss the modal as user clicked submit. Maybe we show toast with text "No changes made to the index" or something?

mabaasit: > (Was trying to run this locally and manually test it but now bootstrap is broken for me. unrelated to this PR, though..)

It frequently fails for me since we updated `lerna` (i think so)

lerouxb: It was just this again:

```
~/mongo/compass % python3

You have not agreed to the Xcode and Apple SDKs license. You must agree to the license below in order to use Xcode.
Press enter to display the license:
```

lerouxb (approved):

PR #4888, opened 2023-09-20
chore(atlas-service): always use mms prod, rename preference for atlas service backend COMPASS-7235

Anemy:
COMPASS-7235

Updates `atlas-service`'s backend to mms prod. Renames the preference for the atlas config as it would be otherwise defaulting to whatever was previously there in Compass (this is a bug in our preference user data loading).

### DIFF
--- a/packages/compass-preferences-model/src/preferences.ts
+++ b/packages/compass-preferences-model/src/preferences.ts
@@ -44,10 +44,6 @@ export type UserConfigurablePreferences = PermanentFeatureFlags &
     maxTimeMS?: number;
     installURLHandlers: boolean;
     protectConnectionStringsForNewConnections: boolean;
-    // This preference is not a great fit for user preferences, but everything
-    // except for user preferences doesn't allow required preferences to be
-    // defined, so we are sticking it here
-    atlasServiceConfigPreset: 'compass-dev' | 'compass' | 'atlas-dev' | 'atlas';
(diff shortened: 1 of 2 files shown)


packages/compass-preferences-model/src/preferences.ts
```diff
@@ -635,25 +631,6 @@ export const storedUserPreferencesProps: Required<{
...
     type: 'boolean',
   },
-  /**
-   * Chooses atlas service backend configuration from preset
-   *  - compass-dev: locally running compass kanopy backend (localhost)
-   *  - compass:    compass kanopy backend (compass.mongodb.com)
-   *  - atlas-dev:  dev mms backend (cloud-dev.mongodb.com)
-   *  - atlas:      mms backend (cloud.mongodb.com)
-   */
-  atlasServiceConfigPreset: {
```
Anemy, line 645: @gribnoysup was this preference intended to be kept in?

mcasimir: This was meant to stay yeah, also definitely a bit premature to do this operation

Anemy: Chatting on slack, we're thinking let's keep the preference, but rename it and default it to `atlas` so that it's consistent for everyone.

mcasimir (approved):

addaleax (approved):

PR #4890, opened 2023-09-20
chore(atlas-service): update ai query response handling with aggregation key COMPASS-7241

Anemy:
Updates how we parse the ai query response when there's an aggregation. Previously we had `aggregation` inside of the `query` object. Now it's at the top level.

Similar update on Compass' backend: https://github.com/10gen/compass-mongodb-com/pull/99


packages/compass-query-bar/src/stores/ai-query-reducer.ts
```diff
@@ -40,7 +40,6 @@ export const enum AIQueryActionTypes {
   AIQueryCancelled = 'compass-query-bar/ai-query/AIQueryCancelled',
   AIQueryFailed = 'compass-query-bar/ai-query/AIQueryFailed',
   AIQuerySucceeded = 'compass-query-bar/ai-query/AIQuerySucceeded',
-  AIQueryReturnedAggregation = 'compass-query-bar/ai-query/AIQueryReturnedAggregation',
```
Anemy, line 43: This wasn't listened to anywhere, so I removed it.

Anemy:
Merged in without reviewers since it's fairly small and we'll want it in a beta tomorrow.
Feel free to revert if anything doesn't look right.

PR #4892, opened 2023-09-21
feat(search-indexes): add a drop down to choose a search index template COMPASS-7173

kmruiz:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-indexes/src/components/search-indexes-modals/base-search-index-modal.tsx
```diff
@@ -141,6 +163,16 @@ export const BaseSearchIndexModal: React.FunctionComponent<
     onSubmit(indexName, indexDefinitionDoc);
   }, [onSubmit, parsingError, indexName, indexDefinition]);
 
+  const onChangeTemplate = useCallback(
+    (template: SearchTemplate) => {
+      editorRef.current?.applySnippet(template.snippet);
+      rafraf(() => {
+        editorRef.current?.focus();
+      });
```
gribnoysup, line 171: Focusing after apply can mess with snippet logic, so you should rather apply after focus, both calls should be done from the next frame:

```suggestion
      rafraf(() => {
        editorRef.current?.focus();
        editorRef.current?.applySnippet(template.snippet);
      });
```

We learned it when integrating codemirror snippets with [select in aggregation builder](https://github.com/mongodb-js/compass/blob/70e22b25fdaf21b15fc45d4873cbb4804c0072c8/packages/compass-aggregations/src/components/stage.tsx#L172-L179)

kmruiz: Will do the change, thanks a lot! I've found that without rafraf it did something weird sometimes, but didn't know that we should focus before applying the snippet.

packages/compass-indexes/src/components/search-indexes-modals/base-search-index-modal.tsx
```diff
@@ -40,6 +44,22 @@ const bodyStyles = css({
   gap: spacing[3],
 });
 
+const templateToolbarStyles = css({
+  display: 'flex',
+  flexDirection: 'row',
+  gap: spacing[3],
+  overflow: 'hidden',
```
gribnoysup, line 51: Because of the overflow hidden this selector gets cut off when switching between the options (and even when nothing is selected URL is not visible):

image

packages/compass-indexes/src/components/search-indexes-modals/base-search-index-modal.tsx
```diff
@@ -141,6 +163,17 @@ export const BaseSearchIndexModal: React.FunctionComponent<
     onSubmit(indexName, indexDefinitionDoc);
   }, [onSubmit, parsingError, indexName, indexDefinition]);
 
+  const onChangeTemplate = useCallback(
+    (template: SearchTemplate) => {
+      rafraf(() => {
+        editorRef.current?.focus();
+        editorRef.current?.applySnippet(template.snippet);
+        editorRef.current?.prettify();
```
gribnoysup, line 171: Why snippets are not formatted already in the constants? Running prettify can mess up with apply selection range focus and there is generally no guarantee that snippet placeholders are valid JS, so this might just not do anything

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const dropdownStyles = css({
+  display: 'flex',
+  overflow: 'hidden',
+  gap: 0,
+});
+
+const overlapDropdownLabelStyles = css({
+  marginLeft: -spacing[3],
+});
```
gribnoysup, line 21: This is a very unconventional way of positioning this icon, flex is not a good way to do it and you are hacking on top of it to align the icon inside the label. While leafygreen types only allow to pass strings to the label, you can actually pass components there, or, if we don't want to do that, leafygreen label is not a good match here and you need to add your own label with an icon inside

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  Option,
+  Icon,
+  css,
+  spacing,
+  Tooltip,
+} from '@mongodb-js/compass-components';
+
+const dropdownStyles = css({
+  display: 'flex',
+  overflow: 'hidden',
```
gribnoysup, line 15: This overflow hidden in combination with a margin hack for the icon makes the right border of the input to disappear. You generally only want to use overflow when you intentionally want to not allow scrolling and hide stuff that is out of the element box

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  gap: 0,
```
gribnoysup, line 16: You don't need to specify gap if it's not needed / zero

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+    (value: string) => {
+      setTemplateValue(value);
+      onTemplate(ATLAS_SEARCH_TEMPLATES[+value]);
+    },
+    [onTemplate]
+  );
+
+  return (
+    <section className={dropdownStyles}>
+      <Select
```
gribnoysup, line 43: If your container is of a fixed size of 40%, this should probably stretch to the whole container, otherwise because some of the labels are shorter, some are longer, you get weird empty space on the side:

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+            {template.name}
+          </Option>
+        ))}
+      </Select>
+      <Tooltip
+        align="right"
+        triggerEvent="hover"
+        trigger={({ children, ...props }) => (
+          <>
+            {children}
```
gribnoysup, line 60: The tooltip will be aligned with the parent whatever place the children are rendered, because you are rendering this in the fragment, it just randomly aligns itself with the section here, is this expected?

image

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
kmruiz: Yeah I'm not sure about this solution either, but the design adds an information icon to a component that doesn't have it. So I isolated it in a single component in case we want to change it in the future or not and implemented the easiest way that shouldn't break the layout (tested in different sizes of the popup).

I can try passing a component for the label itself, that has the flex config inside, and see if it works properly.

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
kmruiz: Did you change anything in the code?  The alignment should be on the right side.

Taking a look at the alignment though.

packages/compass-indexes/src/components/search-indexes-modals/base-search-index-modal.tsx
kmruiz: This was actually is a scenario that Basit and I were testing. We were checking if prettify would help at this point, but you already answered, so I'm removing it.

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
kmruiz: Yeah, I usually make it explicit even if it's not inherited. I do it even in CSS, but if it's a problem I can just remove it.

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
kmruiz: I'll try the suggestion to use a component instead of a string for the label, that should fix it.

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
gribnoysup: Feel free to keep it, but specifying properties that seemingly do nothing just raises questions, especially after some time passed, for whoever will be working with this code or looking at it.

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
kmruiz: With a component it works and the type system does not complain, so I'll go that way. Thanks a lot @gribnoysup!

gribnoysup: Interesting, in their types that I see they only allow strings (but yeah, anyway, the way they handle the prop we can pass components there):

image

mcasimir:
NIT: how come we are excluding this PR from release notes? unless we want to merge this before the next GA we probably want it there

kmruiz:
We decided to leave it for the next GA as we've already done the testing and it would mean doing the whole test matrix again today.

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
kmruiz: Oh, understood. With the changes I've done, as you suggested, now it aligns correctly.

Screenshot 2023-09-22 at 12 11 50

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
kmruiz: I changed a bit the styles with your suggestions and removed the gap: 0 so it's now a bit simpler.

packages/compass-indexes/src/components/search-indexes-modals/base-search-index-modal.tsx
kmruiz: Removed the overflow hidden and tested different sizes, now it should not get cut off. Thanks for pointing out!

mcasimir:
> We decided to leave it for the next GA as we've already done the testing and it would mean doing the whole test matrix again today.

Don't we want this PR appear in the release notes of the GA where it lands?

kmruiz:
You are right, changed it.

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
gribnoysup: > I've checked and it didn't complain for some reason when compiling, but it failed with check-ci, so there is something definitely weird there

That's expected, we don't break the locally running dev build on compilation errors, otherwise you'd have failures all the time while working on a feature locally and this is not productive. This issues are something that for most of us would be highlighted by typescript or eslint plugin in vscode before you catch it by explicitly running check command

packages/compass-indexes/src/components/search-indexes-modals/base-search-index-modal.tsx
gribnoysup: Did you revert this change or something went wrong with the merge from main? The hidden is still here a still causes issues:

image

kmruiz: I think I did something wrong merging with main, I'm merging again (more changes 😄) and I'll remove the hidden.

packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+      value={templateValue}
+      allowDeselect={false}
+      onChange={onChooseTemplate}
+      label={
+        (
+          <SearchIndexTemplateDropdownLabel
+            label="Template"
+            tooltip={tooltip}
+          />
+        ) as unknown as string
```
lerouxb, line 78: 🙈

packages/compass-indexes/src/components/search-indexes-modals/base-search-index-modal.spec.tsx
```diff
@@ -1,14 +1,24 @@
...
 import React from 'react';
 import { getCodemirrorEditorValue } from '@mongodb-js/compass-editor';
 
+function normalizedTemplateNamed(name: string) {
+  const snippet =
+    ATLAS_SEARCH_TEMPLATES.find((t) => t.name === name)?.snippet || '';
+  // code mirror 'changes' the template placeholders, so let's do the same
+  // this regexp removes `tab` markers to their default value, for example:
+  // ${1:default} => default
+  return snippet.replace(/\${\d:([<>a-z ]+)}/gm, '$1');
```
lerouxb, line 19: Shouldn't `\d` be `\d+` in case we ever have more than 9? I haven't really seen these, so no idea if that's possible. 

And maybe `[<>a-z ]` could just be `[^}]`? Wondering what happens with capitals and numbers and other punctuation.

lerouxb (approved):

gribnoysup (approved):

packages/compass/package.json
```diff
@@ -212,7 +212,7 @@
     "@mongodb-js/mocha-config-compass": "^1.3.1",
     "@mongodb-js/mongodb-downloader": "^0.2.1",
     "@mongodb-js/prettier-config-compass": "^1.0.1",
-    "@mongodb-js/sbom-tools": "^0.5.2",
+    "@mongodb-js/sbom-tools": "^0.5.3",
```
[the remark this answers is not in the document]
kmruiz: I think the depalign --fix updated it. I'll revert this change.

packages/compass-indexes/src/components/search-indexes-modals/base-search-index-modal.spec.tsx
kmruiz: I'm not sure we can have more than 9, but let me apply the fixes, as they should be semantically the same, and it's already under tested, so we'll find out if it breaks.

PR #4915, opened 2023-09-26
feat: open search indexes table when a search index is created COMPASS-7247

kmruiz:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-indexes/src/modules/index-view.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  return state;
+}
+
+export const changeIndexView = (view: IndexView): ChangeIndexViewAction => ({
+  type: ActionTypes.ChangeIndexView,
+  view: view,
+});
+
+export const changedToRegularIndexes = () => changeIndexView('regular-indexes');
```
lerouxb, line 30: nit: Not crazy about the naming, but not sure what's better. switchToRegularIndexes and switchToSearchIndexes? Just feels like this should be an action / verb in the present tense given that actions do things, they don't just notify that a thing happened.

lerouxb (approved):

packages/compass-indexes/src/components/indexes-toolbar/indexes-toolbar.tsx
```diff
@@ -299,6 +302,7 @@ const mapState = ({
 const mapDispatch = {
   onCreateRegularIndex,
   onCreateSearchIndex,
+  onChangeIndexView: changeIndexView,
```
mabaasit, line 305: We can call this method directly from the segmented control and remove `onChangeIndexesSegment`

mabaasit (approved):

PR #4941, opened 2023-10-03
chore: add atlas-local to simplify local dev against mms

mcasimir:
`ELECTRON_EXTRA_ARGS="--atlasServiceBackendPreset=atlas-local" npm run start`

### DIFF
--- a/packages/compass/src/main/application.ts
+++ b/packages/compass/src/main/application.ts
@@ -145,6 +145,16 @@ class CompassApplication {
         },
         authPortalUrl: 'https://account-dev.mongodb.com/account/login',
       },
+      'atlas-local': {
+        atlasApiBaseUrl: 'http://localhost:8080',
+        atlasApiUnauthBaseUrl:
+          'https://cloud-dev.mongodb.com/api/private/unauth',
+        atlasLogin: {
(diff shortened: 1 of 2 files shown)


packages/compass/src/main/application.ts
```diff
@@ -145,6 +145,16 @@ class CompassApplication {
         },
         authPortalUrl: 'https://account-dev.mongodb.com/account/login',
       },
+      'atlas-local': {
+        atlasApiBaseUrl: 'http://localhost:8080',
+        atlasApiUnauthBaseUrl:
+          'https://cloud-dev.mongodb.com/api/private/unauth',
```
Anemy, line 151: Should this also be local host?
```suggestion
          'http://localhost:8080/api/private/unauth',
```
I can give it a manual test on local if it helps to confirm.

Anemy (commented):
Added questions on pathing and hello endpoint that we might want to confirm with a manual local test.

packages/compass/src/main/application.ts
mcasimir: that would be awesome!

Anemy (approved):
@mcasimir Ran it against local mms. Looks good 👍

PR #4949, opened 2023-10-04
chore: remove gnome-keyring as a dep [WIP]

lerouxb:
Based on https://github.com/atom/node-keytar/commit/9d6821d8c57ae4744c07bdc10ea2910b86d15866 which is 6 years old it seems like we don't need gnome-keyring as a dep anymore and libsecret is enough.


lerouxb:
Closing for now. In combination with https://github.com/mongodb-js/compass/pull/4962 it would make compass be _installable_ on amazon linux 2023 at least. Whether it would actually be usable I don't know.

PR #4959, opened 2023-10-06
test(search-indexes): increase timeout COMPASS-7303

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-e2e-tests/tests/search-indexes.test.ts
```diff
@@ -233,6 +233,11 @@ describe('Search Indexes', function () {
           return this.skip();
         }
         currentConnectionString = connectionString;
+        /**
+         * Increasing the timeout to 4mins to allow for Search Index
+         * to be created and be queryable on Atlas.
+         */
+        this.timeout(240_000);
```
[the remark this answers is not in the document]
mabaasit: oh i didn't realize there're various timeouts. i added [implicit](https://v5.webdriver.io/docs/timeouts.html#session-implicit-wait-timeout) timeout to wait for elements to be found [77c1d64](https://github.com/mongodb-js/compass/pull/4959/commits/77c1d64f4bcf3957b0dad254395532793c4923ad)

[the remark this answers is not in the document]
mabaasit: it did not seem to work, so i added timeout for individual waits where needed.

packages/compass-e2e-tests/tests/search-indexes.test.ts
```diff
@@ -41,6 +41,11 @@ const INDEX_DEFINITION = JSON.stringify({
   },
 });
 
+// The current timeout (2mins) is not enough for the search indexes to be created
+// and be queryable on Atlas. So we are increasing the timeout to 4mins.
+// This can not be more than 4mins as the mocha timeout is 4mins.
+const WAIT_TIMEOUT = 240_000;
```
lerouxb, line 47: This is the same timeout as mocha so I worry that the mocha test timeout will be hit before the waitUntil below (since the test contains it and is this slower overall). So the test would still fail like it would if the waitUntil times out but then we'd get a worse error message because it wouldn't point at the wait line.

packages/compass-e2e-tests/tests/search-indexes.test.ts
```diff
@@ -114,7 +119,10 @@ async function dropSearchIndex(browser: CompassBrowser, indexName: string) {
   await modal.waitForDisplayed({ reverse: true });
 
   await browser.waitUntil(async () => {
-    return await indexRow.waitForExist({ reverse: true });
+    return await indexRow.waitForExist({
```
lerouxb, line 122: I think the waitUntil wrapping this is superfluous? I get that that was already the case before. But also the waitUntil's using the default timeout and that's shorter than this one used by the waitForExist.

mabaasit: Yep, that makes sense. Will remove it

packages/compass-e2e-tests/tests/search-indexes.test.ts
```diff
@@ -228,6 +235,8 @@ describe('Search Indexes', function () {
 
   for (const { name, connectionString } of connectionsWithSearchSupport) {
     context(`supports search indexes in ${name}`, function () {
+      // Set the mocha timeout to 6mins to accomodate the 4mins wait timeout
+      this.timeout(360_000);
```
lerouxb, line 239: nit: You're hardcoding this one but have the other one as a variable at the top. Would probably make sense to have them both as vars at the top then you can explain it all in one comment.

lerouxb (approved):

PR #4966, opened 2023-10-09
feat(crud): clear values if 'With Field Values' is unchecked /COMPASS-4134

lerouxb:
Needs tests. Just pushing it so people can try it out and see if it feels right.

Screenshot 2023-10-09 at 16 36 04

Screenshot 2023-10-09 at 16 36 10

Screenshot 2023-10-09 at 16 36 20

Screenshot 2023-10-09 at 16 36 32


packages/compass-crud/src/components/insert-json-document.tsx
```diff
@@ -23,34 +34,86 @@ type InsertJsonDocumentProps = {
...
+          onChangeText={onChangeText}
+          initialJSONFoldAll={false}
+          minLines={18}
+        />
+      </div>
+      <Checkbox
+        className={checkboxStyles}
+        data-testid="with-field-values"
+        onChange={onChangeBlank}
+        label="With Field Values"
```
addaleax, line 113: How is a user expected to know what “With Field Values” means? Without playing around with it, I would expect it to modify the behavior of the insertion itself, not the editor above it. I see the conversation in the linked ticket about this, but this just really seems like bad UX, and if we do go with a checkbox here, we should really adjust the phrasing to be better.

packages/compass-crud/src/components/insert-json-document.tsx
```diff
@@ -23,34 +34,86 @@ type InsertJsonDocumentProps = {
...
 
+function blankLeafValues(jsonDoc: string) {
+  return (
+    jsonDoc
+      // strings
+      .replace(/: "[^"]+"/g, ': ""')
+      // numbers
+      .replace(/: -?\d+(\.\d+)?/g, ': ')
+      // booleans
+      .replace(/: (true|false)/g, ': ')
```
addaleax, line 45: Are we using regular expressions here because we want this to work when the input isn’t valid JSON? It’s surprising, given that we already start out with valid JSON

lerouxb: I just hacked it because I agree this is weird UX.

But how else would you turn { foo: 5 } into { foo: }? I guess we can walk the object and serialise it ourselves? Would also require valid JSON and in all cases we seem to have valid JSON formatted at this point anyway.

addaleax: We could walk the object, replace values that we want to get rid of with a placeholder (that is sufficiently unlikely to appear in real-world usage) and then do string replacement to substitute it with an empty string. I think that would be a bit cleaner than regexes, but yeah, it does require us starting off with something that’s reliably parsable.

lerouxb:
Closing because our plan has changed.

PR #4974, opened 2023-10-10
chore: avoid process.nextTick() and setImmediate() COMPASS-7314

addaleax:
Replace, or where unused remove, usage of `process.nextTick()` and `setImmediate()`, both of which are Node.js-specific APIs for async deferral for which browser-compatible replacements with near-identical behavior exist.

Add linting rules to prevent usage of these methods. There might be a way to scope these linting rules to only the packages that we intend to use in browser environments, or exclude them from applying to test files. However, that seems unnecessary, given that the replacements are no harder to use, so a consistent style across the entire Compass codebase seems preferable.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/atlas-service/src/store/atlas-signin-reducer.spec.ts
+++ b/packages/atlas-service/src/store/atlas-signin-reducer.spec.ts
@@ -316,9 +316,16 @@ describe('atlasSignInReducer', function () {
     });
 
     it('should reject if provided signal was aborted', async function () {
+      let resolveSignInCalled = () => {};
+      const signInCalled: Promise<void> = new Promise(
+        (resolve) => (resolveSignInCalled = resolve)
+      );
       const mockAtlasService = {
         isAuthenticated: sandbox.stub().resolves(false),
-        signIn: sandbox.stub().resolves({ sub: '1234' }),
+        signIn: sandbox.stub().callsFake(() => {
+          resolveSignInCalled();
+          return { sub: '1234' };
+        }),
         getUserInfo: sandbox.stub().resolves({ sub: '1234' }),
         emit: sandbox.stub(),
       };
@@ -337,6 +344,9 @@ describe('atlasSignInReducer', function () {
         expect(err).to.have.property('message', 'Aborted from outside');
       }
       expect(store.getState()).to.have.property('state', 'canceled');
+
+      // Ensure that we are not leaving a dangling store operation that would conflict with our mocks being reset.
+      await signInCalled;
     });
   });
 });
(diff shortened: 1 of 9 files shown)


packages/compass-schema/src/stores/store.js
```diff
@@ -195,13 +195,6 @@ const configureStore = (options = {}) => {
 
     onSchemaSampled() {
       this.geoLayers = {};
-
-      process.nextTick(() => {
-        this.globalAppRegistry.emit('compass:schema:schema-sampled', {
-          ...this.state,
-          geo: this.geoLayers,
-        });
-      });
```
addaleax, line 204: The last consumer of this event was removed in 953d9b213bebfeae67c70eb2a01ee6aabd57c3ad which removed the legacy metrics package. We have tracking code for this event next to the call site of this method.

Anemy (approved):
Nice lint add!

alenakhineika (approved):

packages/atlas-service/src/store/atlas-signin-reducer.spec.ts
```diff
@@ -337,6 +344,9 @@ describe('atlasSignInReducer', function () {
         expect(err).to.have.property('message', 'Aborted from outside');
       }
       expect(store.getState()).to.have.property('state', 'canceled');
+
+      // Ensure that we are not leaving a dangling store operation that would conflict with our mocks being reset.
+      await signInCalled;
```
addaleax, line 349: @gribnoysup Just fyi in case you run across this later … I pushed this because it seems that mocha [internally also uses](https://github.com/mochajs/mocha/blob/37deed262d4bc0788d32c66636495d10038ad398/browser-entry.js#L116) `setTimeout`, which is really the only relative timing case where `setTimeout` and `setImmediate` differ (i.e. they have their own lists for callbacks rather than sharing one).

Without this fix, what happened was that the test finished before [the core logic](https://github.com/mongodb-js/compass/blob/b57b2505f156d942f97d32927d3c8676c14e9489/packages/atlas-service/src/store/atlas-signin-reducer.ts#L473) in `signIn()` ran, so the `sandbox.reset()` in the `afterEach()` hook ran _before_ the action called `atlasService.signIn()`, which then returned `undefined` (because the sandbox was reset) and then crashed because `undefined` is not an expected return value for `atlasService.signIn()`.

It would have been possible to work around this by changing the timing again slightly, but I think it’s okay to go this way here and make the test a bit more robust with regard to these subtle timing differences.

PR #4977, opened 2023-10-11
feat: remove NONE and authMechanism=DEFAULT COMPASS-7308

mcasimir:
Before:
Screenshot 2023-10-11 at 16 40 18

After:
Screenshot 2023-10-11 at 16 36 08


packages/connection-form/src/components/advanced-options-tabs/authentication-tab/authentication-tab.tsx
```diff
@@ -43,13 +42,6 @@ interface TabOption {
 }
 
 const options: TabOption[] = [
-  {
-    title: 'None',
-    id: 'AUTH_NONE',
-    component: function None() {
-      return <></>;
-    },
-  },
```
addaleax, line 52: I do think we probably want to mark the username/passwords fields as `Optional` now (unconditionally, because we don’t really adjust that part of the form representation depending on other state in the same form)

mcasimir: thought the same, i've asked in the design channel

addaleax (approved):

PR #4982, opened 2023-10-12
fix(search-indexes): ui fixes COMPASS-7304

mabaasit:
In this PR, implemented following UI fixes:

1. Render small segmented tabs for sub-indexes tabs.
2. In create/update modal, move templates info icon next to its label.
3. Remove spacing (top/bottom) around Index Definition Editor.
4. After create/update/delete of search index, the toast should variant should be 'progress', not success.

<table>
<tr>
 <td> Old
 <td> New
<tr> 
 <td> 
image

 <td> 
image

<tr>
 <td>
image

 <td>
image

<tr>
 <td>
image

 <td>image
<tr>
 <td>
image

 <td>
image

</table>

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-components/package.json
+++ b/packages/compass-components/package.json
@@ -46,6 +46,7 @@
     "@leafygreen-ui/hooks": "^7.3.3",
     "@leafygreen-ui/icon": "^11.21.0",
     "@leafygreen-ui/icon-button": "^15.0.3",
+    "@leafygreen-ui/info-sprinkle": "^1.0.0",
     "@leafygreen-ui/inline-definition": "^6.0.0",
     "@leafygreen-ui/leafygreen-provider": "^3.1.0",
     "@leafygreen-ui/lib": "^10.0.0",
--- a/packages/compass-indexes/src/modules/search-indexes.ts
+++ b/packages/compass-indexes/src/modules/search-indexes.ts
@@ -452,7 +452,7 @@ export const createIndex = (
       title: `Your index ${indexName} is in progress.`,
       dismissible: true,
       timeout: 5000,
-      variant: 'success',
+      variant: 'progress',
     });
 
     void dispatch(switchToSearchIndexes());
@@ -494,7 +494,7 @@ export const updateIndex = (
         title: `Your index ${indexName} is being updated.`,
         dismissible: true,
         timeout: 5000,
-        variant: 'success',
+        variant: 'progress',
       });
       void dispatch(fetchIndexes(SearchIndexesStatuses.REFRESHING));
     } catch (e) {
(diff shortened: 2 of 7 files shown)


packages/compass-components/package.json
```diff
@@ -46,6 +46,7 @@
     "@leafygreen-ui/hooks": "^7.3.3",
     "@leafygreen-ui/icon": "^11.21.0",
     "@leafygreen-ui/icon-button": "^15.0.3",
+    "@leafygreen-ui/info-sprinkle": "^1.0.0",
```
Anemy, line 49: TIL
Probably a lot of other places in Compass we can look at updating to use this over time.

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -633,7 +633,7 @@ export const dropSearchIndex = (
         title: `Your index ${name} is being deleted.`,
         dismissible: true,
         timeout: 5000,
-        variant: 'success',
+        variant: 'progress',
```
Anemy, line 636: Slightly unrelated, would a better flow for the toasts to be showing the `Your index x is being deleted...` until either it completes or an error occurs? It would make the flow similar to how the import/export toasts work: https://github.com/mongodb-js/compass/blob/d87a58e0817aa475e5e4c7c607e65a35c01b6d93/packages/compass-import-export/src/modules/export.ts#L423  https://github.com/mongodb-js/compass/blob/d87a58e0817aa475e5e4c7c607e65a35c01b6d93/packages/compass-import-export/src/components/export-toast.tsx#L52
They share the same id so they are overridden when the error or completion state happens.


Maybe this has been asked before, but does the `await dataService.dropSearchIndex()` method not indicate completion when the function succeeds? Should we be showing this toasat before that function call?

Anemy (approved):
lgtm, one non-blocker question on possible toast showing behavior improvements. It feels like we might be showing the in progress toast now when its already completed.

packages/compass-indexes/src/modules/search-indexes.ts
lerouxb: It doesn't because for create, update and delete of search indexes the command (and therefore method) returns immediately and then the action gets "queued" and it can take any amount of time to complete.

lerouxb: (Which is why we poll in the list view to keep updating the statuses.)

lerouxb (approved):

PR #4999, opened 2023-10-17
fix(data-service): address review comments for previewUpdate

addaleax:
Most notably, address that we want to apply the timeout to the entire transaction (currently, the timeout only applies to the *starting* of the transaction, not the remainder of it).

As drive-by fixes, assign a new log id to the method instead of re-using an old one, and simplify the options extraction a bit.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/data-service/src/data-service.ts
+++ b/packages/data-service/src/data-service.ts
@@ -105,9 +105,6 @@ import type {
 } from '@mongodb-js/devtools-connect';
 import { omit } from 'lodash';
 
-const PREVIEW_TIMEOUT = 1000;
-const PREVIEW_SAMPLE = 10;
-
 function uniqueBy<T extends Record<string, unknown>>(
   values: T[],
   key: keyof T
@@ -146,8 +143,8 @@ export type UpdatePreviewChange = {
 };
 
 export type UpdatePreviewExecutionOptions = ExecutionOptions & {
-  sample: number;
-  timeout: number;
+  sample?: number;
+  timeout?: number;
 };
 
 export type UpdatePreview = {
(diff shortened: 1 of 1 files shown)


packages/data-service/src/data-service.ts
```diff
@@ -105,9 +105,6 @@ import type {
 } from '@mongodb-js/devtools-connect';
 import { omit } from 'lodash';
 
-const PREVIEW_TIMEOUT = 1000;
-const PREVIEW_SAMPLE = 10;
```
addaleax, line 109: No super hard feelings about these, happy to restore them, but if we keep them I’d probably choose a different name that’s more indicative of these applying to *update* preview specifically

packages/data-service/src/data-service.ts
```diff
@@ -2314,25 +2311,21 @@ class DataServiceImpl extends WithLogContext implements DataService {
...
-        sample: PREVIEW_SAMPLE,
-        timeout: PREVIEW_TIMEOUT,
-      };
-    }
-
-    const sample = executionOptions.sample || PREVIEW_SAMPLE;
-    const timeout = executionOptions.timeout || PREVIEW_TIMEOUT;
-
+    const {
+      abortSignal = new AbortController().signal,
```
addaleax, line 2322: (previously, passing `executionOptions` with `sample` and `timeout` would have errored because no session is created if `abortSignal` is nullish)

lerouxb (approved):

PR #5002, opened 2023-10-18
feat: Improve stage wizard discoverability and interaction COMPASS-7350

mcasimir:
Improves the stage wizard discoverability:

- Swap the trigger IconButton with a regular Button with an Icon and a label
- Make the panel open by default and remember its state
- Remove links from the card and use the badge instead
- Make it feel more clickable
- Change the copy in the search box
- Make the card dimmed while dragging
- Prioritize the "click" interaction (insert as last stage)
- Give back focus to the trigger after cancelling the form

https://github.com/mongodb-js/compass/assets/334881/0c5b50aa-4d2a-47f1-8e50-978392387086

Screenshot 2023-10-18 at 10 36 43

### DIFF
--- a/packages/compass-aggregations/src/components/aggregation-side-panel/index.tsx
+++ b/packages/compass-aggregations/src/components/aggregation-side-panel/index.tsx
@@ -63,6 +63,9 @@ const contentStyles = css({
   flexDirection: 'column',
   gap: spacing[2],
   overflow: 'auto',
+  paddingTop: spacing[1],
+  paddingLeft: spacing[1],
+  paddingRight: spacing[1],
   paddingBottom: spacing[3],
 });
 
@@ -132,8 +135,8 @@ export const AggregationSidePanel = ({
       <SearchInput
         value={searchText}
         onChange={handleSearchTextChange}
-        placeholder="How can we help?"
-        aria-label="How can we help?"
+        placeholder="Search a Stage Wizard"
+        aria-label="Search a Stage Wizard"
       />
       <div className={contentStyles} data-testid="side-panel-content">
         {filteredUseCases.map((useCase, index) => {
--- a/packages/compass-aggregations/src/components/aggregation-side-panel/index.tsx
+++ b/packages/compass-aggregations/src/components/aggregation-side-panel/index.tsx
@@ -25,8 +25,7 @@ const { track } = createLoggerAndTelemetry('COMPASS-AGGREGATIONS-UI');
 
 const containerStyles = css({
   height: '100%',
-  paddingLeft: spacing[2],
-  paddingRight: spacing[2],
+
   paddingTop: spacing[1],
   borderBottomRightRadius: 0,
   borderBottomLeftRadius: 0,
@@ -58,14 +57,19 @@ const titleStylesLight = css({
   color: palette.green.dark2,
 });
 
+const headerContainerStyles = css({
+  paddingLeft: spacing[2],
+  paddingRight: spacing[2],
+});
+
 const contentStyles = css({
   display: 'flex',
   flexDirection: 'column',
   gap: spacing[2],
   overflow: 'auto',
   paddingTop: spacing[1],
-  paddingLeft: spacing[1],
-  paddingRight: spacing[1],
+  paddingLeft: spacing[2],
+  paddingRight: spacing[2],
   paddingBottom: spacing[3],
 });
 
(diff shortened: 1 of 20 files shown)


packages/compass-components/src/components/keyline-card.tsx
```diff
@@ -17,23 +20,74 @@ const keylineDarkThemeStyles = css({
...
 
+const lightHoverBoxShadow = `0 4px 20px -4px ${transparentize(
+  0.8,
+  palette.black
+)}`;
+
+const darkHoverBoxShadow = `0 4px 20px -4px ${transparentize(
+  0.3,
+  palette.black
+)}`;
```
Anemy, line 31: Looking closely this hover seems to go out of the boundaries and gets a cut off look:
Screenshot 2023-10-19 at 11 28 16 AM
I see that this is the same as the [clickable card styles](https://github.com/mongodb/leafygreen-ui/blob/cf06988dcef8e2dac83ebc409f3b2e9a905fe9e8/packages/card/src/Card/styles.ts#L22), so I'm assuming we want to keep them.
Maybe we can make the container padding and margins set in the way where the items are not cut off?
Fine as is too, this is a design nit.

packages/compass-components/src/components/keyline-card.tsx
```diff
@@ -17,23 +20,74 @@ const keylineDarkThemeStyles = css({
...
-  return (
-    <div
-      ref={ref}
-      className={cx(
+  const allProps = mergeProps(
+    contentStyle === 'clickable'
+      ? { role: 'button', tabIndex: 0, ...focusRingProps }
+      : {},
+    {
+      className: cx(
```
Anemy, line 79: Related to the above question on box shadows, the clickable `Cards` from leafygreen have a box shadow that's always showing, and expanded when hovered. Would that make this more `clickable` looking? Fine as is too I reckon.

| not hovered | hovered |
| - | - |
| Screenshot 2023-10-19 at 11 47 39 AM | Screenshot 2023-10-19 at 11 47 42 AM |

packages/compass-components/src/components/keyline-card.tsx
```diff
@@ -17,23 +20,74 @@ const keylineDarkThemeStyles = css({
...
+  }
+`;
+
+const clickableStyleDark = css`
+  cursor: pointer;
+
+  &:hover {
+    box-shadow: ${darkHoverBoxShadow};
+  }
+`;
```
Anemy, line 52: To make the hover smoother, I'm thinking lets pull over the transition styles from lg:
https://github.com/mongodb/leafygreen-ui/blob/cf06988dcef8e2dac83ebc409f3b2e9a905fe9e8/packages/card/src/Card/styles.ts#L86C1-L87C43

```
import {
  transitionDuration
} from '@leafygreen-ui/tokens';
```


```suggestion
const clickableStyleLight = css`
  cursor: pointer;
  transition: ${transitionDuration.default}ms ease-in-out;
  transition-property: border, box-shadow;

  &:hover {
    border: 1px solid ${palette.gray.light2};
    box-shadow: ${lightHoverBoxShadow};
  }

  &:active {
    box-shadow: none;
  }
`;

const clickableStyleDark = css`
  cursor: pointer;
  transition: ${transitionDuration.default}ms ease-in-out;
  transition-property: border, box-shadow;

  &:hover {
    box-shadow: ${darkHoverBoxShadow};
  }
`;
```

A total nit, but we might want to standardize on using `css({...})` vs 'css`...`' in these places. Although this is pulled from LG that uses 'css`...`'.

Anemy (commented):
Looking good, a couple styling comments, probably not blockers.

packages/compass-components/src/components/keyline-card.tsx
mcasimir: mmm weird, I'm sure i copied the styles as they were in Card, I'll play around with that

packages/compass-components/src/components/keyline-card.tsx
mcasimir: Sounds great!

packages/compass-aggregations/src/components/pipeline-toolbar/pipeline-settings/hidden-on-narrow-container.ts
```diff
@@ -0,0 +1,8 @@
+import { css } from '@mongodb-js/compass-components';
+import { workspaceContainerQueryName } from '@mongodb-js/compass-components';
```
gribnoysup, line 2: I was thinking about what we discussed yesterday and to make it more contextual, how do you feel about assigning this property to the component constructor? Maybe a good way to make it super obvious what the container query refers to?

```js
import { WorkspaceContainer } from '@mongodb-js/compass-components';

css`
  @container ${WorkspaceContainer.containerQueryName} (...) { ... }
`
```

Super nit though, feel free to ignore, just thinking how to make it  not just a random-ish const export from the components package

mcasimir: Makes sense!

Actually, I was almost about to remove this from WorkspaceContainer :). While doing this I stumbled on a couple of controls for which, if we want to do something nice we may still need something for react, cause hiding labels with css on LG controls is not the easiest thing ever :(

packages/compass-components/src/components/keyline-card.tsx
mcasimir: turns out i've ditched those styles by accident, added them back

packages/compass-components/src/components/keyline-card.tsx
mcasimir: done

packages/compass-components/src/components/keyline-card.tsx
mcasimir: probably done .. i moved the paddings around

packages/compass-aggregations/src/components/aggregation-side-panel/index.tsx
```diff
@@ -132,8 +135,8 @@ export const AggregationSidePanel = ({
       <SearchInput
         value={searchText}
         onChange={handleSearchTextChange}
-        placeholder="How can we help?"
-        aria-label="How can we help?"
+        placeholder="Search a Stage Wizard"
+        aria-label="Search a Stage Wizard"
```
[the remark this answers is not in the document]
mcasimir: ended up with Search for a Stage, under suggestion for claudia and julia

packages/compass-components/package.json
```diff
@@ -34,6 +34,7 @@
     "@dnd-kit/core": "^6.0.7",
     "@dnd-kit/sortable": "^7.0.2",
     "@dnd-kit/utilities": "^3.2.1",
+    "@emotion/css": "^11.11.2",
```
Anemy, line 37: Why do we need this dependency?

Anemy (commented):
Changes look good, just one question on the emotion dependency. Should we always be using the styles from leafygreen emotion?

packages/compass-components/package.json
gribnoysup: [Context for this](https://mongodb.slack.com/archives/G2L10JAV7/p1698163251582629?thread_ts=1698151766.108829&cid=G2L10JAV7)

Anemy: Ah cool, thanks. I wish we could leave a comment in json for on this one. Would it help if we add a comment in the `depcheck.rc`?

Anemy (approved):
lgtm! looks good

packages/compass-components/package.json
mcasimir: sounds good, i'll follow up with that in a different PR, i have more random improvements coming up

PR #5003, opened 2023-10-18
feat: add delete button next to querybar COMPASS-7324

kmruiz:
_no release notes because it's under a feature flag._

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-components/src/components/guide-cue/guide-cue-service.ts
+++ b/packages/compass-components/src/components/guide-cue/guide-cue-service.ts
@@ -101,8 +101,6 @@ export class GuideCueService extends EventTarget {
       // TODO: this seems to be a temporary error happening sometimes during test.
       // In that case, assume the event is not dispatched
     }
-
-    return false;
   }
 
   private validateCueData(groupId: GroupName, step: number) {
--- a/packages/compass-components/src/components/guide-cue/guide-cue-service.ts
+++ b/packages/compass-components/src/components/guide-cue/guide-cue-service.ts
@@ -88,14 +88,21 @@ export class GuideCueService extends EventTarget {
     if (!this._activeCue) {
       return;
     }
-    return this.dispatchEvent(
-      new CustomEvent('show-cue', {
-        detail: {
-          cueId: this._activeCue.cueId,
-          groupId: this._activeCue.groupId,
-        },
-      })
-    );
+    try {
+      return this.dispatchEvent(
+        new CustomEvent('show-cue', {
+          detail: {
+            cueId: this._activeCue.cueId,
+            groupId: this._activeCue.groupId,
+          },
+        })
+      );
+    } catch (ex) {
+      // TODO: this seems to be a temporary error happening sometimes during test.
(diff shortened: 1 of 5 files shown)


packages/compass-components/src/components/guide-cue/guide-cue-service.ts
```diff
@@ -88,14 +88,21 @@ export class GuideCueService extends EventTarget {
...
+    try {
+      return this.dispatchEvent(
+        new CustomEvent('show-cue', {
+          detail: {
+            cueId: this._activeCue.cueId,
+            groupId: this._activeCue.groupId,
+          },
+        })
+      );
+    } catch (ex) {
```
lerouxb, line 100: What error?

But also looking at lime 89 it looks like this function doesn't return anything if it early returns and this case should probably do the same?

kmruiz: For some reason, in some tests, it throws an error trying to map the error and dispatch it. It looks like a race condition, that doesn't really affect the test behaviour, but when it bubbles up, stops the test.

And you are right about line 89, I can just remove the return and leave it return nothing. Fixing it.

packages/compass-crud/src/components/delete-data-menu.tsx
```diff
@@ -0,0 +1,69 @@
...
+};
+
+const DeleteMenuButton: React.FunctionComponent<DeleteMenuButtonProps> = ({
+  isWritable,
+  onClick,
+}) => {
+  const isVisible = usePreference('enableBulkDeleteOperations', React);
+
+  if (!isVisible) {
+    return <></>;
```
lerouxb, line 17: Shouldn't this just be `return null;` I'm no expert, but I think that's what's idiomatic here.

packages/compass-components/src/components/guide-cue/guide-cue-service.ts
lerouxb: I'm just wondering if that's not best fixed in a test rather than the code under test.

kmruiz: It seems to happen on all tests that depend on this service. What I feel it's that there is some race condition between the service being ready and dispatching the event, failing sometimes.

I feel that fixing the underlying issue would be expensive and it doesn't seem to affect production code.

lerouxb: OK That's fair. Let's file a ticket for that and then you can leave a `// TODO(ticket-number): blah` style comment here and merge.

lerouxb (approved):

packages/compass-components/src/components/guide-cue/guide-cue-service.ts
kmruiz: Makes sense, will do. Thanks :heart:

PR #5005, opened 2023-10-19
feat: delete bulk operation modal COMPASS-7326

kmruiz:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-crud/src/components/bulk-delete-modal.tsx
+++ b/packages/compass-crud/src/components/bulk-delete-modal.tsx
@@ -1,0 +1,105 @@
+import React from 'react';
+import {
+  Modal,
+  ModalHeader,
+  ModalBody,
+  ModalFooter,
+  Button,
+  TextInput,
+  KeylineCard,
+  css,
+  spacing,
+} from '@mongodb-js/compass-components';
+import ReadonlyDocument from './readonly-document';
+
+const modalFooterSpacingStyles = css({
+  gap: spacing[2],
+});
+
+const documentHorizontalWrapper = css({
+  display: 'flex',
+  flexDirection: 'row',
+  flex: 'none',
+  flexShrink: 0,
+  overflow: 'auto',
+  marginBottom: spacing[2],
+  gap: spacing[2],
+  maxWidth: '100%',
+});
+
+const documentContainerStyles = css({
+  display: 'flex',
+  flexDirection: 'column',
+  flex: 'none',
+  flexShrink: 0,
+  marginBottom: spacing[2],
+  width: '100%',
+});
+
+const documentStyles = css({
+  flexBasis: '164px',
+  flexGrow: 1,
+  flexShrink: 0,
+  overflow: 'auto',
+  padding: 0,
+  width: '100%',
+});
+
+type BulkDeleteModalProps = {
+  open: boolean;
+  documentCount: number;
+  filterQuery: string;
+  namespace: string;
+  sampleDocuments: Document[];
+  onCancel: () => void;
+  onConfirmDeletion: () => void;
+};
+
+const BulkDeleteModal: React.FunctionComponent<BulkDeleteModalProps> = ({
+  open,
+  documentCount,
+  filterQuery,
+  namespace,
+  sampleDocuments,
+  onCancel,
+  onConfirmDeletion,
+}) => {
+  const preview = (
+    <div className={documentHorizontalWrapper}>
+      {sampleDocuments.map((doc, i) => {
+        return (
+          <KeylineCard key={i} className={documentContainerStyles}>
+            <div className={documentStyles}>
+              <ReadonlyDocument doc={doc} />
+            </div>
+          </KeylineCard>
+        );
+      })}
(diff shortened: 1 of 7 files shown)


packages/compass-crud/src/components/bulk-delete-modal.tsx
```diff
@@ -0,0 +1,159 @@
...
+      />
+      <ModalBody variant={'danger'} className={modalBodySpacingStyles}>
+        <TextInput
+          label={
+            (
+              <QueryLabel
+                label="Query"
+                tooltip="Return to the Documents tab to edit this query."
+              />
+            ) as any as string
```
mabaasit, line 137: or you can either ts-ignore or just keep it as any.

packages/compass-crud/src/components/bulk-delete-modal.tsx
```diff
@@ -0,0 +1,159 @@
...
+  open,
+  documentCount,
+  filterQuery,
+  namespace,
+  sampleDocuments,
+  onCancel,
+  onConfirmDeletion,
+}) => {
+  const preview = (
+    <div className={documentHorizontalWrapper}>
```
mabaasit, line 106: The designs are not clear on this (horizontal scroll), but i struggled a little to find that its a horizontal scroll.

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1584,6 +1600,72 @@ class CrudStoreImpl
...
+      },
+    }); // TODO: COMPASS-7328
+  }
+
+  bulkDeleteFailed(ex: Error) {
+    return ex; // TODO: COMPASS-7328
+  }
+
+  bulkDeleteSuccess() {
+    this.closeBulkDeleteDialog();
```
mabaasit, line 1630: the dialog is already closed when we call `runBulkDelete`.

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1584,6 +1600,72 @@ class CrudStoreImpl
...
+    });
+
+    if (confirmation) {
+      this.bulkDeleteInProgress();
+      try {
+        await this.dataService.deleteMany(
+          this.state.ns,
+          this.state.query.filter
+        );
+        this.bulkDeleteSuccess();
```
mabaasit, line 1663: I am not sure how we want to handle the current list of documents that was rendered before delete. It still is visible on the screen - are we planning to refresh the list or leave it as it?

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1584,6 +1600,72 @@ class CrudStoreImpl
...
+      bulkDelete: {
+        ...this.state.bulkDelete,
+        status: 'closed',
+      },
+    });
+  }
+
+  async runBulkDelete() {
+    const { affected } = this.state.bulkDelete;
+    this.closeBulkDeleteDialog();
```
mabaasit, line 1645: If we close this and user clicks cancel while confirming, should we show it back? This (hiding delete modal and show confirmation modal) creates a flicker. Maybe we should keep showing the delete modal as well?

packages/compass-crud/src/stores/crud-store.ts
mabaasit: its just that since there's no toast yet (which is todo for next ticket), it feels like nothing happened.

packages/compass-crud/src/components/bulk-delete-modal.tsx
kmruiz: When we had the discussion, the idea was to keep what the aggregation screen does, so we are keeping the horizontal scroll.

packages/compass-crud/src/stores/crud-store.ts
kmruiz: That's a good point. We were discussing it yesterday. The issue with refreshing is that a delete can take minutes (or more) and if we refresh while the user is doing something else (we don't block the screen) there might be some weird behaviour. So we will leave it as it is and notify by using the toast, so the user can refresh.

The next task is the in-progress toast so there will be some feedback for the user at that point.

packages/compass-crud/src/stores/crud-store.ts
kmruiz: Discussed on slack, but writing the answer here for reference:

This is as discussed with design, we are not going to use double modals for this flow. We might want to revisit later.

packages/compass-crud/src/stores/crud-store.ts
kmruiz: You are right, removing this.

packages/compass-crud/src/components/bulk-delete-modal.tsx
kmruiz: Adding @ts-ignore :D

PR #5007, opened 2023-10-19
feat(crud): Bulk update modal COMPASS-7325

lerouxb:
Screenshot 2023-10-20 at 14 31 17

Not very many tests yet, but adding the toast will properly test in-progress, done and error states. And we'll add e2e tests in a later ticket.

There's also probably some duplication between this and the delete modal that we can clean up separately as well.


packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1053,6 +1091,130 @@ class CrudStoreImpl
...
+      });
+      return;
+    }
+
+    const { ns } = this.state;
+    const { filter } = this.state.query;
+
+    let preview;
+    try {
+      // TODO(COMPASS-7369): we should automatically retry if the get "Write
```
lerouxb, line 1154: ☝️

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1053,6 +1091,130 @@ class CrudStoreImpl
...
+      // conflict during plan execution and yielding is disabled."
+      preview = await this.dataService.previewUpdate(ns, filter, update, {
+        sample: 3,
+        // TODO(COMPASS-7368): aborting the in-flight operation is still buggy,
```
lerouxb, line 1158: ☝️

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1053,6 +1091,130 @@ class CrudStoreImpl
...
+    try {
+      update = parseShellBSON(this.state.bulkUpdate.updateText);
+    } catch (err) {
+      // If this couldn't parse then the update button should have been
+      // disabled. So if we get here it is a race condition and ignoring is
+      // probably OK - the button will soon appear disabled to the user anyway.
+      return;
+    }
+
+    // TODO(COMPASS-7327): in-progress, success, error in toast
```
lerouxb, line 1214: ☝️

packages/data-service/src/data-service.ts
```diff
@@ -2113,7 +2128,7 @@ class DataServiceImpl extends WithLogContext implements DataService {
       result = await raceWithAbort(start(session), abortSignal);
     } catch (err) {
       if (isCancelError(err)) {
-        void abort();
+        await abort();
```
lerouxb, line 2131: No idea why we didn't do this before.

packages/data-service/src/data-service.ts
```diff
@@ -2357,19 +2372,36 @@ class DataServiceImpl extends WithLogContext implements DataService {
...
           const changes = docsToPreview.map((before, idx) => ({
             before,
             after: changedDocs[idx],
           }));
           return { changes };
-        } finally {
-          await session.abortTransaction();
-          await session.endSession();
+        } catch (err: any) {
+          if (isTransactionAbortError(err)) {
```
lerouxb, line 2386: We _might_ be able to avoid this if we check if the transaction has been aborted or the session otherwise closed after every async call above, but I don't think we can really avoid that. The session can be aborted/killed at any point if we pass in an abortSignal.

packages/compass-crud/src/components/bulk-update-dialog.tsx
```diff
@@ -0,0 +1,298 @@
...
+              </InfoSprinkle>
+            </div>
+            <KeylineCard
+              className={cx(
+                codeContainerStyles,
+                inlineContainerStyles,
+                darkMode ? codeDarkContainerStyles : codeLightContainerStyles
+              )}
+            >
+              <pre>{(toJSString(filter) ?? '').replace(/\s+/g, ' ')}</pre>
```
lerouxb, line 212: Need to test and do something about wide filters that don't fit. I think the plan is to cut them off and show the full thing in a tooltip. As opposed to wrapping or scrolling.

packages/data-service/src/data-service.ts
mabaasit: because we didn't want to throw if abort failed. now that you changed it, should we catch `about` error and then continue to throw `err`, which is relevant in our case?

addaleax: We already wrote `abort` in a way such that it won’t throw, and if it does, that’s a genuine programming error then that we _should_ raise, right?

lerouxb: If your cancellable thing is multiple queries, right now you _will_ get unhandled rejections if you abort it. For example.

packages/compass-crud/src/components/bulk-update-dialog.tsx
lerouxb: I just copied the same pattern/style from the bulk delete modal: just use a text input.

packages/compass-crud/src/components/bulk-update-dialog.tsx
```diff
@@ -0,0 +1,306 @@
...
+  paddingTop: spacing[2],
+  paddingBottom: spacing[2],
+});
+
+const codeDarkContainerStyles = css({
+  background: palette.gray.dark4,
+});
+
+const codeLightContainerStyles = css({
+  backgroundColor: palette.gray.light3,
```
gribnoysup, line 73: Not sure how this correlates with the styles below: this is exactly the same color we use in the editor? Is it only the dark mode that's different? Why?

packages/compass-crud/src/components/bulk-update-dialog.tsx
```diff
@@ -0,0 +1,306 @@
...
+  preview,
+  syntaxError,
+  serverError,
+  closeBulkUpdateDialog,
+  updateBulkUpdatePreview,
+  runBulkUpdate,
+}: BulkUpdateDialogProps) {
+  const darkMode = useDarkMode();
+
+  const [text, setText] = useState(updateText);
```
gribnoysup, line 164: Doesn't look like we reset this when modal is re-opened, is this expected?

packages/compass-crud/src/components/bulk-update-dialog.tsx
```diff
@@ -0,0 +1,306 @@
...
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore the label can be any component, but it's weirdly typed to string
+              label={
+                <QueryLabel
+                  label="Filter"
+                  tooltip="Return to the Documents tab to edit this query."
+                />
+              }
+              disabled={true}
+              value={(toJSString(filter) ?? '').replace(/\s+/g, ' ')}
```
gribnoysup, line 220: If I have a filter for a string value with multiple spaces, this reduces all those to a single one, this is probably not expected, right?

image

image

packages/compass-crud/src/components/bulk-update-dialog.tsx
```diff
@@ -0,0 +1,306 @@
...
+      onSubmit={runBulkUpdate}
+      onCancel={closeBulkUpdateDialog}
+      cancelButtonText="Close"
+      submitButtonText="Update documents"
+      submitDisabled={!!(syntaxError || serverError)}
+    >
+      <div className={columnsStyles}>
+        <div className={queryStyles}>
+          <div className={queryFieldStyles}>
+            <TextInput
```
gribnoysup, line 210: Do we want to switch the font to monospace here? That's what designs seems to expect

packages/compass-crud/src/components/bulk-update-dialog.tsx
lerouxb: To match the design. Otherwise it is white. We do the exact same thing in the aggregation stage editor: https://github.com/mongodb-js/compass/blob/dbf1b0d71c1efcb870e4690e91bac37fc3f1e6db/packages/compass-aggregations/src/components/stage-editor/stage-editor.tsx#L43

There's a ticket to refactor that so we don't keep duplicating it along with the banner area at the bottom.

packages/compass-crud/src/components/bulk-update-dialog.tsx
lerouxb: Yup.

packages/compass-crud/src/components/bulk-update-dialog.tsx
lerouxb: I just copy/pasted from the bulk delete modal. We're doing them in parallel and reinventing every wheel twice. I'd rather have a ticket to just factor these out and clean it up afterwards.

packages/compass-crud/src/components/bulk-update-dialog.tsx
```diff
@@ -0,0 +1,306 @@
...
+                  >
+                    {serverError.message}
+                  </Banner>
+                )}
+              </div>
+            </KeylineCard>
+          </div>
+        </div>
+        <div className={previewStyles}>
+          <Label htmlFor="bulk-update-preview">
```
gribnoysup, line 270: In the designs this preview label has enough padding so that the filter preview on the left and the first doc on the right are aligned:

image

Doesn't seem to be the case here:

image

packages/compass-crud/src/components/bulk-update-dialog.tsx
lerouxb: Oops. Leftover.

packages/compass-crud/src/components/bulk-update-dialog.tsx
```diff
@@ -0,0 +1,306 @@
...
+  updateBulkUpdatePreview,
+  runBulkUpdate,
+}: BulkUpdateDialogProps) {
+  const darkMode = useDarkMode();
+
+  const [text, setText] = useState(updateText);
+
+  const previewDocuments = useMemo(() => {
+    return preview.changes.map(
+      (change) => new HadronDocument(change.after as Record<string, unknown>)
```
gribnoysup, line 168: Feel free to leave this, but Document component does basically the same thing internally if the object is passed, so you should be able to remove this HadronDocument instance creation for here and just pass the after object to the component directly

packages/compass-crud/src/components/bulk-update-dialog.tsx
gribnoysup: Maybe worth leaving a comment that it's on purposes, we have a bunch of modal forms where we reset forms on modal open, I'm worried we can loose this behavior by accident

packages/compass-crud/src/components/bulk-update-dialog.tsx
gribnoysup: Not sure I follow, default multiline background is the same gray color: https://github.com/mongodb-js/compass/blob/c3a174ea2272f21df453d929f0542fd33e675284/packages/compass-editor/src/json-editor.tsx#L141

Sounds good if we're planning to do a follow-up clean up for this 👍

packages/compass-crud/src/components/bulk-update-dialog.tsx
gribnoysup: Got it, a ticket to clean this up afterwards sounds good 👍

gribnoysup (approved):
Left a few small comments, but otherwise looks good!

packages/compass-crud/src/components/bulk-update-dialog.tsx
lerouxb: OK it is that info icon. I made it just "hang out" now rather than push that line bigger because that seems like the simplest fix.

packages/compass-crud/src/components/bulk-update-dialog.tsx
lerouxb: https://jira.mongodb.org/browse/COMPASS-7387

PR #5008, opened 2023-10-19
chore(generative-ai): move accuracy tests to generative-ai package, parallelize tests

Anemy:
Uses the [`p-queue` package](https://github.com/sindresorhus/p-queue) to manage the concurrency of the accuracy tests (currently set to `4`).
Moves the accuracy test script into the `compass-generative-ai` package.


packages/compass-generative-ai/scripts/ai-accuracy-tests.js
```diff
@@ -20,6 +23,11 @@ const os = require('os');
 const assert = require('assert');
 const ejsonShellParser = require('ejson-shell-parser');
 const { MongoClient } = require('mongodb');
+let PQueue;
+(async () => {
+  // ESM package only. This could introduce race conditions.
```
mcasimir, line 28: Any reason for not moving this in `main`?

packages/compass-generative-ai/scripts/ai-accuracy-tests.js
```diff
@@ -33,6 +41,11 @@ const DEFAULT_MIN_ACCURACY = 0.8;
 
 const MAX_TIMEOUTS_PER_TEST = 10;
 
+// There are a limited amount of resources available both on the Atlas
+// and on the ai service side of things, so we want to limit how many
+// requests can be happening at a time.
+const TESTS_TO_RUN_CONCURRENTLY = 4;
```
mcasimir, line 47: Is this the highest number that we can use? If this is all we can get, assuming that we would want to add more tests, should we consider working around the rate limit on Atlas for the test user?

packages/compass-generative-ai/scripts/ai-accuracy-tests.js
Anemy: Nope, good suggestion, moved into `setup`. 👍

packages/compass-generative-ai/scripts/ai-accuracy-tests.js
Anemy: Yeah... I think we'll want to have a workaround for the rate limit for test user for cloud dev and local. Right now we hit it with 3 concurrents. I had to add in a timeout for us not to get rate limited every time. I think that'll be a follow up done separately. These changes at least do speed up the script, I'm thinking that can happen in another pr and will probably require some mms work.

mcasimir (approved):

PR #5015, opened 2023-10-23
chore(compass-e2e-tests): update e2e test db name to have e2e in it

Anemy:
This db is in the `Compass data sets` deployment.

We might want to update this to not be the Compass data sets deployment as well.


Anemy:
Merging as this is a couple lines, can look into which deployment we'll do the connectivity tests to if we want to change it another time I think. Slack thread https://mongodb.slack.com/archives/G2L10JAV7/p1698088546012649

PR #5047, opened 2023-11-01
feat: reset the bulk update modal when re-opened COMPASS-7325

lerouxb:
Follow-up from https://github.com/mongodb-js/compass/pull/5007

### DIFF
--- a/packages/compass-crud/src/components/bulk-update-dialog.tsx
+++ b/packages/compass-crud/src/components/bulk-update-dialog.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useState, useEffect } from 'react';
 import Document from './document';
 import HadronDocument from 'hadron-document';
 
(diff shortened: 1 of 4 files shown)


packages/compass-crud/src/components/bulk-update-dialog.tsx
```diff
@@ -163,6 +163,7 @@ export default function BulkUpdateDialog({
   const darkMode = useDarkMode();
 
   const [text, setText] = useState(updateText);
+  const [wasOpen, setWasOpen] = useState(isOpen);
```
Anemy, line 166: Should this be a `useRef` as it isn't involved in the rendering life cycle?
If someone calls `setWasOpen` I think we don't want to re-render.

lerouxb:
I have pushed a commit that demonstrates the problem I'm trying to work around. If you run this and start typing you will inevitably get the text editor into a state that it oscillates between two values.

packages/compass-crud/src/components/bulk-update-dialog.tsx
lerouxb: I switched to the usePrevious() pattern we already use elsewhere.

Anemy (approved):

PR #5049, opened 2023-11-01
attempt at adding a decorator

lerouxb:


packages/data-service/src/data-service.ts
```diff
@@ -2621,5 +2622,45 @@ function isTransactionAbortError(err: any) {
...
+        try {
+          return await target.call(this, ...args);
+        } catch (err: any) {
+          lastError = err;
+          if (err.codeName === 'WriteConflict') {
+            // retry
+            continue;
+          } else {
+            // immediately throw everything else including abort errors
+            throw err;
```
addaleax, line 2649: Should we maybe be using `session.withTransaction()` instead of `session.startTransaction()` + `session.abortTransaction()`? I think the callback-style API automatically retries on write conflict: https://mongodb.slack.com/archives/C9ZDY9M0E/p1698341325345119?thread_ts=1698304175.048259&cid=C9ZDY9M0E

PR #5076, opened 2023-11-07
feat: keep the delete and update buttons enabled despite query filters COMPASS-7362

kmruiz:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-crud/src/components/update-data-menu.tsx
+++ b/packages/compass-crud/src/components/update-data-menu.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Icon, Button, Tooltip } from '@mongodb-js/compass-components';
+import { Icon, Button } from '@mongodb-js/compass-components';
 import { usePreference } from 'compass-preferences-model';
 
 type UpdateMenuButtonProps = {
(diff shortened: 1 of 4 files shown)


packages/compass-crud/src/components/update-data-menu.tsx
```diff
@@ -31,39 +31,15 @@ const UpdateMenuButton: React.FunctionComponent<UpdateMenuButtonProps> = ({
...
-      isDisabled={isWritable}
-      justify="middle"
-      delay={500}
-    >
-      {disabledTooltip}
-    </Tooltip>
+    <UpdateMenuButton
+      isWritable={isWritable}
+      onClick={onClick}
+    ></UpdateMenuButton>
```
addaleax, line 42: We don’t end up including the limit/skip values in the update/delete (we can’t, right), correct? Do we need to do anything to make users aware of that?

lerouxb: Like I said in slack - the count will take into account the limit and skip and therefore be incorrect in the modal so I don't think we should be doing this.

kmruiz:
We are going to keep the current behaviour, we believe it's more user friendly.

PR #5077, opened 2023-11-07
fix(compass-sidebar): convert to ts, fix save connections bug COMPASS-7418

addaleax:
### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-sidebar/src/modules/connection-info.ts
```diff
@@ -25,11 +40,23 @@ export const INITIAL_STATE = {
...
+  | ChangeConnectionInfoAction
+  | SaveConnectionInfoAction;
+
+async function saveConnectionInfo(
+  connectionInfo: ConnectionInfo,
+  connectionStorage: typeof ConnectionStorage
+) {
   try {
-    await connectionStorage.save(connectionInfo);
+    await connectionStorage.save({ connectionInfo });
```
addaleax, line 57: ^ This is what was preventing connections from being saved

packages/compass-sidebar/src/modules/collection.js
```diff
@@ -1,48 +0,0 @@
-import toNS from 'mongodb-ns';
```
addaleax, line 1: This file was just completely unused, so I just removed it

packages/compass-sidebar/test/setup.js
```diff
@@ -1,15 +0,0 @@
-const hadronApp = require('hadron-app');
```
addaleax, line 1: This file and the ones in test/mocks as well, they didn’t seem to actually be consumed anywhere so I just removed them

packages/compass-sidebar/src/modules/connection-info.ts
mcasimir: Like completely?

mcasimir (approved):

packages/compass-sidebar/src/modules/connection-info.ts
addaleax: Yeah, we don’t have an e2e test for this

PR #5089, opened 2023-11-09
chore(scripts): make scripts/logserver.js output pipeable

baileympearson:
## Description

I found it useful to use `jq` to filter the output of the logserver, but the logserver currently logs with `console.dir` which is not parsable by jq.  This PR auto-detects when the output of the script is piped writes the output accordingly.  I also added colors for the non-jq use case.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/scripts/logserver.js
+++ b/scripts/logserver.js
@@ -2,7 +2,15 @@
 // replies with 200 Ok. This is very useful when debugging telemetry:
 // env HADRON_METRICS_SEGMENT_API_KEY='ignore' HADRON_METRICS_SEGMENT_HOST='http://localhost:8000' npm start
 const http = require('http');
+const { inspect } = require('util');
 
+function write(payload) {
+  if (process.stdout.isTTY) {
+    console.log(inspect(payload, { depth: Infinity, colors: true }));
+  } else {
+    process.stdout.write(JSON.stringify(payload) + '\n');
+  }
+}
 http
   .createServer((req, res) => {
     let body = '';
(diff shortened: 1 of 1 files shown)


scripts/logserver.js
```diff
@@ -2,7 +2,15 @@
 // replies with 200 Ok. This is very useful when debugging telemetry:
 // env HADRON_METRICS_SEGMENT_API_KEY='ignore' HADRON_METRICS_SEGMENT_HOST='http://localhost:8000' npm start
 const http = require('http');
+const { inspect } = require('util');
 
+function write(payload) {
+  if (process.stdout.isTTY) {
+    console.log(inspect(payload, { depth: Infinity, colors: true }));
```
addaleax, line 9: Is there a case in which we want `colors: true` in which we wouldn’t have already gotten color support? `console.dir()` does look at the color support (not just TTY-ness) of the output stream before writing to it to automatically determine the right `colors:` value.

baileympearson: This is how I've always done it in the past, I haven't looked closely into how console.dir works.  And I didn't even notice it already calls `inspect` - useful! 

`console.dir` doesn't output valid json though, so I think it still is worthwhile to leave the pipe detection and use JSON.stringify instead

addaleax (approved):

PR #5090, opened 2023-11-09
fix: pass relaxed: false to EJSON.deserialize() so values won't be promoted COMPASS-7431

lerouxb:


packages/compass-import-export/src/import/import-json.spec.ts
```diff
@@ -132,7 +132,7 @@ describe('importJSON', function () {
           hasUnboundArray: false,
         });
 
-        const docs = await dataService.find(ns, {});
+        const docs = await dataService.find(ns, {}, { promoteValues: false });
```
lerouxb, line 135: Had to add this so I could assert against the values without the values being promoted.

packages/compass-import-export/src/import/import-json.spec.ts
```diff
@@ -158,7 +158,7 @@ describe('importJSON', function () {
           throw err;
         }
 
-        const expectedResult = EJSON.parse(text);
+        const expectedResult = EJSON.parse(text, { relaxed: false });
```
lerouxb, line 161: Same here.

Anemy (approved):
lgtm! Nice test add.

PR #5093, opened 2023-11-09
feat(connections): add filter to saved connections when there are more than 10 saved COMPASS-7439

Anemy:
COMPASS-7439
This was a feature someone I bumped into at MongoDB world two years ago asked for, I saw it pop up again in the survey so I figured I'd spend an hour and add it. Probably won't be needed when we do multi-connection, however I figure it'll be useful for some power users in the short term that have many connections (and aren't familiar with the global cmd+f search).

Asked in slack to check with product if we should include: https://mongodb.slack.com/archives/GDTJXPHD0/p1699974533603229

https://github.com/mongodb-js/compass/assets/1791149/b74de6d6-99e9-4eb5-8f0e-b05c6990a323


addaleax (approved):
Code LGTM!

packages/compass-connections/src/components/connection-list/connection-list.tsx
```diff
@@ -211,8 +234,24 @@ function ConnectionList({
             isVisible={favoriteHeaderHover}
           ></ItemActionControls>
         </div>
+        {showFilteredSavedConnections && (
+          <TextInput
+            data-testid="sidebar-filter-saved-connections-input"
+            placeholder="Filter"
```
Anemy, line 240:
```suggestion
            placeholder="Search"
```

PR #5097, opened 2023-11-10
fix(stage-wizard): remove field selector from $count

mcasimir:
The `$count` accumulator doesn't take fields as parameter like the other accumulators. 
This was a pretty big assumption which now makes that specific case mess a bit the code, not sure if is the case to refactor it already, but in the interest of unblocking a release is probably a good idea to merge this one first.

Screenshot 2023-11-10 at 15 32 03


Anemy (approved):
lgtm, we can make this more generic/refactor down the line if we need similar uses.

PR #5102, opened 2023-11-10
chore(compass-sidebar): convert to new plugin interface COMPASS-7443

addaleax:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-sidebar/src/index.ts
```diff
@@ -1,25 +1,63 @@
...
+import type { MongoDBInstance } from 'mongodb-instance-model';
 
-/**
- * Activate all the components in the Sidebar package.
- * @param {Object} appRegistry - The Hadron appRegisrty to activate this plugin with.
- **/
-function activate(appRegistry: AppRegistry) {
-  appRegistry.registerComponent('Sidebar.Component', SidebarPlugin);
-  appRegistry.registerStore('Sidebar.Store', SidebarStore);
+function activate() {
```
Anemy, line 10: Can we remove this `activate` and `deactivate` now? Or are they supposed to be kept around for the time being?

Digging this new `registerHadronPlugin` usage !

packages/compass-sidebar/src/components/sidebar.tsx
```diff
@@ -219,6 +220,7 @@ const MappedSidebar = connect(mapStateToProps, {
   updateAndSaveConnectionInfo,
   toggleIsGenuineMongoDBVisible,
   setIsExpanded,
+  setConnectionIsCSFLEEnabled,
```
Anemy, line 223: Looks like we still have the `sidebar-toggle-csfle-enabled` event being emitted here, should we remove that and use this function?

packages/compass-sidebar/src/index.ts
addaleax: Right now, we are still loading all of these plugins into the PluginManager in the core compass package, and that interface expects these methods to be available. But that should be something we can get rid off after doing the rest of the conversion work :)

packages/compass-sidebar/src/components/sidebar.tsx
addaleax: Yeah, thanks for catching this … :see_no_evil:

gribnoysup (approved):

PR #5105, opened 2023-11-13
chore: add a button to save an update query from the Bulk Update modal COMPASS-7331

kmruiz:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-query-bar/src/stores/query-bar-reducer.ts
```diff
@@ -214,8 +214,21 @@ type ApplyFromHistoryAction = {
...
+  query: BaseQuery & { update?: Document }
+): QueryBarThunkAction<Promise<void>, ApplyFromHistoryAction> => {
+  return async (dispatch, getState, { globalAppRegistry }) => {
+    dispatch({
+      type: QueryBarActions.ApplyFromHistory,
+      query,
+    } as ApplyFromHistoryAction);
+
+    if (query.update) {
+      globalAppRegistry?.emit('favorites:open-bulk-update-favorite', query);
```
gribnoysup, line 227: This should be `localAppRegistry`, not global one. Emitting this on global means that every tab currently in runtime will receive the event and will open the modal, not only the one from which the action was actually triggered

packages/compass-query-bar/src/components/query-history/query-item/query-item-action-buttons.tsx
```diff
@@ -1,54 +1,85 @@
...
+    }
+  };
+
+export const FavoriteActionButton = ({
+  onClick,
+  visible,
+}: ActionButtonProps) => {
   return (
     <IconButton
+      className={css({
```
gribnoysup, line 26: Let's not use `css` in render, this have perf implications

Also if you need to remove the button completely, why not just render nothing here? (Or from the parent component as this seems to be added to handle the "hovered" state that parent controls)

packages/compass-query-bar/src/components/query-history/query-item/query-item-action-buttons.tsx
```diff
@@ -1,54 +1,85 @@
...
-  onClick: () => void;
+  onClick?: () => void;
+  visible?: boolean;
 };
 
-export const FavoriteActionButton = ({ onClick }: ActionButtonProps) => {
+const VISIBLE_DISPLAY = 'inline';
+const INVISIBLE_DISPLAY = 'none';
+
+const exhaustIfInteractive =
```
gribnoysup, line 12: Why do we need that? How can onClick be optional on a button that should cause an action?

packages/compass-query-bar/src/components/query-history-button-popover.tsx
```diff
@@ -64,6 +64,8 @@ const QueryHistoryButtonPopover = ({
     [onOpenPopover]
   );
 
+  const closePopover = useCallback(() => setIsOpen(false), [setIsOpen]);
```
gribnoysup, line 67: Why are we changing the behavior for update selected but not for the other saved queries types?

packages/compass-saved-aggregations-queries/src/stores/aggregations-queries-items.ts
```diff
@@ -125,7 +125,7 @@ const mapQueryToItem = (query: FavoriteQuery): Item => {
     lastModified: (query._dateModified ?? query._dateSaved).getTime(),
     database,
     collection,
-    type: 'query',
+    type: query.update ? 'updatemany' : 'query',
```
gribnoysup, line 128: I see this mapping, but no corresponding handling seems to be added anywhere in the compass-collection or crud, and so doesn't seem like trying to open an item from "my queries" actually does anything, at least it doesn't seem to work for me locally. Is it working for you?

packages/compass-query-bar/src/components/query-history/favorite-list.tsx
```diff
@@ -54,11 +60,13 @@ const FavoriteItem = ({
...
       header={(isHovered: boolean) => (
-        <QueryItemHeading title={query._name} isHovered={isHovered}>
+        <QueryItemHeading title={query._name} isHovered={true}>
           <CopyActionButton
+            visible={isHovered}
             onClick={() => copyToClipboard(formatQuery(attributes))}
           />
-          <DeleteActionButton onClick={onDeleteClick} />
+          <DeleteActionButton visible={isHovered} onClick={onDeleteClick} />
+          <OpensInModal visible={isUpdateQuery} />
```
gribnoysup, line 69: While it's nice that onClick does the same thing, I'd suggest we pass the callback here to be explicit about what action this button does instead of relying on card click handler for that

packages/compass-query-bar/src/components/query-history/query-item/query-item-action-buttons.tsx
```diff
@@ -1,54 +1,85 @@
...
-        onClick();
-      }}
+      onClick={exhaustIfInteractive(onClick)}
     >
       <Icon glyph="Trash" />
     </IconButton>
   );
 };
+
+export const OpensInModal = ({ onClick, visible }: ActionButtonProps) => {
```
gribnoysup, line 71: For consistency with other action buttons

```suggestion
export const OpenBulkUpdateActionButton = ({ onClick, visible }: ActionButtonProps) => {
```

packages/compass-query-bar/src/components/query-history/query-item/query-item-action-buttons.tsx
```diff
@@ -1,54 +1,85 @@
...
+    <IconButton
+      className={css({
+        display: visible ? VISIBLE_DISPLAY : INVISIBLE_DISPLAY,
+      })}
+      data-testid="query-opens-in-modal-button"
+      aria-label="Open in Modal"
+      title="Open in Modal"
+      onClick={exhaustIfInteractive(onClick)}
+    >
+      <Icon glyph="OpenNewTab" />
```
gribnoysup, line 82: We usually use this glyph to indicate that something will be opened in the new tab, maybe something to check with design: maybe there is a better icon for this?

packages/compass-query-bar/src/components/query-history/query-item/query-item-action-buttons.tsx
gribnoysup: Ah, okay, I see that it's because of https://github.com/mongodb-js/compass/pull/5105#discussion_r1391111472 and https://github.com/mongodb-js/compass/pull/5105#discussion_r1391094629. I think if you change it to use the conditional rendering and being explicit about what behavior action button will trigger, this might remove the need for that here

packages/compass-components/src/components/interactive-popover.tsx
```diff
@@ -182,19 +197,26 @@ function InteractivePopover({
...
+                id={closeButtonId}
+                ref={closeButtonRef}
+              >
+                <Icon glyph="X" />
+              </IconButton>
+            )}
+            {hasCustomCloseButton && (
+              <button id={closeButtonId} className={css({ display: 'none' })}>
+                Not Visible
+              </button>
```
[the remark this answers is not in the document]
kmruiz: This is in the meantime, to make a test pass. The FocusTrap component fails if there is no button that acts as a fallback. In my use case, the close button is custom, so to make sure the test passes while I find an alternative solution (this issue only happens in tests) I added this placeholder.

packages/compass-crud/src/components/bulk-update-dialog.tsx
```diff
@@ -107,6 +114,127 @@ const updatePreviewStyles = css({
...
+  }, [setOpen, setFavoriteName]);
+
+  const onClickSave = useCallback(() => {
+    onSave(favoriteName);
+    cleanClose();
+  }, [onSave, favoriteName, cleanClose]);
+
+  const updateFavoriteName = useCallback(
+    (ev: React.ChangeEvent) => {
+      const target = ev.target as any as { value: string };
```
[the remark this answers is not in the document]
kmruiz: Ah wonderful, thanks a lot! Fixing it.

packages/compass-crud/src/components/bulk-update-dialog.tsx
```diff
@@ -107,6 +114,127 @@ const updatePreviewStyles = css({
...
+      const favoriteName: string = (target.value || '').trim();
```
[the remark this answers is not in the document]
kmruiz: It was added to remove the initial and end spaces in the names when I was using onBlur, it's a left over, thanks for realising!

packages/compass-query-bar/src/components/query-history/favorite-list.tsx
kmruiz: Makes sense, I'll propagate the handler downwards.

packages/compass-saved-aggregations-queries/src/stores/aggregations-queries-items.ts
kmruiz: It is used in `packages/compass-saved-aggregations-queries/src/components/saved-item-card.tsx`, which despite it's name, it's the My Queries section.
Screenshot 2023-11-13 at 16 40 32

packages/compass-query-bar/src/components/query-history-button-popover.tsx
kmruiz: We decided to not add a breaking change, as there is a project in the future for reworking the query bar and this popover will be re-designed. The only reason we are closing here is because, due to some leafy-green hardcoded z-indexes, part of the popup shows over the modal.

packages/compass-query-bar/src/components/query-history/query-item/query-item-action-buttons.tsx
kmruiz: Yeah, I did the changes you suggested and simplified, I'll ask you for another review when I'm done with the rest of the comments 😄

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1970,6 +2003,18 @@ const configureStore = (options: CrudStoreOptions & GridStoreOptions) => {
       void store.refreshDocuments();
     });
 
+    globalAppRegistry.on(
```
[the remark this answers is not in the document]
kmruiz: Changed to the local app registry. It still works. I though that with the local app registry won't work :D.

packages/compass-query-bar/src/components/query-history-button-popover.tsx
gribnoysup: Hmm, if we're not planning to change the current behavior, then you'd definitely want to separate card click actions and open update modal actions in some way because right now it's implemented in a way where opening a recent query is not closing the popover, but opening a favorite (even without `update`) will close it which is just weirdly inconsistent.

Saying that though, I think chaging this popup to autoclose when item is selected _in all cases_ was on the list of things to do for awhile. @Anemy maybe you remember this and even haave a ticket reference?

packages/compass-saved-aggregations-queries/src/stores/aggregations-queries-items.ts
gribnoysup: Yeah, I mean, it is passed to the event too, that's cool, but nothing is handling it in a way that would actually make it work, opening the saved update item doesn't open update modal currently

packages/compass-query-bar/src/components/query-history-button-popover.tsx
kmruiz: If this is the case and there is a ticket, it could be handled in the backlog when the ticket is prioritised. Unless there is a big reason, apart from that the behaviour changed and it has been approved, I would keep the behaviour in the PR and continue with the plan of tackling it during the query bar improvements or in that specific ticket.

packages/compass-query-bar/src/components/query-history/query-item/query-item-action-buttons.tsx
kmruiz: Discussed and we are going to keep this for now. There doesn't seem to be a more clear icon for this behaviour (maybe we should ask for a new icon at some point :D)

packages/compass-query-bar/src/components/query-history-button-popover.tsx
gribnoysup: Sure, then we just need to fix the favorited queries behavior without the update modal showing up that this PR currently breaks 👍

kmruiz: Ah yes, I was testing your suggestion to use the localAppRegistry instead of the globalAppRegistry for opening the update modal, but it doesn't work because the communication has to be between plugins, so the only solution is to use the globalAppRegistry.

I'll revert that change and hopefully evergreen will be green 😄 .

packages/compass-components/src/components/interactive-popover.tsx
```diff
@@ -50,6 +50,8 @@ type InteractivePopoverProps = {
     ref: React.LegacyRef<HTMLButtonElement>;
     children: React.ReactNode;
   }) => React.ReactElement;
+  hasCustomCloseButton?: boolean;
```
gribnoysup, line 53: Can we come up with a better name here? If something has a button, you'd expect it to show and not be hidden?

packages/compass-components/src/components/interactive-popover.tsx
```diff
@@ -50,6 +50,8 @@ type InteractivePopoverProps = {
...
+  customTrapFallback?: string;
```
gribnoysup, line 54: Can we use a better name here? It's custom because you are already providing a prop. What it's supposed to trap?

packages/compass-components/src/components/interactive-popover.tsx
```diff
@@ -148,7 +165,7 @@ function InteractivePopover({
   const closeButtonId = useId('close-button-id');
 
   return trigger({
-    onClick: onClickTrigger,
+    onClick: onClickTrigger as any,
```
gribnoysup, line 168: There is almost never a case where you need to asser any for a good reason, the type of the event argument in this onClickTrigger method should be `React.MouseEventHandler<HTMLButtonElement>`

packages/compass-crud/src/components/bulk-update-dialog.tsx
```diff
@@ -93,6 +100,131 @@ const updatePreviewStyles = css({
...
+        cleanClose();
+      }
+    },
+    [favoriteName, onSave, cleanClose]
+  );
+
+  return (
+    <InteractivePopover
+      containedElements={inlineSaveQueryModalContainedElements}
+      // To prevent popover from closing when confirmation modal is shown
```
gribnoysup, line 176: Is this comment supposed to be for the line above?

packages/compass-crud/src/stores/crud-store.ts
gribnoysup, line 2006: This is still incorrect and causes an issue that triggers the modal to show up on all currently open collection tabs:

Kapture 2023-11-14 at 14 43 14

This event should be scoped to the `localAppRegistry`, not global one

kmruiz: As mentioned, it's between different plugins, so the localAppRegistry doesn't work, unless there is some configuration that is not set up properly, which I don't see anywhere. 

Probably the correct solution would be the tab to ignore the event if it's not visible, which is common behaviour in lots of applications. Is there a way to detect the visible tab on the event handler?

packages/compass-components/src/components/interactive-popover.tsx
kmruiz: Using custom, for specifying custom buttons, is something we use in leafy-green and in the code-base already:

https://www.mongodb.design/component/code/documentation/

And also for additional business logic, like in CodemirrorMultilineEditor, where we have customActions.

(I know these examples because are the ones I used as inspiration).

It could be an array of buttons like in the Code component, but it's more complex than an optional boolean for this case, where we just want to have a custom action for closing.

packages/compass-crud/src/stores/crud-store.ts
gribnoysup: The `localAppRegistry` is _specifically_ a mechanism for different plugins to communicate inside one scope. For plugins registered under `Collection.Tab` role, the localAppRegistry scope is defined. This is how `open-create-index-modal`, `open-*-export-to-language`, `open-explain-plan-modal` events work

packages/compass-components/src/components/interactive-popover.tsx
gribnoysup: That's exactly what I'm saying: both `customActions` and `customActionButtons` **_adds_** additional elements that render inside the component when provided, you added a property called `hasCustomCloseButton` that **_removes_** the element rendered inside the component when provided

packages/compass-crud/src/stores/crud-store.ts
gribnoysup: Can you clarify a bit what's not working for you? I pulled your branch locally and just replaced globalAppRegistry for the emit and on for this event to localAppRegistry with no other changes and it works as expected

packages/compass-crud/src/components/bulk-update-dialog.tsx
kmruiz: Yep, but I'm going to remove it, I don't think this comment adds much information. Ty ❤️ .

packages/compass-components/src/components/interactive-popover.tsx
kmruiz: It could be also a React.TouchEventHandler, if you see the signature of onClickTrigger, it's (event: MouseEvent | TouchEvent) => void;

Unless there is a way to specify that is both of them (or a common parent), it would require changing the typing of all components using the popover, which then I would suggest to have a specific task to clean up the whole InteractivePopover as there are already a few things that might need a refactor.

packages/compass-components/src/components/interactive-popover.tsx
gribnoysup: `customFocusTrapFallback` at least so we know this is not related to the existing `containedElements` prop that also affects trap boundaries, but for close on click and not for focus?

packages/compass-components/src/components/interactive-popover.tsx
kmruiz: I'll be using this type:

`const onClickTrigger: React.EventHandler<React.MouseEvent<HTMLButtonElement> | React.TouchEvent<HTMLButtonElement>> = useCallback(
    (event: React.MouseEvent<HTMLButtonElement> | React.TouchEvent<HTMLButtonElement>) => void
  );`

It's a bit more complex, but doesn't require the any.

gribnoysup: How is the TouchEvent relevant here? Both browser events and React onClick callbacks will always provide a MouseEvent instance on click

packages/compass-components/src/components/interactive-popover.tsx
kmruiz: Yeah, that sounds good, if it's more explicit and it helps is just changing one name.

packages/compass-components/src/components/interactive-popover.tsx
```diff
@@ -91,14 +95,33 @@ function InteractivePopover({
...
-      onClose();
-      return;
-    }
+  const onClickTrigger: React.EventHandler<
+    React.MouseEvent<HTMLButtonElement> | React.TouchEvent<HTMLButtonElement>
+  > = useCallback(
+    (
+      event:
+        | React.MouseEvent<HTMLButtonElement>
+        | React.TouchEvent<HTMLButtonElement>
```
gribnoysup, line 104: Can you clarify why touch event is added here to the type definition? This is passed as an `onClick` listener, both browser click events and corersponding React `onClick` callbacks only return MouseEvent instances for click event

packages/compass-crud/src/stores/crud-store.ts
kmruiz: I think I was doing something wrong when I tested it, as mentioned in the standup. I tried again and it worked flawlessly, as you mentioned, so I'm going to commit the change ❤️ .

packages/compass-components/src/components/interactive-popover.tsx
kmruiz: I could change the name to overrideCloseButton, is it better?

packages/compass-components/src/components/interactive-popover.tsx
kmruiz: I can't because the component was already there, I just adapted it to my needs. Compass is not touch-enabled, so I wouldn't add it, but didn't want to change something that works, that doesn't leak outside and is tested because the typescript compiler complains.

kmruiz: I mean, I can remove the support for the TouchEvent, it shouldn't affect in runtime, but probably I would do it async in another small PR and I would ask whoever did this and way.

gribnoysup: I'm not sure what's causing the hiccup here, this event didn't have any types before, you are adding them, the `trigger` `onClick` callback is expecting only the `MouseEvent` already, so why are we adding a `TouchEvent` where nothing really expects it and it's not correct in the first place? I'm not asking for any major refactoring of existing code, just not to add code that is not needed in the first place

Are you somehow mixing it with the `clickEventListener` below? This is a separate method, doesn't look like you are modifying it in this PR at all, and `clickEventListener` is indeed passed both to `mousedown` and `touchstart` so it's warranted to have a union type there

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -2003,7 +2004,7 @@ const configureStore = (options: CrudStoreOptions & GridStoreOptions) => {
       void store.refreshDocuments();
     });
 
-    globalAppRegistry.on(
+    localAppRegistry?.on(
```
gribnoysup, line 2007: We'd probably want to group this with other localAppRegistry events above for consistency?

kmruiz: Sure, done.

packages/compass-components/src/components/interactive-popover.tsx
gribnoysup: I'd go with hideCloseButton as this is literally what this property does, you're not passing any overrides here, nor you can force this somehow with this property, but feel free to pick up whatever makes sense to you or just resolve this

kmruiz: I'll change it to the name you suggested and resolve.

packages/compass-components/src/components/interactive-popover.tsx
kmruiz: You are right, it was part of a refactor, cleaning up and commiting.

packages/compass-components/src/components/interactive-popover.tsx
```diff
@@ -167,7 +184,7 @@ function InteractivePopover({
           focusTrapOptions={{
             clickOutsideDeactivates: true,
             // Tests fail without a fallback. (https://github.com/focus-trap/focus-trap-react/issues/91)
-            fallbackFocus: `#${closeButtonId}`,
+            fallbackFocus: customFocusTrapFallback || `#${closeButtonId}`,
```
mcasimir, line 187: NIT: do we need a custom fallback? it seems it was only needed for the tests of the popover, probably i'm wrong though. if you don't specify anything it should focus on the first focusable element on its own

kmruiz: TBH, the main reason is testing, the whole plugin fails if not. Another option I had earlier was an empty invisible component when there is no closing button, however, it's even worse.

However, with this behaviour, right now, if someone changes the interactive popover that we are using and doesn't add a focusable component, it will focus on the close button, instead of breaking the UI, which is more resilient. For example, if for some reason we disable the input until we load something (for example to check existing names to avoid overwriting) or any other behaviour we might think of.

packages/compass-query-bar/src/components/query-history/favorite-list.tsx
```diff
@@ -54,11 +63,16 @@ const FavoriteItem = ({
...
+        <QueryItemHeading title={query._name} isHovered={true}>
+          {isHovered && (
+            <CopyActionButton
+              onClick={() => copyToClipboard(formatQuery(attributes))}
+            />
+          )}
+          {isHovered && <DeleteActionButton onClick={onDeleteClick} />}
+          {isUpdateQuery && (
+            <OpenBulkUpdateActionButton onClick={onCardClick} />
+          )}
```
gribnoysup, line 75: Should this account for views (potentially other readonly collections)? We show `update` entries in the list, but selecting them doesn't open the update modal:

image

packages/compass-query-bar/src/stores/query-bar-reducer.ts
```diff
@@ -214,8 +214,21 @@ type ApplyFromHistoryAction = {
   query: BaseQuery;
 };
 
-export const applyFromHistory = (query: BaseQuery): ApplyFromHistoryAction => {
-  return { type: QueryBarActions.ApplyFromHistory, query };
+export const applyFromHistory = (
+  query: BaseQuery & { update?: Document }
+): QueryBarThunkAction<Promise<void>, ApplyFromHistoryAction> => {
```
gribnoysup, line 219: There is no async operations in this actions, why is it async?

packages/compass-saved-aggregations-queries/src/stores/aggregations-queries-items.ts
gribnoysup, line 1: The `updatemany` items are visible in the list even if I disable the feature flag, should this plugin take feature flag that should control the feature into account?

packages/compass-query-bar/src/components/query-history/favorite-list.tsx
gribnoysup, line 1: The `updatemany` items are visible in the list even if I disable the feature flag, should this plugin take feature flag that should control the feature into account?

packages/compass-query-bar/src/stores/query-bar-reducer.ts
```diff
@@ -214,8 +214,21 @@ type ApplyFromHistoryAction = {
...
+  return async (dispatch, getState, { localAppRegistry }) => {
+    dispatch({
+      type: QueryBarActions.ApplyFromHistory,
+      query,
+    } as ApplyFromHistoryAction);
```
gribnoysup, line 224: Why are we asserting here? The action shape already matches `ApplyFromHistoryAction` type

packages/compass-saved-aggregations-queries/src/stores/aggregations-queries-items.ts
kmruiz: You can't save it unless the feature flag is on, which is only off because the feature is disabled. We are not expecting to disable it.

packages/compass-query-bar/src/components/query-history/favorite-list.tsx
kmruiz: I don't understand the question, can you elaborate? Clicking on a favourite update opens the modal, unless you've found a bug, can you share how you reproduce it?

https://github.com/mongodb-js/compass/assets/1822138/fb5a5db7-d3fe-4f3a-946f-fdd4306ea3f7

gribnoysup: Sure! Here's one of the scenarios to repro this:

* Connect with a user that has readwrite privileges
* Save a bulkUpdate
* Disconnect and connect with a user that has readonly privileges or in a way that makes the connection readonly
* You see the bulkUpdate item in the drop down, but clicking it doesn't show update modal (probably because something correctly checking if readonly is true for either connection or namespace elsewhere in the plugin, but not here in any way)

Saved items here are filtered only by namespace, they are not attached to a specific connection (on purpose, you might want to use then on different connection, like dev and prod), so there is a decent chance that saved bulk update will show up in case where collection or cluster is readonly.

The same thing also applies to "My Queries" screen.

Should we account for that or is this expected behavior that we show the item that does nothing on click?

packages/compass-saved-aggregations-queries/src/stores/aggregations-queries-items.ts
gribnoysup: A bit annoying that it stays there and does nothing when you disable the options, but sure, if that's how we want to handle that 👍

packages/compass-query-bar/src/components/query-history/favorite-list.tsx
kmruiz: This is a corner case that we didn't take into account when we design the feature, so probably we should discuss it further with Misba and Le Roux. I personally would grey-out update queries in readonly connections, because you might want still to read the query even if you can't run it.

Do you mind if I open a ticket so we can track this specific use case? I think it's important to get design input, because we might want to have another flow for readonly collections (IDK, a copy button instead of opening the modal?).

gribnoysup: Sure, if we address this before the feature is ready for public release, that's fine by me

kmruiz: I created the ticket and mentioned you in the conversation in Slack. I'm resolving here so we can continue the discussion with Misba and Le Roux.

gribnoysup (approved):

PR #5107, opened 2023-11-13
fix(data-service): don't abort the transaction from 'outside' the transaction COMPASS-7368

lerouxb:
Long story, but this is sorta related https://github.com/mongodb/node-mongodb-native/pull/3916 and I investigated the issue here https://github.com/lerouxb/unhandled-rejection

Should prevent this:
image (8)

### DIFF
--- a/packages/data-service/src/data-service.ts
+++ b/packages/data-service/src/data-service.ts
@@ -2421,11 +2421,13 @@ class DataServiceImpl extends WithLogContext implements DataService {
           throw err;
         }
       },
-      async (session) => {
-        if (session.inTransaction()) {
-          await session.abortTransaction();
-          await session.endSession();
-        }
+      async () => {
+        // Rely on the session being killed when we cancel the operation. It can
+        // take a few seconds before the driver reacts, but the Promise.race()
+        // against the abort signal should cause the UI to immediately be
+        // notified that the operation was cancelled. We don't abort the
+        // transaction or end the session as well because the code we run inside
+        // session.withTransaction() above would experience race conditions.
       },
       abortSignal
     );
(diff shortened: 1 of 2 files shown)


packages/data-service/src/data-service.ts
```diff
@@ -2421,11 +2421,13 @@ class DataServiceImpl extends WithLogContext implements DataService {
...
-          await session.abortTransaction();
-          await session.endSession();
-        }
+      async () => {
+        // Rely on the session being killed when we cancel the operation. It can
+        // take a few seconds before the driver reacts, but the Promise.race()
+        // against the abort signal should cause the UI to immediately be
+        // notified that the operation was cancelled. We don't abort the
+        // transaction or end the session as well because the code we run inside
+        // session.withTransaction() above would experience race conditions.
```
addaleax, line 2430: Should we also end the session in the `isTransactionAbortError()` error case in the try/catch block above?

addaleax (approved):

packages/data-service/src/data-service.ts
lerouxb: It can now only be ever be aborted by the block:

```
              if (session.inTransaction()) {
                await session.abortTransaction();
                await session.endSession();
              }
```

So that `if (session.inTransaction()) {` is probably unnecessary and if `isTransactionAbortError()` ever returns true then you definitely made it past that block because that's the only way the transaction could have been aborted.

Well. Technically the error "cannot use a session that has ended" would also count as a transaction abort error, but in that case by definition the session is already ended anyway ;)

addaleax: > So that `if (session.inTransaction()) {` is probably unnecessary

… maybe add a comment about that in that block, because its existence implies that it is something that _can_ happen? :slightly_smiling_face:

PR #5111, opened 2023-11-14
fix(compass-components): do not keep fallback content when not needed COMPASS-7417

himanshusinghs:
## Description
The fallback content kept rendering css keyframes with opacity 0 which ended up consuming a lot of CPU as seen in the Activity monitor. As I understand, we probably wanted to workaround the inability to animate css display property here and although with the change in place, there is a slight difference in the output, it is probably worth getting to normal CPU usage.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-components/src/components/content-with-fallback.tsx
+++ b/packages/compass-components/src/components/content-with-fallback.tsx
@@ -113,16 +113,12 @@ const fallbackContainer = css({
   right: 0,
   bottom: 0,
   left: 0,
-  opacity: 0,
-  transition: 'opacity display .16s ease-out',
   pointerEvents: 'none',
   display: 'none',
 });
 
 const fallbackContainerVisible = css({
-  opacity: 1,
   display: 'block',
-  transitionTimingFunction: 'ease-in',
 });
 
 type FadeInPlaceholderProps = {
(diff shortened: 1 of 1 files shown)


packages/compass-components/src/components/content-with-fallback.tsx
```diff
@@ -114,12 +114,14 @@ const fallbackContainer = css({
   bottom: 0,
   left: 0,
   opacity: 0,
-  transition: 'opacity .16s ease-out',
+  transition: 'opacity display .16s ease-out',
```
mcasimir, line 117: i almost feel this can just be confusing at this point, should we just drop it?

mcasimir: unless i'm missing something ..

himanshusinghs: this has a nice side-effect that the visibility(coming into view) of fallback content still gets animated nicely.

himanshusinghs: actually there very little difference (just checked again), maybe i will just remove it to cut down any confusion here.

himanshusinghs: done 👍

mcasimir (approved):

PR #5115, opened 2023-11-15
feat(connection-storage): migrate from keytar COMPASS-6854

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/connection-storage/src/connection-storage.ts
```diff
@@ -98,11 +107,128 @@ export class ConnectionStorage {
...
+      }
+    );
+
+    if (numFailedToMigrate > 0) {
+      track('Keytar Secrets Migration Failed', {
+        num_saved_connections:
+          unmigratedConnections.length - numFailedToMigrate,
+        num_failed_connections: numFailedToMigrate,
+      });
+    }
```
himanshusinghs, line 197: Nothin critical, just wanted to point out that in case of a few failed connections we will be tracking both `Migration complete` and `Keytar Secrets Migration Failed`. I wonder if it would be fine just to track `Keytar Secrets Migration Failed`? I don't know the tracking requirements so pardon me if I did not take that into account 😄

packages/connection-storage/src/connection-storage.spec.ts
```diff
@@ -75,6 +87,238 @@ describe('ConnectionStorage', function () {
...
+        await ConnectionStorage.migrateToSafeStorage();
+
+        expect(
+          encryptSecretsSpy.called,
+          'it does not try to encrypt'
+        ).to.be.false;
+        expect(
+          getKeytarCredentialsSpy.called,
+          'it does not try to get secrets from keychain'
+        ).to.be.false;
```
himanshusinghs, line 128: TIL! Didn't know before that we could add custom message in `expect` for the failure cases. Thanks for this ⭐

himanshusinghs (approved):
Looks great! ⭐

himanshusinghs (approved):
Looks great! ⭐

packages/connection-storage/src/connection-storage.ts
himanshusinghs: Sorry I just re-read the highlighted code, one is tracking and one is logging. Please ignore.

packages/connection-storage/src/connection-storage.spec.ts
mabaasit: yeah, makes it easier to understand the expectation when there're many.

packages/connection-storage/src/connection-storage.ts
```diff
@@ -98,11 +107,128 @@ export class ConnectionStorage {
...
-    storedConnectionInfo: ConnectionInfo,
+  static async migrateToSafeStorage() {
+    // If user lands on this version of Compass with legacy connections (using storage mixin),
+    // we will ignore those and only migrate connections that have connectionInfo.
+    const connections = (await this.getConnections()).filter(
+      (x) => x.connectionInfo
+    );
+    if (connections.length === 0) {
+      log.info(
+        mongoLogId(1_001_000_200),
```
addaleax, line 118:
```
$ node scripts/check-logids.js 
Next available log id: 1_001_000_270 (1001000270)
```

We’re not re-using old log ids unless they actually refer to the same thing, `1_001_000_200` was previously `Failed to parse connection` (removed in 2c91060412b4).

packages/connection-storage/src/connection-storage.ts
```diff
@@ -98,11 +107,128 @@ export class ConnectionStorage {
...
+      );
+      return;
+    }
+
+    if (
+      connections.filter((x) => 'version' in x && x.version === this.version)
+        .length === connections.length
+    ) {
+      log.info(
+        mongoLogId(1_001_000_205),
```
addaleax, line 130: ditto

packages/connection-storage/src/connection-storage.ts
```diff
@@ -98,11 +107,128 @@ export class ConnectionStorage {
...
+  private static mapStoredConnectionToConnectionInfo({
+    connectionInfo: storedConnectionInfo,
+    connectionSecrets: storedConnectionSecrets,
+  }: ConnectionProps): ConnectionInfo {
+    let secrets: ConnectionSecrets | undefined = undefined;
+    try {
+      secrets = this.decryptSecrets(storedConnectionSecrets);
+    } catch (e) {
+      log.error(
+        mongoLogId(1_001_000_208),
```
addaleax, line 225: ditto

packages/connection-storage/src/connection-storage.ts
```diff
@@ -98,11 +107,128 @@ export class ConnectionStorage {
     this.calledOnce = true;
   }
 
-  private static mapStoredConnectionToConnectionInfo(
-    storedConnectionInfo: ConnectionInfo,
+  static async migrateToSafeStorage() {
+    // If user lands on this version of Compass with legacy connections (using storage mixin),
+    // we will ignore those and only migrate connections that have connectionInfo.
+    const connections = (await this.getConnections()).filter(
```
addaleax, line 113: If this fails – it could if the connections directory does not exist or is not readable – the entire app wouldn’t start anymore, because this is called early during startup. That does not seem okay.

packages/connection-storage/src/connection-storage.ts
```diff
@@ -98,11 +107,128 @@ export class ConnectionStorage {
...
+        num_secrets: Object.keys(secrets).length,
+      }
+    );
+
+    let numFailedToMigrate = 0;
+
+    for (const { connectionInfo } of unmigratedConnections) {
+      // We have connectionInfo as we are limiting this migration to
+      // connections that have connectionInfo (not legacy connections)
+      const _id = connectionInfo!.id;
```
addaleax, line 159: Would we still want to provide a more explicit error message in this case than something like `cannot read property id of undefined`?

packages/connection-storage/src/connection-storage.ts
mabaasit: That's a good point. However, we resolved this in [user-data](https://github.com/mongodb-js/compass/blob/eda417439a906b9d70656b8a7020005547febc37/packages/compass-user-data/src/user-data.ts#L89), where as we try to read from a path, we ensure that the directory exists. 

And if the directory is not readable for some reason, we have [ignoreErrors](https://github.com/mongodb-js/compass/blob/main/packages/compass-user-data/src/user-data.ts#L256) prop which defaults to true and this method returns an empty list.

packages/connection-storage/src/connection-storage.ts
mabaasit: ah! didn't know about this. thanks.

packages/connection-storage/src/connection-storage.ts
mabaasit: the connections list is already being [filtered](https://github.com/mongodb-js/compass/pull/5115/files/7cd05a37bf6c506bd960844c80f514573745ed85#diff-11d5f2650641bd452506da099a11965f742934b623ee0aebcd6028658934a419R113) as we fetch it ensuring that we only consider connections that have `connectionInfo` property.

packages/connection-storage/src/connection-storage.ts
mabaasit: fixed in [7d055e0](https://github.com/mongodb-js/compass/pull/5115/commits/7d055e0ceb031894ffd1eb55dea4fe0225dd9f38)

packages/connection-storage/src/connection-storage.ts
addaleax: > And if the directory is not readable for some reason, we have [ignoreErrors](https://github.com/mongodb-js/compass/blob/main/packages/compass-user-data/src/user-data.ts#L256) 

It doesn't ignore errors from `readdir`, though, which is what I was thinking of

> That's a good point. However, we resolved this in [user-data](https://github.com/mongodb-js/compass/blob/eda417439a906b9d70656b8a7020005547febc37/packages/compass-user-data/src/user-data.ts#L89), where as we try to read from a path, we ensure that the directory exists.

A directory existing doesn't strictly mean that it's readable:

```js
> fs.mkdirSync('/root', {recursive: true})
undefined
> fs.readdirSync('/root')
Uncaught Error: EACCES: permission denied, scandir '/root'
```

I don't think this is something that is likely to happen in Compass, but still, maybe the top-level call to the migration function could be wrapped in a `try/catch` just to make sure?

packages/connection-storage/src/connection-storage.ts
```diff
@@ -98,11 +107,128 @@ export class ConnectionStorage {
     this.calledOnce = true;
   }
 
-  private static mapStoredConnectionToConnectionInfo(
-    storedConnectionInfo: ConnectionInfo,
+  static async migrateToSafeStorage() {
+    // If user lands on this version of Compass with legacy connections (using storage mixin),
+    // we will ignore those and only migrate connections that have connectionInfo.
+    const connections = (await this.getConnections()).filter(
+      (x) => x.connectionInfo
```
addaleax, line 114: very optional, but makes it clear below that the filtering has already actually happened:

```suggestion
      (x): x is ConnectionWithLegacyProps & Required<Pick<ConnectionProps, 'connectionInfo'>> => !!x.connectionInfo
```

packages/connection-storage/src/connection-storage.ts
addaleax: Yeah, good point – I’ve left a comment above that allows you to get rid of the non-null assertion here by fully typing this out, if you like, otherwise feel free to just resolve

packages/connection-storage/src/connection-storage.ts
mabaasit: Yes, now that all makes sense - `readdir` being outside try/catch, maybe we should also handle it with `options.ignoreErrors`? 

I'll wrap the migration in try/catch.

addaleax: > maybe we should also handle it with options.ignoreErrors?

Yes, I think that would also make sense :)

packages/connection-storage/src/connection-storage.spec.ts
```diff
@@ -75,6 +87,238 @@ describe('ConnectionStorage', function () {
...
+          version: 1,
+        });
+
+        const encryptSecretsSpy = sandbox.spy(
+          ConnectionStorage,
+          <any>'encryptSecrets'
+        );
+        const getKeytarCredentialsSpy = sandbox.spy(
+          ConnectionStorage,
+          <any>'getKeytarCredentials'
```
addaleax, line 152:
```suggestion
          'getKeytarCredentials' as any
```

we generally don't do the `<...>` syntax because that's a) inconsistent with the rest of the codebase and b) has broken some scripts in the past (depcheck at least, iirc)

mabaasit: fixed in [df48b1b](https://github.com/mongodb-js/compass/pull/5115/commits/df48b1bffc7ac8136fe35a45425aca2e190d7a9c)

PR #5119, opened 2023-11-15
chore(preferences-model): remove generative ai feature flag

Anemy:


mcasimir:
Just double checking, this is not going to enable it on DE too?

Anemy:
@mcasimir No this wouldn't enable it, it's still dependent on a `enableGenAIFeatures` preference. That being said I'm thinking we should add something like `showGenerativeAI`, similar to what we do with `showExportToLanguageButton` and `showExplainButton`, so that this feature enablement isn't fully reliant on Compass' preferences assumptions.

https://github.com/mongodb-js/compass/blob/eda417439a906b9d70656b8a7020005547febc37/packages/compass-query-bar/src/components/query-bar.tsx#L342
https://github.com/mongodb-js/compass/blob/eda417439a906b9d70656b8a7020005547febc37/packages/compass-preferences-model/src/utils.ts#L51

mcasimir (approved):

PR #5128, opened 2023-11-16
WIP: chore(build): rename hadron-build and move to typescript

mcasimir:
Turned out to be a pretty big change, nothing urgent to deal with, opened the PR to start testing on evg


packages/compass/scripts/check-bundled-app-size.ts
```diff
@@ -1,7 +1,7 @@
 import { promises as fs } from 'fs';
 import path from 'path';
 
-const MAX_SIZE_MB = 200;
+const MAX_SIZE_MB = 400;
```
mcasimir, line 4: NOTE, this is just a WIP, debugging if there is a real issue, but this script checks the size of `./compass/dist` which is not the bundle size but the sum of all the artifacts created by one task.

I'm going to move this check in release.ts, so it runs as part of the packaging on the artifacts created

PR #5140, opened 2023-11-21
feat(bulk-update): render document count and restyle query preview 

mcasimir:
Screenshot 2023-11-21 at 08 27 01
Screenshot 2023-11-21 at 08 27 41
Screenshot 2023-11-21 at 08 28 54

### DIFF
--- a/packages/compass-crud/src/components/bulk-update-dialog.tsx
+++ b/packages/compass-crud/src/components/bulk-update-dialog.tsx
@@ -306,7 +306,7 @@ export default function BulkUpdateDialog({
     return `Update ${count} documents`;
   }, [count]);
 
-  const bulkUpdateUpdateId = useId('bulk-update-update');
+  const bulkUpdateUpdateId = useId();
   return (
     <Modal
       open={isOpen}
--- a/packages/compass-crud/src/components/bulk-update-dialog.tsx
+++ b/packages/compass-crud/src/components/bulk-update-dialog.tsx
@@ -293,6 +293,18 @@ export default function BulkUpdateDialog({
     }
   }, [isOpen, wasOpen, updateText]);
 
+  const modalTitleAndButtonText = useMemo(() => {
+    if (typeof count !== 'number') {
+      return 'Update documents';
+    }
+
+    if (count === 1) {
+      return `Update 1 document`;
+    }
+
+    return `Update ${count} documents`;
+  }, [count]);
+
   return (
     <Modal
       open={isOpen}
@@ -300,10 +312,7 @@ export default function BulkUpdateDialog({
       size="large"
       data-testid="bulk-update-dialog"
     >
-      <ModalHeader
-        title={`Update ${count ?? ''} document${count === 1 ? '' : 's'}`}
-        subtitle={ns}
-      />
+      <ModalHeader title={modalTitleAndButtonText} subtitle={ns} />
       <ModalBody>
         <div className={columnsStyles}>
           <div className={queryStyles}>
@@ -317,7 +326,10 @@ export default function BulkUpdateDialog({
             <div className={cx(queryFieldStyles, updateFieldStyles)}>
               <Label htmlFor="bulk-update-update">Update</Label>
               <Description className={descriptionStyles}>
-                <Link href="https://www.mongodb.com/docs/manual/reference/method/db.collection.updateMany/#std-label-update-many-update">
+                <Link
+                  tabIndex={0}
+                  href="https://www.mongodb.com/docs/manual/reference/method/db.collection.updateMany/#std-label-update-many-update"
+                >
                   Learn more about Update syntax
                 </Link>
               </Description>
@@ -398,7 +410,7 @@ export default function BulkUpdateDialog({
             onClick={runBulkUpdate}
             data-testid="update-button"
           >
-            Update {count ?? ''} document{count === 1 ? '' : 's'}
+            {modalTitleAndButtonText}
           </Button>
         </div>
       </ModalFooter>
--- a/packages/compass-crud/src/components/bulk-update-dialog.tsx
+++ b/packages/compass-crud/src/components/bulk-update-dialog.tsx
@@ -23,6 +23,7 @@ import {
   Icon,
   InteractivePopover,
   TextInput,
+  useId,
 } from '@mongodb-js/compass-components';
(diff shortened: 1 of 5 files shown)


lerouxb (approved):

packages/compass-crud/src/components/readonly-filter.tsx
```diff
@@ -50,20 +51,19 @@ export function ReadonlyFilter({
...
+        label={queryLabel}
+        tooltip="Return to the Documents tab to edit this query."
+      />
+      <KeylineCard
+        id="readonly-filter"
+        data-testid="readonly-filter"
+        disabled={true}
+        className={textInputStyles}
+      >
+        <code>{filterQuery}</code>
```
gribnoysup, line 65: This is not using leafygreen monospace font styles

mcasimir: fixed, thanks for pointing it out

packages/compass-crud/src/components/readonly-filter.tsx
```diff
@@ -33,11 +32,15 @@ const QueryLabel: React.FunctionComponent<QueryLabelProps> = ({
...
-const textInputStyles = css({
-  input: {
-    fontFamily: fontFamilies.code,
-  },
+const readOnlyFilterStyles = css({
+  padding: spacing[2],
+  overflow: 'scroll',
   width: '100%',
+  whiteSpace: 'nowrap',
+  fontFamily: fontFamilies.code,
```
gribnoysup, line 40: You'd need to apply it to the code tag here because user-agent styles applied to it take precedence over parent styles, it won't inherit them by default:

image

mcasimir: I wonder .. can it be that we just need to wrap everything in a <Body> element?

gribnoysup: I'm not sure what you mean by that. The `code` element has default, browser-applied, font-family styles. You can't override them by applying styles to parent elements, you'd need to actually target this element with a selector to change it's styles, that's just how css works

You can also not use `code` here at all, it doesn't make any difference and doesn't add any accessibility features

kmruiz (approved):

packages/compass-crud/src/components/bulk-update-dialog.tsx
```diff
@@ -293,17 +294,28 @@ export default function BulkUpdateDialog({
...
+    }
+
+    if (count === 1) {
+      return `Update 1 document`;
+    }
+
+    return `Update ${count} documents`;
+  }, [count]);
+
+  const bulkUpdateUpdateId = useId('bulk-update-update');
```
gribnoysup, line 309: Passing a default id just returns this id, is that the behavior we want? Why do we need to use the hook then?

packages/compass-crud/src/components/readonly-filter.tsx
```diff
@@ -49,21 +43,23 @@ export function ReadonlyFilter({
...
+      <div className={queryLabelStyles}>
+        <Label htmlFor={readOnlyFilterId}>{queryLabel}</Label>
+        <InfoSprinkle align="right">
+          Return to the Documents tab to edit this query.
+        </InfoSprinkle>
+      </div>
+      <KeylineCard
+        id={readOnlyFilterId}
+        data-testid="readonly-filter"
+        disabled={true}
```
gribnoysup, line 58: This is probably a leftover from this being a textinput? There is no meaning in passing disabled on a div

packages/compass-crud/src/components/readonly-filter.tsx
mcasimir: no what i mean is, shouldn't the `<Body as=div>` or some other root LG element be doing that? I would expect everything to have the right line height, base font size and font families from an element on. We could do that on our end, but if feels a bit strange

packages/compass-crud/src/components/bulk-update-dialog.tsx
mcasimir: well .. TIL, `useId` doesn't take the id prefix as argument and always used it wrong :)

packages/compass-crud/src/components/readonly-filter.tsx
gribnoysup: That's not exactly how css works. There are two things to take into account here: by default if no value is set for a css property, whether or not parent will affect it will depend on whether or not property is inherited. In case of `font` is the property is inherited, yes, but another thing, as I said, `code` element has a specific style set for font-family by the browser, so whatever it inherits from the parent, is overriden by an explicit type set by the user agent

I'm not sure what you mean by font-size, I see you set it to 14px for this specific element. That overrides the 13px value inherited from a parent element that sets it to 13px value as expected for an inherited property

mcasimir: I mean, we don't need anything for this PR, but I really don't think we are understanding each other, Lg has 2 base font sizes, 13 shouldn't be one of them.

My knowledge of css is a bit outdated, but I should know how a property propagates from the parent to children, that is not a recent feature. What I'm saying is that I would expect this css to be provided by LG somewhere (i thought as part of <Body>):

```
.lg-body-classname code {
   font-family: the right monospace one
}
```

which, in my book would override the default browser styles, cause the selector is more specific. The reasoning here is that having font families and sizes misaligned accidentally in a portion of the app isn't really good.

gribnoysup: Ah, yes, then we are talking about different things. Leafygreen Body is a paragraph and I don't think it's expected to be used the way we are using them, this is not a root component and there is no concept for something like that in lg and so I kinda get why they are not doing what you expect them to do. They do have an [InlineCode](https://www.mongodb.design/component/typography/) component exported from typography it seems

gribnoysup: To clarify, I think what they expect us to do is not 

```
<Body>
  <code> ...
  <p> ...
  <a> ....
</Body>
```

and expect it to provide styles for children, but just use their typography components directly where applicable

```
<div>
  <InlineCode> ...
  <Body> ...
  <Link> ...
</div>
```

mcasimir: Ahh yeah, ok that's what i was looking for thanks!

PR #5147, opened 2023-11-21
Fix dnd conflict with container layout

mcasimir:


packages/compass-components/src/components/workspace-container.tsx
```diff
@@ -6,20 +6,20 @@ import { rgba } from 'polished';
...
   display: 'flex',
   flexDirection: 'column',
   overflow: 'hidden',
-  containerName: workspaceContainerQueryName,
-  containerType: 'inline-size',
 });
 
 const toolbarStyles = css({
   flex: 'none',
+  containerName: workspacetoolbarContainerQueryName,
```
himanshusinghs, line 21: TIL! This is amazing 😮

himanshusinghs (commented):
I don't entirely understand how this fixes the problem but learnt of two new css properties, thanks for that.

PR #5162, opened 2023-11-24
chore(connection-form): split connection form from compass preferences

mcasimir:
@himanshusinghs @Anemy this should allow to use the form in VSCode without extra deps and have it reflect potential changes from VSCode preferences. Feel free to do what you deem best with it, so this is not a blocker while I'm out in any way.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,9 +11,6 @@
         "configs/*",
         "scripts"
       ],
-      "dependencies": {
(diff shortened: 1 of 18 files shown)


himanshusinghs (approved):
This looks great, thanks for this <3

gribnoysup (approved):

Anemy (approved):

PR #5174, opened 2023-11-29
feat(atlas-service): migrate from keytar COMPASS-7491

mabaasit:
In this PR, we made following two changes:
1. Use `compass-user-data` to read/write to the Atlas Config
2. Use `electron.safeStorage` to encrypt/decrypt the OIDC state and save it locally

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/atlas-service/src/secret-store.ts
```diff
@@ -1,28 +1,40 @@
...
 export class SecretStore {
-  async getItem(key: string): Promise<string | undefined> {
+  private readonly userData: UserData<typeof AtlasPluginStateSchema>;
+  private readonly fileName = 'AtlasPluginState';
+  constructor(basePath?: string) {
+    this.userData = new UserData(AtlasPluginStateSchema, {
+      subdir: '',
+      basePath,
+    });
+  }
```
mabaasit, line 14: Now that we are saving this data on fs, already logged in users will have to login again as we are not migrating the data (from keytar). I think its fine to keep it that way instead of writing a migration (and eventually coming back to remove the code).

If you feel strongly otherwise, let me know.

gribnoysup (approved):

packages/atlas-service/src/secret-store.ts
mabaasit: I'll check with product.

PR #5183, opened 2023-12-01
feat(components): add file input backend provider to allow different file backend environments  VSCODE-484

Anemy:
VSCODE-484

This pr adds a `FileInputBackendProvider` which provides a backend for the `FileInput` component to use for opening and saving files. The `connection-form` in VSCode is rendered in a `WebView`, [WebViews](https://code.visualstudio.com/api/extension-guides/webview) have limited permissions and require file access to be done by the main extension, so we will do some messaging for file system access. This new provider gives us the ability to support the different environments.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44701,7 +44701,6 @@
       "version": "0.17.1",
       "license": "SSPL",
       "dependencies": {
-        "@electron/remote": "^2.1.0",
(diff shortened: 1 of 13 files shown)


himanshusinghs (approved):
Super neat change! Looking good 🚀

PR #5195, opened 2023-12-05
fix(connection-form): use ejson shell parser with strict mode on csfle COMPASS-7496

Anemy:
COMPASS-7496


packages/connection-form/src/utils/csfle-handler.ts
```diff
@@ -213,7 +214,9 @@ export function textToEncryptedFieldConfig(text: string): Document | undefined {
 
   let parsed: Document = {};
   try {
-    parsed = parseFilter(text);
```
Anemy, line 216: Parse filter adds a layer of indirection, going to merge this so it bumps connection-form for vscode, feel free to revert if anything comes up.

PR #5197, opened 2023-12-06
chore(connection-form): use mongodb dep only with types VSCODE-488

Anemy:
VSCODE-488

This is so we can exclude it from the webview webpack in VSCODE without erroring. We could alternatively add a lot of polyfills (os-browserify path-browserify stream-http timers-browserify url util browserify-zlib) and keep it included. Excluding it seems a bit simpler and probably reduces bundle size.

### DIFF
--- a/packages/connection-form/src/components/advanced-options-tabs/advanced-tab/advanced-tab.tsx
+++ b/packages/connection-form/src/components/advanced-options-tabs/advanced-tab/advanced-tab.tsx
@@ -12,7 +12,6 @@ import {
 } from '@mongodb-js/compass-components';
 import type ConnectionStringUrl from 'mongodb-connection-string-url';
 import type { MongoClientOptions, ReadPreferenceMode } from 'mongodb';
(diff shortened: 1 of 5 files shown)


gribnoysup (approved):
Looks good, but I wonder how do we make sure we don't cause any regression on that. Maybe there is a linter rule we can add to enforce only type imports from certain packages?

himanshusinghs (approved):

packages/connection-form/src/components/advanced-options-tabs/advanced-tab/advanced-tab.tsx
```diff
@@ -29,23 +28,23 @@ interface ReadPreference {
...
   },
   {
     title: 'Secondary Preferred',
-    id: MongoReadPreference.SECONDARY_PREFERRED,
+    id: 'secondaryPreferred',
   },
   {
     title: 'Nearest',
-    id: MongoReadPreference.NEAREST,
+    id: 'nearest',
```
addaleax, line 47: :blue_heart:

PR #5209, opened 2023-12-07
feat: add ChangeView component to bulk update preview

lerouxb:
To review: Maybe start with `unifyDocuments()` in `unified-document.ts`. Or alternatively just the `ChangeView` component in `change-view.tsx`. **Don't be intimidated by the line count** - it is almost all test fixtures' expected results.

Screenshot 2023-12-12 at 11 41 38


packages/compass-crud/src/components/change-view/unified-document.ts
```diff
@@ -0,0 +1,529 @@
...
+export function getImplicitChangeType(obj: UnifiedBranch) {
+  if (['added', 'removed'].includes(obj.implicitChangeType)) {
+    // these are "sticky" as we descend
+    return obj.implicitChangeType;
+  }
+
+  return obj.changeType;
+}
+
+export function unifyDocuments(
```
lerouxb, line 486: Maybe start here.

packages/compass-crud/src/components/change-view/change-view.tsx
```diff
@@ -0,0 +1,618 @@
...
+
+const changeViewStylesDark = css({
+  color: palette.gray.light2,
+});
+
+const changeViewStylesLight = css({
+  color: palette.gray.dark2,
+});
+
+export function ChangeView({
```
lerouxb, line 589: GitHub collapsed this file, but this is a good entry-point for the component.

packages/compass-crud/src/components/change-view/change-view.tsx
```diff
@@ -0,0 +1,587 @@
...
+  const darkMode = useDarkMode();
+
+  const leftValue = includeLeft
+    ? lookupValue((obj.left as Branch).path, left)
+    : undefined;
+  const rightValue = includeRight
+    ? lookupValue((obj.right as Branch).path, right)
+    : undefined;
+
+  // TODO: BSONValue does not always show the bson type, so you can't spot bson type changes
```
lerouxb, line 522: TODO is not really the right word, but I did want to call attention to this. Since we're using the same components as the Document component the BSON values behave the same. So a Double with the value of `1` or an Int32 with value `1` or a Long with the value `1` all look the same.

This might not be much of a problem. As long we keep the before/after values with their types they will be stringified as `new Double(1)`, `new Int32(1)` and `new Long(1)` respectively before being diffed and therefore it will show up as changed. But the "red" and "green" values will then look identical.

I just bring this up because during tech design some people mentioned bson type changes.

lerouxb: (I removed the TODO)

packages/compass-crud/src/components/change-view/shape-utils.ts
```diff
@@ -0,0 +1,16 @@
+export function isSimpleObject(value: any) {
+  return (
+    Object.prototype.toString.call(value) === '[object Object]' &&
+    !value._bsontype
+  );
+}
+
+export function getValueShape(value: any) {
```
kmruiz, line 8:
```suggestion
export function getValueShape(value: any): 'array' | 'object' | 'leaf' {
```

For autocompletion and compiler checks.

kmruiz (approved):

PR #5237, opened 2023-12-14
chore(crud): Optimise bulk update previews

lerouxb:


packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1123,27 +1125,31 @@ class CrudStoreImpl
   }
 
   async updateBulkUpdatePreview(updateText: string) {
+    let lastTime = Date.now();
+    const previewNumber = this.bulkUpdatePreviewNumber++;
+    function took(note: string, data?: any) {
```
kmruiz, line 1130: Do we want to keep this after the profiling?

lerouxb: I'm in two minds about it. debug() is cheap, tagged as verbose and we have lots of it everywhere and it is easy enough to ignore. But I might as well also remove it..

kmruiz: I'm fine with any option, just wanted to make sure is not something we just missed from the code. Let me resolve the comment 😄

kmruiz (approved):

PR #5251, opened 2023-12-19
fix(bulk-update): remove timeout on error toast

Anemy:
Otherwise folks might not know things updated if they looked away. I'm thinking this timeout here may have been unintended, there isn't one for the success state. 

I also feel showing errors in a toast without giving more information on the error might be a rough ux pattern, but that's another discussion to have another time.

### DIFF
--- a/packages/compass-crud/src/components/bulk-actions-toasts.tsx
+++ b/packages/compass-crud/src/components/bulk-actions-toasts.tsx
@@ -95,7 +95,6 @@ export function openBulkDeleteFailureToast({
     title: '',
     variant: 'warning',
     dismissible: true,
-    timeout: 6_000,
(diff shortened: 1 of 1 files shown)


gribnoysup (approved):
> I also feel showing errors in a toast without giving more information on the error might be a rough ux pattern

I think usually we will use toast title for what is being passed as text here, and then the body would show the actual error message, maybe worth updating this too

And totally makes sense to remove the timeout, I almost feel like when `variant` is `warning` we never should allow the timeout to be applied by the component

Anemy:
Merged in the toast timeout updates. For the error messaging improvements I started a slack thread: https://mongodb.slack.com/archives/G2L10JAV7/p1703081795674249

PR #5252, opened 2023-12-20
feat(crud): disable the save button if the update query is invalid

lerouxb:
I suppose it is inevitable that someone's going to _want_ to save the broken query they are working on, but oh well.


gribnoysup (approved):
Code-wise looks good, but I gotta say the reason we allow to save the "broken" pipelines was exactly this, people wanting to save work in progress 😅

PR #5270, opened 2023-12-22
[wip] signtool with ssh

mabaasit:
This PR is intended to be closed.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-signtool/src/ssh-client.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  async connect() {
+    if (this.connected) {
+      return Promise.resolve();
+    }
+    return new Promise(async (resolve, reject) => {
+      this.sshConnection.connect({
+        ...this.sshClientOptions,
+        privateKey: readFileSync(this.sshClientOptions.privateKey),
```
mcasimir, line 46: this is a really silly and optional one: can we use async I/O everywhere? Nowadays is so easy that we can just do it by default and avoid having code that mixes both sync and async, in case we would end up replicating or copy/pasting fragments of code and accidentally introduce sync calls where we don't want to.

packages/compass-signtool/package.json
```diff
@@ -0,0 +1,84 @@
+{
+  "name": "@mongodb-js/compass-signtool",
+  "description": "Signtool to sign Compass windows artifacts",
```
mcasimir, line 3: from the sync we had, i understood we want do have this also for signing locally for linux machines, and not only for Compass, ultimately have it in devtools-shared, can we name this differently and adjust the description?

packages/compass-signtool/package.json
```diff
@@ -0,0 +1,84 @@
...
+    "test": "mocha",
+    "test-cov": "nyc --compact=false --produce-source-map=false -x \"**/*.spec.*\" --reporter=lcov --reporter=text --reporter=html npm run test",
+    "test-watch": "npm run test -- --watch",
+    "test-ci": "npm run test-cov",
+    "reformat": "npm run eslint . -- --fix && npm run prettier -- --write .",
+    "build": "npm run compile && pkg package.json",
+    "postinstall": "npm run build"
+  },
+  "bin": "dist/cli.js",
+  "pkg": {
```
mcasimir, line 55: are we sure we want to create an exe like this, we can assume our CI machines to have node.js installed, and we only need the signtool for wininstaller and compass, i think is 1000 times easier and safer (with less unknowns) to have this package not export any binary, and just modify the `signtool.go` that is in hadron-build, and just make it calls `npx run @mongodb-js/this-package sign`, rather than completely change the way we are doing it, and generalize it right away.

packages/compass-signtool/src/utils.ts
```diff
@@ -0,0 +1,14 @@
+export const debug = console.log;
+
+export function assertRequiredVars() {
+  [
+    'GARASIGN_USERNAME',
```
mcasimir, line 5: are we sure we want to check if these are set on the machine where the js runs? it seems that this is just an EVG quirk. To me it makes more sense to make this check on the bash script itself.

packages/compass-signtool/src/sign.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const getSshClient = async () => {
+  const hostname = await getSpawnnedHostName();
+  const sshClient = new SSHClient({
+    host: hostname,
+    port: 22,
+    username: 'ubuntu',
+    // todo: this will fail?
+    privateKey: path.resolve('/.ssh/mcipacker.pem'),
```
mcasimir, line 28: A thought: later we probably want to make this one, and some of the other stuff that are directly depending on how the ssh host is configured, or how our CI is setup to be arguments passed to this library.

The reasoning here is that if this package is going to be shared and live outside of the compass and mongosh repos, if we have to change the signing host, or paths change in evergreen, we don't need to also change the package at all.

packages/compass-signtool/src/utils.ts
mcasimir: scrap that, i thought there was a way to spawn an host and set enviroment variables directly .. which isn't true :)

packages/compass-signtool/src/sign.ts
```diff
@@ -0,0 +1,49 @@
...
+import { SSHClient } from './ssh-client';
+import { SigningClient } from './signing-client';
+import { assertRequiredVars } from './utils';
+
+const execAsync = promisify(exec);
+
+async function getSpawnnedHostName() {
+  // todo: this will fail
+  // return 'ec2-3-254-76-71.eu-west-1.compute.amazonaws.com';
+  const command = `tr -d '"[]{}' < ${path.resolve(
```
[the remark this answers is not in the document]
mabaasit: i copied this from evergreen docs. but I do not intend to use this. 
following on your other comment on `privateKey`, in order to avoid this, i would pass these both options to the script. following options (initially):
1. host
2. username
3. port
4. privateKey (path)

PR #5273, opened 2023-12-27
chore: add streams operation methods to DataService COMPASS-7553

shaketbaby:
## Description
Update DataService to support some streams operations required by VS Code extension.

### Checklist

## Motivation and Context

- [ ] New feature

## Open Questions

- Do the new methods need to annotated with @op and what would be the `logId` for them if so?

## Types of changes

### DIFF
--- a/packages/data-service/src/data-service.ts
+++ b/packages/data-service/src/data-service.ts
@@ -153,6 +153,28 @@ export type UpdatePreview = {
   changes: UpdatePreviewChange[];
 };
 
+export type StreamProcessor = {
+  id: string;
+  name: string;
+  state:
+    | 'CREATING'
+    | 'CREATED'
+    | 'VALIDATING'
+    | 'PROVISIONING'
+    | 'RECEIVED_ON_DISPATCHER'
+    | 'STARTING'
+    | 'STARTED'
+    | 'STOPPING'
+    | 'STOPPED'
+    | 'RELEASING'
+    | 'DROPPING'
+    | 'DROPPED'
+    | 'FAILED';
+  pipeline: Document[];
+  lastStateChange: Date;
+  lastModified: Date;
+};
+
 export interface DataService {
   // TypeScript uses something like this itself for its EventTarget definitions.
   on<K extends keyof DataServiceEventMap>(
@@ -786,6 +808,34 @@ export interface DataService {
     update: UpdateFilter<Document>,
     options?: UpdateOptions
   ): Promise<UpdateResult>;
+
+  /*** Streams ***/
+
+  /**
+   * List all the named stream processors.
+   */
+  listStreamProcessors(filter?: Document): Promise<StreamProcessor[]>;
+
+  /**
+   * Start the specified stream processor
+   *
+   * @param name processor name
+   */
+  startStreamProcessor(name: string): Promise<void>;
+
+  /**
+   * Stop the specified stream processor
+   *
+   * @param name processor name
+   */
+  stopStreamProcessor(name: string): Promise<void>;
+
+  /**
+   * Drop the specified stream processor
+   *
+   * @param name processor name
+   */
+  dropStreamProcessor(name: string): Promise<void>;
 }
 
 const maybePickNs = ([ns]: unknown[]) => {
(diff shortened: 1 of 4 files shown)


packages/data-service/src/data-service.ts
```diff
@@ -153,6 +153,28 @@ export type UpdatePreview = {
...
+    | 'STARTED'
+    | 'STOPPING'
+    | 'STOPPED'
+    | 'RELEASING'
+    | 'DROPPING'
+    | 'DROPPED'
+    | 'FAILED';
+  pipeline: Document[];
+  lastStateChange: Date;
+  lastModified: Date;
```
himanshusinghs, line 175: Are we sure that we will be receiving date object from the command run? If not, maybe we should type these as string or number (in case of timestamp)?

packages/data-service/src/data-service.ts
```diff
@@ -2645,6 +2695,31 @@ class DataServiceImpl extends WithLogContext implements DataService {
...
+    await this._runStreamCmd({ stopStreamProcessor: name });
+  }
+
+  async dropStreamProcessor(name: string): Promise<void> {
+    await this._runStreamCmd({ dropStreamProcessor: name });
+  }
+
+  private async _runStreamCmd(cmd: Document, options: Document = {}) {
+    const result = await this._database('admin', 'CRUD').command(cmd, options);
+    return result.ok === 1 ? result : Promise.reject(result);
```
gribnoysup, line 2720: I don't think we ever just "throw" raw results from the driver if driver itself doesn't throw. Usually driver will throw if result is not ok. Is is not the case here? If not, can we throw a meaningful error that is constructed as error, and not just driver response?

packages/data-service/src/data-service.ts
```diff
@@ -2645,6 +2695,31 @@ class DataServiceImpl extends WithLogContext implements DataService {
...
+  async stopStreamProcessor(name: string): Promise<void> {
+    await this._runStreamCmd({ stopStreamProcessor: name });
+  }
+
+  async dropStreamProcessor(name: string): Promise<void> {
+    await this._runStreamCmd({ dropStreamProcessor: name });
+  }
+
+  private async _runStreamCmd(cmd: Document, options: Document = {}) {
+    const result = await this._database('admin', 'CRUD').command(cmd, options);
```
gribnoysup, line 2719: We purposefully only use `runCommand` method when running commands directly against driver for the reasons mentioned in the method description. We probably should use the same approach here, or is it a deliberate decision not to?

packages/data-service/src/data-service.ts
```diff
@@ -2645,6 +2695,31 @@ class DataServiceImpl extends WithLogContext implements DataService {
...
 
+  async listStreamProcessors(filter?: Document): Promise<StreamProcessor[]> {
+    const { streamProcessors } = await this._runStreamCmd({
+      listStreamProcessors: 1,
+      filter,
+    });
+    return streamProcessors;
+  }
+
+  async startStreamProcessor(name: string): Promise<void> {
```
gribnoysup, line 2706: > Do the new methods need to annotated with @op and what would be the logId for them if so?

Yes, all methods that run driver commands should have an `op` decorator applied so that the logging and errors are normalized. The id is the next available one as returned by `npm run check-logids` or using eslint fix suggestions if you have integration implemented.

packages/data-service/src/data-service.ts
shaketbaby: Good call, they are shown as below in playground result but doesn't seem to be an instance of `Date`. Will change to `string`.
```
    "lastModified": {
      "$date": "2023-12-04T04:06:49.506Z"
    },
```

packages/data-service/src/data-service.ts
shaketbaby: Didn't know driver throws when `ok !== 1`. Will remove this check.

packages/data-service/src/data-service.ts
shaketbaby: Have changed to use `runCommand`.

packages/data-service/src/data-service.ts
shaketbaby: Have annotated with `@op`.

packages/data-service/src/data-service.ts
gribnoysup: Hmmm, I think the format you're seeing here is not a `string` but a EJSON representation of a bson Date value. I think you should be able to pass `promoteValues` to the command options and then the driver should actually promote this value to a JavaScript Date:

image

This is something we also do for some other operations, so maybe would make sense here too so that the consumers of this method don't need to deal with EJSON. Don't have a preference though if we want to do this. If not, this should probably be a Document type, not a string though.

packages/data-service/src/run-command.ts
```diff
@@ -223,6 +223,10 @@ export type ListCollectionsResult<CollectionType> = {
   cursor: { firstBatch: CollectionType };
 };
 
+export type listStreamProcessorsOptions = {
```
gribnoysup, line 226: Very tiny nit, we usually use upper camel case for types:

```suggestion
export type ListStreamProcessorsOptions = {
```

gribnoysup (approved):
One tiny nit and a note about how to type the return value of the list stream processors command, otherwise looks pretty good!

packages/data-service/src/run-command.ts
shaketbaby: Ah, good spot. I missed that.

packages/data-service/src/data-service.ts
shaketbaby: It looks like `promoteValues` is `true` by default and I can confirm that the value returned from driver is actually an instance of `Date`. So I've changed the type back to `Date`.

shaketbaby:
@gribnoysup thanks for your feedback. Is it possible to release a new version of `mongodb-data-service` after this is merged? These changes are required to add Streams support to VS Code Extension.

packages/data-service/src/data-service.ts
gribnoysup: Awesome, thanks for double-checking

gribnoysup:
@shaketbaby sure thing! As soon as you merge, I'll merge https://github.com/mongodb-js/compass/pull/5268 and this will kick-off the packages release process

PR #5279, opened 2023-12-28
ci(signtool): sign artifacts with garasign COMPASS-7549

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-signtool/src/signing-clients/local-signing-client.ts
```diff
@@ -0,0 +1,48 @@
+import path from 'path';
+import { exec } from 'child_process';
```
mcasimir, line 2: `execFile` to make CodeQL happy?

.evergreen/compass_package.sh
```diff
@@ -2,6 +2,14 @@
 
 set -e
 
+echo "WINDOWS_SIGNING_SERVER_HOSTNAME: ${WINDOWS_SIGNING_SERVER_HOSTNAME}"
+echo "WINDOWS_SIGNING_SERVER_PRIVATE_KEY: ${WINDOWS_SIGNING_SERVER_PRIVATE_KEY}"
```
addaleax, line 6: This is a private key identifier and not the actual contents of the private key, right?

packages/compass-signtool/src/garasign.sh
```diff
@@ -0,0 +1,48 @@
...
+echo "File to be signed: $file"
+echo "Working directory: $directory"
+
+docker run \
+  --env-file=signing-envfile \
+  --rm \
+  -v $directory:$directory \
+  -w $directory \
+  ${ARTIFACTORY_HOST}/release-tools-container-registry-local/garasign-gpg \
+  /bin/bash -c "gpgloader && gpg --yes -v --armor -o $file.sig --detach-sign $file"
```
addaleax, line 45:
```suggestion
  /bin/bash -c "gpgloader && gpg --yes -v --armor -o '$file.sig' --detach-sign '$file'"
```

iiuc this script is supposed to be able to take a generic filename, would be nice to avoid the double bash expansion that could happen here if the filename is not just a single word

packages/compass-signtool/src/signing-clients/local-signing-client.ts
```diff
@@ -0,0 +1,56 @@
...
+import type { SigningClient, SigningClientOptions } from '.';
+
+const execAsync = promisify(exec);
+
+export class LocalSigningClient implements SigningClient {
+  constructor(private options: SigningClientOptions) {}
+
+  private async init() {
+    const remoteScript = `${this.options.rootDir}/garasign.sh`;
+    await execAsync(`mkdir -p ${this.options.rootDir}`);
```
addaleax, line 14: Most of these `execAsync` calls could be `fs` calls, e.g. `fs.mkdir()` with `{ recursive: true }`, `fs.chmod()`, `fs.copyFile()`, etc. (and they probably should be unless we really intentionally want the bash extensions to apply here, which I'm pretty sure we don't)

packages/compass-signtool/src/signing-clients/local-signing-client.ts
```diff
@@ -0,0 +1,56 @@
...
+    const remotePath = path.join(this.options.rootDir, path.basename(file));
+
+    try {
+      await this.init();
+
+      await this.copyFile(file, remotePath);
+      debug(`LocalSigningClient: Copied file ${file} to ${remotePath}`);
+
+      await execAsync(
+        `cd ${this.options.rootDir} && ./garasign.sh ${path.basename(file)}`,
```
addaleax, line 35: `child_process.execFile` takes a `cwd` option so you don't need to manually `cd` into a specific directory first

packages/compass-signtool/src/signing-clients/remote-signing-client.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Initialize the signing client and setup remote machine to be ready for signing
+   * the files. This will do following things:
+   * - Create a working directory on the remote machine
+   * - Copy the signing script to the remote machine
+   */
+  private async init() {
+    this.sftpConnection = await this.sshClient.getSftpConnection();
+    await this.sshClient.exec(`mkdir -p ${this.options.rootDir}`);
```
addaleax, line 23: Here I guess we can't do fs calls instead, but let's at least try to do some minimal escaping?
```suggestion
    await this.sshClient.exec(`mkdir -p '${this.options.rootDir}'`);
```

packages/compass-signtool/src/ssh-client.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (!this.connected) {
+      throw new Error('Not connected to ssh server');
+    }
+    return new Promise((resolve, reject) => {
+      this.sshConnection.exec(command, (err, stream) => {
+        if (err) {
+          return reject(err);
+        }
+        let data = '';
+        stream.on('data', (chunk: string) => {
```
addaleax, line 70:
```suggestion
        stream.setEncoding('utf8').on('data', (chunk: string) => {
```

packages/compass-signtool/src/ssh-client.ts
```diff
@@ -0,0 +1,110 @@
...
+          data += chunk;
+        });
+        stream.stderr.on('data', (chunk) => {
```
addaleax, line 73:
```suggestion
        stream.stderr.setEncoding('utf8').on('data', (chunk) => {
```

.evergreen/compass_package.sh
mabaasit: Correct, its just the absolute path to the file.

packages/compass-signtool/src/garasign.sh
mabaasit: thanks for these comments, but we are housing this package in devtools-shared in https://github.com/mongodb-js/devtools-shared/pull/190.

I'll ask Bailey to check these comments.

mabaasit:
Closing this in favour of individual PR for gpg and jsign.

PR #5286, opened 2023-12-29
chore: build signtool.exe in postinstall script

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/hadron-build/signtool/install.js
```diff
@@ -18,5 +19,16 @@
   throw new Error('Original signtool.exe to be replaced not found');
 }
 
+const buildScriptDir = __dirname;
+const buildScriptName = 'build.sh';
+const buildScriptCmd = `cd "${buildScriptDir}" && ${path.resolve(__dirname, buildScriptName)}`;
+console.info(`Running preinstall script in hadron-build. Command: ${buildScriptCmd}`);
+console.info('Building signtool.exe');
+console.info(execSync(buildScriptCmd).toString());
```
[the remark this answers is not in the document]
mcasimir: `execFileSync`?

packages/hadron-build/signtool/install.js
```diff
@@ -18,5 +19,16 @@ if (!fs.existsSync(originalSigntool)) {
   throw new Error('Original signtool.exe to be replaced not found');
 }
 
+const buildScriptDir = __dirname;
```
baileympearson, line 22: Adding a postinstall script to build signtool is 1) unnecessary for local development and 2) adds a dependency on the correct Go version to local dev environments.  It's also not necessary for the majority of CI.

Is it possible to just build signtool in CI when necessary instead of always building it?

baileympearson (changes requested):

packages/hadron-build/signtool/install.js
mabaasit: yeah that's a good idea. only build it when we need to sign something.

packages/hadron-build/lib/signtool.js
```diff
@@ -1,12 +1,50 @@
...
+  // Replaces signtool.exe in electron-winstaller with the emulated version
+  // that uses the notary service.
+  const originalSigntool = require.resolve('electron-winstaller/vendor/signtool.exe');
+  const originalSigntoolBackup = originalSigntool + '.bkp';
+
+  try {
+    await fs.access(originalSigntoolBackup);
+  } catch (err) {
+    debug('Original signtool.exe to be replaced not found');
+    return;
```
mcasimir, line 31: Isn't this always reaching this line? On a fresh install `signtool.bkp` will not be there.

Made me think, what happens if we forget to replace `signtool.exe`? Whatever route with take here, let's make sure that we cannot end up with unsigned binaries. Perhaps now that you found out about git-attributes we could go back to push the pre-build version?

PR #5308, opened 2024-01-08
chore(telemetry): update analytics-node to latest version

mcasimir:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass/src/main/telemetry.ts
```diff
@@ -169,7 +172,6 @@ class CompassTelemetry {
           event: 'Telemetry Disabled',
           properties: {},
         });
-        void this._flushTelemetryAndIgnoreFailure();
```
mcasimir, line 172: this can't be done anymore without closing the client

packages/compass/src/main/telemetry.ts
```diff
@@ -121,7 +121,7 @@ class CompassTelemetry {
   }
 
   private static _flushTelemetryAndIgnoreFailure() {
-    return this.analytics?.flush().catch(() => Promise.resolve());
+    return this.analytics?.closeAndFlush().catch(() => Promise.resolve());
```
himanshusinghs, line 124: Out of curiosity - it won't be any different from just having this -
```
return this.analytics?.closeAndFlush().catch(() => {});
```
right?

packages/compass/src/main/telemetry.ts
himanshusinghs: Just making sure - does this mean that there is a chance that we might lose some telemetry that isn't synced yet? Probably not, right?

himanshusinghs (approved):

packages/compass/src/main/telemetry.ts
mcasimir: none that i can tell yeah

packages/compass/src/main/telemetry.ts
mcasimir: That should be covered by the other flush, i think the reason this was here before  is to "send telemetry one last time" and then stop sending anything after it has been disabled. I don't remember any strict requirement about this, I think that it should be fine to let the flush happen with a bit of delay here.

PR #5326, opened 2024-01-10
feat(connection-form): add show favorite actions preference, make save prop required VSCODE-406

Anemy:
Part of VSCODE-406

This adds a `showFavoriteActions` preferences for the connection form. When set to false the actions around favoriting a connection are hidden. These actions are Compass specific and something we don't want shown in VSCode.


Anemy:
Merging to unblock VSCode connection form work, feel free to revert if something is up.

PR #5332, opened 2024-01-12
chore: re-enable the CLI connections tests on windows

lerouxb:
Seeing how this goes.


lerouxb:
Seems to be fine except for one instance of "compass-e2e-tests__Database_collections_tab.can_create_a_collection_with_custom_collation" so I think we can merge.

PR #5335, opened 2024-01-12
chore(e2e): rename and refactor beforeTests, afterTest and afterTests helpers; use them properly

lerouxb:
The idea is to have all our log files named after the test suite, test or "step inside a test". ie. every time we open compass we give it a relevant name. Then we can easily find the relevant log files for the test.

Hopefully renaming the helpers init(), cleanup() and screenshotIfFailed() is much clearer and reflects what it actually does nowadays. This should fit better into the patterns of how we start and stop compass in our tests nowadays.

I also added a lot of try {} finally {} for when we init compass inside tests because whenever we init() we have to cleanup(). I'm still considering keeping track of how many we created and cleaned up and erroring out in case someone forgot to close the previous one. Maybe also make sure that we don't reuse names because then we'll have logs overwriting each other..

Just running this to test that it really does actually use the right names and doesn't just end up with timestamps anyway.

### DIFF
--- a/packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
+++ b/packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
@@ -3,9 +3,9 @@ import type { Element } from 'webdriverio';
 import { promises as fs } from 'fs';
 import type { CompassBrowser } from '../helpers/compass-browser';
 import {
-  beforeTests,
-  afterTests,
-  afterTest,
+  init,
+  cleanup,
+  screenshotIfFailed,
   outputFilename,
   serverSatisfies,
 } from '../helpers/compass';
@@ -150,7 +150,7 @@ describe('Collection aggregations tab', function () {
   let browser: CompassBrowser;
 
   before(async function () {
-    compass = await beforeTests();
+    compass = await init(this.test?.fullTitle());
     browser = compass.browser;
   });
 
@@ -176,11 +176,11 @@ describe('Collection aggregations tab', function () {
   });
 
   after(async function () {
-    await afterTests(compass, this.currentTest);
+    await cleanup(compass);
   });
 
   afterEach(async function () {
-    await afterTest(compass, this.currentTest);
+    await screenshotIfFailed(compass, this.currentTest);
   });
 
   it('supports the right stages for the environment', async function () {
--- a/packages/compass-e2e-tests/tests/connection.test.ts
+++ b/packages/compass-e2e-tests/tests/connection.test.ts
@@ -7,9 +7,9 @@ import ConnectionString from 'mongodb-connection-string-url';
 import resolveMongodbSrv from 'resolve-mongodb-srv';
 import type { CompassBrowser } from '../helpers/compass-browser';
 import {
-  beforeTests,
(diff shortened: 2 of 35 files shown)


packages/compass-e2e-tests/tests/collection-aggregations-tab.test.ts
```diff
@@ -150,7 +150,7 @@ describe('Collection aggregations tab', function () {
   let browser: CompassBrowser;
 
   before(async function () {
-    compass = await beforeTests();
+    compass = await init(this.test?.fullTitle());
```
Anemy, line 153: Since this runs in the `before`, does that mean that the `this.test?.fullTitle()` will only be the first tests name?

packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -705,8 +705,8 @@ describe('System CA access', function () {
       assertNotError(result);
       expect(result).to.have.property('ok', 1);
     } finally {
-      // make sure the browser gets closed otherwise if this fails the process wont exit
-      await afterTests(compass, this.currentTest, 'no-connect');
+      // make sure the browser gets closed otherwise if this fails the process won't exit
+      await cleanup(compass);
```
Anemy, line 709: We use two different patterns for how we manage these `cleanup`s in these changes. We use init in the test and `try.... finally -> cleanup` and then we also use the `cleanup` in the afterEach. Would it make sense to stick to one pattern there? I'm thinking folks when adding more tests might miss the need for having the `finally` and will bring over code, should we use the `cleanup` in the afterEach as a standard? Is there a nice way we can prevent the process not exiting without the future developer knowing this pattern?
Not a blocker

lerouxb: It is supposed to just be before() with init and after() with cleanup but then people started doing init in tests, sometimes multiple times in the same test.

What is the pattern you would prefer?

lerouxb: Not much you can do about the need for try / finally if you const compass = await init() inside a test, I think. Especially for the tests that do it multiple times.

I'm not really a fan of having let compass; in the describe() and then compass = await init() inside an it() and then cleanup(compass) in afterEach because the control/ownership of the compass variable becomes split all over the place. I think if we have to create compass inside a test I'd rather keep the var there because the afterEach can be who knows where in relation to the test and then you have to keep track of whether compass is undefined, an initialised compass var or one that was already cleanedup..

Anemy: I with you on the preferences order. Moving the init out of the test seems like a good pattern where we can.

Anemy (approved):

packages/compass-e2e-tests/tests/connection.test.ts
lerouxb: We usually do. I think the cases where we have it inside the test are usually because we have to pass different options to init() for each test and the tests are each testing a different option. For the final case I think the only tests that do that are the connection import/export ones and it kinda makes sense once you read through them. By definition those tests involve multiple compass sessions.

PR #5336, opened 2024-01-12
chore(compass-web): remove types export to unblock publishing packages

Anemy:
The `check-exports-exist` was failing as we're not currently building a index.d.ts in this package. We probably want to, but that'll take a bit more investigation. This pr will unblock publishing package releases for now, so we can update them in VSCode.


Anemy:
Created https://github.com/mongodb-js/compass/pull/5337 to fix type declaration.

PR #5342, opened 2024-01-15
chore(e2e): scroll to the custom collation field option before clicking and take a screenshot, fix css selector COMPASS-7569

lerouxb:
Debugging and maybe fixing the e2e test flake where we click on a menu option to select it yet it never seems to have an effect.

As a drive-by I'm also fixing our screenshotting:

* browser.saveScreenshot() is webdriver's command and in order to use it properly we have to give it a full "absolute" path to the screenshot so that the file ends up in a directory (probably .logs) that we'll upload to S3.
* browser.screenshot() is our command that calls browser.saveScreenshot(). It takes a filename only and path.join()s that together with the path where screenshots are supposed to go so you don't have to import that and join it together everywhere.
* browser.clickVisible(selector, options) takes screenshot as an option and it is supposed to just be a filename because it gets passed to browser.screenshot. This is for convenience so you can screenshot the state of the world at the moment we click.

So.. we should probably not use browser.saveScreenshot() unless we want to put the file somewhere non-standard. After that passing a screenshot option to browser.clickVisible() is often the most convenient and useful. For everything else please use browser.screenshot()

I also removed the capturePage() helper in compass.ts because that was just redundant at this point.

### DIFF
--- a/packages/compass-e2e-tests/helpers/commands/add-collection.ts
+++ b/packages/compass-e2e-tests/helpers/commands/add-collection.ts
@@ -1,8 +1,5 @@
-import path from 'path';
 import type { CompassBrowser } from '../compass-browser';
 import * as Selectors from '../selectors';
-
-import { LOG_PATH } from '../compass';
 
 export type AddCollectionOptions = {
   capped?: {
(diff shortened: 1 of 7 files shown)


mcasimir (approved):

lerouxb:
I see timeouts on mac, but those seem to be unrelated as they are in different tests. Merging.

PR #5346, opened 2024-01-15
chore: we apparently never emit 'refresh-data' on localAppRegistry

lerouxb:


gribnoysup (approved):
Was double-checking and noticed that we have another one that doesn't do anything in crud too: https://github.com/mongodb-js/compass/blob/da32854d5ad02e3f85fca59c6da8fe3d38df0beb/packages/compass-crud/src/stores/crud-store.ts#L2084

Otherwise looks good!

PR #5349, opened 2024-01-16
chore(build): sign all artifacts COMPASS-7549

mabaasit:
In this PR, we are signing all the remaining compass artifacts - for windows, macos (zip only) and rhel. Steps to verify a files are similar to the one in https://github.com/mongodb-js/compass/pull/5334. 
For windows `.exe` and `.msi`, the files should have a digital signature, chrome (when downloading) and os (when installing) should not complain.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/.evergreen/functions.yml
+++ b/.evergreen/functions.yml
@@ -80,6 +80,10 @@ variables:
     MONGODB_RUNNER_LOG_DIR: ${workdir}/src/.testserver/
     E2E_TESTS_ATLAS_CS_WITHOUT_SEARCH: ${e2e_tests_atlas_cs_without_search}
     E2E_TESTS_ATLAS_CS_WITH_SEARCH: ${e2e_tests_atlas_cs_with_search}
+    GARASIGN_USERNAME: ${garasign_username}
+    GARASIGN_PASSWORD: ${garasign_password}
+    ARTIFACTORY_USERNAME: ${artifactory_username}
+    ARTIFACTORY_PASSWORD: ${artifactory_password}
 
 # This is here with the variables because anchors aren't supported across includes
 post:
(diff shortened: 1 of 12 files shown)


.evergreen/functions.yml
```diff
@@ -310,6 +310,85 @@ functions:
...
+  spawn-signing-server:
+    # spawn
+    - command: host.create
+      type: setup
+      params:
+        provider: ec2
+        distro: ubuntu2004-large
+        security_group_ids:
+          - sg-097bff6dd0d1d31d0 # Magic string that's needed for SSH'ing.
+    # write host info (this file will be read by sigingtool when connection to ssh server)
```
himanshusinghs, line 322:
```suggestion
    # write host info (this file will be read by signingtool when connection to ssh server)
```

.evergreen/functions.yml
```diff
@@ -310,6 +310,85 @@ functions:
...
+    # copy ssh key (this key will be used to connect to ssh server)
+    - command: shell.exec
+      params:
+        shell: bash
+        script: |
+          set -e
+          {
+          set +x
+          echo '${__project_aws_ssh_key_value}' > ~/.ssh/mcipacker.pem
+          chmod 0600 ~/.ssh/mcipacker.pem
```
himanshusinghs, line 338: nothing major - but maybe 0400 is enough for this since we are only reading the file, at-least in this script.

himanshusinghs (approved):
Amazing work 🚀

packages/hadron-build/lib/signtool.js
```diff
@@ -1,22 +1,103 @@
...
+  }
+
+  const remoteServerOptions = {
+    host: process.env.SIGNING_SERVER_HOSTNAME,
+    username: process.env.SIGNING_SERVER_USERNAME,
+    port: process.env.SIGNING_SERVER_PORT,
+    privateKey: process.env.SIGNING_SERVER_PRIVATE_KEY,
+  };
+
+  if (variant === 'ubuntu') {
```
mcasimir, line 75: NIT: feels that the variant here just dictates the 'local' / 'remote', and the extension is determining the method, this is not immediately understandable from the code, since we have if conditions combining both the variant and the filename.

I was thinking this code would represent reality a bit better, and be simpler with something like this:

``` js
const clientOptions = variant === 'ubuntu' ? { 
    client: 'local'
} : {
    client: 'remote',  host: process.env.SIGNING_SERVER_HOSTNAME,
    username: process.env.SIGNING_SERVER_USERNAME,
    port: process.env.SIGNING_SERVER_PORT,
    privateKey: process.env.SIGNING_SERVER_PRIVATE_KEY
}

const signingMethod = path.extname(src) === '.exe' || path.extname(src) === '.msi' ? 'jsign' : 'gpg';

return await garasign(src, {
  ...clientOptions,
  signingMethod,
});
```

packages/hadron-build/lib/signtool.js
```diff
@@ -1,22 +1,103 @@
...
+ * |----------|--------|--------|----------------|--------------|
+ *
+ * ‡ On rhel we have docker, but currently we are using rhel7.6
+ *  for packaging and docker does not support login using --password-stdin
+ *
+ * @param {string} src
+ * @returns {Promise<void>}
+ */
+async function sign(src, garasign = _garasign) {
+  const variant = process.env.EVERGREEN_BUILD_VARIANT;
```
mcasimir, line 60: We have similar branching based on build variants in mongosh, and personally i find hard to digest and maintain, it creates obscure dependencies between the CI and the code that may not be immediately clear, and relies on assumptions about the variants that may not stand true over time.

Could we do something different to detect where to use ssh and where to use docker? For example, we could directly set this as a variable / expansion in CI: https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files#build-variants

```
name: ubuntu
expansions:
    garasign_client_type: "remote"
```

mcasimir (approved):
Looks good! I left 2 comments that may be worth looking at if you have some time

packages/hadron-build/lib/signtool.js
mabaasit: added in [d1ffa71](https://github.com/mongodb-js/compass/pull/5349/commits/d1ffa71970219314752073db1bea737a6a434951)

mcasimir (approved):

PR #5354, opened 2024-01-17
chore(build): notarize dmg COMPASS-7489

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/hadron-build/lib/notary-service.js
```diff
@@ -0,0 +1,87 @@
+const download = require('download');
```
baileympearson, line 1: to avoid confusion (since there's also a notary service) - could we rename the file `macos-notary-service`?

packages/hadron-build/lib/notary-service.js
```diff
@@ -0,0 +1,87 @@
...
+    '--verify',
+    ...(notarizeOptions.macosEntitlements ? ['-e', notarizeOptions.macosEntitlements] : [])
+  ], {
+    cwd: path.dirname(src),
+    encoding: 'utf8'
+  });
+  debug('macnotary result:', macnotaryResult.stdout, macnotaryResult.stderr);
+  debug('ls', (await execFile('ls', ['-lh'], { cwd: path.dirname(src), encoding: 'utf8' })).stdout);
+
+  // Step:3 - clean up. remove existing src, unzip signedArchive to src, remove signedArchive and unsignedArchive
```
baileympearson, line 70: q: do you think we need to be concerned about clean up resources (step 3 here) if step 1/step 2 fails?  as it's written, if anything throws (including any step in the cleanup), we won't clean up any temporary files.

baileympearson (changes requested):

packages/hadron-build/lib/notary-service.js
mabaasit: if for some reason any step here fails, the whole build process will fail.

packages/hadron-build/lib/notary-service.js
mabaasit: the other notary-service will be removed as we are using garasign now. i've another upcoming PR to remove it completely and then we are left with this only.

packages/hadron-build/lib/notary-service.js
mcasimir: I know we imply that we only run this on CI and is not a big deal, but since is not a big effort, it may be good to cleanup afterwards.

packages/hadron-build/lib/target.js
```diff
@@ -1,25 +1,22 @@
 // eslint-disable-next-line strict
 'use strict';
 const chalk = require('chalk');
-const childProcess = require('child_process');
```
mcasimir, line 4: Nice seeing some line of code leaving this file! :)

mcasimir (approved):

packages/hadron-build/lib/notary-service.js
mabaasit: Renamed!

packages/hadron-build/lib/mac-notary-service.js
```diff
@@ -0,0 +1,95 @@
+const download = require('download');
```
addaleax, line 1:
```suggestion
'use strict';
const download = require('download');
```

(this is not an ES/TS module so not strict by default)

packages/hadron-build/lib/mac-notary-service.js
```diff
@@ -0,0 +1,95 @@
...
+    // Step:3 - remove existing src, unzip signedArchive to src
+    debug('removing existing directory contents');
+    await execFile('rm', ['-r', fileName], {
+      cwd: path.dirname(src)
+    });
+    debug(`unzipping with "unzip -u ${signedArchive}"`);
+    await execFile('unzip', ['-u', signedArchive], {
+      cwd: path.dirname(src),
+      encoding: 'utf8'
+    });
```
addaleax, line 82: If you want to, I think it wouldn't be a bad idea to store `{ cwd: path.dirname(src), encoding: 'utf8' }` in a variable once and then pass it to all the `execFile()` calls, completely optional suggestion though

addaleax (approved):

baileympearson (approved):

PR #5359, opened 2024-01-18
chore: ignore inflight vuln COMPASS-7571

Anemy:
COMPASS-7571


Anemy:
Closing as we don't want to ignore this as there is no remediation.

PR #5365, opened 2024-01-19
chore(build): verify packaged artifacts

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


.evergreen/verify-artifacts.sh
```diff
@@ -0,0 +1,65 @@
...
+  verify_using_powershell $WINDOWS_EXE_NAME
+  verify_using_powershell $WINDOWS_MSI_NAME
+
+  echo "Skipping verification for Windows artifacts using gpg: $WINDOWS_ZIP_NAME, $WINDOWS_NUPKG_NAME"
+  exit 0
+fi
+
+
+echo "Importing Compass public key"
+curl https://pgp.mongodb.com/compass.asc | gpg --import > "$TMP_FILE" 2>&1
```
addaleax, line 50: This modifies global state on the machine on which it runs – that's probably fine most of the time, but it's not ideal, and it's probably fairly easy to work around it by passing `--homedir` to all GPG invocations with some sort of temporary directory (a new one created for the script or even just $PWD)?

.evergreen/verify-artifacts.sh
```diff
@@ -0,0 +1,65 @@
...
+
+if [ "$IS_RHEL" = true ]; then
+  verify_using_gpg $RHEL_RPM_NAME
+  verify_using_gpg $RHEL_TAR_NAME
+fi
+
+if [ "$IS_OSX" = true ]; then
+  verify_using_gpg $OSX_ZIP_NAME
+  verify_using_codesign $OSX_DMG_NAME
+fi
```
addaleax, line 65: Random suggestion: Turn this into an if/elif/elif/elif/else, with the else case/fallthrough case erroring out? That way there's no chance of this script "accidentally" passing because the `IS_...` environment variables weren't set (e.g. if their names changed at some point and we missed this file)

.evergreen/verify-artifacts.sh
mabaasit: added in [ef6c717](https://github.com/mongodb-js/compass/pull/5365/commits/ef6c717d2ff509b48e1bb874520c42cc21130d52)

addaleax (approved):

PR #5381, opened 2024-01-26
chore(compass-preferences-model): remove compass specific imports from the provider entrypoint in compass-preferences-model

himanshusinghs:
## Description
The provider export in compass-preferences-model when imported in a plugin (in this case - databases-collections plugin) was also bundling together compass-user-data and hadron-ipc. This PR separates the compass-specific exports into their own files that not included in any way when importing from `compass-preferences-model/provider`

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-preferences-model/src/in-memory-storage.ts
```diff
@@ -0,0 +1,35 @@
+import { getDefaultsForStoredPreferences } from './preferences-schema';
+import type { AllPreferences, StoredPreferences } from './preferences-schema';
+
+export interface BasePreferencesStorage {
+  setup(): Promise<void>;
+  getPreferences(): StoredPreferences;
+  updatePreferences(attributes: Partial<StoredPreferences>): Promise<void>;
+}
+
+export class InMemoryStorage implements BasePreferencesStorage {
```
mcasimir, line 10: this is a mega NIT, probably unrelated and on code that was already there, however i find `BasePreferencesStorage` not really idiomatic here.

`BasePreferencesStorage` makes me think of a base abstract class that is incomplete and that we are extending. Here instead we are providing a completely new implementation of an interface, if that is the case i would rather expect the interface name to be something like `PreferencesStorage` instead of `BasePreferencesStorage`.

packages/compass-preferences-model/src/storage.ts
```diff
@@ -7,43 +7,11 @@ import {
...
-      ...this.preferences,
-      ...attributes,
-    };
-  }
-
-  async setup() {
-    // noop
-  }
-}
+import type { BasePreferencesStorage } from './in-memory-storage';
```
mcasimir, line 14: this whole naming and relationship between types looks confusing at first to me. 
looks like you are going for something like:`BasePreferencesStorage` is the main interface that every storages must implement. Then you have 2 implementation, one "environment agnostic" `InMemoryStorage` and one compass-only `PersistentStorage`.

It seems you are avoiding to change the existing persistent code, but i think that if we don't we would model this in an unclear way, making it look like there is a relationship between the in memory storage and the persistent one when there is none.

What about we extract `BasePreferencesStorage` to be outside of `./in-memory-storage` in its own file (and possibly just call it `PreferencesStorage`) and we require this interface in both `./in-memory-storage` and `/src/storage.ts`.

Also since now the class is called PersistentStorage, i think it would be clearer if we make the file name reflect that and rename `/src/storage.ts` to persistent-storage.ts`.

packages/databases-collections/.depcheckrc
```diff
@@ -8,3 +8,5 @@ ignores:
  - "@types/chai-dom"
  - "@types/react"
  - "@types/react-dom"
+ignore-patterns:
+  - 'dist'
```
mcasimir, line 12: why does this matter? if it does should we create a ticket to fix it everywhere?

himanshusinghs: I will be removing this change from this PR and instead follow up with what we discussed regarding webpack plugin.

packages/compass-preferences-model/src/storage.ts
himanshusinghs: Addressed this in my recent commit

addaleax (approved):

PR #5384, opened 2024-01-26
chore(my-queries-storage): implement provider for my-queries-storage COMPASS-7411

himanshusinghs:
## Description
This PR implement providers for PipelineStorage, FavoriteQueryStorage and RecentQueryStorage.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/test/configure-store.ts
```diff
@@ -35,6 +36,9 @@ export default function configureStore(
       localAppRegistry: new AppRegistry(),
       workspaces: {} as any,
       logger: createNoopLoggerAndTelemetry(),
+      pipelineStorage: {
+        loadAll: () => Promise.resolve([]),
+      } as unknown as PipelineStorage,
```
himanshusinghs, line 41: Passing a bare minimum here only because it was the easiest thing to do. Let me know we would rather prefer having a full mock implementation of PipelineStorage.

packages/compass-query-bar/src/index.tsx
```diff
@@ -38,6 +42,8 @@ const QueryBarPlugin = registerHadronPlugin(
     instance: mongoDBInstanceLocator,
     preferences: preferencesLocator,
     logger: createLoggerAndTelemetryLocator('COMPASS-QUERY-BAR-UI'),
+    locateFavoriteQueryStorage: createFavoriteQueryStorageLocator,
+    locateRecentQueryStorage: createRecentQueryStorageLocator,
```
himanshusinghs, line 46: Would like to call out - that initialisation of FavoriteQueryStorage/RecentQueryStorage depended on namespace to be provided [during store initialisation](https://github.com/mongodb-js/compass/pull/5384/files#diff-d17431ff802f04d48d10966a42c04c20550413052e79f9ecf57ee95243ec2b80L112-L113) and hence these are not direct services, instead the service locators that accepts the same parameter as that of the constructors (FavoriteQueryStorage/RecentQueryStorage) and returns a service for that namespace.

This diverges a bit from the general pattern we have where the registerHadronPlugin is provided a service locator which returns a service implementation which is why I am unsure of how I feel about this. Let me know if I could've approached this in a different way.

himanshusinghs:
Converted this to draft to fix some implementation issues but please feel free to drive by.

packages/compass-aggregations/test/configure-store.ts
himanshusinghs: I think I was trying to avoid a @tslint-ignore here since we [don't allow imports](https://github.com/mongodb-js/compass/blob/COMPASS-7411-provider-for-my-queries-storage/configs/eslint-config-compass/plugin.js#L23) from `@mongodb-js/my-queries-storage`.
I understand that this should be fine for test files/test helpers here but as per our [override definition](https://github.com/mongodb-js/compass/blob/COMPASS-7411-provider-for-my-queries-storage/configs/eslint-config-compass/index.js#L40-L58) this is not a test file. How about I expand that definition to turn the rule off also for the entire test directory, including something like `**/test/**/*`?

packages/my-queries-storage/src/provider.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export type FavoriteQueryStorageAccess = {
+  createStorage(
+    ...options: ConstructorParameters<typeof FavoriteQueryStorage>
+  ): FavoriteQueryStorage;
+};
+
+export type RecentQueryStorageAccess = {
+  createStorage(
+    ...options: ConstructorParameters<typeof RecentQueryStorage>
```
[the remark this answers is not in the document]
himanshusinghs: On it 👍

packages/compass-aggregations/test/configure-store.ts
addaleax: > How about I expand that definition to turn the rule off also for the entire test directory, including something like `**/test/**/*`?

Yes, that makes sense to me and is definitely in line with the intentions of those definitions 👍

configs/eslint-config-compass/index.js
```diff
@@ -57,6 +57,13 @@ const testTsOverrides = {
   },
 };
 
+const testHelpersOverrides = {
+  files: ['**/test/**/*'],
```
addaleax, line 61: Would it make sense to add `**/test/**/*.js` and `**/test/**/*.ts` patterns to the existing `testJsOverrides` and `testTsOverrides` file patterns, respectively?

packages/my-queries-storage/src/provider.ts
```diff
@@ -0,0 +1,51 @@
...
+  RecentQueryStorage,
+  QueryStorageOptions,
+} from './query-storage';
+
+export type FavoriteQueryStorageAccess = {
+  createStorage(options?: QueryStorageOptions): FavoriteQueryStorage;
+};
+
+export type RecentQueryStorageAccess = {
+  createStorage(options?: QueryStorageOptions): RecentQueryStorage;
```
addaleax, line 14: Two things:
- This is just a really tiny thing, but it should be transparent to the caller whether `createStorage()` genuinely returns a new object or not, it never actually creates a resource, so should these be called `getStorage()` to make that a bit clearer?
- Right now the return types for these methods are class types, that means that this is not super usable as an _interface_ (that e.g. Data Explorer could implement differently), would it be possible to turn this into one? I know that that adds a bit of complexity, but otherwise dependency injection doesn't make a whole lot of sense here if we're already being very prescriptive about the implementation

configs/eslint-config-compass/index.js
himanshusinghs: Yea - I initially did that but it started creating other eslint problems in multiple packages. From the initial analysis it looked like the pattern that I added was conflicting somehow with other patterns. Some errors were quite cryptic to debug and I wanted a quick turn around so I decided to keep it separate here. I can try to debug those issues further if we should definitely do the other one.

packages/my-queries-storage/src/provider.ts
himanshusinghs: For the first point - It actually does create an instance of the class [here](https://github.com/mongodb-js/compass/pull/5384/files#diff-b90c9f0d64dcf60e3a6638e888d2175e05cf9dff8f7ad68f4945619030e36ec6R193-R199) but at the same time I agree that `getStorage` would convey it much clearer that a new object is being returned.

> would it be possible to turn this into one?
I think something got missed here? What were you suggesting for this to be turned into?

packages/compass-query-bar/src/stores/query-bar-store.ts
```diff
@@ -108,11 +112,17 @@ export function activatePlugin(
...
+    favoriteQueryStorageAccess,
+    recentQueryStorageAccess,
     atlasService = new AtlasService(),
-    recentQueryStorage = new RecentQueryStorage({ namespace }),
-    favoriteQueryStorage = new FavoriteQueryStorage({ namespace }),
   } = services;
 
+  const favoriteQueryStorage = favoriteQueryStorageAccess.createStorage({
+    namespace,
+  });
```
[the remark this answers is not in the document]
himanshusinghs: Thanks for pointing this out @gribnoysup. Addressed this finally in [9ea5597](https://github.com/mongodb-js/compass/pull/5384/commits/9ea55977b09743e06699490329f4d89baf6e6335)

packages/compass-query-bar/src/index.tsx
himanshusinghs: Not a concern anymore because we are now injecting access interfaces to create instance of the desired class when needed in plugin stores.

packages/my-queries-storage/src/provider.ts
addaleax: > `getStorage` would convey it much clearer that a new object is being returned

Well yeah, my point here was, it shouldn't matter whether it's a new object or not to callers :)

> What were you suggesting for this to be turned into?

Right now, the return type of `createStorage()`/`getStorage()` is `RecentQueryStorage`, which is a class, so from a TypeScript point of view it's not just a type, it's a *specific* implementation of that type. If we leave it like that, we cannot have alternate implementations of this service, which makes service injection a bit pointless – you're already stuck with a single, fixed implementation.

(Here's [a TS playground](https://www.typescriptlang.org/play?#code/MYGwhgzhAECSC2AHEBGaBvAUNaAzA9vgBQCUAXNAG74CWAJhgL7bSIBONlYALgKbR1e3MDRCkK1ek0zNMoSDATIATBhYFi5KrQbpmOdpx79Bw0eO1S9MzJgD0d6ABF8vCADsA5N2gB3fGwA1gA00ADmQkpiJNDAYO7QAFT47iAAnonQbEIArmwJ8dA07hDC7sD8+LhwSKiYxXxsuGAV0ABihAAKbPiU9LxsaFg4EdxRFlEoANw2cimlRbUoHfjdvf2DFCtrfYKD0AC8GOGRtaTH2dx5Ce68vjXIKKRT0IzMwPM+NLXK2z27AxQWy6-w2aCO6BOYzOMUhl2u0Fu9yiymer1kDmgAHUAoEIKFuAALfi8AAeiF4wD4DAaA2arRoMDJyBowBo3Hq7ka9P4AGFwFAsezCSs1DgNBZJHQZrJaU0Wvw-us9qphlDxlp+QohUSVjLbB8Sl8lkqAWxlMDVqCVYdjqMNRdcvlEXcHqg0W85kbFipTRsLe0QcqBqoIeqYY6rs6kW7USQXp6gA) that shows the problem as a concrete error, in case that helps clarify what I mean 🙂)

Very broadly speaking, we shouldn't rely on types from the "Compass"/"default" implementation in our provider.ts files, we should define interfaces here in the provider.ts file and then just make sure that the implementation implements that interface

himanshusinghs: Ahh now it clicked to me! I definitely overlooked that aspect, thank you so much for pointing this out and for the playground also, definitely helped 🙌

configs/eslint-config-compass/index.js
addaleax: Hm, it works fine for me (diff I tried: https://gist.github.com/addaleax/86a632e15db6805b787bf99e95f3911c)

Do you remember the actual error message here? This probably isn't something you need to spend time on for this PR specifically, no, but I'm happy to take a deeper look if I find some time to do so

packages/my-queries-storage/src/provider.ts
himanshusinghs: Addressed both in my recent commit. Now the compass specific implementation is separated from the interface itself for both pipeline storage and query storage. I renamed the compass specific implementation to be prefixed with `Compass` and the normal one is for the interface itself.

packages/my-queries-storage/src/provider.ts
```diff
@@ -0,0 +1,50 @@
+import { createContext, useContext } from 'react';
+import type { QueryStorageOptions } from './compass-query-storage';
+import type { PipelineStorage } from './pipeline-storage';
+import type { FavoriteQueryStorage, RecentQueryStorage } from './query-storage';
```
himanshusinghs, line 4: These are not implementations but the interfaces. I renamed the implementation to be prefixed with `Compass` and moved them to separate files also prefixed with `compass-`.

configs/eslint-config-compass/index.js
himanshusinghs: Ok so the issue was that I was using `**/test/**/*` which was most likely also including non-js/ts files and for that reason the eslint parser was failing on those files with this cryptic error. With your pattern it all works fine. Thank you 🙌 I just added this commit.

packages/compass-crud/src/stores/crud-store.spec.ts
```diff
@@ -188,6 +188,16 @@ describe('store', function () {
...
+          favoriteQueryStorageAccess: {
+            getStorage() {
+              return new CompassFavoriteQueryStorage();
+            },
+          },
+          recentQueryStorageAccess: {
+            getStorage() {
+              return new CompassRecentQueryStorage();
+            },
+          },
```
addaleax, line 200: Fwiw, if you want to add something like

```js
const favoriteQueryStorageAccess = {
  getStorage() {
    return new CompassFavoriteQueryStorage();
  },
};
const recentQueryStorageAccess = {
  getStorage() {
    return new CompassRecentQueryStorage();
  },
};
```

at he top of the file and then juts use

```js
          favoriteQueryStorageAccess,
          recentQueryStorageAccess,
```

here instead of providing a new definition each time, I think that would be okay and maybe reduce the verbosity level a bit 🙂 

Alternatively, maybe we could even just export `favoriteQueryStorageAccess` and `recentQueryStorageAccess` (with `compass...` name prefixed) from `'@mongodb-js/my-queries-storage'` as utilities so that consumers in test files don't need to repeat this bit at all?

addaleax (approved):
very nice! one optional suggestion, but this is good to go as far as I am concerned 🙂

packages/compass-crud/src/stores/crud-store.spec.ts
himanshusinghs: Ahh yea that makes sense, looks simple enough so I will get it in right away. Thank you :)

addaleax (approved):

PR #5410, opened 2024-02-02
feat(bson-transpiler)!: bump bson and remove unused inputs

mcasimir:
- Bump bson in `bson-transpilers`
- removes "support" and code for the unused `python` and `javascript` as input in `bson-transpiler`
- expand `shell` input tests cases to match the `javascript` input tests cases
- fixes failing tests and missing bson types for shell 
- deletes C# grammars, which where never used as we don't have a C# input

The motivation for doing this is:
- This is a first step towards cleaning this package up a bit and reducing the maintenance overhead so we can at least keep driver dependencies up-to-date. I added a note in the readme pointing to a commit that still has Python.
- Grammars and symbols are bundled with VSCode and Compass and brings a significant amount of code.


packages/bson-transpilers/symbols/javascript/templates.yaml
```diff
@@ -220,24 +220,12 @@ Templates:
     CodeCodeArgsTemplate: &CodeCodeArgsTemplate null
     CodeScopeTemplate: &CodeScopeTemplate null
     CodeScopeArgsTemplate: &CodeScopeArgsTemplate null
-    ObjectIdToStringTemplate: &ObjectIdToStringTemplate !!js/function >
```
mcasimir, line 223: this was not necessary, only breaks some of the transformations that are correctly handled by default

packages/bson-transpilers/test/yaml/bson-methods.yaml
```diff
@@ -273,50 +236,12 @@ tests:
             '$id' => BSON::ObjectId('5bfeee5ffa63a31985db9e5e')
           ).id
   Long-methods:
-    - input:
-        shell: NumberLong(429496729601).top
```
mcasimir: this syntax is not valid anymore

packages/bson-transpilers/test/yaml/non-idiomatic.yaml
```diff
@@ -1,209 +1,181 @@
-runner: !!js/function >
-    (it, type, expect, input, output, transpiler, test) => {
-        it(test.description, () => {
-            expect(
-                transpiler[input][output].compile(test.input[input], false)
-            ).to.equal(test.output[output]);
-        });
-    }
+runner: !<tag:yaml.org,2002:js/function> |-
```
kmruiz, line 1: I'm sorry, but what is this? Is this to have syntax highlighting?

kmruiz (approved):

packages/bson-transpilers/test/yaml/non-idiomatic.yaml
mcasimir: Ah yeah, i don't know :). I did most of the changes with a script that parsed the yaml files, changes / replaces the tests, and wrote the file back with Yaml.dump. The library generated that, i'm assuming is a fancier way to write this: `!!js/function`

PR #5413, opened 2024-02-04
feat(compass-indexes): display vector search type in search indexes table COMPASS-7509

Anemy:
COMPASS-7509

This only works on MongoDB version 6.0.11, 7.0.2 in Atlas. For testing I use the `analytics-node-test` deployment in our Compass team project.

Docs: https://www.mongodb.com/docs/atlas/atlas-vector-search/create-index/

*Vector Search Index on the table*
Screenshot 2024-02-04 at 9 47 04 AM

Clicking the `Aggregate` button it opens a new tab on the aggregations page with the first stage filled:
Screenshot 2024-02-04 at 9 48 08 AM

*Index Details*
Screenshot 2024-02-04 at 10 04 47 AM


packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -659,6 +660,29 @@ export function getInitialSearchIndexPipeline(name: string) {
   ];
 }
 
+export function getInitialVectorSearchIndexPipelineText(name: string) {
+  return `[
+  {
+    $vectorSearch: {
+      // Name of the Atlas Vector Search index to use.
+      index: "${name}",
```
addaleax, line 668:
```suggestion
      index: ${JSON.stringify(name)},
```

Hopefully not something where characters that need to be escaped show up frequently, but probably also no reason not to make this bit resilient against that

addaleax (approved):

packages/compass-indexes/src/components/search-indexes-table/search-indexes-table.tsx
```diff
@@ -215,18 +255,30 @@ export const SearchIndexesTable: React.FunctionComponent<
...
-                initialPipeline: getInitialSearchIndexPipeline(name),
                 newTab: true,
+                ...(isVectorSearchIndex
+                  ? {
+                      initialPipelineText:
+                        getInitialVectorSearchIndexPipelineText(name),
+                    }
+                  : {
+                      initialPipeline: getInitialSearchIndexPipeline(name),
+                    }),
```
gribnoysup, line 266: Seems a bit weird that we are passing text for one, but parsed version for another. Can we normalize this?

packages/compass-indexes/src/components/search-indexes-table/search-indexes-table.tsx
```diff
@@ -118,14 +125,33 @@ const searchIndexFieldStyles = css({
...
-  definition,
-}: {
-  indexName: string;
-  definition: Document;
-}) {
+function VectorSearchIndexDetails({ definition }: { definition: Document }) {
+  return (
+    <>
+      {!definition.fields || definition.fields.length === 0
+        ? '[empty]'
```
gribnoysup, line 132: Is this in the designs? Should we just not allow expanding when there is nothing to show?

image

packages/compass-indexes/src/components/search-indexes-table/search-indexes-table.tsx
Anemy: I'll chat with product on what we want to show here, text gives the option to give an invalid pipeline and include code comments. We can do something similar to what we do for regular search indexes.

packages/compass-indexes/src/components/search-indexes-table/search-indexes-table.tsx
Anemy: No, it's not in the designs, however it shouldn't be something that's ever encountered. Trying to create a vector search index without any fields results in an error:
Screenshot 2024-02-05 at 1 07 08 PM
I don't think it's necessarily worth removing the empty case although we could remove the expanding altogether like you suggested. Since it shouldn't ever happen I think I'd prefer to just leave it as is.

packages/compass-indexes/src/components/search-indexes-table/search-indexes-table.tsx
Anemy: Also, a bit related, started a thread on slack on what we want the pipeline text to look like. Based on that I'll align them. If no one gets back I'll just update them both to text with a comment. Might wrap it into the follow up branch / pr for urls.

packages/compass-indexes/src/components/search-indexes-table/search-indexes-table.tsx
gribnoysup: One of the indexes that you created has this state, how did we manage to create it if server should throw an error? Why are we not showing an error in this case? My concern is [empty] looks like an unexplainable placeholder that will be confusing for the users and we already handle it differently for normal indexes, so why are we breaking the pattern? If this is a totally unexpected case, why are we handling this like that?

Anemy: Ah I was wrong that's an error from something else, it can happen. Also it's worth mentioning this is also what we show when there are no mappings or fields in the mappings for regular search indexes.
How does this look instead?
Screenshot 2024-02-06 at 12 00 20 PM

Also created slack thread: https://mongodb.slack.com/archives/C05C8MVJJQ7/p1707239370305669
Thanks for bringing this up!

gribnoysup: Nice, this covers it very well, thanks for handling this!

gribnoysup (approved):

PR #5442, opened 2024-02-13
feat: track gen-ai use in Intercom COMPASS-7116

paula-stacho:
https://jira.mongodb.org/browse/COMPASS-7116

## Description

It seems feasible that the intercom track would be reused in the future, so I added it to utils - let me know if that's not the best place. The intercom setup is currently in the main compass package.
Also please consider if I missed any other place where NL is used 😅 

The new tests for `GenerativeAIInput` are regression, I couldn't test the new behaviour because I couldn't put a spy on `intercomTrack` (not without moving it directly to the utils/index file or somewhere else completely).

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions
There are several logs associated with intercom setup. Do we want to log this event as well? https://github.com/mongodb-js/compass/blob/07194863a3c383631043eed68dd8a5d668edac93/packages/compass/src/app/intercom/intercom-script.ts#L55

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-utils/src/intercom-track.ts
+++ b/packages/compass-utils/src/intercom-track.ts
@@ -19,7 +19,14 @@ export function intercomTrack(
   if (typeof win === 'undefined') {
     return;
   }
-  if (!win.Intercom || typeof win.Intercom !== 'function') {
+
+  if (
+    !win.Intercom ||
+    typeof win.Intercom !== 'function' ||
+    // the last check is because the packages might be embedded in an environment where an external Intercom lives
+    // this way we ensure we don't polute such external Intercom with events intended for Compass Intercom
+    !process.env.HADRON_METRICS_INTERCOM_APP_ID
+  ) {
     return;
   }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -48700,6 +48700,9 @@
       "name": "@mongodb-js/compass-utils",
       "version": "0.5.11",
       "license": "SSPL",
+      "dependencies": {
+        "@mongodb-js/compass-logging": "^1.2.12"
+      },
       "devDependencies": {
         "@mongodb-js/eslint-config-compass": "^1.0.15",
         "@mongodb-js/mocha-config-compass": "^1.3.6",
@@ -48721,6 +48724,9 @@
       "optionalDependencies": {
         "@electron/remote": "^2.1.1",
         "electron": "^28.2.1"
+      },
+      "peerDependencies": {
+        "@mongodb-js/compass-logging": "^1.2.12"
       }
     },
     "packages/compass-utils/node_modules/sinon": {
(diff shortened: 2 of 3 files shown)


lerouxb (approved):

packages/compass-utils/src/intercom-track.ts
```diff
@@ -0,0 +1,24 @@
+type WindowWithIntercomGlobals = Window &
+  typeof globalThis & {
+    Intercom?: (...args: any[]) => any;
+    attachEvent?: (...args: any[]) => any;
+  };
+
+export enum IntercomTrackingEvent {
+  submittedNlPrompt = 'submitted-nl-prompt',
```
lerouxb, line 8: Is this really the first and only event we're tracking in intercom?

paula-stacho: but I'm pretty noob in this codebase, so perhaps someone else can confirm 😅

paula-stacho: duplicate conversation in https://mongodb.slack.com/archives/G2L10JAV7/p1707914433425239

paula-stacho:
adding a `wip` label as I'm addressing some comments from @mcasimir

paula-stacho:
@lerouxb added some changes after your review - a catch with a debug log, and an extra check ensuring we don't pollute external (Atlas) Intercom.

another suggestion from @mcasimir was to move the all the intercom code together. I will do this as a follow up in a separate PR

packages/compass-utils/src/intercom-track.ts
```diff
@@ -0,0 +1,38 @@
...
+    // this way we ensure we don't polute such external Intercom with events intended for Compass Intercom
+    !process.env.HADRON_METRICS_INTERCOM_APP_ID
+  ) {
+    return;
+  }
+
+  try {
+    win.Intercom('track', event, metadata);
+  } catch (error) {
+    debug('intercom track error', error);
```
lerouxb, line 36: Just FYI: I don't think debug() would end up in a log file. If you're interested in seeing this in logs if it happens you would have to use log() on it.

paula-stacho: hmm, I thought it'd make sense to use the same level as when the intercom fails to load -> https://github.com/mongodb-js/compass/blob/20a6ff86654818b70348e61ee611a29afade7277/packages/compass/src/app/intercom/intercom-script.ts#L37.
but tbh I'm not sure why that level was chosen in that instance either 🤔

paula-stacho: probably because it's not an user facing failure, so noone would look into the logs unless debugging anyway? or are there other cases where you'd imagine we'd need this in the logs?

lerouxb: The problem with debug() is that you can only see it if you have debug enabled and you're looking at the console or stdout. So it isn't something a user can spot in a log file and send it to us and we'd have to already know how to recreate the intercom failure in order to generate any debug output locally.

Which might be fine. Just thought I'd bring it up.

PR #5445, opened 2024-02-14
chore: add main process unhandled rejection handler COMPASS-7268

lerouxb:
We have an unhandled exception handler in compass, but no unhandled rejection handler. This adds one.

### DIFF
--- a/packages/compass/src/main/handle-unhandled-rejection.ts
+++ b/packages/compass/src/main/handle-unhandled-rejection.ts
@@ -1,0 +1,48 @@
+import { app, dialog, clipboard } from 'electron';
+import cleanStack from 'clean-stack';
+import ensureError from 'ensure-error';
+import COMPASS_ICON from './icon';
+
+async function handleUnhandledRejection(err: Error): Promise<void> {
+  // eslint-disable-next-line no-console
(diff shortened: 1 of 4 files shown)


packages/compass/src/main/handle-unhandled-rejection.ts
```diff
@@ -0,0 +1,48 @@
+import { app, dialog, clipboard } from 'electron';
+import cleanStack from 'clean-stack';
+import ensureError from 'ensure-error';
+import COMPASS_ICON from './icon';
+
+async function handleUnhandledRejection(err: Error): Promise<void> {
```
lerouxb, line 6: Right now this is just a copy of the uncaught exception handler with literally only the console.error line changed.

Anemy: Should we share/reduce the duplicated code somehow? Bind the uncaught type to one method that changes the console error?

Anemy (approved):
Lgtm, would be nice to reduce the code duplication if we can before merging I think.

lerouxb:
aaan I accidentally merged before we do the GA. REverting.

PR #5454, opened 2024-02-15
chore: fix for compass-web sandbox reporting about missing providers

himanshusinghs:
## Description
A followup of #5384 in which I missed taking into account how the injection of PipelineStorage and QueryStorage will behave on compass-web. As it turn out we have these features disabled on DE and currently its not in the plan to have them enabled at any point.

This PR marks `PipelineStorage` and `QueryStorage` optional and every part of code which depends on them are modified to deal with the possibility of them being not there.

Additionally we have removed the preference `enableSavedAggregationsQueries` to simplify the dependency between the preference itself and the presence of the storages in contexts. Now we will impose the UI restrictions that earlier were done using the preference, now with the presence/absence of storage value in the context.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51294,8 +51294,7 @@
       "dependencies": {
         "@mongodb-js/compass-editor": "^0.20.3",
         "@mongodb-js/compass-user-data": "^0.1.15",
-        "bson": "^6.3.0",
-        "compass-preferences-model": "^2.17.4"
+        "bson": "^6.3.0"
       },
       "devDependencies": {
         "@mongodb-js/eslint-config-compass": "^1.0.15",
@@ -63091,7 +63090,6 @@
         "@types/sinon-chai": "^3.2.5",
         "bson": "^6.3.0",
         "chai": "^4.3.6",
-        "compass-preferences-model": "^2.17.4",
         "depcheck": "^1.4.1",
         "eslint": "^7.25.0",
         "gen-esm-wrapper": "^1.1.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -51294,7 +51294,8 @@
       "dependencies": {
         "@mongodb-js/compass-editor": "^0.20.3",
         "@mongodb-js/compass-user-data": "^0.1.15",
-        "bson": "^6.3.0"
+        "bson": "^6.3.0",
+        "compass-preferences-model": "^2.17.4"
       },
       "devDependencies": {
         "@mongodb-js/eslint-config-compass": "^1.0.15",
@@ -63090,6 +63091,7 @@
         "@types/sinon-chai": "^3.2.5",
         "bson": "^6.3.0",
         "chai": "^4.3.6",
+        "compass-preferences-model": "^2.17.4",
         "depcheck": "^1.4.1",
         "eslint": "^7.25.0",
         "gen-esm-wrapper": "^1.1.0",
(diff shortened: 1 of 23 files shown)


lerouxb (approved):

package-lock.json
```diff
@@ -46395,7 +46395,7 @@
         "react": "^17.0.2",
         "react-dom": "^17.0.2",
         "reflux": "^0.4.1",
-        "reflux-state-mixin": "git+ssh://git@github.com/mongodb-js/reflux-state-mixin.git#e050454cb3be029c3e7fd2ee6a08111e4d15161f",
```
himanshusinghs, line 46398: These are not related to my changes, but will this update cause any issue? Seems like we are pinning it down to a particular commit?

packages/compass-saved-aggregations-queries/src/stores/aggregations-queries-items.ts
```diff
@@ -62,22 +60,9 @@ const reducer: Reducer<State, Actions | EditItemActions | DeleteItemActions> = (
         items: newItems,
       };
     }
-    case EditItemActionTypes.EditItemUpdated: {
```
himanshusinghs, line 65: We don't need to do this, instead we will simply reload all the queries again from the storage (doing that in edit-item.ts down below).

packages/my-queries-storage/src/provider.ts
```diff
@@ -13,38 +14,51 @@ export type RecentQueryStorageAccess = {
...
-  };
-
-export const recentQueryStorageAccessLocator = (): RecentQueryStorageAccess => {
-  const recentQueryStorageAccess = useContext(RecentQueryStorageContext);
-  if (!recentQueryStorageAccess) {
-    throw new Error('No RecentQueryStorage available in this context');
+export const favoriteQueryStorageAccessLocator = ():
+  | FavoriteQueryStorageAccess
+  | undefined => {
+  const storageIsRequired = usePreference('enableSavedAggregationsQueries');
```
himanshusinghs, line 45: One way to use the preferences was this, to use the hook and assume that we are in a tree below the PreferencesProvider. The other way was to provider preferencesInstance alongside these storage values in the providers so that we can access that instance here in locator, similar to how this is done for LoggerAndTelemetryProvider.

I opted for this one only because it will keep the CompassWeb render tree free of these Providers, otherwise we will have to have them there as well passing just the preferences in. LMK if its alright with you all.

package-lock.json
himanshusinghs: Outdated, its gone now

packages/my-queries-storage/src/provider.ts
himanshusinghs: Summarising my discussion with @addaleax:
We can simplify this peace here by not having this preference altogether because at this point this preference is not giving much value on top of the presence/absence of storage values in providers and we would be better off by just checking the presence of the storage in the provider and doing the same that we were earlier doing with the preference.

himanshusinghs: Addressed this in [5325252](https://github.com/mongodb-js/compass/pull/5454/commits/532525211d7723217019504e3525e41dc75056f3)

addaleax (approved):

PR #5470, opened 2024-02-20
chore(ci): use node v20.11.1 in CI

lerouxb:
With this PR:

* On RHEL 7.6, use an unofficial build.
* We package compass for ubuntu on 16.04 and there we also use the same unofficial build.
* On windows, mac and other ubuntu tasks we use the official build.

### DIFF
--- a/.evergreen/preinstall.sh
+++ b/.evergreen/preinstall.sh
@@ -43,11 +43,21 @@ if [ -n "$IS_WINDOWS" ]; then
     cd ..
     .evergreen/node-gyp-bug-workaround.sh
 else
-    echo "Installing nodejs v${NODE_JS_VERSION} for ${PLATFORM} on ${ARCH}..."
+    if [ -n "$IS_RHEL" ]; then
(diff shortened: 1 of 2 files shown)


.evergreen/preinstall.sh
```diff
@@ -43,11 +43,21 @@ if [ -n "$IS_WINDOWS" ]; then
     cd ..
     .evergreen/node-gyp-bug-workaround.sh
 else
-    echo "Installing nodejs v${NODE_JS_VERSION} for ${PLATFORM} on ${ARCH}..."
+    if [ -n "$IS_RHEL" ]; then
+        echo "Installing unoffocial nodejs compiled for glibc 2.17 v${NODE_JS_VERSION} for ${PLATFORM} on ${ARCH}..."
```
addaleax, line 47:
```suggestion
        echo "Installing unofficial nodejs compiled for glibc 2.17 v${NODE_JS_VERSION} for ${PLATFORM} on ${ARCH}..."
```

addaleax (approved):

PR #5479, opened 2024-02-21
chore(ci): node 18, npm 10 in CI

lerouxb:
Replaces https://github.com/mongodb-js/compass/pull/5470

Some tests fail on node 20, so I'm doing node 18 for now. Going with npm 10 since that's what you get with node 18 so it is convenient. And it seems to work.


addaleax (approved):

packages/compass-e2e-tests/tests/connection-form.test.ts
```diff
@@ -101,7 +101,7 @@ describe('Connection form', function () {
   });
 
   it('parses and formats a URI for multiple hosts', async function () {
-    const connectionString = 'mongodb://localhost:27017,localhost:27091/';
+    const connectionString = 'mongodb://localhost:27017,127.0.0.1:27091/';
```
lerouxb, line 104: Many of these aren't technically necessary, but should be harmless. I was _trying_ to surgically find&replace. We really only need to change the places that actually connects and since compass still has the dns resolution order hack we actually only have to change the places where we connect from the e2e tests process.

lerouxb: I guess the only places that really matter would be things like our insert-data.ts helpers and a handful of other places where we connect to mongodb directly rather than via compass.

PR #5481, opened 2024-02-21
chore(e2e): Unskip web e2e tests COMPASS-7600

lerouxb:
You should be able to search for `skipForWeb(this,` to see all the tests we're still skipping including reasons. And then there are a few assertions skipped due to telemetry not working yet or things checking recent queries which you can be found with `if (!TEST_COMPASS_WEB`.

The rest should all be unskipped now.

I cherry-picked one test from https://github.com/mongodb-js/compass/pull/5479 because I was testing this on node 18, so I'll probably merge that first.

### DIFF
--- a/packages/compass-e2e-tests/tests/atlas-login.test.ts
+++ b/packages/compass-e2e-tests/tests/atlas-login.test.ts
@@ -4,7 +4,7 @@ import {
   cleanup,
   screenshotIfFailed,
   Selectors,
-  TEST_COMPASS_WEB,
+  skipForWeb,
 } from '../helpers/compass';
 import type { Compass } from '../helpers/compass';
 import type { OIDCMockProviderConfig } from '@mongodb-js/oidc-mock-provider';
@@ -41,10 +41,8 @@ describe('Atlas Login', function () {
   let stopMockAtlasServer: () => Promise<void>;
 
   before(async function () {
-    if (TEST_COMPASS_WEB) {
-      // atlast-login not supported in compass-web
-      this.skip();
-    }
+    skipForWeb(this, 'atlas-login not supported in compass-web');
+
     // Start a mock server to pass an ai response.
     const { endpoint, stop } = await startMockAtlasServiceServer();
     stopMockAtlasServer = stop;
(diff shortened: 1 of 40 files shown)


packages/compass-e2e-tests/tests/atlas-login.test.ts
```diff
@@ -125,9 +124,7 @@ describe('Atlas Login', function () {
   });
 
   after(async function () {
-    if (TEST_COMPASS_WEB) {
-      return;
-    }
+    skipForWeb(this);
```
addaleax, line 127: What exactly is being skipped when doing this in an `after()` (or `afterEach()`) hook?

lerouxb: The after() or afterEach() hook. for every before() or beforeEach() hook even if you skipped in there the corresponding after() or afterEach() hook still executes. By design because they can't know how far you got before you called this.skip().

lerouxb: I just realised I am relying on skip working the same as early return there. Didn't actually check if it is even supported in after/afterEach hooks 😬

lerouxb: it seems to throw and get caught. typescript somehow even knows that code after it won't execute.

lerouxb: grumble: `Error: `this.skip` forbidden`. I guess I'll change those back to early return

packages/compass-e2e-tests/tests/atlas-login.test.ts
```diff
@@ -41,9 +42,8 @@ describe('Atlas Login', function () {
   let stopMockAtlasServer: () => Promise<void>;
 
   before(async function () {
-    if (TEST_COMPASS_WEB) {
-      this.skip();
-    }
+    skipForWeb(this, 'atlas-login not supported in compass-web');
```
addaleax, line 45: There's still a couple of places where early returns are missing for `skipForWeb()` inside `before()` hooks :)

lerouxb: you can call skip inside before hooks and it is implemented with a throw. you only have to early return in after hooks where skip is not available. Or am I misunderstanding you?

lerouxb: Basically: The PR won't be green if we don't skip over bits that will break in compass web.

mabaasit (approved):

addaleax (approved):

PR #5497, opened 2024-02-26
fix: don't include the version number in process.title because it shows in the menubar in macOS Sonoma COMPASS-7513

lerouxb:
Looks like this was harmless before the release of macOS Sonoma.

Screenshot 2024-02-26 at 11 42 39

### DIFF
--- a/packages/compass/src/main/index.ts
+++ b/packages/compass/src/main/index.ts
@@ -24,7 +24,7 @@ initializeElectronRemote();
 installEarlyLoggingListener();
 installEarlyOpenUrlListener();
 
-process.title = `${app.getName()} ${app.getVersion()}`;
+process.title = app.getName();
 
 void main();
 


packages/compass/src/main/index.ts
```diff
@@ -24,7 +24,7 @@ initializeElectronRemote();
 installEarlyLoggingListener();
 installEarlyOpenUrlListener();
 
-process.title = `${app.getName()} ${app.getVersion()}`;
+process.title = app.getName();
```
kmruiz, line 27: Why can't we just remove it? Shouldn't take the title from the Info.plist?

kmruiz (approved):

packages/compass/src/main/index.ts
addaleax: https://github.com/mongodb-js/compass/commit/01838395c179e7ea9b9109ff771012d8c7e91c2c

`process.title` also overrides the actual command line, so that tools like `ps` don't display it (because the command line could include e.g. a connection string with credentials or a `--password` argument). We do the same in mongosh.

PR #5540, opened 2024-03-05
chore: zip e2e coverage report and upload it

lerouxb:
No idea why I didn't do this from the start.


mabaasit (approved):

.evergreen/functions.yml
```diff
@@ -88,14 +88,15 @@ post:
       working_dir: src
       shell: bash
       script: |
-        tar czvf all-e2e-logs.tgz packages/compass-e2e-tests/.log
+        tar czf all-e2e-logs.tgz packages/compass-e2e-tests/.log
```
lerouxb, line 91: This runs right at the end and was running with -v so every file was listed and recently a gazillion little .json files were added for each test's output..

PR #5543, opened 2024-03-06
chore(ci): better snyk error COMPASS-7693

mcasimir:


himanshusinghs (approved):

lerouxb (approved):

scripts/snyk-test.js
```diff
@@ -26,7 +26,10 @@ async function snykTest(cwd) {
           '--dev',
           `--json-file-output=${tmpPath}`,
         ],
-        { cwd }
+        {
+          cwd,
+          maxBuffer: 1024 * 1024 * 5, // 5 MB (default is 1)
```
addaleax, line 31: [nit] We could as well raise this to be 50 MB :) Or use `spawn` with I/O passed through instead of `execFile`

addaleax (approved):

scripts/snyk-test.js
mcasimir: i thought about both: 
- using spawn and forwarding stdio, we silenced the output recently cause it kept confusing folks about what are the actual vulnerabilities that needs attention
- 50 MB that we can do yeah, any reason why we may want to have a smaller limit instead?

addaleax: > any reason why we may want to have a smaller limit instead?

Not really, I don't think 50 MB is something where we'd need to worry about memory exhaustion or anything like that

scripts/snyk-test.js
```diff
@@ -6,12 +6,25 @@ const { glob } = require('glob');
...
+}
+
 async function snykTest(cwd) {
   const tmpPath = path.join(os.tmpdir(), 'tempfile-' + Date.now());
 
   let execErr;
 
   try {
+    if (!(await fileExists(path.join(cwd, `package.json`)))) {
+      return;
```
lerouxb, line 25: You're pushing `null` onto results again.

mcasimir: nope, that's ok, the `"null"` issue, only happens with the monorepo root, but in the monorepo root a package.json is granted to be there, cause that package.json is being read to fetch workspaces. The issue is not in this function in any case.

mcasimir: https://github.com/mongodb-js/compass/blob/594d5049270abd08869276da331732010b792afd/scripts/snyk-test.js#L81

PR #5544, opened 2024-03-06
chore(e2e): Rewrite Instance Databases tab -> can refresh the list of databases using refresh controls COMPASS-7663

lerouxb:
Rather than wait for one of https://jira.mongodb.org/browse/COMPASS-7695 or https://jira.mongodb.org/browse/COMPASS-7696 or disabling the test in the meantime I just rewrote it to do less and just check that the button works in the shortest number of steps for now.

I think there was also a race condition in there _somewhere_ but I'm pretty confident that it won't affect real users and I've spent weeks debugging this off and on now. I really just want this test to stop flaking regularly because it keeps failing CI.


gribnoysup (approved):

packages/compass-e2e-tests/tests/instance-databases-tab.test.ts
```diff
@@ -172,10 +178,11 @@ describe('Instance databases tab', function () {
       await mongoClient.close();
     }
 
-    // looks like if you refresh too fast the database appears in the list but
-    // the stats never load, so just pause for bit first
-    await browser.pause(1000);
-    await browser.clickVisible(Selectors.InstanceRefreshDatabaseButton);
+    // Refresh again and the database card should disappear.
+    await browser.clickVisible(Selectors.InstanceRefreshDatabaseButton, {
```
lerouxb, line 182: Sometimes the card goes away. This is now the only time we hit the refresh button so the only thing I can think of is that the initial load is still somehow in progress by the time we hit the refresh button.

packages/compass-e2e-tests/tests/instance-databases-tab.test.ts
```diff
@@ -137,32 +137,38 @@ describe('Instance databases tab', function () {
...
+      Selectors.DatabasesTable,
+      dbSelector,
+      'grid'
+    );
+    await browser.$(dbSelector).waitForDisplayed();
+
+    // Wait for the page to finish loading as best as we can
+    await browser.waitUntil(async () => {
+      const placeholders = await browser.$$(Selectors.DatabaseStatLoader);
+      return placeholders.length === 0;
```
lerouxb, line 158: If only we (or a user) could reliably tell in the UI once everything is done loading. 🤔

PR #5551, opened 2024-03-07
chore(aggregations): remove preferences maxTimeMS from store COMPASS-7550

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/stores/store.ts
```diff
@@ -176,15 +175,6 @@ export function activateAggregationsPlugin(
     )
   );
 
-  store.dispatch(
-    preferencesMaxTimeMSChanged(preferences.getPreferences().maxTimeMS)
-  );
```
addaleax, line 181: Do we still apply the initial maxTimeMS value when the store was created somewhere?

addaleax: Otherwise, I think the `maxTimeMS` value from the preferences might not end up being applied, e.g. for the explain call in modules/insights.ts?

mabaasit: added [3ac15a9](https://github.com/mongodb-js/compass/pull/5551/commits/3ac15a9898f1244d9848f951cfa04d968b816f97) to apply initial maxTimeMS from preferences (or the default one).

packages/compass-aggregations/src/stores/store.ts
```diff
@@ -237,6 +229,12 @@ export function activateAggregationsPlugin(
 
   store.dispatch(updatePipelinePreview());
 
+  store.dispatch(
+    maxTimeMSChanged(
+      preferences.getPreferences().maxTimeMS ?? DEFAULT_MAX_TIME_MS
+    )
+  );
```
addaleax, line 236: In pipeline-collation.tsx we were so far differentiating the cases of "user has not set a maxTimeMS", "user inherits maxTimeMS from preferences/default" and "user has explicitly set a maxTimeMS for this aggregation"

That goes away if we approach it this way, I'd maybe figure out some way to keep the current behavior, but if we can't, then we should probably also refactor the pipeline-options.tsx component

addaleax (approved):

PR #5552, opened 2024-03-07
chore: show the list of saved connections in the new sidebar COMPASS-7708

kmruiz:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-connections/src/index.ts
```diff
@@ -1,3 +1,5 @@
 import Connections from './components/connections';
 
+export { useConnections } from './stores/connections-store';
```
gribnoysup, line 3: This should not be exported from the main package export

packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection.tsx
```diff
@@ -0,0 +1,260 @@
...
+import type { ItemSeparator } from '@mongodb-js/compass-components/lib/components/item-action-controls';
+
+const TOAST_TIMEOUT_MS = 5000; // 5 seconds.
+
+const iconStyles = css({
+  fontSize: 14,
+  minWidth: '14px',
+  maxWidth: '14px',
+  minHeight: '14px',
+  maxHeight: '14px',
```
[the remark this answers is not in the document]
gribnoysup: `Icon` components also have `size` property that you might want to use instead of this maybe

packages/compass-web/webpack.config.js
```diff
@@ -40,6 +40,11 @@ module.exports = async (env, args) => {
           '@mongodb-js/devtools-connect'
         ),
 
+        // They are unlikely to be used in a web environment as this information will
+        // be provided somehow else.
+        'resolve-mongodb-srv': false,
+        'mongodb-cloud-info': false,
+
```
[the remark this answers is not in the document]
kmruiz: AFAIK, the shouldn't be used in the web environment.

* resolve-mongodb-srv is only used for telemetry (getHostnameForConnection in compass-connections/.../telemetry.ts)
* mongodb-cloud-info is only used for telemetry too (getPublicCloudInfo in same file).

I personally don't think they should be packed by default as this information can be gathered in the web environment from the Atlas endpoints if we want them for telemetry.

packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection.tsx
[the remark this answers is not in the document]
kmruiz: The design is explicitly saying 14px and I don't see a spacing token of 14px in leafygreen:

https://www.mongodb.design/component/tokens/example/

What I can do is to use the spacing of 16px we have there, it will be 2px bigger, but I don't think that would be a big issue. WDYT?

packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection.tsx
```diff
@@ -0,0 +1,260 @@
...
+      onToggleFavoriteConnection,
+      onDuplicateConnection,
+      onDeleteConnection,
+    ]
+  );
+  return (
+    <li
+      {...hoverProps}
+      style={{
+        backgroundColor: connectionColorToHex(connectionInfo.favorite?.color),
```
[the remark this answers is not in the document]
kmruiz: The background colour will be different in case of light/dark theme and they are designed for accessibility.
Right now it's partially implemented, but it will be finished in COMPASS-7717:

https://jira.mongodb.org/browse/COMPASS-7717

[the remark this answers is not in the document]
gribnoysup: Ah, so it's not implemented yet, got it

packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection.tsx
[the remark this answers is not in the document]
kmruiz: That works for the icons, now I'll need to create a custom icon for the SVG so I can use small there too. Trying it.

packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection.tsx
```diff
@@ -0,0 +1,267 @@
...
+      onEditConnection,
+      copyConnectionString,
+      maybeProtectConnectionString,
+      onToggleFavoriteConnection,
+      onDuplicateConnection,
+      onDeleteConnection,
+    ]
+  );
+  return (
+    <li
```
gribnoysup, line 235: This item is part of the compass navigation, but is completely uninteractive when using a keyboard. Are we planning to address this?

kmruiz: Yes, the idea is to reuse `useVirtualNavigationTree` to implement keyboard navigation because we don't have anything specified in the design. You made me realise that I didn't create a task to apply keyboard accessibility so I'll create one for that and I'll discuss with Ben if there is any specific case we should consider on the keyboard navigation.

packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection-list.tsx
```diff
@@ -0,0 +1,114 @@
...
+const savedConnectionListStyles = css({
+  width: '100%',
+  height: '320px',
+  flex: 'none',
+  padding: spacing[3],
+  marginTop: 'auto',
+});
+
+const savedConnectionListPaddingStyles = css({
+  paddingLeft: spacing[2],
```
paula-stacho, line 21: I don't see this padding in figma, is it intentional?

kmruiz: Screenshot 2024-03-12 at 11 34 08

Is this little padding on the left of the server icon. It's tiny.

paula-stacho: in this one it's not visible because there is no color, but on the colored ones the padding is inside the items, not on the list level
Screenshot 2024-03-12 at 11 42 19

kmruiz: Ah, I see now the difference, yeah the padding should be inside the list element, not outside. Great catch, fixing it.

packages/compass-logging/src/provider.ts
```diff
@@ -44,6 +44,8 @@ const LoggerAndTelemetryContext = React.createContext<{
 
 export const LoggerAndTelemetryProvider = LoggerAndTelemetryContext.Provider;
 
+export { createLoggerAndTelemetry } from './ipc-logger';
```
gribnoysup, line 47: You are adding a platform specific dependency export to the provider, this is incorrect, if you are trying to use logger in the render, you should use the hook already exported from there, if you need to inject it to the plugin activate, you should use a locator

packages/compass-sidebar/src/components/multiple-connections/header/sidebar-header.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Subtitle, spacing, css } from '@mongodb-js/compass-components';
+
+const sidebarHeaderStyles = css({
+  maxHeight: '64px',
+  height: '64px',
```
gribnoysup, line 6:
```suggestion
  height: spacing[6],
```

packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection-list.tsx
```diff
@@ -0,0 +1,113 @@
...
+  width: '100%',
+  height: '320px',
+  flex: 'none',
+  padding: spacing[3],
+  marginTop: 'auto',
+});
+
+const savedConnectionListPaddingStyles = css({
+  overflowY: 'auto',
+  maxHeight: '100%',
```
gribnoysup, line 22: This max-height is causing the container to not scroll fully because you're basically forcing it to be the full height of the parent, not accounting for the header:

image

image

You probably want this container to take all the available space sans header, you can either hardcode the header height here to account for that (a bit clunky, hard to change and maintain) or actually make the parent a flex container to make sure that the list can stretch to all available space, but no more than that.

(the bottom padding of the parent should also probably be moved, otherwise the content weirdly "disappears" in parent when scrolling)

packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection.tsx
```diff
@@ -0,0 +1,264 @@
...
+        {getConnectionTitle(connectionInfo)}
+      </div>
+      <div
+        style={{ visibility: isHovered ? 'visible' : 'hidden' }}
+        className={savedConnectionToolbarStyles}
+      >
+        <IconButton
+          onClick={() => onConnect(connectionInfo)}
+          data-testid="connect-button"
+          aria-label="Connect"
```
gribnoysup, line 250: Maybe worth adding a `title` here too so that it can show up when you hover over the button, I couldn't figure out what this icon does without looking at the source

packages/compass-sidebar/src/components/multiple-connections/header/sidebar-header.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Subtitle, spacing, css } from '@mongodb-js/compass-components';
+
+const sidebarHeaderStyles = css({
+  maxHeight: '64px',
```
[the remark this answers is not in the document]
kmruiz: It should be fixed to 64px. It's a header with fixed content (right now is just a placeholder).

packages/compass-home/src/modules/telemetry.spec.ts
```diff
@@ -35,8 +39,10 @@ const dataService: Pick<DataService, 'instance' | 'getCurrentTopologyType'> = {
 
 describe('connection tracking', function () {
   let trackUsageStatistics: boolean;
+  let loggerAndTelemetry: LoggerAndTelemetry;
 
   before(async function () {
+    loggerAndTelemetry = createLoggerAndTelemetry('TEST-CONNECTION');
     // TODO(COMPASS-7397): Proper dependency injection for logger + telemetry would be nice here!
```
gribnoysup, line 46: You now have "dependency injection" added for those methods (through function params) and can actually fix this todo and detach tests from the implementation

gribnoysup (approved):

PR #5553, opened 2024-03-07
feat(query-bar): remove the filter label/link from the querybar COMPASS-7070

lerouxb:


packages/compass-query-bar/src/components/query-bar.tsx
```diff
@@ -71,7 +71,7 @@ const queryBarFirstRowStyles = css({
   // to account for their height individually.
   alignItems: 'flex-start',
   gap: spacing[2],
-  paddingLeft: spacing[2],
+  paddingLeft: spacing[1],
```
lerouxb, line 74: I have never understood what this 8px was doing there and it looks even weirder to me now. I tried removing it entirely then second-guessed myself and now I have it at 4px.. We could also just revert this line.

packages/compass-query-bar/src/components/query-history-button-popover.tsx
```diff
@@ -19,7 +19,6 @@ const openQueryHistoryButtonStyles = css(
     display: 'inline-flex',
     alignItems: 'center',
     padding: spacing[2] - 2, // -2px for border.
-    marginLeft: spacing[1],
```
lerouxb, line 22: There is nothing before it to space away from anymore.

packages/compass-query-bar/src/components/query-option.tsx
```diff
@@ -33,7 +33,7 @@ const queryOptionLabelStyles = css({
 });
 
 const documentEditorQueryOptionLabelStyles = css(queryOptionLabelStyles, {
-  minWidth: spacing[5] * 3,
+  minWidth: spacing[5] * 2,
```
lerouxb, line 36: I reduced this in line with this thread https://mongodb.slack.com/archives/GDTJXPHD0/p1709576196037729

We don't have to align with the Filter label anymore because it is not there. There's often also no queries dropdown.

Anemy (approved):
lgtm! So much more space !

PR #5555, opened 2024-03-08
chore(ci): stop testing with EOL mongodb servers

lerouxb:
This came out of a discussion in a [slack thread](https://mongodb.slack.com/archives/GDTJXPHD0/p1709821986595719).


kmruiz (approved):

addaleax (changes requested):
I wouldn't do this without
- confirmation from product that this we want to drop support for 4.2 and in particular 4.4 (I don't know if Julia's "we don't really have to worry about anything before 5.0" really means that)
- a COMPASS ticket with Documentation Changes Needed so that https://www.mongodb.com/docs/compass/current/install/#software-requirements gets updated

PR #5565, opened 2024-03-12
feat: information panels for multiple connections

paula-stacho:
## Description

Looks like more changes than it is, mainly copying and moving stuff.

Changes in nutshell:
- added the 2 info panels with updated design to connection-form, behind multiple connections feature toggle
- the original info panels are hidden when the toggle is on
- separated atlas-help to make future changes & cleanup easier, as this will be the only one of the original 'form-help' that remains in the welcome screen (might be completely useless if we don't stay within compass-connections)

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-connections/src/components/atlas-help/atlas-help.tsx
+++ b/packages/compass-connections/src/components/atlas-help/atlas-help.tsx
@@ -1,0 +1,98 @@
+import React from 'react';
+import {
+  Button,
+  ButtonSize,
+  ButtonVariant,
+  Subtitle,
+  Body,
+  Link,
+  spacing,
+  palette,
+  css,
+  cx,
+  useDarkMode,
+} from '@mongodb-js/compass-components';
+import { useLoggerAndTelemetry } from '@mongodb-js/compass-logging/provider';
+
+const sectionContainerStyles = css({
+  margin: 0,
+  padding: spacing[4],
+  paddingBottom: 0,
+});
+
+const atlasContainerStyles = css({
+  backgroundColor: palette.green.light3,
+  paddingBottom: spacing[4],
+});
+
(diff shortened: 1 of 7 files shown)


himanshusinghs (approved):
Noice, looks great 🚀

kmruiz:
Screenshot 2024-03-12 at 11 23 47

This info box is visible in the initial screen with the toggle on, should this be hidden when the multiple connections toggle is on?

kmruiz (approved):
I think it's great, I'm approving the PR because there is only one thing that I don't think it's blocker.

paula-stacho:
> Screenshot 2024-03-12 at 11 23 47
> This info box is visible in the initial screen with the toggle on, should this be hidden when the multiple connections toggle is on?

Depends? This one is not moved to the modal, and since we said to hide the other two, I imagined we're keeping this part and morphing it to the Welcome screen. That info box will be included in the Welcome screen. Perhaps I got it wrong and it will be morphed into something else or discarded? Either way I separated the box so that it can be moved or copied to the new position later.

kmruiz:
> > Screenshot 2024-03-12 at 11 23 47
> > This info box is visible in the initial screen with the toggle on, should this be hidden when the multiple connections toggle is on?
> 
> Depends? This one is not moved to the modal, and since we said to hide the other two, I imagined we're keeping this part and morphing it to the Welcome screen. That info box will be included in the Welcome screen. Perhaps I got it wrong and it will be morphed into something else or discarded? Either way I separated the box so that it can be moved or copied to the new position later.

That sounds a good approach, it works for me :D.

PR #5570, opened 2024-03-13
fix: don't allow the readonly filter to grow in width indefinitely COMPASS-7728

lerouxb:
I'm pretty sure we focused on this detail during development and that I worked on it. I must have broken it at the last moment. See the [ticket](https://jira.mongodb.org/browse/COMPASS-7728) for how bad this can get. It renders the preview pretty much useless for anything but the simplest filter queries.

after screenshots:

Screenshot 2024-03-13 at 09 24 30

Screenshot 2024-03-13 at 09 29 48

(see the ticket for the before screenshot)

### DIFF
--- a/packages/compass-crud/src/components/bulk-update-modal.tsx
+++ b/packages/compass-crud/src/components/bulk-update-modal.tsx
@@ -45,7 +45,10 @@ const columnsStyles = css({
   display: 'grid',
   width: '100%',
   gap: spacing[4],
-  gridTemplateColumns: '2fr 3fr',
+  // make sure the readonly filter field starts scrolling for large/complicated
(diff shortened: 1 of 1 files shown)


gribnoysup (approved):

paula-stacho (approved):
I haven't used the fr unit much, do you know why it doesn't work in this case?
Note: looks like it would work if the child had explicit `overflow: auto;`

lerouxb:
Explicit overflow: auto _might_ work but another problem was that the text editor tended to get too big anyway. Feels like you're more likely to want the preview to be big. At least in my own usage so far.

Having a dividing resizier line thing so users can choose themselves and being able to "maximize" the modal like the person that filed the ticked asked for would be ideal, I think.

PR #5573, opened 2024-03-13
fix(compass-web): min-width COMPASS-7614

lerouxb:
Here I limited the width of CompassWeb simply by changing the CSS in the sandbox wrapper without making any changes to anything inside CompassWeb:

Screenshot 2024-03-28 at 16 45 00
Screenshot 2024-03-28 at 16 44 54

I think this proves that we can achieve this entirely from MMS.


packages/compass-workspaces/src/components/workspaces.tsx
```diff
@@ -54,6 +54,7 @@ const workspacesContainerStyles = css({
   height: '100%',
   gridTemplateColumns: '100%',
   gridTemplateRows: 'auto 1fr',
+  minWidth: '1024px',
```
lerouxb, line 57: 1024 is probably too wide given that his excludes the sidebar and atlas chrome. Just for demonstration purposes at this stage.

We'll have to deal with the modals separately.

PR #5575, opened 2024-03-14
chore: allow to connect using the new sidebar COMPASS-7710

kmruiz:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


alenakhineika:
I have noticed that bootstrap fails in your branch when downloading fonts, when in main, no such issue.

```
FetchError: request to https://cloud.mongodb.com/static/font/EuclidCircularA-Semibold-WebXL.woff2 failed, reason: getaddrinfo ENOTFOUND cloud.mongodb.comDKpuz/5A9Fh4CEpyLScKR4Vcx8PqZAOQWapHQFcG4Z93jnNvMUUR00ZbC/YEgKSGkn+lcw== Completed in 56ms
    at ClientRequest.<anonymous> (/Users/alena.khineika/www/compass/node_modules/minipass-fetch/lib/index.js:97:14)
    at ClientRequest.emit (node:events:519:28)
    at TLSSocket.socketErrorListener (node:_http_client:500:9)
    at TLSSocket.emit (node:events:531:35)
    at emitErrorNT (node:internal/streams/destroy:169:8)
    at emitErrorCloseNT (node:internal/streams/destroy:128:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  code: 'ENOTFOUND',
  errno: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'cloud.mongodb.com',
  type: 'system'
}
```

kmruiz:
> I have noticed that bootstrap fails in your branch when downloading fonts, when in main, no such issue.

I didn't see the issue neither in CI or in my local machine, it seems like a temporary network issue in your laptop, maybe a VPN or something. Can you try again?

The URL seems correct and was unchanged (https://cloud.mongodb.com/...)  so it doesn't seem related to this PR.

packages/compass-connections/src/hooks/use-connection-status.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { type ConnectionInfo } from '@mongodb-js/connection-info';
+import { useEffect, useState } from 'react';
+
+export function useConnectionStatus(connectionInfo?: ConnectionInfo): {
+  status: ConnectionStatus;
+} {
+  if (!connectionInfo) {
+    connectionInfo = useConnectionInfo();
+  }
```
gribnoysup, line 16: It's not allowed to use hooks dynamically

packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection.tsx
```diff
@@ -57,15 +59,33 @@ type Action =
   | 'duplicate-connection'
   | 'remove-connection';
 
+const WarningIcon = () => {
```
gribnoysup, line 62: There is a leafygreen warning icon, why are we making a copy?

packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
```diff
@@ -22,45 +32,139 @@ const sidebarStyles = css({
...
+    </span>
+  );
+}
 // Having props here is useful as a placeholder and we will fix it with the first props.
 // eslint-disable-next-line
-export function MultipleConnectionSidebar({}: MultipleConnectionSidebarProps) {
+export function MultipleConnectionSidebar({
+  appName,
+}: MultipleConnectionSidebarProps) {
+  const { openToast, closeToast } = useToast('multiple-connection-status');
```
gribnoysup, line 78: Once again, just so we continue to be on the same page: this code doesn't belong in the sidebar and should be refactored out to a connection plugin at some point. If you are still planning to use only COMPASS-7397 as an overarching ticket for this work, it would be good to add a todo here

packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
```diff
@@ -22,45 +32,139 @@ const sidebarStyles = css({
...
+  const cancelCurrentConnectionRef = useRef<(id: string) => Promise<void>>();
+
+  const [isExpanded, setIsExpanded] = useState(true);
+  const [isConnectionFormOpen, setIsConnectionFormOpen] = useState(false);
+
+  const onConnected = useCallback(() => {
+    closeToast('connection-status');
```
gribnoysup, line 85: This probably should have an id based on a connection info id, otherwise one connection closes toasts for other connections. As the process is async and is not even visually blocked by the modal of any kind, this is very confusing: you can see progress or errors for multiple ongoing connections

packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
```diff
@@ -22,45 +32,139 @@ const sidebarStyles = css({
...
+      openToast('connection-status', {
+        title: `${error.message}`,
+        description: (
+          <ConnectionErrorToastBody
+            info={info}
+            onReview={reviewAndCloseToast}
+          />
+        ),
+        variant: 'warning',
+        timeout: 5000,
```
gribnoysup, line 125: Very unexpected that error toast has a timeout, that's not a pattern we use anywhere else in the app

packages/compass-components/src/components/with-status-marker.tsx
```diff
@@ -0,0 +1,117 @@
...
+  | 'connected'
+  | 'disconnected'
+  | 'connecting'
+  | 'failed';
+export type StatusMarkerProps = {
+  status: StatusMarker;
+  children: React.ReactNode;
+};
+
+function ConnectedStatusMarker(): React.ReactElement {
```
gribnoysup, line 16: We usually don't put components in compass-components that are very specific and are only used by a single place in the whole application

packages/compass-connections/src/hooks/use-connection-status.ts
```diff
@@ -0,0 +1,44 @@
...
+import { useConnectionsManagerContext } from '../provider';
+import {
+  ConnectionStatus,
+  ConnectionsManagerEvents,
+} from '../connections-manager';
+
+import { type ConnectionInfo } from '@mongodb-js/connection-info';
+import { useEffect, useState } from 'react';
+
+export function useConnectionStatus(connectionInfo?: ConnectionInfo): {
```
gribnoysup, line 11: This interface seems too broad: it requires us to get the whole connection info before being able to access the status while the only required value here is an id, can we limit this to actually required parameters?

packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
```diff
@@ -87,32 +191,39 @@ export function MultipleConnectionSidebar({}: MultipleConnectionSidebarProps) {
...
     (info: ConnectionInfo) => {
       void removeConnection(info);
     },
     [removeConnection]
   );
 
-  const onDuplicateConnection = useCallback(
+  const onEditConnection = useCallback(
     // Placeholder for when we implement it
     // eslint-disable-next-line
```
gribnoysup, line 203: You probably can remove this comment now that you implemented it?

packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
```diff
@@ -22,45 +32,139 @@ const sidebarStyles = css({
...
-    onConnectionAttemptStarted: noop_tmp,
-    onConnectionFailed: noop_tmp,
-    appName: '', // TODO: COMPASS-7710
-    getAutoConnectInfo: noop_tmp, // TODO: COMPASS-7710
+    onConnected: onConnected,
+    onConnectionAttemptStarted: onConnectionAttemptStarted,
+    onConnectionFailed(info, error) {
+      void onConnectionFailed(info, error);
+    },
+    appName: appName,
```
gribnoysup, line 148: It doesn't seem to be used by anything in the `useConnections`, why are we adding this?

packages/compass-components/src/components/with-status-marker.tsx
```diff
@@ -0,0 +1,117 @@
...
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+`;
+
+export const connectingIconStyle = css`
+  animation: ${connectingIconAnimation} 1.5s linear infinite;
+`;
```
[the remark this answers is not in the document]
kmruiz: Screenshot 2024-03-20 at 12 21 07

It's slightly different, it's green and has a background, as it overlaps with an icon. Look how the current spinner looks:

Screenshot 2024-03-20 at 12 24 50

packages/compass-components/src/components/with-status-marker.tsx
```diff
@@ -0,0 +1,117 @@
...
+} as const;
+
+const withStatusMarkerStyles = css({
+  position: 'relative',
+});
+
+const withStatusMarkerMarkerStyles = css({
+  position: 'absolute',
+  right: 0,
+  bottom: 0,
```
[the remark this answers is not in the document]
kmruiz: Didn't realise it was a bit different, thanks! I'll fix it.

packages/compass-connections/src/hooks/use-connection-status.ts
kmruiz: Ah right, I always forget about that rule.

I'll make the parameter mandatory.

packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
kmruiz: I think I'll use this ticket for this change, as it should be small, instead of using COMPASS-7397.

packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
kmruiz: Yeah Maurizio also was a bit surprised on this behaviour, but it's in the design. If the pattern is that error toasts are persistent, I think it's safe to remove the timeout and force the user to close it.

packages/compass-components/src/components/with-status-marker.tsx
kmruiz: Well it's going to be used in different places in the sidebar, and it's a relatively small and simple, but I can move it to the sidebar.

packages/compass-connections/src/hooks/use-connection-status.ts
kmruiz: Is there any reason to do so? I think it's more safe to just:

```
const connection = useConnectionInfo()
const status = useConnectionStatus(connection)
```

than

```
const connection = useConnectionInfo()
const status = useConnectionStatus(connection.id)
```

as requiring just the id is just an implementation detail of useConnectionStatus.

packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
kmruiz: Ah, it was used before and probably kept it after the merge with Himanshu's PR and forgot to remove it. Thanks for pointing out!

packages/compass-connections/src/hooks/use-connection-status.ts
gribnoysup: Can you elaborate what does it mean for id to be an implementation detail in this case? How else other parts of the UI should be able to communicate that they are attached to a certain connection?

The reason to do so is that we don't want to make a requirement for some parts of the application to depend on the whole connection info object (that they generally don't care about), potentially causing unneccessary subscriptions to the state updates, heavy deep check to bypass referencial equality issues, and complex state propagation scenarios in cases where passing an id that uniquely identifies a connection should be enough.

packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection.tsx
kmruiz: Didn't realise. Fixing it.

packages/compass-components/src/components/with-status-marker.tsx
gribnoysup: Small and simple is not a deciding factor for putting something in compass-components, using across multiple packages is

packages/compass-connections/src/hooks/use-connection-status.ts
kmruiz: > The reason to do so is that we don't want to make a requirement for some parts of the application to depend on the whole connection info object (that they generally don't care about), potentially causing unneccessary subscriptions to the state updates, heavy deep check to bypass referencial equality issues, and complex state propagation scenarios in cases where passing an id that uniquely identifies a connection should be enough.

That's not the case, and using objects as an argument to a function is not a problem. If you have a hook that has a dependency on the whole object in the dependency array (thanks React for that) yes, it can be problematic, but there is no subscription in this case anywhere. However, by passing the whole object, you gain the benefit of:

1. Encapsulation, if the object changes or the hook depends on other fields of ConnectionInfo, you just change the hook.
2. Intention is clear: you get the status of a connection, that is what you want to achieve, and the code clearly makes that explicit (doesn't care about the id).
3. We already have a reference to an object when we use useConnectionInfo (the only way to get the connection info), and in JavaScript mostly everything (unless some optimisations) is passed by reference. In this case, there is no additional impact.

If I'm missing something obvious I can change the code obviously, but I don't think it will generate additional subscriptions to anything by just using an object as an argument instead of a string.

gribnoysup: > That's not the case, and using objects as an argument to a function is not a problem.

This is normal for functions, but you are not writing a function, you are writing a hook and hooks have a very spefic use-case: to be called inside React components.

> If you have a hook that has a dependency on the whole object in the dependency array (thanks React for that) yes, it can be problematic, but there is no subscription in this case anywhere

You are ignoring that to get that value to React rendering context to be able to pass it to the hook, it needs to pass through the props first (or get pulled from context), both cases mean that you are binding the component to state of a connection info

> Encapsulation, if the object changes or the hook depends on other fields of ConnectionInfo, you just change the hook

Realistically speaking, do we expect an id on the connecion info object not to be a sole identifier of a connection info? We should have a clear, documented way to identify a connection in the list of others. This is an id right now. Do we have any plans for connection info to be identified by something else? If yes, why would we design a system for ourselves that is poorly adapted to the other parts of the application code?

> We already have a reference to an object when we use useConnectionInfo (the only way to get the connection info), and in JavaScript mostly everything (unless some optimisations) is passed by reference. In this case, there is no additional impact.

Referential equality is currently preserved ONLY in the case of the old, single connection flow that this project will remove. Your multi connection code right now is written like this: connection info changes > code re-reads the list of connections from disk as the only existing way of updating the state > connection info object is completely overriden, you lost your reference (and due to the fact that the stores and hooks are all using local UI React state you have no referential equality between them). Same thing is implied by redux: changes in object require to create a new object reference, and state passed to the UI should be granular. As soon as we move this state to redux where it belongs, you will not have referential equality preserved even if we fix how the connectionInfo is updated. Relying on this is very brittle and will lead to hard-to-debug issues.

kmruiz: Well the React magic on hooks arguments is something that I wasn't aware of and it's pretty unexpected. I appreciate the explanation because it's another design trap from the library.

I'll use the connectionInfoId as a parameter and resolve this comment.

Thanks a lot!

gribnoysup: It's not about the hook function arguments, there is no magic there. To get the value to the hook you need to get it inside the render method from somewhere first. For example in this scenario currently it's the `useConnectionInfo` hook that is "dumb" enough in the current single connection flow implementation that the value set in the application root never changes after connection. But this will not hold up for the multiple connections scenario, it's already not holding up, the code that's currently written doesn't give any referential equality guarantees.

packages/compass-connections/src/hooks/use-connection-status.ts
```diff
@@ -0,0 +1,35 @@
...
+  status: ConnectionStatus;
+} {
+  const connectionManager = useConnectionsManagerContext();
+  const [status, setStatus] = useState<ConnectionStatus>(
+    connectionManager.statusOf(connectionInfoId)
+  );
+
+  const updateStatus = () => {
+    setStatus(connectionManager.statusOf(connectionInfoId));
+  };
```
gribnoysup, line 20: This needs to be wrapped in useCallback, or you should move the definition inside effect. You are recreating this function every render unnecessarily adding and removing subscriptions from the connection manager instead of doing it once

packages/compass-connections/src/hooks/use-connection-status.ts
```diff
@@ -0,0 +1,35 @@
...
+    for (const event of Object.values(ConnectionsManagerEvents)) {
+      connectionManager.on(event, updateStatus);
+    }
+
+    return () => {
+      for (const event of Object.values(ConnectionsManagerEvents)) {
+        connectionManager.off(event, updateStatus);
+      }
+    };
+  }, [updateStatus, connectionInfoId]);
```
gribnoysup, line 32: This doesn't currently have a dependency on the connection info id

gribnoysup (approved):
Please fix the hook, otherwise lgtm

packages/compass-connections/src/hooks/use-connection-status.ts
```diff
@@ -0,0 +1,35 @@
...
+    for (const event of Object.values(ConnectionsManagerEvents)) {
+      connectionManager.on(event, updateStatus);
+    }
+
+    return () => {
+      for (const event of Object.values(ConnectionsManagerEvents)) {
+        connectionManager.off(event, updateStatus);
+      }
+    };
+  }, []);
```
gribnoysup, line 32: You are now missing dependencies here, it should include `connectionManager` and `connectionInfoId` in hooks deps.

PR #5578, opened 2024-03-14
feat: connection form as a modal COMPASS-7709

paula-stacho:
https://jira.mongodb.org/browse/COMPASS-7709

## Description

Adding connection-form as a modal, to be opened from the sidebar.

Hint: Do hide whitespaces for the diff.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
+++ b/packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
@@ -74,7 +74,7 @@ export function MultipleConnectionSidebar({}: MultipleConnectionSidebarProps) {
 
   const onNewConnectionConnect = useCallback(
     (connectionInfo) => {
-      connect({
+      void connect({
         ...cloneDeep(connectionInfo),
       }).then(() => setIsConnectionFormOpen(false));
     },
--- a/packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
+++ b/packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
@@ -73,11 +73,10 @@ export function MultipleConnectionSidebar({}: MultipleConnectionSidebarProps) {
   );
 
   const onNewConnectionConnect = useCallback(
-    async (connectionInfo) => {
-      await connect({
+    (connectionInfo) => {
+      connect({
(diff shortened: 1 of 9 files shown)


packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
```diff
@@ -46,9 +59,35 @@ export function MultipleConnectionSidebar({}: MultipleConnectionSidebarProps) {
...
+  const onNewConnectionToggle = useCallback(
+    (open: boolean) => setIsConnectionFormOpen(open),
+    []
+  );
+
+  const onNewConnectionConnect = useCallback(
+    (connectionInfo) => {
+      connect({
+        ...cloneDeep(connectionInfo),
+      }).then(() => setIsConnectionFormOpen(false));
```
paula-stacho, line 79: we had no error handling in https://github.com/mongodb-js/compass/blob/af95fc323d063384200ada17ff7f80ebdb8df890/packages/compass-connections/src/components/connections.tsx#L245, I assume this is because we'll have the connection progress bar @kmruiz

kmruiz: Yep, we have error handling in the task I'm working on to show the error toast and a way to open the modal on edit mode.

paula-stacho:
> LGTM. There is some scrolling behaviour that we might want to check when we open the modal and the window is relatively small, because at that point we have two scrolls.

> However it's not blocking as it's not in the design how to scale down to smaller screens if necessary. Let's keep it in our mind so we can discuss it with Ben.

Yes, I wasn't sure how to resolve this. Should I create a follow up ticket, or do we have other place where we store such open design questions?

kmruiz:
> Yes, I wasn't sure how to resolve this in a better way. Should I create a follow up ticket, or do we have other place where we store such open design questions?

I think we can cover it on Slack, in the project channel, it's faster to gather feedback from Ben there. If we conclude that we want to handle the modal differently, we can create a follow up ticket.

PR #5592, opened 2024-03-19
fix(compass): set up local mms COMPASS-7767

alenakhineika:
## Description

Follow up to https://github.com/mongodb-js/compass/pull/4941

Add `localhost` to the allowed cloud URLs to authorize requests to MMS running locally.

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass/src/main/application.ts
+++ b/packages/compass/src/main/application.ts
@@ -312,6 +312,7 @@ class CompassApplication {
         '*://cloud.mongodb.com/*',
         '*://cloud-dev.mongodb.com/*',
         '*://compass.mongodb.com/*',
+        '*://localhost/*',
       ],
     };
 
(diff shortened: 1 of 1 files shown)


packages/compass/src/main/application.ts
```diff
@@ -312,6 +312,7 @@ class CompassApplication {
         '*://cloud.mongodb.com/*',
         '*://cloud-dev.mongodb.com/*',
         '*://compass.mongodb.com/*',
+        '*://localhost/*',
```
mabaasit, line 315: probably only for dev mode?

mabaasit: but now that i am think about it, do we want to run (prod) compass with local atlas? if not, then we should only add this URL for development purposes.

alenakhineika: Aligned with the injectCSP logic where we add localhost dynamically.

mabaasit (approved):

PR #5593, opened 2024-03-19
feat: show list of active connections COMPASS-7654

paula-stacho:
https://jira.mongodb.org/browse/COMPASS-7771

## Description

Included:
- adds useActiveConnections hook
- displays list of connections sorted alphabetically
- the connections can be collapsed/expanded

Will handled be in separate subtasks:
- filter (we'll need to extend the existing filter so that it works on connections level)
- it does not work with multiple connections (needs connection scope)
- integrate with nested list of databases/collections (this one is tricky as the current databases menu is calculating it's height with `react-virtualized-auto-sizer` and `react-window`)

Not included in this task:
- the 3 dot menu

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-sidebar/src/components/multiple-connections/icons/server-icon.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import { css, palette, useDarkMode } from '@mongodb-js/compass-components';
+
+const iconStyles = css({
+  flex: 'none',
+});
+
+const ServerIcon = () => {
```
paula-stacho, line 8: This is a copy of the icon in saved-connections. We should be able to re-use more of the icons between the two, but I didn't want to create conflicts between the PRs, so I'd rather address this separately.

packages/compass-sidebar/src/components/multiple-connections/active-connections/active-connection-list.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+const activeConnectionsListStyles = css({
+  listStyle: 'none',
+  height: '100%',
+});
+
+export function ActiveConnectionList({
+  testUseActiveConnections,
+}: {
+  testUseActiveConnections?: () => ConnectionInfo[];
```
paula-stacho, line 42: I know this is unusual, but I really wanted to avoid pulling too much of the connection logic into the component test.

packages/compass-sidebar/src/components/multiple-connections/active-connections/active-connection.tsx
```diff
@@ -0,0 +1,144 @@
...
+});
+
+const toggleBtnStyles = css({
+  color: palette.gray.dark2,
+  backgroundColor: 'none',
+  boxShadow: 'none',
+  '&:hover': {
+    backgroundColor: 'none',
+    boxShadow: 'none',
+  },
```
paula-stacho, line 24: Here I'm trying my best to suppress the background/shadow/something on LG buttons - it is not sufficient. Ideas welcome, otherwise I'll come back to this within https://jira.mongodb.org/browse/COMPASS-7773

packages/compass-sidebar/src/components/multiple-connections/active-connections/active-connection-list.tsx
gribnoysup: We have a pattern for using `__TEST_<something>` props for passing values to the components that are otherwise impossible to mock, but as you are dealing with services here, the injection mechanism for them is a provider, and I'd suggest to follow this approach and set up providers with mock services that return values you need for that purpose

packages/compass-sidebar/src/components/multiple-connections/active-connections/active-connection-list.tsx
```diff
@@ -0,0 +1,108 @@
...
+          <ActiveConnection
+            key={connection.id}
+            connection={connection}
+            isExpanded={!collapsed.includes(connection.id)}
+            onToggle={(isExpanded: boolean) =>
+              onConnectionToggle(connection.id, isExpanded)
+            }
+          />
+        ))}
+      </ul>
```
gribnoysup, line 105: This is already better than the current saved connections list as your ActiveConnection component is following some of the accessibility guidelines, but this should be using our navigation tree component instead: you are basically just partially re-implementing some parts of it instead of making it work here. Is this temporary and we are planning to replace this with our navigation tree later or what's the plan here? Asking because once again we seem to add a lot of code that we'll just throw away later and have to rewrite.

packages/compass-connections/src/stores/active-connections.ts
```diff
@@ -0,0 +1,40 @@
...
+      ...(await connectionRepository.listNonFavoriteConnections()),
+    ].filter(({ id }) => connectionManager.statusOf(id) === 'connected');
+    setActiveConnections(list);
+  }, []);
+
+  useEffect(() => {
+    void updateList();
+
+    for (const event of Object.values(ConnectionsManagerEvents)) {
+      connectionManager.on(event, () => void updateList());
```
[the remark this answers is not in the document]
paula-stacho: good point, I didn't think about that!

packages/compass-sidebar/src/components/multiple-connections/active-connections/active-connection-list.tsx
paula-stacho: Yeah actually I planned to talk to you about this - originally I thought we could just implement this top level list and link the nested database lists, but I've realised this won't be that simple and we might need to redo the `compass-databases-navigation`, or extend it to include the new highest level. For this purpose I separated https://jira.mongodb.org/browse/COMPASS-7773.

packages/compass-connections/src/stores/active-connections.ts
```diff
@@ -0,0 +1,40 @@
...
+    []
+  );
+
+  const updateList = useCallback(async () => {
+    const list = [
+      ...(await connectionRepository.listFavoriteConnections()),
+      ...(await connectionRepository.listNonFavoriteConnections()),
+    ].filter(({ id }) => connectionManager.statusOf(id) === 'connected');
+    setActiveConnections(list);
+  }, []);
```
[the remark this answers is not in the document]
paula-stacho: useConnections as it is is already doing a lot of things and this would make it even more complex, our approach is to have smaller hooks, @kmruiz is doing something similar with the `useConnectionStatus`.
 the point about the list updating more often than necessary is valid though. perhaps I could include a check that compares the arrays before updating the state?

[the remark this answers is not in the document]
gribnoysup: Right, I forgot that useConnections is not actually a hook that gives you connections, but all the connections, local UI state, and the connect logic, you're correct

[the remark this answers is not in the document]
gribnoysup: To your second point: yes, We'd need to do some sort of deep compare here to avoid unnecessary re-renderes. The way it is currently implmented where connections are re-read from disk every time they are updated we can't really rely on referential equality a-la redux for those kinds of checks.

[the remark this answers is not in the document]
kmruiz: Another option is to render the child components always, but display them only when they are active. It would work essentially as a virtualised list and we would isolate rendering reasons.

1. When you change the list of connections, it will just change the amount of list elements. If the keys are the same, this is really cheap.
2. When you change the status of the connection it will only rerender the list element for that connection.

You could do that by just using useConnectionStatus in the child component and using useConnections in the parent (or having a hook that does that independently but is reactive to changes on the storage like this https://github.com/mongodb-js/compass/pull/5584/files#diff-5f0d1b5898d50c4399d08b4b53a9f4d06a8e65aacdc585183dc78fa3cd418b65R301

packages/compass-sidebar/src/components/multiple-connections/active-connections/active-connection-list.tsx
gribnoysup: Sure, we can chat about this, feel free to set up a meeting or reach out in Slack. Some code that gives you an exact container size (like autosizer library does) is needed every time you are doing any sort of virtualization like this in a container of variable size, it's not in any way library dependent.

It's good to know that we have a follow up ticket, but I'm still confused that we are doing more work that we will be immediately re-doing.

packages/compass-connections/src/stores/active-connections.ts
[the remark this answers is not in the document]
gribnoysup: As always with delegating rendering logic to the child components themselves, you are essetially removing the ability to rely on React component lifecycles, this always requires more work to work consistently in the long run. Especially in this case where the rendering will be delegated eventually to another component, this would definitely lead to conflicts in which component owns the rendering. I would not pursue it.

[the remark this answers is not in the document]
paula-stacho: I'll go with the deep comparison, the ConnectionInfo isn't that complex and this would only happen if the connections or connection status changes. Apart of the lifecycle point, the object references not being the same might still cause unnecessary re-renders down the line.

kmruiz (approved):

packages/compass-connections/src/stores/active-connections.ts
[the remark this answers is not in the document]
gribnoysup: Connection info options can contain bson values, so I'd double-check that those are not tripping the comparison. There are some helpers in the bson packages to make the deep eq check IIRC

packages/compass-connections/src/stores/active-connections.ts
```diff
@@ -0,0 +1,76 @@
...
+  return isEqual(
+    listA.map((a: any) => BSON.serialize(a)),
+    listB.map((b: any) => BSON.serialize(b))
+  );
+}
+
+export function useActiveConnections(): ConnectionInfo[] {
+  // TODO(COMPASS-7397): services should not be used directly in render method,
+  // when this code is refactored to use the hadron plugin interface, storage
+  // should be handled through the plugin activation lifecycle
```
paula-stacho, line 31: copied this TODO from connections-store, so that we don't miss this in the refactor

packages/compass-connections/src/stores/active-connections.ts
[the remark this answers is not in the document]
paula-stacho: added connectionStorage listener bellow

packages/compass-sidebar/src/components/multiple-connections/active-connections/active-connection-list.tsx
paula-stacho: updated the test

packages/compass-connections/src/stores/active-connections.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const updateList = useCallback(async () => {
+    const list = [
+      ...(await connectionRepository.listFavoriteConnections()),
+      ...(await connectionRepository.listNonFavoriteConnections()),
+    ].filter(({ id }) => connectionManager.statusOf(id) === 'connected');
+    if (!areConnectionsEqual(activeConnections, list)) {
+      setActiveConnections(list);
+    }
+  }, [activeConnections]);
```
gribnoysup, line 48: When you're dealing with React setState that is derived from the previous state, the guideline is to use the setState callback instead of current value from the render scope, this guarantees you are avoiding any chance of running your checks against the stale value (there are also actual hook deps missing here, even though they are stable)

```suggestion
  const updateList = useCallback(async () => {
    const newList = [
      ...(await connectionRepository.listFavoriteConnections()),
      ...(await connectionRepository.listNonFavoriteConnections()),
    ].filter(({ id }) => connectionManager.statusOf(id) === 'connected');
    setActiveConnections(prevList => {
      return areConnectionsEqual(prevList, newList) ? prevList : newList;
    });
  }, [connectionRepository, connectionManager]);
```

packages/compass-connections/src/stores/active-connections.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    // reacting to connection status updates
+    for (const event of Object.values(ConnectionsManagerEvents)) {
+      connectionManager.on(event, () => void updateList());
+    }
+
+    // reacting to connection info updates
+    connectionStorage.events?.on(
+      ConnectionStorageEvents.ConnectionsChanged,
+      () => void updateList
```
gribnoysup, line 61: This should be called, not just return a function, right?

```suggestion
      () => void updateList()
```

packages/compass-connections/src/stores/active-connections.ts
```diff
@@ -0,0 +1,76 @@
...
+    );
+
+    return () => {
+      for (const event of Object.values(ConnectionsManagerEvents)) {
+        connectionManager.off(event, () => void updateList());
+      }
+      connectionStorage.events?.off(
+        ConnectionStorageEvents.ConnectionsChanged,
+        () => void updateList
```
gribnoysup, line 70: Ditto

```suggestion
        () => void updateList()
```

packages/compass-connections/src/stores/active-connections.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    // reacting to connection info updates
+    connectionStorage.events?.on(
+      ConnectionStorageEvents.ConnectionsChanged,
+      () => void updateList
+    );
+
+    return () => {
+      for (const event of Object.values(ConnectionsManagerEvents)) {
+        connectionManager.off(event, () => void updateList());
```
gribnoysup, line 66: You are trying to remove a subscription for the function that was just created, this will do nothing. If you are wrapping your functions, they need to be defined first (can be inside effect scope), so that you can correctly clean up

packages/compass-connections/src/stores/active-connections.ts
```diff
@@ -0,0 +1,76 @@
...
+      }
+      connectionStorage.events?.off(
+        ConnectionStorageEvents.ConnectionsChanged,
+        () => void updateList
+      );
+    };
+  }, [updateList]);
```
gribnoysup, line 73: Missing deps

```suggestion
  }, [connectionManager, connectionStorage, updateList]);
```

packages/compass-connections/src/stores/active-connections.ts
paula-stacho: ah, thanks! not the first time I've done this mistake when adding the `() => void` part 🤦

package-lock.json
gribnoysup, line 1: It's a tiiiny bit suspicious to see a 800 loc change to a package-lock file in a PR that only adds 3 deps that are already used in the monorepo. Do you mind checking out package-lock from main and re-running npm install just to comfirm that it's actually how it should be?

```
git checkout origin/main -- package-lock.json
npm install
```

paula-stacho: makes sense, will do!

paula-stacho: hmm, now the diff has changed but it's not really smaller 🤔 can it be I have wrong version of node/npm?

gribnoysup: You haven't merged the latest main changes in, right? We just removed a package and your branch still has it so the diff if weird. To be more specific, when I mentioned that this should help resolve merge conflict, what I meant is:

- Update your refs
- Start merge with main
- Checkout package-lock from origin
- Run npm install to update the lock
- Stage the file and finish merge

```
git fetch
git merge origin/main
git checkout origin/main -- package-lock.json
npm install
git add package-lock.json
git commit
```

I did this locally and seeing a 3 line diff in package-lock in this branch compared to main, which looks more reasonable for the amount of changes in package.json file you have here 🙂

packages/compass-connections/src/stores/active-connections.ts
paula-stacho: thank you!

package-lock.json
paula-stacho: Oh right, I checkout the file and immediately rewrote the changes with install. evening brain 🤦

addaleax (approved):

PR #5608, opened 2024-03-21
fix(connections): do not save the connection string on connect COMPASS-7751

kmruiz:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-connections/src/stores/connections-store.ts
+++ b/packages/compass-connections/src/stores/connections-store.ts
@@ -272,7 +272,10 @@ export function useConnections({
   recentConnections: ConnectionInfo[];
   favoriteConnections: ConnectionInfo[];
   cancelConnectionAttempt: (connectionInfoId: string) => Promise<void>;
-  connect: (connectionInfo: ConnectionInfo) => Promise<void>;
+  connect: (
+    connectionInfo: ConnectionInfo,
+    shouldSaveConnectionInfo?: boolean
+  ) => Promise<void>;
   createNewConnection: () => void;
   saveConnection: (connectionInfo: ConnectionInfo) => Promise<void>;
   setActiveConnectionById: (newConnectionId: string) => void;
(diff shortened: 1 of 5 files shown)


packages/connection-storage/src/connection-repository.spec.ts
```diff
@@ -41,7 +41,7 @@ function mockStorage(): StorageContext {
   return mockStorageWithConnections([]);
 }
 
-describe('CompassConnectionProvider', function () {
+describe.only('CompassConnectionRepository', function () {
```
himanshusinghs, line 44:
```suggestion
describe('CompassConnectionRepository', function () {
```

paula-stacho:
Last used is a legacy thing, or? Perhaps we could name it such?

packages/compass-connections/src/stores/connections-store.ts
```diff
@@ -272,7 +272,10 @@ export function useConnections({
   recentConnections: ConnectionInfo[];
   favoriteConnections: ConnectionInfo[];
   cancelConnectionAttempt: (connectionInfoId: string) => Promise<void>;
-  connect: (connectionInfo: ConnectionInfo) => Promise<void>;
```
paula-stacho, line 275: it's a bit unsettling that TS didn't complain about the actual implementation having two parameters, just because the second was optional 🤔

addaleax: That's kind of just how TS works, a function that takes 1 required parameter + 1 optional parameter can be interacted with in the same way as a function that only takes 1 required parameter, so there's no way for TS to tell that this wasn't intentional

addaleax (approved):

packages/compass-connections/src/stores/connections-store.ts
paula-stacho: Yes I get the reasoning. I was wondering if there is some other way this could've been defined so that the mistake would've been caught earlier. But I guess the only way would've been to make the parameter required, which doesn't make sense in this use case.

paula-stacho (approved):

PR #5613, opened 2024-03-22
chore(ci): use the same node and npm version as we do in EVG ci

himanshusinghs:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/.github/workflows/authors-and-third-party-notices.yaml
+++ b/.github/workflows/authors-and-third-party-notices.yaml
@@ -23,14 +23,14 @@ jobs:
           fetch-depth: '0'
 
 
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
-          node-version: ^16.20.2
+          node-version: ^18.19.1
           cache: 'npm'
 
-      - name: Install npm@8
+      - name: Install npm@10.2.4
         run: |
-          npm install -g npm@8
+          npm install -g npm@10.2.4
 
       - name: Install Dependencies
         run: |
--- a/.github/workflows/authors-and-third-party-notices.yaml
+++ b/.github/workflows/authors-and-third-party-notices.yaml
@@ -25,7 +25,7 @@ jobs:
 
       - uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/authors-and-third-party-notices.yaml
```diff
@@ -23,14 +23,14 @@ jobs:
           fetch-depth: '0'
 
 
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
-          node-version: ^16.20.2
+          node-version: ^18.19.1
```
gribnoysup, line 28: We use exact, not range

himanshusinghs: Ahh wow, I didn't check that. Thanks - I will check and fix.

lerouxb: wasn't it `^16.20.2` before which is a range?

gribnoysup: It's not a range in evergreen

.github/workflows/bump-packages.yaml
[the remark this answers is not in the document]
himanshusinghs: Noice! Thanks for the pointer, I just pushed an update. Didn't look deeper if the params have changed but if I see the action failing. I will fix it.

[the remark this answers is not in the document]
gribnoysup: Thanks!

gribnoysup (approved):

himanshusinghs:
Merging now since all the GH checks went fine. The EVG checks are not related.

.github/workflows/authors-and-third-party-notices.yaml
lerouxb: Oh you mean in evergreen. Forgot this is to make it the same as there and not just to bump it. Carry on, then ;)

PR #5615, opened 2024-03-22
Compass 7773

paula-stacho:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-connections-navigation/src/connection-item.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+const iconStyles = css({
+  flex: 'none',
+});
+
+const expanded = css({
+  transform: 'rotate(90deg)',
+});
+
+const ExpandButton: React.FunctionComponent<{
```
gribnoysup, line 56: This seems to exactly copy the same code in database item. Can we move this out to the tree-item so that we don't need to maintain the same code in two places here?

paula-stacho:
Note: I made the namespace parameter for databases workspace required, forcing a bunch of `namespace: ''` placeholders. I have since started to doubt this decision, especially knowing that DE will be working with single connections for the time being. I'll probably change it to optional.

packages/compass-workspaces/src/provider.tsx
```diff
@@ -145,9 +149,9 @@ export const WorkspacesServiceProvider: React.FunctionComponent<{
           openWorkspaceAction({ type: 'My Queries' }, tabOptions)
         );
       },
-      openDatabasesWorkspace: (tabOptions) => {
+      openDatabasesWorkspace: (namespace, tabOptions) => {
         return store.dispatch(
-          openWorkspaceAction({ type: 'Databases' }, tabOptions)
+          openWorkspaceAction({ type: 'Databases', namespace }, tabOptions)
```
gribnoysup, line 154: I'm not sure how this makes sense. Namespace can't be applied to "Databases" workspace, this is an instance connection level workspace, not db / coll level

paula-stacho: I thought it would make sense for namespaces in the future to include the connection level as well. This is of course not implemented here, I only added this bit.
'connection1' is an item in the sidebar and will be highlighted when the user is on the list of 'connection1'. and 'db1.collection1' will be insufficient for multiconnection sidebar to correctly recognize which namespace to highlight, so the most straightforward way to me seemed to change namespaces so that they start from the connection level. I might be missing some context on what a namespace means outside of the sidebar.

gribnoysup: `namespace` has a defined meaning, it's specifically <db>.<coll> for Compass and generally when people talk about namespaces in MongoDB ([see glossary](https://www.mongodb.com/docs/manual/reference/glossary/)), we shouldn't mix connection id into the namespace

gribnoysup: If you want to start adding connection ids to the tab metadata in this PR, feel free to, but this should be a separate property on the tab options and added to all of them

paula-stacho: Oh I didn't know that, thank you! 

@kmruiz We were saying for the sidebar store that we could just extend the ids to include connection. We could still do it, but we'll also need to distinguish `namespace` and `connectionId` and send both in the actions:
```typescript
onNamespaceAction(
  evt.currentTarget.dataset.id as string,
  'select-connection'
);
```

paula-stacho: I removed the connection from namespace and left some TODOs for COMPASS-7775 to add activeConnection, but it can be also a separate task, or part of reworking the workspaces - what do you think @kmruiz ?

gribnoysup: In the sidebar list items it can be whatever you need it to be, this is just catching a DOM event with an id set on an html node and sending it up to be handled. But something inside the sidebar code should parse it to parts and provide further as two distinct values: connectionId and namespace. This is an implementation detail that should not affect the whole application

paula-stacho: Absolutely, as I said the `id` can still be combined. The part that needs to change is that `id` is not sent as `namespace`, it needs to be clear that the two are not interchangeable. Since we need to support both versions for a while, I'd separate them from the beginning to avoid confusion. Either having the `namespace` and `connectionId` as their own parameters, or use a different separator - e.g. `id: connection_db.coll`

kmruiz: Currently the ConnectionInfo in single connection is global and can be acquired by using `useConnectionInfo`, and in multiple connections, we have the list of connections as parameters of the hook. I think we could adapt the store to always use a connectionId and relatively easily use the connectionInfo.id from whenever it comes as a separate parameter, as in the patch I shared, and we might not need to change the namespace. After taking a look, I'm leaning more to have a separate parameter for the connectionId, not changing the namespace (because we would overload it semantically, it would be either `db.coll` or `conn.db.coll`) and making the connectionId mandatory in the store in the form of `Record<ConnectionInfo['id'], whatever>`.

paula-stacho: Yes, changing the namespace apparently isn't an option. I have already reverted the changes. We could still have combined ids -> `db.coll` (collection item id in single connection usage) and `conn.db` (db item id in multi connection usage). Alternatively we could rename the ids to namespace, as they're not really identifiers for the menu item if they only contain `db.coll` level.

packages/compass-connections-navigation/src/constants.tsx
```diff
@@ -0,0 +1,21 @@
+import { spacing } from '@mongodb-js/compass-components';
+
+// TODO: Currently we show placeholder for every collection/database item in the list, but
+// do we want to / need to?
```
gribnoysup, line 4: This was a deliberate design decision to prevent collection list and the content below it from "jumping" when collections finished loading. Why we'd want to remove that?

packages/compass-connections-navigation/src/database-item.tsx
```diff
@@ -92,18 +48,25 @@ export const DatabaseItem: React.FunctionComponent<
 > = ({
   id,
   name,
+  level,
   posInSet,
   setSize,
   isExpanded,
   isActive,
   isReadOnly,
+  isLegacy,
```
gribnoysup, line 57: Maybe there is a better name for this taking into account that we will need to support this in shared code path for a single connection in compass-web use-case

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -0,0 +1,541 @@
...
+
+const navigationTree = css({
+  flex: '1 0 auto',
+});
+
+const ConnectionsNavigationTree: React.FunctionComponent<{
+  connections?: Connection[];
+  databasesLegacy?: Database[];
+  expanded?: Record<string, false | Record<string, boolean>>;
+  expandedLegacy?: Record<string, boolean>;
```
gribnoysup, line 343: This feels like an unnecessary fork in logic here. We will need to support single connection in compass-web with the new sidebar: the logic is pretty straightforward, if there is only one connection and only one collection is allowed, this tree should render as before, can we just incorporate this logic into the component and not have a hard fork on this logic based on props? It's low effort and less cleaning-up afterwards when the multiple connections are enabled in Compass

packages/compass-connections-navigation/src/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+    case 3:
+      paddingLeft = spacing[3] + spacing[5];
+      break;
+  }
+
+  if (isPlaceholder && (level === 1 || (!isLegacy && level === 2))) {
+    paddingLeft += spacing[2];
+  }
+
+  return css({ paddingLeft });
```
gribnoysup, line 34: We should avoid `css` calls in render, this causes class name re-generation and style tag re-insertion on every render. If we have highly dynamic styles like that, we should use style prop instead, or pre-defined actual classNames outside of function scope and just return generated class names here

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
kmruiz: I agree, I'll do the change when I take this ticket.

paula-stacho: not sure if I follow here.. is the idea just to keep the `databases` name and `expanded` type will depend on whether `connections` or `databases` are set? we can do that for sure, it just doesn't seem like a "fork in logic" for me, it's really just the props layer. that's why I'm not sure if I understand correctly.
when I started on this I wasn't aware of the compass-web single connection scenario, so I thought the cleanup would be easiest if we clearly mark the legacy bits.

paula-stacho: also relevant note - the compass-web will likely keep using a separate sidebar because the new sidebar structure doesn't make sense for single connection. whether we call it legacy or single-connection (likely the latter, but I'm afraid to rename the sidebar folder at this moment, there are too many conflicts with the main branch already and at least one rebase last week apparently went wrong 🙈  )

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -0,0 +1,541 @@
...
+        .map((database, databaseIndex) =>
+          databaseToItems({
+            database,
+            databaseIndex,
+            databasesLength: databasesLegacy?.length || 0,
+            expanded: expandedLegacy,
+            level: 1,
+          })
+        )
+        .flat();
```
addaleax, line 409: Not a huge difference, but just fyi, there's also `.flatMap()` which does `.map().flat()` as a single call :)

packages/compass-connections-navigation/src/placeholder-item.tsx
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable react/prop-types */
```
addaleax, line 1: I've seen this disable directive in a couple of files here, it doesn't seem to trigger anything locally though?

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
paula-stacho: update: I've changed the props and also internally removed the legacy notation in favour of single connection (SC)

packages/compass-connections-navigation/src/connection-item.tsx
```diff
@@ -0,0 +1,165 @@
...
+  onNamespaceAction,
+  onConnectionExpand,
+  onConnectionSelect,
+}) => {
+  const [hoverProps, isHovered] = useHoverState();
+
+  const isLocalhost =
+    connectionInfo.connectionOptions.connectionString.startsWith(
+      'mongodb://localhost'
+    );
```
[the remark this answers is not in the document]
paula-stacho: yes a follow up sounds good. I'll create one
cc @kmruiz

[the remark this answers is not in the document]
paula-stacho: https://jira.mongodb.org/browse/COMPASS-7832

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -0,0 +1,541 @@
...
+  name: string;
+  collectionsStatus: string;
+  collectionsLength: number;
+  collections: Collection[];
+};
+
+export type Connection = {
+  connectionInfo: ConnectionInfo;
+  name: string;
+  databasesStatus: string;
```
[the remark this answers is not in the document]
paula-stacho: absolutely. I've added a string union type as I'm not sure where exactly this is originating and I don't want to grow the scope of this PR any more 😅

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
paula-stacho: neat!

packages/compass-connections-navigation/src/constants.tsx
paula-stacho: sorry, what did we want to revert here? or is it reverted already? @gribnoysup @kmruiz

paula-stacho: note: this file is moved from `compass-databases-navigation`

packages/compass-connections-navigation/src/placeholder-item.tsx
paula-stacho: yup, might not be needed anymore. I'll try removing it and see if it triggers anything in the CI

packages/compass-connections-navigation/src/constants.tsx
kmruiz: I think nothing needs to be reverted, looking at what the code does, it still keeps the placeholders as expected.

kmruiz (approved):

PR #5628, opened 2024-03-27
fix(ci): give the create pull request step an id so subsequent steps can use it

lerouxb:
no idea how this worked before: https://github.com/mongodb-js/compass/blob/56707e8fcab13b1c99e89650ae8ba97900f95574/.github/workflows/authors-and-third-party-notices.yaml#L60


addaleax:
https://github.com/mongodb-js/compass/pull/5627?

PR #5630, opened 2024-03-28
chore(data-service): log maxTimeMS along with find(), aggregate() and estimatedCount(), always set a default maxTimeMS COMPASS-7747

lerouxb:
So we can prove that DataService did get these values and (presumably) passed them on.

```
{"t":{"$date":"2024-03-28T10:21:50.952Z"},"s":"I","c":"COMPASS-DATA-SERVICE","id":1001000060,"ctx":"Connection 0","msg":"Running find","attr":{"ns":"config.collections","maxTimeMS":60000}}
```

```
{"t":{"$date":"2024-03-28T10:21:52.211Z"},"s":"I","c":"COMPASS-DATA-SERVICE","id":1001000181,"ctx":"Connection 0","msg":"Running aggregate","attr":{"ns":"Corporate-Registrations.corporations","stages":["$match","$count"],"maxTimeMS":5000}}
```

and now (we didn't specify this maxTimeMS before):

```
{"t":{"$date":"2024-03-28T10:54:30.824Z"},"s":"I","c":"COMPASS-DATA-SERVICE","id":1001000034,"ctx":"Connection 0","msg":"Running estimatedCount","attr":{"ns":"timeseries.random","maxTimeMS":5000}}
```

### DIFF
--- a/packages/compass-aggregations/src/modules/input-documents.ts
+++ b/packages/compass-aggregations/src/modules/input-documents.ts
@@ -122,14 +122,23 @@ export const refreshInputDocuments = (): PipelineBuilderThunkAction<
         | undefined,
     };
 
+    // maxTimeMS defaults to null here because the aggregation options field's
+    // maxTimeMS defaults to empty and the preference defaults to undefined.
+    // We need a timeout on count because for timeseries estimatedCount() seems
+    // to just do a colscan and we need that timeout to be low compared to the
+    // aggregation one anyway. For aggregations it is less critical because we
+    // $limit to the first few records.
+    const countOptions = { ...options, maxTimeMS: 5000 };
+    const aggregateOptions = { ...options };
+
     const exampleDocumentsPipeline = [{ $limit: sampleSize }];
 
     dispatch(loadingInputDocuments());
 
     try {
       const data = await Promise.allSettled([
-        dataService.estimatedCount(ns, options),
-        dataService.aggregate(ns, exampleDocumentsPipeline, options),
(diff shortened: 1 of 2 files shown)


addaleax (approved):

packages/compass-aggregations/src/modules/input-documents.ts
```diff
@@ -122,14 +122,23 @@ export const refreshInputDocuments = (): PipelineBuilderThunkAction<
         | undefined,
     };
 
+    // maxTimeMS defaults to null here because the aggregation options field's
+    // maxTimeMS defaults to empty and the preference defaults to undefined.
+    // We need a timeout on count because for timeseries estimatedCount() seems
+    // to just do a colscan and we need that timeout to be low compared to the
+    // aggregation one anyway. For aggregations it is less critical because we
+    // $limit to the first few records.
+    const countOptions = { ...options, maxTimeMS: 5000 };
```
lerouxb, line 131: I thought about removing this again now that I also changed the method in DataService to fill it in if it isn't set, but I think the comment is useful documentation and it might still be a good idea to decouple this timeout from the hardcoded default in DataService. Yes right now it is the same, but now it is easy for someone to change the value we use in aggregations.

lerouxb: Oh and it also means that we print a value for maxTimeMS in the logs rather than null so no one will get a heart attack.

lerouxb: https://jira.mongodb.org/browse/COMPASS-7798

PR #5635, opened 2024-03-28
chore(compass-generative-ai): add request id to telemetry and ai API requests COMPASS-7793

Anemy:
COMPASS-7793

Adds a `request_id` url parameter to our generative AI API requests. This parameter currently unused in Atlas, however will be useful for us tracking telemetry and may have use in cloud with https://jira.mongodb.org/browse/CLOUDP-242020


packages/compass-aggregations/package.json
```diff
@@ -96,7 +96,8 @@
     "react-redux": "^8.1.3",
     "redux": "^4.2.1",
     "redux-thunk": "^2.4.2",
-    "semver": "^5.7.2"
+    "semver": "^5.7.2",
+    "uuid": "^8.3.2"
```
addaleax, line 100: Would we potentially want to move away from `uuid` as a separate dependency and use `new UUID().toString()` from `bson` instead? We currently have a couple of uses of the latter, and one usage of `uuid` in compass-connections

addaleax (approved):

packages/compass-aggregations/package.json
Anemy: Good thinking! I'll do a follow up pr and remove it fully from the code base.

PR #5567, opened 2024-03-12
feat(compass-app-stores): makes CompassInstanceStore initialisation based on the possibility of multiple connections COMPASS-7721

himanshusinghs:
## Description
This PR modifies the CompassInstanceStore plugin such that it can instantiate and provide different MongoDBInstances for multiple connected connections.

Please use "Hide Whitespaces" while reviewing, would be much easier.

### Checklist
- [ ] New tests and/or benchmarks are included
- [ ] Documentation is changed or added
- [ ] I have signed the MongoDB Contributor License Agreement (https://www.mongodb.com/legal/contributor-agreement)

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*
- [ ] Patch (non-breaking change which fixes an issue)
- [ ] Minor (non-breaking change which adds functionality)
- [ ] Major (fix or feature that would cause existing functionality to change)

PR #5644, opened 2024-04-02
feat(compass-field-store): makes FieldStorePlugin aware of multiple connections and remove dependency on events for updating the store COMPASS-7722

himanshusinghs:
## Description

This PR:
- Makes the FieldStorePlugin's store aware of multiple connections. Essentially we are nesting the existing store one level under the connectionInfoId.
- introduces `FieldStoreService` and its locator `fieldStoreServiceLocator` for other plugins to update the store on changes in documents/schema, directly using the service. This change removes the FieldStore plugin's dependency on globalAppRegistry events.
- updates compass-crud and compass-schema plugin to use the `FieldStoreService` for updating the store.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*
This replaces the events being listened to in FieldStorePlugin. Other
plugins that were earlier depending on events to update field store
should now use the FieldStoreService to do the same.


himanshusinghs:
Evergreen Patch [here](https://spruce.mongodb.com/version/660bf1c6ceb6880007189d9b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) until I target this to main.

kmruiz (approved):

packages/compass-field-store/src/stores/context.ts
```diff
@@ -1,29 +1,104 @@
...
+  dispatch: ReturnType<typeof useDispatch>,
+  connectionInfoAccess: ConnectionInfoAccess
+): FieldStoreService {
+  return {
+    async updateFieldsFromDocuments(
+      ns: string,
+      documents: Record<string, any>[]
+    ) {
+      try {
+        const { fields } = await parseSchema(documents);
```
gribnoysup, line 47: This probably should be somehow part of the store logic, a separate action, not this service. In the current implementation you basically have two different ways of modifying state of the store, but they are split between two places in the code, this just makes it a bit harder to discover for people who will not be aware of this detail

packages/compass-field-store/src/stores/store.ts
```diff
@@ -1,57 +1,14 @@
...
+import reducer from '../modules';
+import type { ActivateHelpers } from 'hadron-app-registry';
 import { FieldStoreContext } from './context';
 
 export function activatePlugin(
   _initialProps: unknown,
-  { globalAppRegistry }: { globalAppRegistry: AppRegistry },
-  { on, cleanup }: ActivateHelpers
+  _: unknown,
+  { cleanup }: ActivateHelpers
```
gribnoysup, line 9: You don't need to pass cleanup here if you are not setting up any subscriptions, so you can clean this up a bit if you want (but also see my other comment, you might want to set something here listening to connection manager)

packages/compass-field-store/src/stores/store.ts
```diff
@@ -1,57 +1,14 @@
...
-  on(
-    globalAppRegistry,
-    'schema-analyzed',
-    ({ ns, schema }: { ns: string; schema: Schema }) => {
-      store.dispatch(changeFields(ns, schema.fields));
-      emitFieldsChanged(ns);
-    }
-  );
-
   return { store, deactivate: cleanup, context: FieldStoreContext };
```
gribnoysup, line 13: Now that this plugin is hoisted outside of the "connected" tree, would it make sense to set up something to clean up the store when disconnecting? Otherwise we might be accumulating too much data in memory that is not being used otherwise

packages/compass-crud/src/stores/crud-store.spec.ts
```diff
@@ -190,6 +196,7 @@ describe('store', function () {
           logger: createNoopLoggerAndTelemetry(),
           favoriteQueryStorageAccess: compassFavoriteQueryStorageAccess,
           recentQueryStorageAccess: compassRecentQueryStorageAccess,
+          fieldStoreService: mockFieldStoreService,
```
gribnoysup, line 199: If we're not testing it here, maybe we can provide some fallbacks for test env so that we don't need to override it (no strong preference for this though)

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -1417,6 +1426,11 @@ class CrudStoreImpl
         multiple: false,
         docs: [doc],
       };
+      void this.fieldStoreService.updateFieldsFromDocuments(this.state.ns, [
+        doc,
+      ]);
+      // Need to keep it around until compass-app-stores stop depending on it to
+      // update the instance
```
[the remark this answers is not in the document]
himanshusinghs: Not really but I will create one right away.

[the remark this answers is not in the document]
himanshusinghs: Ohh sorry forgot to clarify, the app-store was already converted to support multiple connections in [this PR](https://github.com/mongodb-js/compass/pull/5567) but it still depend on these events. I created a different ticket to remove dependency on those events here - https://jira.mongodb.org/browse/COMPASS-7815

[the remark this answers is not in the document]
gribnoysup: Ah, right, I forgot we did the conversion already 🙈 Thanks for opening another ticket!

packages/compass-crud/src/stores/crud-store.spec.ts
himanshusinghs: We are calling the plugin activate directly here so the locator does not come into play for this part which is why I think we would still need to pass the mocked service.

gribnoysup: Ah, didn't notice that, then it's kinda unavoidable, yeah

gribnoysup (approved):

PR #5645, opened 2024-04-02
feat(sidebar): remove collapsing/expanding the sidebar COMPASS-7812

lerouxb:
This removes the ability to collapse the sidebar as we discussed. The minimum width right now is (still) 210px, but we can change that. There's also a maximum of 600px for the moment. A follow-up might tweak that because it steals width from the main workspace.


packages/compass-components/src/components/resizeable-sidebar.tsx
```diff
@@ -58,30 +58,20 @@ const containerStylesLight = css({
...
 const ResizableSidebar = ({
-  collapsable = false,
-  expanded = true,
-  setExpanded = () => {
-    return;
-  },
   initialWidth = defaultSidebarWidth,
   minWidth = 210,
-  collapsedWidth = 48,
+  // TODO: maxWidth
```
lerouxb, line 63: actually I'll just remove this line before merging. But I'm leaving CI to run for the moment

PR #5651, opened 2024-04-04
feat(generative-ai): add passing of sample documents with setting COMPASS-7758

Anemy:
COMPASS-7758

Adds a new setting `enableGenAISampleDocumentPassing` for Compass' ai query and aggregation generation. When enabled (off by default), it the passes sample documents to the endpoint to improve results. 

Added one test to our automated accuracy tests that ensures the sample docs are used in the generation of results.

Increased the size cap of the items we send to the ai since the token validation happens on the backend and we should air on the side of not validating on the frontend.

Opened as a draft as this is adding a setting and we're refactoring the settings a good bit in the ai settings pr, so this will land after that as that's already approved.

### DIFF
--- a/packages/compass-generative-ai/scripts/ai-accuracy-tests.ts
+++ b/packages/compass-generative-ai/scripts/ai-accuracy-tests.ts
@@ -196,11 +196,13 @@ const generateMQL = async ({
   databaseName,
   collectionName,
   userInput,
+  includeSampleDocuments,
 }: {
   type: string;
   databaseName: string;
   collectionName: string;
   userInput: string;
+  includeSampleDocuments?: boolean;
 }) => {
   const collection = mongoClient.db(databaseName).collection(collectionName);
   const schema = await getSimplifiedSchema(collection.find());
@@ -211,7 +213,8 @@ const generateMQL = async ({
       schema: schema,
       collectionName,
       databaseName,
-      sampleDocuments: USE_SAMPLE_DOCS ? sample : undefined,
+      sampleDocuments:
+        includeSampleDocuments || USE_SAMPLE_DOCS ? sample : undefined,
       userInput,
     });
   }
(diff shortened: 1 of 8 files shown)


packages/compass-settings/src/components/settings/gen-ai-settings.tsx
```diff
@@ -21,9 +21,12 @@ export const GenAISettings: React.FunctionComponent = () => {
...
 
       {aiFeatureEnabled && (
-        <div className={atlasSettingsContainerStyles}>
-          <ConnectedAtlasLoginSettings></ConnectedAtlasLoginSettings>
-        </div>
+        <>
+          <div className={atlasSettingsContainerStyles}>
+            <ConnectedAtlasLoginSettings></ConnectedAtlasLoginSettings>
+          </div>
+          <SettingsList fields={['enableGenAISampleDocumentPassing']} />
```
Anemy, line 28: Going to update this setting to always show when the `enableGenAIFeatures` is enabled (even when it is not yet saved to true).

addaleax (approved):

packages/compass-settings/src/components/settings/gen-ai-settings.tsx
alenakhineika: I guess this is an outdated comment and we do it an a separate PR?

packages/compass-generative-ai/scripts/ai-accuracy-tests.ts
```diff
@@ -211,7 +213,8 @@ const generateMQL = async ({
       schema: schema,
       collectionName,
       databaseName,
-      sampleDocuments: USE_SAMPLE_DOCS ? sample : undefined,
+      sampleDocuments:
+        includeSampleDocuments || USE_SAMPLE_DOCS ? sample : undefined,
```
alenakhineika, line 217: Can we rename `USE_SAMPLE_DOCS` to `AI_TESTS_USE_SAMPLE_DOCS` so it is more clear what it does?

packages/compass-settings/src/components/settings/gen-ai-settings.tsx
Anemy: Yup, we'll do this in https://jira.mongodb.org/browse/COMPASS-7800

PR #5653, opened 2024-04-04
chore(e2e): test Atlas connections in compass-web sandbox

lerouxb:
This factors out the connection string tests from the connection form tests. Then fills in some more connection string based tests. Then we only skip the form based tests in compass-web and we do run the connection string based tests.


packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -247,7 +247,273 @@ async function assertCannotCreateCollection(
...
+  });
+
+  afterEach(async function () {
+    await screenshotIfFailed(compass, this.currentTest);
+    await disconnect(browser); // disconnect AFTER potentially taking a screenshot
+  });
+
+  it('can connect using connection string', async function () {
+    await browser.connectWithConnectionString();
+    if (!TEST_COMPASS_WEB) {
```
lerouxb, line 270: `await browser.connectWithConnectionString()` already checks that we connected so we don't _really_ need this additional assertion.

packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -543,96 +752,15 @@ describe('Connection screen', function () {
...
-  });
 });
 
 // eslint-disable-next-line mocha/max-top-level-suites
 describe('SRV connectivity', function () {
   before(function () {
-    skipForWeb(this, 'not applicable to compass-web');
+    skipForWeb(
+      this,
+      'we do not support the either connectionStatus param in compass-web yet'
```
lerouxb, line 762: See https://github.com/mongodb-js/compass/blob/4eb3befc9c07e7e115aa7412cac7f8d6898847ef/packages/compass-e2e-tests/helpers/commands/wait-for-connection-result.ts#L12

addaleax: okay, but even if we did support `connectionStatus` in compass-web, this would still not be applicable to it?

packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -752,7 +880,7 @@ describe('FLE2', function () {
   let browser: CompassBrowser;
 
   before(async function () {
-    skipForWeb(this, 'connect form not available in compass-web');
+    skipForWeb(this, 'FLE2 not available in compass-web');
```
addaleax, line 883:
```suggestion
    skipForWeb(this, 'QE not available in compass-web');
```

(FLE2 is no longer a name we use) or

```suggestion
    skipForWeb(this, 'client-side encryption not available in compass-web');
```

(to cover both CSFLE/QE)

addaleax (approved):

PR #5667, opened 2024-04-08
feat(compass-workspaces): adapt WorkspacePlugin to support tabs with multiple DataServices COMPASS-7718

himanshusinghs:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
+++ b/packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
@@ -47,13 +47,13 @@ function SidebarDatabasesNavigation({
       // TODO: COMPASS-7718 to use connectionId for new tabs
       switch (action) {
         case 'select-database':
-          openCollectionsWorkspace(ns);
+          openCollectionsWorkspace(connectionId, ns);
           return;
         case 'select-collection':
-          openCollectionWorkspace(ns);
+          openCollectionWorkspace(connectionId, ns);
           return;
         case 'open-in-new-tab':
-          openCollectionWorkspace(ns, { newTab: true });
+          openCollectionWorkspace(connectionId, ns, { newTab: true });
           return;
         case 'modify-view': {
           const coll = findCollection(
@@ -62,7 +62,7 @@ function SidebarDatabasesNavigation({
           );
 
           if (coll && coll.sourceName && coll.pipeline) {
-            openEditViewWorkspace(coll._id, {
+            openEditViewWorkspace(connectionId, coll._id, {
               sourceName: coll.sourceName,
               sourcePipeline: coll.pipeline,
               newTab: true,
--- a/packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
+++ b/packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
@@ -44,7 +44,6 @@ function SidebarDatabasesNavigation({
     preferencesReadOnly || connection.isDataLake || !connection.isWritable;
   const onNamespaceAction = useCallback(
     (connectionId: string, ns: string, action: Actions) => {
-      // TODO: COMPASS-7718 to use connectionId for new tabs
       switch (action) {
         case 'select-database':
           openCollectionsWorkspace(connectionId, ns);
--- a/packages/compass-workspaces/src/components/index.tsx
+++ b/packages/compass-workspaces/src/components/index.tsx
@@ -10,8 +10,15 @@ import {
 import Workspaces from './workspaces';
 import { connect } from '../stores/context';
 import { WorkspacesServiceProvider } from '../provider';
+import type { ConnectionInfo } from '@mongodb-js/compass-connections/provider';
 
 type WorkspacesWithSidebarProps = {
+  /**
+   * ConnectionInfo of the connection that Compass is connected to while in
+   * single connection mode. This is a temporary addition until we change My
+   * Queries plugin to be not dependent on DataService and MongoDBInstance.
+   */
+  singleConnectionConnectionInfo?: ConnectionInfo;
   /**
    * Current active workspace tab
    */
@@ -87,6 +94,7 @@ const sidebarStyles = css({
 const WorkspacesWithSidebar: React.FunctionComponent<
   WorkspacesWithSidebarProps
 > = ({
+  singleConnectionConnectionInfo,
   activeTab,
   activeTabCollectionInfo,
   openOnEmptyWorkspace,
@@ -112,7 +120,10 @@ const WorkspacesWithSidebar: React.FunctionComponent<
           {renderSidebar && React.createElement(renderSidebar)}
         </div>
         <div className={workspacesStyles}>
-          <Workspaces openOnEmptyWorkspace={openOnEmptyWorkspace}></Workspaces>
+          <Workspaces
+            singleConnectionConnectionInfo={singleConnectionConnectionInfo}
+            openOnEmptyWorkspace={openOnEmptyWorkspace}
+          ></Workspaces>
         </div>
       </div>
       {renderModals && React.createElement(renderModals)}
--- a/packages/compass-workspaces/src/provider.tsx
+++ b/packages/compass-workspaces/src/provider.tsx
@@ -44,7 +44,7 @@ export type WorkspacesService = {
    */
   openDatabasesWorkspace(
     this: void,
-    connectionInfoId: string,
+    connectionId: string,
     tabOptions?: TabOptions
   ): void;
   /**
@@ -53,7 +53,7 @@ export type WorkspacesService = {
    */
   openPerformanceWorkspace(
     this: void,
-    connectionInfoId: string,
+    connectionId: string,
     tabOptions?: TabOptions
   ): void;
   /**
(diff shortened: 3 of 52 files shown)


packages/compass/src/app/components/workspace.tsx
```diff
@@ -86,7 +92,7 @@ export default function Workspace({
           ]}
         >
           <WorkspacesPlugin
-            initialWorkspaceTabs={[{ type: 'My Queries' }]}
+            initialWorkspaceTabs={[{ type: 'Welcome' }]}
```
paula-stacho, line 95: Shouldn't this be behind a feature toggle?

packages/compass/src/app/components/home.tsx
```diff
@@ -287,17 +287,10 @@ function Home({
...
-                    scopeName="Connected Application"
-                  >
-                    <Workspace
-                      connectionInfo={connectionInfo}
-                      onActiveWorkspaceTabChange={onWorkspaceChange}
-                    />
-                  </AppRegistryProvider>
-                )}
+                <Workspace
+                  connectionInfo={connectionInfo}
```
kmruiz, line 291: This should be removed right? Connection Info won't be available at the beginning.

packages/compass-workspaces/src/stores/workspaces.ts
```diff
@@ -466,11 +468,12 @@ export const getActiveTab = (state: WorkspacesState): WorkspaceTab | null => {
 };
 
 export type OpenWorkspaceOptions =
-  | Pick<Workspace<'My Queries'>, 'type'>
-  | Pick<Workspace<'Databases'>, 'type'>
-  | Pick<Workspace<'Performance'>, 'type'>
-  | Pick<Workspace<'Collections'>, 'type' | 'namespace'>
-  | (Pick<Workspace<'Collection'>, 'type' | 'namespace'> &
+  | Pick<Workspace<'Welcome'>, 'type'>
+  | Pick<Workspace<'My Queries'>, 'type' | 'connectionInfoId'>
```
paula-stacho, line 472: From the designs, "My Queries" is the only one that won't be per connection

paula-stacho: (Noticed because I've been adding it to the top level just now 😄 )

packages/compass-workspaces/src/provider.tsx
```diff
@@ -38,22 +38,35 @@ export type WorkspacesService = {
    * Open "My Queries" workspace showing a list of aggregations and queries
    * saved by the user
    */
-  openMyQueriesWorkspace(this: void, tabOptions?: TabOptions): void;
+  openMyQueriesWorkspace(
+    this: void,
+    connectionInfoId: string,
```
paula-stacho, line 43: same as below

packages/compass-workspaces/src/provider.tsx
```diff
@@ -140,32 +155,45 @@ export const WorkspacesServiceProvider: React.FunctionComponent<{
       getActiveWorkspace: () => {
         return getActiveTab(store.getState());
       },
-      openMyQueriesWorkspace: (tabOptions) => {
+      openMyQueriesWorkspace: (connectionInfoId, tabOptions) => {
         return store.dispatch(
-          openWorkspaceAction({ type: 'My Queries' }, tabOptions)
+          openWorkspaceAction(
+            { type: 'My Queries', connectionInfoId },
```
paula-stacho, line 161: and here.

paula-stacho:
aren't connectionInfoId and connectionId the same thing?

himanshusinghs:
> aren't connectionInfoId and connectionId the same thing?
yeap they are the same - I renamed just now the instances of connectionInfoId to connectionId in this PR. There are a few more instances of this but those are from old PRs, I will rename them at a later point.

packages/compass-workspaces/src/stores/workspaces.ts
himanshusinghs: It is now not per connection. Thank you for pointing this out. 🙌

packages/compass/src/app/components/home.tsx
himanshusinghs: Yea it won't be but this is to preserve the behaviour of My Queries tab without the feature flag. This will be removed when we move My Queries tab to work without the DS.

packages/compass/src/app/components/workspace.tsx
himanshusinghs: Yea it should be but you shouldn't see this anymore. I was more like experimenting with this.

himanshusinghs:
Besides the points I mentioned there is one other point I need to confirm which are the changes for compass-web. In its current state the CompassWeb tab router does not add ConnectionInfo to the returned WorkspaceOptions. There are two ways to approach it:
- make ConnectionId part of URL path so that it resolves alongside the workspace options but I am not able to fully understand what it would mean for mms
- silently add ConnectionId to the workspace options but this felt a bit hacky

So excluding this, the PR should be ready for review.

packages/compass/src/app/components/workspace.tsx
```diff
@@ -87,31 +92,38 @@ export default function Workspace({
...
       <div className={shellContainerStyles}>
-        <CompassShellPlugin />
+        <ConnectionInfoProvider
+          connectionInfoId={singleConnectionConnectionInfo?.id}
+        >
+          <CompassShellPlugin />
+        </ConnectionInfoProvider>
       </div>
     </div>
   );
```
paula-stacho, line 129: Do we need similar changes in compass-web?

paula-stacho:
@kmruiz do we already have a follow up task to fix the sidebar with the new activeWorkspace.connectionId? If we don't I can create it

packages/compass/src/app/components/workspace.tsx
kmruiz: I think compass-web injects it's own `ConnectionInfoProvider` so I think this should still work, but it's worth to test.

kmruiz:
> @kmruiz do we already have a follow up task to fix the sidebar with the new `activeWorkspace.connectionId`? If we don't I can create it

I believe this was part of this task, but if we want to do it in another please feel free to create a ticket and we can play it when we merge this PR.

packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
```diff
@@ -47,13 +47,13 @@ function SidebarDatabasesNavigation({
       // TODO: COMPASS-7718 to use connectionId for new tabs
```
paula-stacho, line 47:
```suggestion
```

packages/compass-workspaces/src/components/index.tsx
```diff
@@ -10,8 +10,15 @@ import {
...
 import { WorkspacesServiceProvider } from '../provider';
+import type { ConnectionInfo } from '@mongodb-js/compass-connections/provider';
 
 type WorkspacesWithSidebarProps = {
+  /**
+   * ConnectionInfo of the connection that Compass is connected to while in
+   * single connection mode. This is a temporary addition until we change My
+   * Queries plugin to be not dependent on DataService and MongoDBInstance.
+   */
+  singleConnectionConnectionInfo?: ConnectionInfo;
```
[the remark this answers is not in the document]
himanshusinghs: Yea its exactly to preserve that.

packages/compass/src/app/components/workspace.tsx
himanshusinghs: So my note above was exactly about this - I wanted to confirm with Sergey if there is any relation between what we're doing in our sandbox and what will be done in mms. Thankfully there is not much. The keypoint is that mms will always know cluster name and will use that as connectionId to be resolved alongside the ConnectionInfo so I can safely rely on that passed info.

packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
himanshusinghs: I thought I removed it - maybe I did it in the new sidebar.

himanshusinghs:
@paula-stacho All the usage of WorkspaceService already pass connectionId in this PR so that part is covered.

kmruiz (approved):
LGTM.

kmruiz:
> Re @himanshusinghs and @kmruiz
> 
> ConnectionId is passed to the new sidebar, but it doesn't trickle all the way down to ConnectionsNavigationTree. So the highlighting and expanding behaviour of connection items isn't complete. We weren't sure what the new structure of Workspace will be, so we didn't prepare that part. I'd say it can easily be a follow up.

Then please feel free to create a task, we can play it when we merge this PR. From what you say, when we merge this passing down the connection id should be pretty straightforward.

paula-stacho (approved):
Looks good!

packages/compass/src/app/components/workspace.tsx
himanshusinghs: Confirmed and fixed the issues with CompassWeb. The render tree for CompassWeb didn't change much as we're doing the single connection flow there.

PR #5677, opened 2024-04-10
chore: sidebar multiple connection support COMPASS-7775

kmruiz:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-sidebar/src/components/sidebar-databases-navigation.tsx
```diff
@@ -87,22 +88,31 @@ function SidebarDatabasesNavigation({
   );
 }
 
-function mapStateToProps(state: RootState) {
+function mapStateToProps(
+  state: RootState,
+  { connectionId }: { connectionId: ConnectionInfo['id'] }
```
paula-stacho, line 93: I think we just use string in other places

packages/compass-sidebar/src/modules/databases.ts
```diff
@@ -45,7 +46,8 @@ export type Database = Pick<
     '_id' | 'name' | 'type' | 'sourceName' | 'pipeline'
   >[];
 };
-export interface DatabaseState {
+export type DatabaseState = Record<ConnectionInfo['id'], SingleDatabaseState>;
+export interface SingleDatabaseState {
```
paula-stacho, line 50: This sounds very much like there is a single Database, but it's actually a list 😄. Perhaps we can use something like `AllDatabasesState` and `ConnectionDatabasesState`?

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -378,68 +390,52 @@ const ConnectionsNavigationTree: React.FunctionComponent<
...
 
   const listRef = useRef<List | null>(null);
   const id = useId();
 
   useEffect(() => {
     if (activeNamespace) {
-      onDatabaseExpand(activeNamespace, true);
+      for (const connection of connections) {
+        onDatabaseExpand(connection.connectionInfo.id, activeNamespace, true);
+      }
```
paula-stacho, line 406: we won't need to call this multiple times, or? perhaps we can keep it now to not break the legacy version. but I'd add a TODO for when we have active connection, then we can just `onDatabaseExpand(activeConnection, activeNamespace, true)`.
similar to the todo below, where I put the wrong ticket number 🙈 . It will rather be part of @himanshusinghs's workspace changes

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -378,68 +390,52 @@ const ConnectionsNavigationTree: React.FunctionComponent<
...
-            database,
-            databaseIndex,
-            databasesLength: data.length || 0,
-            expanded: expanded as SCConnectionsNavigationTreeProps['expanded'],
-            level: 1,
-          })
+      const connection = connections[0];
+      return connection.databases.flatMap((database, databaseIndex) =>
+        databaseToItems({
+          connectionId: connection.connectionInfo.id, // 'TOFIX'
```
paula-stacho, line 425: what's wrong with this?

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -378,68 +390,52 @@ const ConnectionsNavigationTree: React.FunctionComponent<
...
-            connectionIndex,
-            connectionsLength: data.length,
-            expanded: expanded as MCConnectionsNavigationTreeProps['expanded'],
-          })
+      return connections.flatMap((connection, connectionIndex) =>
+        connectionToItems({
+          connection,
+          connectionIndex,
+          connectionsLength: connections.length,
+          expanded: expanded,
```
paula-stacho, line 418:
```suggestion
          expanded,
```

packages/compass-sidebar/src/modules/is-genuine-mongodb-visible.ts
```diff
@@ -12,15 +13,20 @@ export const TOGGLE_IS_GENUINE_MONGODB_VISIBLE =
...
   type: typeof TOGGLE_IS_GENUINE_MONGODB_VISIBLE;
+  connectionId: ConnectionInfo['id'];
   isVisible: boolean;
 }
 export type IsGenuineMongoDBVisibleAction = ToggleIsGenuineMongoDBVisibleAction;
-export type IsGenuineMongoDBVisibleState = boolean;
+export type IsGenuineMongoDBVisibleState = Record<
+  ConnectionInfo['id'],
+  boolean
+>;
```
paula-stacho, line 23: this reminds me that we don't have this functionality in the new sidebar - do we have a ticket for it?

if I get this correctly, this state is for the modal, not the marker. I assume that with multiple connections there might be multiple markers, but only one modal at a time -> so this state doesn't need to be a record. it can be `IsGenuineMongoDBVisibleState = ConnectionInfo['id'] | undefined`.

see:
```
isGenuineMongoDBVisible:
      state.isGenuineMongoDBVisible[initialConnectionInfo.id], // this is for the modal
isGenuine:
      state.instance[initialConnectionInfo.id]?.genuineMongoDB.isGenuine, // this would be for the marker
```

packages/compass-sidebar/src/components/connection-info-modal.tsx
```diff
@@ -46,6 +48,7 @@ export function ConnectionInfoModal({
   close,
   infos,
 }: {
+  initialConnectionInfo: ConnectionInfo;
```
paula-stacho, line 51: this file has it's own `ConnectionInfo`, but in this prop we send the original - `ConnectionStorageConnectionInfo`. this is what is forcing the "as any" in the sidebar

paula-stacho: I can fix this in my PR, just saying :)

packages/compass-sidebar/src/modules/is-genuine-mongodb-visible.ts
kmruiz: Yeah so I added a follow up ticket to add the non genuine marker. Right now it's a Record<string, boolean> because multiple connections could be using non genuine MongoDB servers, so we would want to mark both of them.

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
kmruiz: Nothing anymore, I had this marker because the connectionInfo was hardcoded during development :D.

packages/compass-sidebar/src/modules/databases.ts
kmruiz: yeah, I totally agree, I think I'll rename it as you suggested.

paula-stacho:
I think now would be a good time to add more tests to `active-connection-navigation`. I only wrote a very minimalistic test suite, because it was basically a placeholder at that point. totally can be a follow up though (saying as someone who'll be unblocked when this is merged 😂 )

packages/compass-sidebar/src/components/connection-info-modal.tsx
kmruiz: Ah yeah I can fix it, you are right. It takes a minute. Thanks for pointing out!

packages/compass-connections-navigation/src/sc-connections-navigation-tree.spec.tsx
```diff
@@ -1,458 +0,0 @@
...
-    name: 'bar',
-    collectionsStatus: 'ready',
-    collectionsLength: 3,
-    collections: [
-      { _id: 'bar.meow', name: 'meow', type: 'collection' },
-      { _id: 'bar.woof', name: 'woof', type: 'timeseries' },
-      { _id: 'bar.bwok', name: 'bwok', type: 'view' },
-    ],
-  },
-];
```
paula-stacho, line 37: I'm not sure if we want to remove this, because we want to be careful not to break the legacy functionality + there are still differences between the two versions. Replacing the databases and expanded props should fix the tests.

kmruiz: You mean the file?

packages/compass-sidebar/src/modules/is-genuine-mongodb-visible.ts
paula-stacho: for marking multiple connections we use `isGenuine`.  `isGenuineMongoDBVisible` is used for `NonGenuineWarningModal`, which is a single piece:
https://github.com/mongodb-js/compass/pull/5677/files#diff-c49bde1f435db7e728d2d67930f0cc4bce9ca7c919c25a6f89c0dd1d768c0deaL206
the variable name is confusing, it should probably be `isNonGenuineWarningModalOpen`

packages/compass-connections-navigation/src/sc-connections-navigation-tree.spec.tsx
paula-stacho: Yes, I left this test suite as regression tests for the SC version.

paula-stacho: Actually it will need also the preferences and the connectionIds in the actions, so perhaps not that easy to fix. I don't know if it's worth it, I'm just feeling cautious around the SC version, since this is the part that's not behind the feature toggle but also we give less attention to it

packages/compass-sidebar/src/modules/is-genuine-mongodb-visible.ts
kmruiz: Ah, I get it now. Yeah you are right, probably the isNonGenuineWarningModalOpen can be a single variable. I need to think about this, because even if the naming is kind of misleading (I can change it, it's fine), I personally like the idea of being it one variable per connection, in case later we don't show a modal but something else (like a marker in the connection name or something)

packages/compass-sidebar/src/modules/databases.ts
```diff
@@ -105,64 +118,78 @@ export default function reducer(
...
-      filterRegex: filterRegex,
-    });
+
+    for (const connectionId of Object.keys(getState().databases)) {
+      dispatch({
+        type: CHANGE_FILTER_REGEX,
+        connectionId,
+        filterRegex,
+      });
+    }
```
paula-stacho, line 192: feels like this could be a single action, but we can refactor that in https://jira.mongodb.org/browse/COMPASS-7772

packages/compass-sidebar/src/modules/is-genuine-mongodb-visible.ts
paula-stacho: we do! this modal opens when we click on one of the markers. but perhaps you're right, in the future maybe the modal will contain something specific to the connection, then we could use the id. right now it's just a disclaimer with a link to https://www.mongodb.com/docs/compass/master/faq/#why-am-i-seeing-a-warning-about-a-non-genuine-mongodb-server-

packages/compass-sidebar/src/modules/databases.ts
kmruiz: Sounds good. My initial approach was this one because if for some reason one of the databases is not ready, it shouldn't affect others (as they are different dispatches). But I think we can achieve also this by using one single action.

packages/compass-connections-navigation/src/connection-item.tsx
```diff
@@ -95,7 +95,7 @@ export const ConnectionItem: React.FunctionComponent<
 
   const onAction = useCallback(
     (action: Actions) => {
-      onNamespaceAction(id, action);
+      onNamespaceAction(id, id, action);
```
himanshusinghs, line 98: Just confirming, both of these are meant to be `id`?

packages/compass-sidebar/src/components/legacy/navigation-items.spec.tsx
```diff
@@ -9,6 +9,8 @@ import reducer from '../../modules';
 import { NavigationItems } from './navigation-items';
 import { WorkspacesProvider } from '@mongodb-js/compass-workspaces';
 
+const CONNECTION_ID = 'webscale';
```
himanshusinghs, line 12: 🤣

himanshusinghs: I am gonna have to watch this again, this triggered me 🤣

packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
```diff
@@ -80,46 +86,74 @@ function SidebarDatabasesNavigation({
...
+        connectionInfo,
+        databasesLength: filteredDatabases?.length || 0,
+        databasesStatus: (status ??
+          'fetching') as Connection['databasesStatus'],
+        databases: filteredDatabases ?? [],
+      },
+    ],
+    expanded: {
+      [connectionId]: expanded,
+    },
```
himanshusinghs, line 146: We probably should do this inside of the component with a memo. This will pass new connections and expanded as props for every state update in redux which will make the underlying component re-render everytime.
(We can surely do a follow up for this - to avoid holding this PR for too long)

packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
```diff
@@ -54,7 +56,11 @@ function SidebarDatabasesNavigation({
           openCollectionWorkspace(ns, { newTab: true });
           return;
         case 'modify-view': {
-          const coll = findCollection(ns, databases);
+          const coll = findCollection(
+            ns,
+            (connections.find((cn) => cn.connectionInfo.id === connectionId)
+              ?.databases as Database[]) || []
+          );
```
paula-stacho, line 63: since we're already working with a single connection, we could simplify -> 
```suggestion
          const coll = findCollection(
            ns, connection.databases as Database[],
          );
```

packages/compass-sidebar/src/components/multiple-connections/active-connections/active-connection-navigation.tsx
```diff
@@ -200,34 +198,65 @@ function ActiveConnectionNavigation({
...
 
   return {
-    isReady,
-    isDataLake,
-    isWritable,
-    databases: filteredDatabases,
-    expanded,
+    isReady: true,
+    connections,
+    expanded: expandedResult,
```
himanshusinghs, line 254: The same concern for this - about doing this inside of component
(can be done as a follow up)

packages/compass-sidebar/src/components/legacy/sidebar.tsx
```diff
@@ -125,7 +127,7 @@ export function Sidebar({
       if (action === 'copy-connection-string') {
         void copyConnectionString(
           maybeProtectConnectionString(
-            connectionInfo.connectionOptions.connectionString
+            initialConnectionInfo?.connectionOptions?.connectionString ?? ''
```
paula-stacho, line 130: initialConnectionInfo is not optional
```suggestion
            initialConnectionInfo.connectionOptions.connectionString
```

packages/compass-sidebar/src/components/legacy/sidebar.tsx
```diff
@@ -147,12 +149,12 @@ export function Sidebar({
       onSidebarAction,
       openToast,
       maybeProtectConnectionString,
-      connectionInfo.connectionOptions.connectionString,
+      initialConnectionInfo?.connectionOptions?.connectionString,
```
paula-stacho, line 152:
```suggestion
      initialConnectionInfo.connectionOptions.connectionString,
```

packages/compass-sidebar/src/plugin.tsx
```diff
@@ -17,13 +17,12 @@ const errorBoundaryStyles = css({
 
 export interface SidebarPluginProps {
   showConnectionInfo?: boolean;
-  // TODO(COMPASS-7397): the need for passing this directly to sidebar should go
-  // away with refactoring compoass-conneciton to a plugin
```
himanshusinghs, line 21: I think this is still relevant?

packages/compass-sidebar/src/modules/data-service.ts
```diff
@@ -1,46 +1,21 @@
...
-    dispatch({ type: SET_CSFLE_ENABLED, enable });
+): SidebarThunkAction<void, never> => {
+  return (_dispatch, _getState, { globalAppRegistry, connectionsManager }) => {
+    const dataService =
+      connectionsManager.getDataServiceForConnection(connectionId);
+
+    if (!dataService) {
+      // This should be unreachable because this is only visible when
+      // we are connected.
+      return;
```
paula-stacho, line 15: wouldn't it be safer to throw then? to avoid a silent failure in the future?

paula-stacho:
I only noticed now - the active connections used to be ordered alphabetically by title and not by favorite/unfavorite. can be a follow up.

packages/compass-sidebar/src/stores/store.ts
```diff
@@ -1,180 +1,57 @@
...
-        (err as MongoServerError).codeName === 'AtlasError' ? false : true;
-      store.dispatch(setIsPerformanceTabSupported(isSupported));
-    }
+  on(
+    instancesManager,
+    MongoDBInstancesManagerEvents.InstanceRemoved,
+    (connectionId) => store.dispatch(closeInstance(connectionId))
   );
 
   return {
```
paula-stacho, line 57: we're missing tests for InstanceCreated and InstanceRemoved events

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -378,68 +394,56 @@ const ConnectionsNavigationTree: React.FunctionComponent<
...
 
   useEffect(() => {
     if (activeNamespace) {
-      onDatabaseExpand(activeNamespace, true);
+      for (const connection of connections) {
+        onDatabaseExpand(connection.connectionInfo.id, activeNamespace, true);
+      }
     }
-  }, [activeNamespace, onDatabaseExpand]);
+  }, [activeNamespace, onDatabaseExpand, connections]);
```
paula-stacho, line 411: this `connections` dependency is still causing issues in the legacy version. it is not possible to collapse active connection, because that changes the connections array and triggers this expand.
Let's restrict it to the first connection, this way the legacy version works as before, and we can fix the multi-connection version once we have active connection.
```suggestion
  }, [activeNamespace, onDatabaseExpand, connections[0].connectionInfo.id]);
```

packages/compass-connections-navigation/src/connection-item.tsx
kmruiz: Yeah, because both the namespace and connection are the same (the connection id itself).

packages/compass-sidebar/src/components/legacy/navigation-items.spec.tsx
kmruiz: image

packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
kmruiz: Maybe we can pair on this one, because I'm not sure how to move it to useMemo as it depends on the redux state.

packages/compass-sidebar/src/plugin.tsx
kmruiz: Ah yes, you are right. Adding it back.

paula-stacho (approved):
Nice! 🎉

packages/compass-sidebar/src/modules/databases.ts
```diff
@@ -105,64 +118,78 @@ export default function reducer(
...
+    databaseId: string,
     forceExpand: boolean
   ): SidebarThunkAction<void, DatabasesAction> =>
   (dispatch, getState, { globalAppRegistry }) => {
-    const { database } = toNS(id);
-    const { databases } = getState();
-    const expanded = forceExpand ?? !databases.expandedDbList[database];
+    const { database } = toNS(databaseId);
+    const { expandedDbList } = getState().databases[connectionId];
+    const expanded = forceExpand ?? !expandedDbList[database];
```
addaleax, line 166: You may want to be careful here, afaict `getState().databases` and `expandedDbList` are not initialized as null-prototype objects anymore, so e.g. a database called `toString` would always be expanded now (something the previous code handled)?

kmruiz: So you mean that because the initial state is not null, if a database is called toString it will be always expanded? Just to understand.

To be fair I didn't see any code handling this specific issue before, but I might missed it.

packages/compass-sidebar/src/components/legacy/navigation-items.spec.tsx
paula-stacho: oh wow, I haven't seen this before and now I'm dying 😂

packages/compass-sidebar/src/modules/databases.ts
addaleax: > So you mean that because the initial state is not null, if a database is called toString it will be always expanded? 

Err, sorry, I mixed that up because of the toggling – it's _never_ expanded here unless `forceExpand` is set, right? Because `!expandedDbList['toString']` will evaluate to `false` because `expandedDbList` now inherits from `Object` (unless I'm missing something)

> To be fair I didn't see any code handling this specific issue before, but I might missed it.

This is what I was referring to: https://github.com/mongodb-js/compass/blob/4ae8d14e4e2d5333a7f9ace82d7d01910b2f0a3a/packages/compass-sidebar/src/modules/databases.ts#L61

kmruiz: Hmm, I get it now, but wouldn't it recreate a normal object when we do Object.fromEntries?

https://github.com/mongodb-js/compass/blob/4ae8d14e4e2d5333a7f9ace82d7d01910b2f0a3a/packages/compass-sidebar/src/modules/databases.ts#L100

And also when shallow copying here:

https://github.com/mongodb-js/compass/blob/4ae8d14e4e2d5333a7f9ace82d7d01910b2f0a3a/packages/compass-sidebar/src/modules/databases.ts#L81

Asking because I don't know.

addaleax: Yeah, it would ... sigh. :) The shallow copy at least should have probably included a `__proto__: null`.

Anyway, yeah, nothing this PR breaks then. Let's merge it, I can do a small follow-up to fix this :)

kmruiz: I have a task for this kind of small follow ups. No worries, I'll add this topic there so we can fix it later haha.

himanshusinghs (approved):

PR #5690, opened 2024-04-15
wip: export schema dialog

mcasimir:


packages/compass-schema/src/components/export-schema/export-schema-modal.tsx
```diff
@@ -0,0 +1,321 @@
...
+  settings: T,
+  converter: (schema: Schema, settings: T) => string
+): Export {
+  const exportInfo = useMemo(() => {
+    try {
+      return {
+        code: converter(schema, settings),
+      };
+    } catch (e) {
+      const errorMessage = (e as Error)?.message || '';
```
addaleax, line 122:
```suggestion
      const errorMessage = (e as Error)?.message || String(e);
```

not that it should really ever matter, of course

packages/compass-schema/src/components/export-schema/export-schema-modal.tsx
```diff
@@ -0,0 +1,321 @@
...
+      return { error: `Conversion failed: ${errorMessage}` };
+    }
+  }, [converter, schema, ...(settings && Object.values(settings))]);
```
addaleax, line 125:
```suggestion
  }, [converter, schema, ...(settings && Object.entries(settings).flat())]);
```

Otherwise `{ a: 1 }` and `{ b: 1 }` are considered to be the same, right?

packages/compass-schema/src/components/export-schema/export-schema-modal.tsx
```diff
@@ -0,0 +1,321 @@
...
+  Option,
+  Select,
+  spacing,
+  TextInput,
+  WorkspaceContainer,
+} from '@mongodb-js/compass-components';
+import type { Schema } from 'mongodb-schema';
+import React, { useMemo, useState, useCallback } from 'react';
+
+import { JSONEditor } from '@mongodb-js/compass-editor';
```
addaleax, line 16: This needs to be declared as a dependency in package.json :)

packages/compass-schema/src/components/export-schema/export-schema-modal.tsx
```diff
@@ -0,0 +1,321 @@
...
+      JSON.stringify(exportMongodbJSONSchema(schema, settings), null, 2)
+  );
+
+  return (
+    <ExportPanel
+      exportInfo={exportInfo}
+      settings={
+        <>
+          <Checkbox
+            label="Include _id"
```
addaleax, line 152: Would it make sense to highlight `_id` as code here?

```suggestion
            label={<>Include <InlineCode>_id</InlineCode></>}
```

packages/compass-schema/src/components/export-schema/export-schema-modal.tsx
```diff
@@ -0,0 +1,321 @@
...
+          aria-label="Select an export format"
+          onChange={(newVal: string) => {
+            setSchemaType(newVal as ExportFormat);
+          }}
+        >
+          <Option value="mongodbJsonSchema">
+            MongoDB Validation Schema ($jsonSchema)
+          </Option>
+          <Option value="jsonSchema">JSON Schema (EJSON)</Option>
+          <Option value="internalSchema">Compass Internal Schema</Option>
```
addaleax, line 304: We're providing this for compatibility with the legacy feature that used this format until now? "Internal" sounds a bit like users aren't really supposed to rely on it :)

packages/compass-schema/src/components/export-schema/export-schema-modal.tsx
```diff
@@ -0,0 +1,321 @@
...
+          value={schemaType}
+          aria-label="Select an export format"
+          onChange={(newVal: string) => {
+            setSchemaType(newVal as ExportFormat);
+          }}
+        >
+          <Option value="mongodbJsonSchema">
+            MongoDB Validation Schema ($jsonSchema)
+          </Option>
+          <Option value="jsonSchema">JSON Schema (EJSON)</Option>
```
addaleax, line 303: I admittedly didn't understand what this meant until I looked at the source code, i.e. that this is the JSON schema _for_ the EJSON representation of the BSON objects, kind of makes sense but maybe we can have a line below the Select to explain what the currently selected format actually means?

mcasimir: That's probably a great idea yeah, awesome that you reviewed these changes! We probably want to rewrite this entire thing, maybe i should try to rebase

packages/compass-schema/src/components/export-schema/export-schema-modal.tsx
mcasimir: good point yeah, not sure what we could call that instead, or if we want to just drop it, maybe we can add an "x-samples: [...]" to the json schema and just get rid of this one

packages/compass-schema/src/components/export-schema/export-schema-modal.tsx
mcasimir: true

packages/compass-schema/src/components/export-schema/export-schema-modal.tsx
mcasimir: Good catch!

packages/compass-schema/src/components/export-schema/export-schema-modal.tsx
addaleax: (this applies for the other format as well)

PR #5691, opened 2024-04-15
chore(settings): use sandbox preference for showing ai settings COMPASS-7800

Anemy:
COMPASS-7800

Initially we were thinking this might be the ticket where we update Compass' settings to autosave. Instead this pr updates showing/hiding the ai preferences based off of the sandbox preference. Chatted with Alena to see if we want to go this route for now and defer updating our settings to autosave until we have more of a requirement for it. Ticket for auto-saving settings: COMPASS-7866


Anemy:
@alenakhineika Got the ok from Ben: https://mongodb.slack.com/archives/GDTJXPHD0/p1713287477962769?thread_ts=1711635817.479729&cid=GDTJXPHD0

PR #5695, opened 2024-04-16
chore(generative-ai): add accuracy test

Anemy:
COMPASS-7850


packages/compass-generative-ai/scripts/ai-accuracy-tests.ts
```diff
@@ -512,17 +512,6 @@ async function pushResultsToDB({
 }
 
 const tests: TestOptions[] = [
```
mcasimir, line 514: do we need to remove this test? this one was targeting a find query, and it would be nice to keep testing simple cases and make sure we get consistently a nearly perfect score here.

Anemy: Don't need to, no. Some of the more complex finds have near perfect scores, so I was thinking it was a bit unnecessary. No harm in keeping it, I'll add it back in.

mcasimir (approved):

PR #5704, opened 2024-04-18
fix(generative-ai): parse ai response correctly COMPASS-7780

mabaasit:
### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/atlas-service/src/atlas-service.ts
+++ b/packages/atlas-service/src/atlas-service.ts
@@ -50,6 +50,17 @@ export class AtlasService {
           ...init?.headers,
         },
       });
+      this.logger.log.info(
+        this.logger.mongoLogId(1_001_000_309),
+        'AtlasService',
+        'Received API response',
+        {
+          url,
+          status: res.status,
+          statusText: res.statusText,
+          contentLength: res.headers.get('content-length'),
+        }
+      );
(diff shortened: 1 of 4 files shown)


packages/atlas-service/src/atlas-service.ts
```diff
@@ -50,6 +50,17 @@ export class AtlasService {
...
       });
+      this.logger.log.info(
+        this.logger.mongoLogId(1_001_000_309),
+        'AtlasService',
+        'Received API response',
+        {
+          url,
+          status: res.status,
+          statusText: res.statusText,
+          contentLength: res.headers.get('content-length'),
```
alenakhineika, line 61: Not sure if we need to log `contentLength`. If the server won't set it explicitly we will always log 0.

alenakhineika (approved):

packages/atlas-service/src/atlas-service.ts
mabaasit: removed in [55d222f](https://github.com/mongodb-js/compass/pull/5704/commits/55d222fe2e65ea79ff5b01a52141fca1de693c42)

PR #5707, opened 2024-04-18
feat(compass-connections-navigation, compass-sidebar): active connection actions (part 1)

paula-stacho:
## Description
Two actions are just placeholders and will be added in separate tasks:
- View Performance metrics https://jira.mongodb.org/browse/COMPASS-7873
- Disconnect https://jira.mongodb.org/browse/COMPASS-7874

Note: In `ConnectionsNavigationTree`, I replaced the __TEST props with a check for `process.env.NODE_ENV === 'test'`. This way don't have to think about the autosizer implementation details in the tests (which is especially helpful when testing higher level components - it took me hours to figure out why the active connections list was only containing placeholders 🤦 )

Note 2: That the active connections are not found in theory shouldn't happen, since even idle connections should stay on the list. But the copy action was already using the error toast, and the favorite action is also relatively subtle, so I figured extra visibility of failure wouldn't hurt.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-connections-navigation/src/connection-item.tsx
```diff
@@ -101,14 +101,40 @@ export const ConnectionItem: React.FunctionComponent<
...
       },
     ];
-  }, []);
+
+    actions.push({
+      action: 'connection-disconnect',
+      icon: 'Disconnect',
+      label: 'Disconnect',
+      variant: 'destructive',
+    });
```
addaleax, line 134: This can just be part of the array above, right?

packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
```diff
@@ -78,15 +83,45 @@ function ConnectionErrorToastBody({
...
+async function copyConnectionString(connectionString: string) {
+  try {
+    await navigator.clipboard.writeText(connectionString);
+    openToast('copy-to-clipboard', {
+      title: 'Success',
+      description: 'Copied to clipboard.',
+      variant: 'success',
+      timeout: TOAST_TIMEOUT_MS,
+    });
+  } catch (err) {
```
addaleax, line 95: Do we want to include `err.message` in the description here, or otherwise log it somehow? Otherwise you can do
```suggestion
  } catch {
```

packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
```diff
@@ -78,15 +83,45 @@ function ConnectionErrorToastBody({
...
 }: MultipleConnectionSidebarProps) {
   const { openToast, closeToast } = useToast('multiple-connection-status');
   const cancelCurrentConnectionRef = useRef<(id: string) => Promise<void>>();
   const activeConnections = useActiveConnections();
+  const maybeProtectConnectionString = useMaybeProtectConnectionString();
 
   const [isConnectionFormOpen, setIsConnectionFormOpen] = useState(false);
+  const [isConnectionInfoModalOpen, setIsConnectionInfoModalOpen] = useState<
+    string | undefined
+  >();
```
addaleax, line 118: If `isConnectionInfoModalOpen` is a string, should this maybe be named differently to reflect that? If you want to keep the naming consistent, you can still use `isConnectionInfoModalOpen`, just maybe as inferred state in that case?

```suggestion
  const [connectionModalConnectionId, setConnectionModalConnectionId] = useState<
    string | undefined
  >();
  const isConnectionInfoModalOpen = !!connectionModalConnectionId;
```

(`connectionModalConnectionId` is just an example, there might be a better name for it 🙂 )

packages/compass-connections-navigation/src/connection-item.tsx
paula-stacho: ah, yes, forgot to merge this. thanks!

packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
paula-stacho: you're right, I will rename it

paula-stacho: I did try this as a fixup, but the code was actually touched by several commits so I had conflicts and gave up.. another time 😄

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -497,6 +488,8 @@ const ConnectionsNavigationTree: React.FunctionComponent<
     return data.items[index].key;
   }, []);
 
+  const isTestEnv = process.env.NODE_ENV === 'test';
```
kmruiz, line 491: Now that I see this, is this going to be a problem in compass-web? Are the tests for compass-web working with this change?

paula-stacho: do they have a different env value?
for e2e tests - the `height` & `width` shouldn't apply. they are only used as default, if there is no "real" size available. the problem with integration tests was autosize requires a container with fixed size. in it's absence it doesn't take all available space, but collapses to nothing. this is what the defaults are preventing.
the only change for e2e tests would be `overscanCount`, which I don't understand yet and need to check.

himanshusinghs (approved):
Looking good. Though you might wanna rethink the point about `isTestEnv` - we just started a thread asking for clarification on the code split.

PR #5711, opened 2024-04-18
chore(query-bar): prefer ai generated aggregations over query fields when supplied COMPASS-7852

Anemy:
COMPASS-7852

For an example of a prompt that would generate a good valid aggregation, but the ai also includes projections and sorts into the query try:
```
whats the count of each bed count, sorted by lowest bed count first
```
On the `airbnb_sample`.`listingsAndReviews` dataset.

### DIFF
--- a/packages/compass-query-bar/src/stores/ai-query-reducer.spec.ts
+++ b/packages/compass-query-bar/src/stores/ai-query-reducer.spec.ts
@@ -46,6 +46,7 @@ describe('aiQueryReducer', function () {
         const mockDataService = {
           sample: sandbox.stub().resolves([{ _id: 42 }]),
         };
+        const emitStub = sandbox.stub();
 
         const store = createStore(
           {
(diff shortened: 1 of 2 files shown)


alenakhineika:
I think we had this conversation previously, and it was the initial idea to prefer aggregations over queries, but it was decided the other way around. I believe most of the team was against redirecting users to aggregations for simple cases like sort (even tho this is a valid agg stage). If we plan to introduce visual indicators of changes made by AI in the query bar, do we really need these extra redirects? I have a concern that we will move users around too much. cc @mcasimir

mcasimir:
That's a good point @alenakhineika. I'd assume that redirecting from query to agg if not necessary is not the best. If we do this change we know we could improve some of the cases where an aggregation was really needed, but do we know how many times we would go to an Agg when we could have stayed on find?  


Maybe better to cover this with an accuracy test @Anemy and try to improve the situation in the prompt?  This is really something  the model should not be doing.

mcasimir (commented):
Btw, not urgent, but why is the api returning these superimposed results that are a bit of this and that? Is there a way to clean up the result on the server and always return a clear response to clients?

Anemy:
Going to investigate some prompt changes. I'll open a separate pr involving an accuracy test for this. We can do this logic on the server side if we want, and prompt changes don't help, which I'm thinking could be a better solution regardless. I'll close this in favor of that. Thanks for bringing it up! I think I had forgotten what happened with the last pr 😅

PR #5720, opened 2024-04-22
feat(auto-updates): add support to notify users for an update on linux/msi COMPASS-7686

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass/src/main/auto-update-manager.ts
```diff
@@ -461,6 +500,55 @@ const STATE_UPDATE: Record<
...
+          newVersion: updateInfo.to,
+        });
+        return;
+      }
+
+      const answer = await dialog.showMessageBox({
+        icon: COMPASS_ICON,
+        title: 'New version available',
+        message: 'A new version of Compass is available to install',
+        detail: `Compass ${updateInfo.to} is available. You are currently using ${updateInfo.from}. Would you like to download it now?`,
```
addaleax, line 533: Can we use the same text here for the new toast as well? It reads a bit nicer tbh :)

addaleax (approved):

packages/compass/src/main/auto-update-manager.ts
mabaasit: I reached out to Misba regarding the toast content and I'd share this idea (to have same for both) with her as well.

lerouxb (approved):
Looks good. We should probably manually test that this works, though. How would we do that? Multiple releases?

mabaasit:
> Looks good. We should probably manually test that this works, though. How would we do that? Multiple releases?

Yes, at least 2 betas - first with these changes and then the next one should notify user of updates.

PR #5722, opened 2024-04-22
fix: add missing `'use strict';` directives

addaleax:
##### fix: add missing `'use strict';` directives

`'use strict';` is 15 years old, it's time we catch up :)

Commit created via

```sh
for f in $( git ls-files | grep '.js$' ); do
  grep -q 'use strict' "$f" || \
  ( head -n5 "$f" | grep -Eq '(im|ex)port ' ) || \
  ( \
    (( head -n1 "$f" | grep -q '#!' ) && \
      ( cat "$f" | sed -e '/#!/s/$/\n'"'use strict';"'/' ) || \
      ( echo "'use strict';" && cat "$f" ) )  > "${f}~" && \
    mv -v "${f}~" "$f" &&  \
    rm -f "${f}~") ;
  done ; npm run bootstrap && (cd packages/bson-transpilers && npm run compile)
```

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


alenakhineika (approved):

mcasimir:
👍 should we add `strict: ["error", "global"]` to our eslint config for .js files? probably `eslint --fix` would also add that

addaleax:
@mcasimir In theory yes, in practice eslint doesn't seem to be having a great time telling modules apart from scripts in our source tree, if I add that rule to e.g. eslint-config-devtools eslint tries to remove it for all files with a hashbang (when it should definitely be present in those)

PR #5725, opened 2024-04-23
chore(e2e): timeout calls to browser.saveScreenshot() and catch the error

lerouxb:


packages/compass-e2e-tests/helpers/commands/screenshot.ts
```diff
@@ -2,12 +2,37 @@ import path from 'path';
 import type { CompassBrowser } from '../compass-browser';
 import { SCREENSHOTS_PATH } from '../compass';
 
+const withTimeout = (millis: number, promise: Promise<any>) => {
+  let timeoutPid: NodeJS.Timeout;
+  const timeout = new Promise(
+    (resolve, reject) =>
+      (timeoutPid = setTimeout(
+        () => reject(`Timed out after ${millis} ms.`),
```
addaleax, line 10: Since this is named `withTimeout` which kind of implies that it can also be a generic utility and not just something for this specific use case, should this `reject` with an error?

lerouxb: ok.

addaleax (approved):

PR #5726, opened 2024-04-23
chore(hadron-app-registry): update README with plugin development docs COMPASS-7563

addaleax:
Nicer to view in GH markdown: https://github.com/mongodb-js/compass/blob/7563-dev/packages/hadron-app-registry/README.md


packages/hadron-app-registry/README.md
```diff
@@ -1,36 +1,165 @@
...
+
+```tsx
+import {
+  globalAppRegistry,
+  AppRegistry,
+  AppRegistryProvider,
+  registerHadronPlugin,
+} from 'hadron-app-registry';
+import CompassLogging from '@mongodb-js/compass-logging';
+import { LoggingProvider, logging } from '@mongodb-js/compass-logging/provider';
```
mabaasit, line 59: Just to make it a bit clear, we would import `loggingLocator` and then in `registerHadronPlugin` would pass it as `logging: loggingLocator`.

mabaasit (approved):

packages/hadron-app-registry/README.md
```diff
@@ -1,36 +1,165 @@
...
+> Some legacy plugins may still use **Reflux** or a similar mechanism where the
+> state is provided to the top-level component of the plugin as a plain object.
+
+Other than tracking plugin lifetimes, app registries provide a communication
+channel between plugins by being event emitters.
+If possible, new code should avoid this communication channel, as it is untyped
+and the specific method of passing messages should be an implementation detail.
+Alternative methods of passing messages between plugins include using React
+contexts to provide an API that can be used by nested plugins, or if that is
+not possible, your plugin can expose methods that other plugins can then access.
```
himanshusinghs, line 44: What do you feel about including WorkspaceService as an example here?

packages/hadron-app-registry/README.md
```diff
@@ -1,36 +1,165 @@
...
 ```
-npm install --save hadron-app-registry
+
+## Writing a service
+
+Services are consumed by plugins through **service locators**, which are
+functions that return the instance of the service that the plugin is
+intended to use.
+
+Typically, these functions are implemented using React contexts.
```
himanshusinghs, line 89: We might want to add a note along the lines:
Service locator functions are supposed to return values that behave closer to React's refs because these values are injected only once in the plugin when it activates and then even if they change in the "hook" that is acting as service locator it won't reflect the change in the plugin.

Only because this is what I continued to do wrong with the service locators until Sergey pointed it out to me 😅

himanshusinghs (commented):
This is an amazing value addition, thank you for this <3

gribnoysup (approved):
Noticed that some parts of this working on plugins is very out of date, I have two suggestions, but maybe if you feel like it can be part of this ticket, it's worth re-reading through this section in contributing and at least removing stuff that's maybe completely not a thing anymore.

Otherwise looks awesome, thanks for writing this down

packages/hadron-app-registry/README.md
addaleax: True, that's probably close to what the naming here would look like :) Done in 536f2e9c4ce372a52ba2689630ba3dbf8afa3f76

packages/hadron-app-registry/README.md
addaleax: I've added a short note in 36a52c08c0871196358653d525b19b140c52a372 but I don't really want to optimize for this (hopefully) rather uncommon case here :)

packages/hadron-app-registry/README.md
addaleax: Yeah, added something along those lines in 36a52c08c0871196358653d525b19b140c52a372 👍

PR #5734, opened 2024-04-24
fix(compass-connections-navigation, compass-sidebar): connection items workspace behaviour COMPASS-7880

paula-stacho:
## TODO

- [x] add tests

## Description
Incorporating the new `activeWorkspace.connectionId` to the `ConnectionsNavigationTree`.

Note: I have only noticed now that we didn't wrap most tests in `connections-navigation-tree.spec.tsx` with preferences when we added that dependency 🙈 , so I've done that now.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions

I added a TODO related to the currentTabbable. The logic apparently needs to be adjusted as it's using activeNamespace, but not activeConnectionId. But I didn't notice anything not working with the tabbing and I'd need to dig deeper to understand that part - so I'll create a follow up. If anyone knows what currentTabbable is supposed to do, let me know so that we can fill more description for the follow up 😉  @gribnoysup 

https://jira.mongodb.org/browse/COMPASS-7887

## Types of changes

- [ ] *Backport Needed*


packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -391,14 +401,37 @@ const ConnectionsNavigationTree: React.FunctionComponent<
...
+          onDatabaseExpand(activeConnectionId, activeNamespace, true);
+        }
       }
     }
-  }, [activeNamespace, onDatabaseExpand, connections[0]?.connectionInfo.id]);
-  // TODO(COMPASS-7775): the we'll need something similar to expand the active connection
+    // onConnectionExpand is excluded from the deps intentionally -
+    // we only want to expand as a reaction to workspace change, not to a collapse action
+    // otherwise active connections are re-expanded
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
himanshusinghs, line 433: Would it then make sense to wrap onConnectionExpand in a useRef?

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -454,7 +487,7 @@ const ConnectionsNavigationTree: React.FunctionComponent<
   const [rootProps, currentTabbable] = useVirtualNavigationTree<HTMLDivElement>(
     {
       items: items as NavigationTreeData,
-      activeItemId: activeNamespace,
+      activeItemId: activeNamespace || '', // TODO
```
himanshusinghs, line 490: What will this todo be?

himanshusinghs (approved):
Left a small note but otherwise it looks good 🚀

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
paula-stacho: See the open questions in description ;) tldr; I didn't include this bit in the original ticket scope and only stumbled on it during implementation

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -277,14 +278,24 @@ const NavigationItem = memo<{
...
 
+  const {
+    connectionId: activeConnectionId,
+    namespace: activeNamespace,
+    type: activeWorkspaceType,
+  } = activeWorkspace as {
+    connectionId?: string;
+    namespace?: string;
+    type?: string;
+  };
```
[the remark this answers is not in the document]
paula-stacho: I have started out with checks for the type, but in the end I felt like that makes the code unnecessarily dependent on the workspaces. For example I could blacklist `My Queries` because it's the only one that doesn't have connectionId, but later we add `My Tests` and now this doesn't compile. Meanwhile for the logic we only care if the workspace is connection scoped or not.

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
paula-stacho: only for this part - the autoexpanding. but the function is also used for intentional collapse/expand

paula-stacho: hmm. yeah, we could do it. the intentional collapse/expand most likely doesn't have another effect depending on it, and it could always use the current. let me try it.

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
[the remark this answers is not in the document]
paula-stacho: I'll give it another try and perhaps I can only do assertions where the type doesn't matter and still rely on the types where the type is checked anyway

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
paula-stacho: works like a charm! thank you @himanshusinghs !

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
paula-stacho: added ticket number in the comment

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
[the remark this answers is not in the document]
paula-stacho: updated - now the casting is restricted to the autoexpanding bit

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -388,17 +400,43 @@ const ConnectionsNavigationTree: React.FunctionComponent<
...
 
+  const onConnectionExpandRef = useRef(onConnectionExpand);
+  useEffect(() => {
+    onConnectionExpandRef.current = onConnectionExpand;
+  }, [onConnectionExpand]);
+
+  const onDatabaseExpandRef = useRef(onDatabaseExpand);
+  useEffect(() => {
+    onDatabaseExpandRef.current = onDatabaseExpand;
+  }, [onDatabaseExpand]);
```
himanshusinghs, line 411: I think we can simplify this as well, probably something like this?
We won't need effect because having it or not having it will have the same result (you modifying the current of the ref). But regardless it remains the same for the consumer because the consumer is dependent on only the ref and can be sure that at any point in time it will refer to the up-to-date version of current.

```suggestion
  const onConnectionExpandRef = useRef();
  onConnectionExpandRef.current = onConnectionExpand;

  const onDatabaseExpandRef = useRef();
  onDatabaseExpandRef.current = onDatabaseExpand;
```

himanshusinghs (changes requested):

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
himanshusinghs: Ahh got it, thank you 😄

paula-stacho:
@gribnoysup moved the auto-expanding bit to ActiveConnections, checking for workspace.type. With this the only remaining typecast is the one left for [COMPASS-7887](https://jira.mongodb.org/browse/COMPASS-7887)

packages/compass-sidebar/src/components/multiple-connections/active-connections/active-connection-navigation.spec.tsx
```diff
@@ -150,4 +216,111 @@ describe('<ActiveConnectionNavigation />', function () {
...
+              />
+            </Provider>
+          </ConnectionsManagerProvider>
+        </PreferencesProvider>
+      );
+
+      expect(screen.getByText('turtleDB1Coll1')).to.be.visible;
+    });
+
+    it.only('should collapse and expand database', async function () {
```
himanshusinghs, line 301:
```suggestion
    it('should collapse and expand database', async function () {
```

himanshusinghs (approved):
Looking good to me :D

packages/compass-sidebar/src/components/multiple-connections/active-connections/active-connection-navigation.spec.tsx
paula-stacho: oops 😂

paula-stacho:
aaaah, now it hit me that we have to do the auto-expanding for the legacy sidebar 😅 . will prepare right away.

PR #5739, opened 2024-04-25
chore(aggregations): Use document_count from collection stats in the Aggregations input documents' total COMPASS-7799

lerouxb:
This way we don't risk kicking off a colScan on a timeseries collection that will then time out.


packages/compass-aggregations/src/components/pipeline-builder-input-documents.tsx
```diff
@@ -101,12 +102,14 @@ function PipelineBuilderInputDocuments({
...
           ></Icon>
         </IconButton>
-        <Body className={headerTextStyles}>
-          <b>
-            {count} Document{count === 1 ? '' : 's'}
-          </b>{' '}
-          in the collection
-        </Body>
+        {count !== undefined && (
+          <Body className={headerTextStyles}>
```
lerouxb, line 106: Might tweak this. Not sure what to do in the unlikely event that the count becomes available after the docs.

packages/compass-aggregations/src/modules/collection-stats.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export type CollectionStats = Pick<
+  Collection,
+  | 'document_count'
+  | 'index_count'
+  | 'index_size'
+  | 'status'
+  | 'avg_document_size'
+  | 'storage_size'
+  | 'free_storage_size'
```
mabaasit, line 12: do we need to store all this here?

mabaasit (approved):

packages/compass-aggregations/src/modules/collection-stats.ts
lerouxb: No. Changing this.

PR #5755, opened 2024-04-30
chore(compass-aggregations): adapt CreateViewPlugin to work with dynamically provided connectionId for supporting multiple connections COMPASS-7861

himanshusinghs:
## Description
In this PR we modified CreateViewPlugin to work with ConnectionsManager and expect connectionId to be passed alongside the eventParameters for 'open-create-view' event, to support creating views for multiple different connections.

The event 'view-created' that gets emitted after successfully creating view also gets passed with connectionId as extra args so that compass-app-stores can react only when instance's connectionId matches that of event's connectionId.

Notes for reviewers:
---
The first two commits (7f45bc593bdfb3db8a6b0d44e1d61e97f5c1bc45, f6c4baea880b3b680201a65470dc5b65b1e28b95) do not bring any functional change and are meant only to align the module's code with the rest of the modules.

So, for an easy diff experience you can check [those individually](https://github.com/mongodb-js/compass/compare/main...f6c4baea880b3b680201a65470dc5b65b1e28b95) and for the rest maybe check [this](https://github.com/mongodb-js/compass/compare/f6c4baea880b3b680201a65470dc5b65b1e28b95...5dd4643e9b5ba6f31ecc9a0b19bea086da08e388).

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-app-stores/src/stores/instance-store.ts
+++ b/packages/compass-app-stores/src/stores/instance-store.ts
@@ -323,7 +323,15 @@ export function createInstancesStore(
         }
       );
 
-      on(globalAppRegistry, 'view-created', maybeAddAndRefreshCollectionModel);
+      on(
+        globalAppRegistry,
+        'view-created',
+        (ns: string, { connectionId }: { connectionId?: string } = {}) => {
+          if (connectionId === instanceConnectionId) {
+            void maybeAddAndRefreshCollectionModel(ns);
+          }
+        }
+      );
 
       on(
         globalAppRegistry,
--- a/packages/compass-aggregations/src/components/create-view-modal/create-view-modal.tsx
+++ b/packages/compass-aggregations/src/components/create-view-modal/create-view-modal.tsx
@@ -9,11 +9,7 @@ import {
   spacing,
   TextInput,
 } from '@mongodb-js/compass-components';
-import {
-  createView,
-  changeViewName,
-  toggleIsVisible,
-} from '../../modules/create-view';
+import { createView, changeViewName, close } from '../../modules/create-view';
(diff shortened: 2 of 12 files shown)


packages/compass-app-stores/src/stores/instance-store.ts
```diff
@@ -323,7 +323,15 @@ export function createInstancesStore(
         }
       );
 
-      on(globalAppRegistry, 'view-created', maybeAddAndRefreshCollectionModel);
+      on(
+        globalAppRegistry,
+        'view-created',
+        (ns: string, { connectionId }: { connectionId?: string } = {}) => {
+          if (connectionId === instanceConnectionId) {
```
addaleax, line 330: Maybe I'm misremembering (and I admittedly haven't followed each detail of the project), but weren't we supposed to have helpers that specifically avoid the problem of listening to events on `globalAppRegistry`, then having to filter for the specific connection we're interested in?

(If I'm misremembering: That's fine :slightly_smiling_face: If I'm not, and this is a TODO, would it be possible to add a reference to the relevant ticket?)

himanshusinghs: IIRC we purposely kept aside the idea of having the possibility of listening via this ConnectionScopedAppRegistry and limited that only to acts as event forwarder. But there is indeed a ticket (COMPASS-7815) in this project that aims to remove the dependency on globalAppRegistry events in compass-app-stores.

addaleax (approved):

PR #5757, opened 2024-04-30
chore(generative-ai): add accuracy test for recent date, $size, $unwind, and query to agg, refactor test files

Anemy:
COMPASS-7896

Screenshot 2024-04-30 at 1 07 08 PM


packages/compass-generative-ai/scripts/ai-accuracy-tests/ai-accuracy-tests.ts
```diff
@@ -294,10 +234,18 @@ const runOnce = async (
         (type === 'query' &&
           acceptAggregationResponse &&
           aggregation.pipeline &&
-          aggregation.pipeline !== '[]')
+          aggregation.pipeline !== '[]' &&
+          // When we don't have a query, we use the aggregation pipeline.
+          !hasQueryFields(query))
```
alenakhineika, line 239: Isn't it the case that we have right now in Compass? So that we have the `{}` query and are not being redirected to aggregation? hasQueryFields check all props explicitly, what probably hides in tests such cases.

Anemy: I'm not totally sure I understand your question. This check ensures that we don't run a pipeline when there are query fields attached. That's how Compass currently behaves, it prefers query fields over aggregations. We do that so that we're sure the model is returning an aggregation in the situations when only an aggregation will fulfill the objective.

alenakhineika (approved):

PR #5758, opened 2024-04-30
chore(settings): show gen ai sample docs setting when dev feature flags are enabled

Anemy:
Exposes the hidden setting when dev feature flags are enabled so folks can try it out. 
We could make this more generalized and have `devOnly` be a property of a setting, but this is more of a one off case. In future cases I think we'd hide new settings that aren't to be brought to production immediately behind feature flags.


mcasimir:
not a big deal, but i was also expecting a regular "dev" feature flag here that enables showing the other "checkbox" for the setting. 

I get that it may feel weird to have a setting that only enables another setting, but I don't see why this should be different from any other unreleased feature we have ever had. When we are ready to release to everybody we will move the FF from 'dev' to 'released' as usual and then remove it.

PR #5761, opened 2024-05-01
chore(compass-import-export): adapt ImportPlugin to work with dynamically provided connectionId for supporting multiple connections COMPASS-7859

himanshusinghs:
## Description
In this PR we modified ImportPlugin to work with dynamically provided DataService. The event `open-import` now expects `connectionId` to be provided as extra event args and will start the import for the specified connectionId. Places where we emit `open-import` now also provides connectionId as extra args.

When import is finished we emit `import-finished` event, now also with connectionId. The places where we were listening to this event now handles responding to it only when the connectionId matches that of event's connectionId.

As a drive by I also updated the events emitted by compass-crud to be emitted via ConnectionScopedAppRegistry. This is so that those events always carry a connectionId as extra args. In response to this change, compass-app-stores is also modified to only respond when the connectionId matches that of Instance's connectionId.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-aggregations/src/stores/store.ts
+++ b/packages/compass-aggregations/src/stores/store.ts
@@ -202,12 +202,21 @@ export function activateAggregationsPlugin(
     refreshInput();
   });
 
-  on(globalAppRegistry, 'import-finished', ({ ns }) => {
-    const { namespace } = store.getState();
-    if (ns === namespace) {
-      refreshInput();
+  on(
+    globalAppRegistry,
+    'import-finished',
+    (
+      { ns }: { ns: string },
+      { connectionId }: { connectionId?: string } = {}
+    ) => {
+      const { id: currentConnectionId } =
+        connectionInfoAccess.getCurrentConnectionInfo();
+      const { namespace } = store.getState();
+      if (currentConnectionId === connectionId && ns === namespace) {
+        refreshInput();
+      }
     }
-  });
+  );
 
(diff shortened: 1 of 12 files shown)


paula-stacho:
Exporting documents in multi-connections doesn't work, is this part of another task?

paula-stacho:
I'm wondering if the ConnectionScopedAppRegistry couldn't also offer `on` method with pre-filtered events for connection. What do you think @himanshusinghs ? Definitely not for this PR, more of a general question.

himanshusinghs:
> Exporting documents in multi-connections doesn't work, is this part of another task?
Yea that's right Paula. Its part of another task.

> ConnectionScopedAppRegistry couldn't also offer on method with pre-filtered events for connection.
There was a discussion about ConnectionScopedAppRegistry in the very beginning of the project and the decision was to keep the scope limited to connection scoped events. The reason was that communication via GlobalAppRegistry is not recommended and we ideally should step away from it as much as possible. ConnectionScopedAppRegistry is only meant to make it easy until we bridge the gap and come to an ideal state so making it feature rich would have been against that idea.
Having said that, after going through last few PRs, I feel like discussing this again with the rest of the team.

paula-stacho (approved):

PR #5763, opened 2024-05-02
feat(query-bar, aggregations): add visual effect for ai generated fields, aggregations, and loaded events COMPASS-7835

Anemy:
COMPASS-7835

Adds a background flashing effect for when a query is generated or loaded from history. Also for aggregations. As a drive-by I fixed the container background on darkmode for the aggregation editors and made it so a query opened from query history with extra options opens the options tray, similar to what we do with generated queries.

Implementation wise this does add a good bit of code for a visual effect, 2 flags and a clearing method, I'm open to any less intrusive or more minimal implementation ideas.

Here's how it looks:
___
**Query bar**

https://github.com/mongodb-js/compass/assets/1791149/f4a000b1-30ae-4654-b972-7f8ad1719fe6

**Aggregations**

https://github.com/mongodb-js/compass/assets/1791149/428e58c4-d281-464b-bea8-eba9c48f6d69 

https://github.com/mongodb-js/compass/assets/1791149/e06fb641-90ec-40f5-9f11-9ad7308432b6

### DIFF
--- a/packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-editor.spec.tsx
+++ b/packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-editor.spec.tsx
@@ -23,9 +23,8 @@ const renderPipelineEditor = (
         serverVersion="4.2"
         onChangePipelineText={() => {}}
         num_stages={1}
-        isPipelineLoadedFromExternal={false}
-        pipelineLoadedFromExternalKey={0}
-        onClearIsPipelineLoadedFromExternal={() => {}}
+        loadedFromExternalSourceId={null}
+        onClearIsPipelineLoadedFromExternalSource={() => {}}
         {...props}
       />
     </Provider>
--- a/packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-editor.spec.tsx
+++ b/packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-editor.spec.tsx
@@ -23,6 +23,9 @@ const renderPipelineEditor = (
         serverVersion="4.2"
         onChangePipelineText={() => {}}
         num_stages={1}
+        isPipelineLoadedFromExternal={false}
+        pipelineLoadedFromExternalKey={0}
+        onClearIsPipelineLoadedFromExternal={() => {}}
         {...props}
       />
     </Provider>
--- a/packages/compass-aggregations/src/modules/is-pipeline-loaded-from-external.ts
+++ b/packages/compass-aggregations/src/modules/is-pipeline-loaded-from-external.ts
@@ -1,0 +1,103 @@
+import type { Reducer } from 'redux';
+
+import type { NewPipelineConfirmedAction } from './is-new-pipeline-confirm';
+import { ActionTypes as ConfirmNewPipelineActions } from './is-new-pipeline-confirm';
+import { ActionTypes as PipelineModeActionTypes } from './pipeline-builder/pipeline-mode';
+import type { PipelineModeToggledAction } from './pipeline-builder/pipeline-mode';
+import type {
+  LoadGeneratedPipelineAction,
+  PipelineGeneratedFromQueryAction,
+} from './pipeline-builder/pipeline-ai';
+import { AIPipelineActionTypes } from './pipeline-builder/pipeline-ai';
+import type { RestorePipelineAction } from './saved-pipeline';
+import { RESTORE_PIPELINE } from './saved-pipeline';
+import { isAction } from '../utils/is-action';
+import type { EditorValueChangeAction } from './pipeline-builder/text-editor-pipeline';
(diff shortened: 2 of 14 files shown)


packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-editor.spec.tsx
```diff
@@ -23,6 +23,9 @@ const renderPipelineEditor = (
         serverVersion="4.2"
         onChangePipelineText={() => {}}
         num_stages={1}
+        isPipelineLoadedFromExternal={false}
+        pipelineLoadedFromExternalKey={0}
```
alenakhineika, line 27: When `pipelineLoadedFromExternalKey` is `0`, does it always mean that `isPipelineLoadedFromExternal` `false`? I am wondering if we need `isPipelineLoadedFromExternal` at all, or if we could get the same usage from just `pipelineLoadedFromExternalKey`.

packages/compass-aggregations/src/modules/is-pipeline-loaded-from-external.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export function clearIsPipelineLoadedFromExternal(): ClearIsPipelineLoadedFromExternalAction {
+  return {
+    type: PipelineLoadedFromExternalActionTypes.ClearIsPipelineLoadedFromExternal,
+  };
+}
+
+export const INITIAL_STATE: IsPipelineLoadedFromExternal = {
+  isPipelineLoadedFromExternal: false,
+  pipelineLoadedFromExternalId: 0,
```
alenakhineika, line 52: Can this be also named `pipelineLoadedFromExternalKey`? This is the same thing, right?

packages/compass-aggregations/src/components/pipeline-builder-workspace/pipeline-as-text-workspace/pipeline-editor.spec.tsx
mcasimir: usual comment about naming, what does "PipelineLoadedFromExternal" means? is there a name that would relate to this feature better?

Also more in general, i would have expected a bit less code, especially a bit less, and less spread around code for this feedback feature. Are we sure is worth?

Anemy: I've renamed it to `LoadedFromExternalSourceId`, how do you feel about that?

I removed the extra boolean, and hopefully cleaned up the interfaces a bit and reduces some of the code overhead.

I can't really think of a cleaner way apply an effect like this. We have to keep a flag around outside of rendering and ensure we clear it. We could do something hacky with applying a css class name when this happens directly from where it happens but I think that could be worse and coupled.

In terms of if it's worth it I'm not sure. I can bring it up with design, it does feel like a small visual impact that other things like the animated border and disabled inputs now assist in. Do you feel with the recent code changes this is unwieldily and not something we want to maintain?

alenakhineika: Maybe Maurizio was asking not necessarily about the `id` ending, for me it is not exactly clear what `LoadedFromExternalSource` means. Would we apply the same styles in the future when we load a query from the saved queries? Could you please also point out why we need this id? Because we can have several queries open on multiple tabs and we need to know which one we are working with?

mcasimir: yeah, i have questions about "what is a source here", and what is an external source, is not really a concept we talk about outside of code, so i'm not sure if there is some naming that would speak more domain language, like `isPiplelineAIGenerated`, if that is what we are trying to track here.

Anemy: > Do we apply the same styles when we load a query from the saved queries
Yes
> Could you please also point out why we need this id?
We need this id in order for the frontend to know a new query has been applied. We need the effect to re-apply if a user run the same prompt twice and gets the same response back. A boolean could work but it would introduce more roundtrips in the component life cycle and each component that listens to the boolean would be attempting to un-set it once it has been set.

Anemy: > `isPipelineAIGenerated`
The effect is also shown when we load a pipeline from saved pipelines or query history. It's not something we need to do, we could not show it for those events and then we could use a naming like this. I was thinking something more generic in order to cover those extra use cases.

alenakhineika: Maybe something like `appliedChangesVersion`?

Anemy:
Chatted with Julia and Maurizio on this a bit more. Going to close. We can revisit and do this later if we feel it's still an issue. The updates to the in progress state, highlight & disabled inputs, likely already make the experience better. It's also nice to avoid code that will have some maintenance costs for a small visual thing.

PR #5774, opened 2024-05-03
chore(ci): add empty .yml for ai-accuracy test evergreen project to use to pick up recent commits

Anemy:
Slack thread: https://mongodb.slack.com/archives/C0V896UV8/p1714745238558629

Error that happens without a .yml to point to:
Screenshot 2024-05-03 at 3 17 21 PM

### DIFF
--- a/.evergreen/generative-ai-accuracy-test-empty.yml
+++ b/.evergreen/generative-ai-accuracy-test-empty.yml
@@ -1,0 +1,9 @@
+# This evergreen .yml is intentionally left empty.
+# We want the compass-generative-ai-accuracy evergreen project to run
+# daily on the latest commit. This file is here so that we have a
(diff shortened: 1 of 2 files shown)


addaleax (approved):
Just to make sure I understand correctly: The empty file is so that mainline evergreen versions don't fail?

Anemy:
@addaleax The empty file is used for us to set the config path to in the evergreen project settings, so we can successfully track the most recent commit with the repotracker. We need the repotracker so that when the periodic build happens, it uses the most recent commit. There has to be a .yml set in the project settings for it to follow commits, otherwise that error in the screenshot in the description happens. It's a bit hacky, and I think it should be something that we can avoid, which is why I asked in the channel, it appears like this is an alright approach.

PR #5781, opened 2024-05-06
chore(settings): add show gen ai sample docs setting feature flag

Anemy:
Enables showing the sample field value setting behind a dev feature flag.

### DIFF
--- a/packages/compass-preferences-model/src/feature-flags.ts
+++ b/packages/compass-preferences-model/src/feature-flags.ts
@@ -19,6 +19,7 @@ export type FeatureFlags = {
   showInsights: boolean;
   enableRenameCollectionModal: boolean;
   enableNewMultipleConnectionSystem: boolean;
+  showGenAIPassSampleDocumentsSetting: boolean;
 };
 
 export const featureFlags: Required<{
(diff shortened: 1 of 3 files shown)


packages/compass-preferences-model/src/feature-flags.ts
```diff
@@ -73,4 +74,17 @@ export const featureFlags: Required<{
...
+  /**
+   * Feature flag for showing the option to pass sample documents with our query and aggregation generation requests.
+   * Enables showing the `enableGenAISampleDocumentPassing` setting in the settings UI so folks can turn it on.
+   * Epic: COMPASS-7584
+   */
+  showGenAIPassSampleDocumentsSetting: {
+    stage: 'development',
+    description: {
+      short: 'Enable showing the sample document gen ai setting.',
+      long: 'Allows users to show a setting to enable the passing of sample field values with our query and aggregation generation requests.',
```
alenakhineika, line 87:
```suggestion
      long: 'Allows sending sample documents from a user's collection to AI when generating queries and aggregations for a natural language .',
```
Maybe something like this? The idea is that the end goal is to enable the sample documents feature, not pointing directly to the setting, but thinking broader.

alenakhineika (approved):

PR #5788, opened 2024-05-07
feat(compass-sidebar): padding around add connection button, don't show zero count in multiple connections

lerouxb:
The button was already there, it just needed some styling. As a drive-by I'm also hiding the zero count like the design specifies.

before:
before

after with 0 connections:
after_0

after with 1 connection:
after_1

### DIFF
--- a/packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection-list.spec.tsx
+++ b/packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection-list.spec.tsx
@@ -113,8 +113,8 @@ describe('SavedConnectionList Component', function () {
     });
 
     it('should not show the empty connections message', function () {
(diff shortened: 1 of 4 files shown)


himanshusinghs (approved):

packages/compass-sidebar/src/components/multiple-connections/saved-connections/saved-connection-list.spec.tsx
```diff
@@ -113,8 +113,8 @@ describe('SavedConnectionList Component', function () {
     });
 
     it('should not show the empty connections message', function () {
-      expect(screen.queryByText('You have not connected to any deployments')).to
-        .be.null;
+      expect(screen.queryByText('You have not connected to any deployments.'))
```
lerouxb, line 116: Unit tests that cannot fail 😆

PR #5812, opened 2024-05-13
feat(compass): update close window hotkey to be cmd+shift+w to avoid conflict with close tab cmd+w COMPASS-7301

Anemy:
COMPASS-7301

The issue reported in the ticket is that the code mirror captures the `Cmd+w`, which makes the close tab hotkey not run, however the close window electron hotkey still runs. This can be frustrating when there's something happening in other tabs. This pr mitigates the issue by updating close window to be `Cmd+Shift+w`.

### DIFF
--- a/packages/compass/src/main/menu.spec.ts
+++ b/packages/compass/src/main/menu.spec.ts
@@ -354,7 +354,7 @@ describe('CompassMenu', function () {
           label: 'Window',
           submenu: [
             { label: 'Minimize', accelerator: 'Command+M', role: 'minimize' },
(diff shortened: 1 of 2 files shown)


paula-stacho (approved):
do we need to update some docs?

Anemy:
@paula-stacho Yup! We'll update the https://www.mongodb.com/docs/compass/current/keyboard-shortcuts/ page. When closing this ticket I'll be marking the `Documentation Changes` section on the jira ticket as `Needed` with a description of what's changing.

PR #5820, opened 2024-05-20
chore(ci): fix snyk test hanging COMPASS-7947

mcasimir:
Tests only the production dependencies calling the snyk api. This should be needed only temporarily, while we replace it with the silk based system.


lerouxb (approved):

addaleax:
Do you know why the test was hanging before?

> This should be needed only temporarily, while we replace it with the silk based system.

We'll still be using Snyk to scan dependencies, just to be clear

mcasimir:
> Do you know why the test was hanging before?

On my machine it was taking hours on a couple of packages, it became really slow over time, not sure why, initially was only taking 10-20 minutes.

> We'll still be using Snyk to scan dependencies, just to be clear

Isn't that what Silk should be doing, getting vulnerabilities for dependencies and adding them to the sbom?

Happy to keep this code, in fact is much quicker and works so much better than the other alternative that we should probably just replace it in sbom-tools if it needs to stay.

addaleax:
> Isn't that what Silk should be doing, getting vulnerabilities for dependencies and adding them to the sbom?

I agree that it _should_ be doing that, but it isn't. It's just going to integrate the vulnerabilities found through Snyk into the SBOM, but we still need to run the Snyk scanning ourselves. (I don't know _why_ this can't be simpler, no.)

mcasimir:
>  we still need to run the Snyk scanning ourselves.

how do we report the vulnerabilities that we find?

addaleax:
@mcasimir My understanding is that we use Snyk to discover vulnerabilities and then Silk's integration with Snyk will take care of augmenting the SBOM, creating JIRA tickets, etc.

PR #5824, opened 2024-05-21
fix(connection-form): don't force the modal larger, prevent it from going too big

lerouxb:
Screenshot 2024-05-21 at 18 06 12
Screenshot 2024-05-21 at 18 06 09
Screenshot 2024-05-21 at 18 05 58
Screenshot 2024-05-21 at 18 05 55

### DIFF
--- a/packages/connection-form/src/components/connection-form.tsx
+++ b/packages/connection-form/src/components/connection-form.tsx
@@ -65,7 +65,7 @@ const formContainerStyles = css({
 const formContentStyles = css({
   display: 'flex',
   columnGap: spacing[3],
-  height: `calc(100vh - 310px)`,
+  maxHeight: `calc(100vh - 310px)`,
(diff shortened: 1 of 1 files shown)


packages/connection-form/src/components/connection-form.tsx
```diff
@@ -65,7 +65,7 @@ const formContainerStyles = css({
 const formContentStyles = css({
   display: 'flex',
   columnGap: spacing[3],
-  height: `calc(100vh - 310px)`,
+  maxHeight: `calc(100vh - 310px)`,
```
lerouxb, line 68: We can also just make this maxHeight: 720px. In theory this calc means we could have a suuuper skinny modal, but something would have to try and make it that big to begin with.

himanshusinghs (approved):

PR #5837, opened 2024-05-24
chore(virtual-list): refactor and fix tabbable in the sidebar COMPASS-7887	 

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-connections-navigation/src/utils.ts
```diff
@@ -1,44 +1,12 @@
...
-  }
-
-  return { paddingLeft };
+  maxNestingLevel,
+}: Pick<SidebarTreeItem, 'level' | 'maxNestingLevel'>): React.CSSProperties => {
+  const isExpandable = level < maxNestingLevel;
+  const defaultPadding = 20;
+  return {
+    paddingLeft: (level - 1) * defaultPadding + (!isExpandable ? 30 : 0),
+  };
```
paula-stacho, line 11: single connection needs 8px extra, because the legacy list goes all the way to the left:
Screenshot 2024-05-24 at 13 09 01

himanshusinghs (commented):
Great refactor, looking good 🚀 Left small suggestions, let me know what you think?

packages/compass-connections-navigation/src/tree-data.ts
```diff
@@ -0,0 +1,288 @@
...
+};
+
+type Status = 'initial' | 'fetching' | 'refreshing' | 'ready' | 'error';
+
+type Database = {
+  _id: string;
+  name: string;
+  collectionsStatus: Status;
+  collectionsLength: number;
+  collections: Collection[];
```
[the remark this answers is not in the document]
himanshusinghs: nevermind - just noticed that this is coming from mapState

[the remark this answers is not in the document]
gribnoysup: It's either number from db stats or actual collection length, so yes, it will be different while we're fetching actual collections list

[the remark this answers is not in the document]
himanshusinghs: ahh i see, thank you for the context 👍

packages/compass-connections-navigation/src/navigation-item.tsx
```diff
@@ -0,0 +1,129 @@
...
+      return <Icon glyph="Visibility" />;
+    }
+    if (item.type === 'timeseries') {
+      return <Icon glyph="TimeSeries" />;
+    }
+    if (item.type === 'connection') {
+      const isLocalhost =
+        item.connectionInfo.connectionOptions.connectionString.startsWith(
+          'mongodb://localhost'
+        ); // TODO(COMPASS-7832)
```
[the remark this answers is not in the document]
himanshusinghs: resolving this as we'll tackle this in the mentioned PR

himanshusinghs (approved):

packages/compass-connections-navigation/src/tree-data.ts
```diff
@@ -0,0 +1,288 @@
...
+  // of items in the list. So we will generate the list first and then map it to have correct setSize and posInSet.
+  const setSize = treeList.length;
+  return treeList.map((item, index) => {
+    const posInSet = index + 1;
+    return {
+      ...item,
+      setSize,
+      posInSet,
+    };
+  });
```
gribnoysup, line 287: I don't think this is correct. The setsize and posinset should be set relative to the tree level, not for the whole tree, this is what aria spec expects, it's not just a total number of items in the list

mabaasit: fixed it in [a082a00](https://github.com/mongodb-js/compass/pull/5837/commits/a082a003b94ee45ab865f60e74ef0066311a7169)

mabaasit: nice catch :)

packages/compass-connections-navigation/src/tree-item.tsx
```diff
@@ -23,7 +21,7 @@ const expandButton = css({
   // Not using leafygreen spacing here because none of them allow to align the
   // button with the search bar content. This probably can go away when we are
   // rebuilding the search also
-  padding: 7,
+  padding: 6,
```
gribnoysup, line 24: This doesn't match the design (along with other styles here), but probably worth opening another ticket to align those because there is a bunch of small inconsistencies in the list when compared to what's in figma:

This padding should be 8:

image

The icon next to it should be 14, not 16 px:

image

Padding between icons and text is 8:

image


There is a radius on a selected item background:

image

We should also probably just use the ticket we will open to make the same styles apply in "single connection mode" so that we don't need to continuosly fix style issues in two places for no reason

packages/compass-connections-navigation/src/virtual-list/virtual-list.tsx
```diff
@@ -0,0 +1,217 @@
...
+function TreeItem<T extends VirtualItem>({
+  index,
+  data,
+  style,
+}: ListChildComponentProps<VirtualItemData<T>>) {
+  const { renderItem, items } = data;
+  const item = useMemo(() => items[index], [items, index]);
+  const focusRingProps = useFocusRing();
+
+  const Component = useMemo(() => {
```
gribnoysup, line 185: Nit: uppercase is usually used for component constructors, this is already an element that you "rendered"
```suggestion
  const component = useMemo(() => {
```

packages/compass-connections-navigation/src/tree-data.ts
```diff
@@ -0,0 +1,289 @@
...
+  isReady: boolean;
+  isDataLake: boolean;
+  isWritable: boolean;
+  isPerformanceTabSupported: boolean;
+};
+
+type PlaceholderTreeItem = VirtualPlaceholderItem & {
+  colorCode?: string;
+  id: string;
+  maxNestingLevel: number;
```
gribnoysup, line 43: Feel free to keep, but this belongs more in the virtual tree item data not in the item itself, tree item data is specifically a mechanism that allows us to pass shared values to all items.

gribnoysup (approved):
A few things caught my eye, so I left a few notes, but overall a great improvement and much simpler to navigate this code now, nice work!

packages/compass-connections-navigation/src/tree-data.ts
mabaasit: good idea. will make this change.

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -416,144 +46,141 @@ const ConnectionsNavigationTree: React.FunctionComponent<
...
-      activeItemId:
-        (activeWorkspace as { namespace?: string })?.namespace || '', // TODO(COMPASS-7887)
-      onExpandedChange,
-      onFocusMove,
-    }
+  const onItemAction = useCallback(
+    (item: SidebarActionableItem, action: Actions) => {
+      const args =
+        item.type === 'connection'
+          ? ([item.connectionInfo.id, item.connectionInfo.id, action] as const)
```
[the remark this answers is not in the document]
mabaasit: the existing behaviour. i think this (the second id) is used as `ns` in [onNamespaceAction](https://github.com/mongodb-js/compass/blob/2791b3324dd2dda47261228bbdd1829a713badc0/packages/compass-sidebar/src/components/multiple-connections/active-connections/active-connection-navigation.tsx#L162). It should be an optional param though. I can create a follow up for this to clean it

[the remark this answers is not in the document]
himanshusinghs: Will also take into account what you mentioned Sergey and just have one approach as I am still mixing things up.

mabaasit:
As a follow up, I created [COMPASS-8000](https://jira.mongodb.org/browse/COMPASS-8000) to address UX issues. Other issues are addressed in #5886

PR #5849, opened 2024-05-28
feat(data-service): expose close event VSCODE-536

alenakhineika:
## Description

Expose the `close` so we can listen on it in VSCode.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/data-service/src/data-service.ts
+++ b/packages/data-service/src/data-service.ts
@@ -2278,7 +2278,7 @@ class DataServiceImpl extends WithLogContext implements DataService {
       });
 
       client.on('close', () => {
-        this._logger.info(mongoLogId(1_001_000_288), 'Client is closed');
+        this._logger.info(mongoLogId(1_001_000_315), 'Client is closed');
         void this.disconnect();
         this._emitter.emit('close');
       });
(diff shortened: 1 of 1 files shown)


packages/data-service/src/data-service.ts
```diff
@@ -2280,6 +2281,7 @@ class DataServiceImpl extends WithLogContext implements DataService {
         this._logger.info(mongoLogId(1_001_000_020), 'Server closed', {
           address: evt.address,
         });
+        this._emitter.emit('serverClosed');
```
addaleax, line 2284: How will we be using this event in VSCode?

For https://github.com/mongodb-js/vscode/pull/734, we'll need the `client.on('close')` event, not `serverClosed` (which can be _any_ server of the topology, e.g. a replica set member being taken down for maintenance while the replica set as a whole remains available)

alenakhineika: I think we can do `this._activeDataService?.once('serverClosed', ...)`?

addaleax (approved):

PR #5857, opened 2024-06-03
feat(settings): enable sample documents setting COMPASS-7931

Anemy:
COMPASS-7931

The setting is still disabled by default, this enables showing it to all users by removing the feature flag.

### DIFF
--- a/packages/compass-preferences-model/src/feature-flags.ts
+++ b/packages/compass-preferences-model/src/feature-flags.ts
@@ -81,7 +81,7 @@ export const featureFlags: Required<{
    * Epic: COMPASS-7584
    */
   showGenAIPassSampleDocumentsSetting: {
-    stage: 'development',
+    stage: 'released',
     description: {
       short: 'Enable showing the sample document gen ai setting.',
       long: 'Allows users to show a setting to enable the passing of sample field values with our query and aggregation generation requests.',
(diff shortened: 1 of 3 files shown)


packages/compass-preferences-model/src/feature-flags.ts
```diff
@@ -81,7 +81,7 @@ export const featureFlags: Required<{
    * Epic: COMPASS-7584
    */
   showGenAIPassSampleDocumentsSetting: {
-    stage: 'development',
+    stage: 'released',
     description: {
       short: 'Enable showing the sample document gen ai setting.',
```
addaleax, line 86:
```suggestion
      short: 'Enable showing the sample document Generative AI setting.',
```

Should this even be a feature flag? To me this makes sense as just a regular setting that we could expose

Anemy: There is a setting that this feature flag, when enabled, shows. So this feature flag is for showing the setting 😅 
Screenshot 2024-06-03 at 1 48 39 PM

mcasimir: We could probably drop the FF entirely in this case? or we think we would roll this one back?

mcasimir (approved):

packages/compass-preferences-model/src/feature-flags.ts
addaleax: I mean, we can keep it, just feels a bit like we don't need two separate settings in a situation like this 🙂

Anemy: We went for a feature flag so that product folks could try out the setting before we released it. In general I agree, a bit of a specialized case here. I think dropping the feature flag makes sense, I don't see it being rolled back. I'll push up that change momentarily.

PR #5868, opened 2024-06-05
fix: publish key prefix to download update correctly

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


.evergreen/functions.yml
```diff
@@ -452,14 +453,14 @@ functions:
           <<: *compass-env
         script: |
           eval $(.evergreen/print-compass-env.sh)
-          .evergreen/publish-dev-release.sh LATEST
+          .evergreen/publish-dev-release.sh release.json
```
mcasimir, line 456: nit: i noticed this while looking for artifacts yesterday, but both the script name "publish-dev-release" and the task name publish-dev-release do not actually publish the dev release, which is kinda unexpected for someone who wasn't too involved.

How about we rename the script as "build-dev-release-info" and the task `dev-release-info` or something?

.evergreen/publish-dev-release.sh
```diff
@@ -13,9 +13,22 @@ if [[ "${EVERGREEN_BRANCH_NAME}" != "main" ]]; then
   exit 0;
 fi
 
+JSON_CONTENT=$( jq -n \
+  --arg id "$DEV_VERSION_IDENTIFIER" \
+  --arg key "${EVERGREEN_REVISION}_${EVERGREEN_REVISION_ORDER_ID}" \
+  '{version: $id, bucket_key_prefix: $key}'
+)
+
 URL="https://mciuploads.s3.amazonaws.com/${EVERGREEN_PROJECT}/compass/dev/$1"
```
mcasimir, line 22: NIT: is it really necessary to parametrize this script, will we ever use it to produce multiple files?

mcasimir (approved):

.evergreen/publish-dev-release.sh
mabaasit: no, but i did that so that when we use script in [functions.yml](https://github.com/mongodb-js/compass/blob/main/.evergreen/functions.yml#L455), user would know the expected file name and then use same in the function to upload the file to s3, or else filename is spread.

.evergreen/functions.yml
mabaasit: made the change in [e0b94e6](https://github.com/mongodb-js/compass/pull/5868/commits/e0b94e64cdbaba312aeb0a59eb1d8b4c9e0fcc92)

PR #5889, opened 2024-06-10
feat(query-bar): expand options when applied from query history with options

Anemy:
Previously you could click on something from the history and it wouldn't open options. This could cause some confusion if a user doesn't see something like a sort or limit applied.

https://github.com/mongodb-js/compass/assets/1791149/8d26e491-5951-4121-b976-29317124f753

Made a thread on slack to discuss: 
Will wait to merge for folks to respond there.

### DIFF
--- a/packages/compass-query-bar/src/stores/query-bar-reducer.spec.ts
+++ b/packages/compass-query-bar/src/stores/query-bar-reducer.spec.ts
@@ -226,6 +226,21 @@ describe('queryBarReducer', function () {
         '{ _id: -1 }'
       );
     });
+
+    it('should auto expand when the query contains extra options', function () {
+      const queryNoExtraOptions = {
+        filter: { _id: 2 },
+      };
+      store.dispatch(applyFromHistory(queryNoExtraOptions));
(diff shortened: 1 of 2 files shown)


packages/compass-query-bar/src/stores/query-bar-reducer.spec.ts
```diff
@@ -226,6 +226,21 @@ describe('queryBarReducer', function () {
         '{ _id: -1 }'
       );
     });
+
+    it('should auto expand when the query contains extra options', function () {
```
mcasimir, line 230: do we need a counter test to make sure the qb doesn't expand if options aren't present?

mcasimir (approved):

packages/compass-query-bar/src/stores/query-bar-reducer.spec.ts
Anemy: There's a counter test inside of this test already, it would have been better practice to explicitly have it as a separate test though, good callout.

PR #5897, opened 2024-06-11
fix(ci): run static analysis reporting only once per version COMPASS-7998

addaleax:


.evergreen.yml
```diff
@@ -1,4 +1,3 @@
-unset_function_vars: true
```
addaleax, line 1: Drive-by: This field does not exist anymore

.evergreen/functions.yml
```diff
@@ -331,7 +331,6 @@ functions:
         wait: true
     # copy ssh key (this key will be used to connect to ssh server)
     - command: shell.exec
-      silent: true
```
addaleax, line 334: Drive-by: `silent` would need to be set on the `params:`, not the command, if we wanted it to have an effect – but this should be fine either way, so let's just drop the invalid extra key

mabaasit (approved):

mcasimir (approved):

PR #5905, opened 2024-06-13
fix: finish readOnly logic COMPASS-7972

paula-stacho:
## Description
Fixes:
- changes in preferences weren't applied immediately (something wasn't re-rendering)

Other changes:
- disable shell when preferences readOnly mode is on, or the connection isn't writable (which isn't the same as the connection item being readOnly! we allow the shell when `instance.isDataLake`)

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-connections-navigation/src/connections-navigation-tree.tsx
+++ b/packages/compass-connections-navigation/src/connections-navigation-tree.tsx
@@ -66,8 +66,13 @@ const ConnectionsNavigationTree: React.FunctionComponent<
   const id = useId();
 
   const treeData = useMemo(() => {
-    return getVirtualTreeItems(connections, isSingleConnection, expanded);
-  }, [connections, isSingleConnection, expanded]);
+    return getVirtualTreeItems({
+      connections,
+      isSingleConnection,
+      expandedItems: expanded,
+      preferencesReadOnly,
+    });
+  }, [connections, isSingleConnection, expanded, preferencesReadOnly]);
 
   const onDefaultAction: OnDefaultAction<SidebarActionableItem> = useCallback(
     (item, evt) => {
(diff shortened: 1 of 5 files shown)


himanshusinghs (approved):
Noice <3

packages/compass-connections-navigation/src/connections-navigation-tree.tsx
```diff
@@ -112,7 +117,8 @@ const ConnectionsNavigationTree: React.FunctionComponent<
             item.connectionInfo?.savedConnectionType === 'favorite';
           if (item.connectionStatus === ConnectionStatus.Connected) {
             return connectedConnectionItemActions({
-              isReadOnly: preferencesReadOnly || item.isConnectionReadOnly,
+              isReadOnly: item.isReadOnly,
```
mcasimir, line 120: NIT: as you have probably noticed "readonly" is overloaded with meanings, and even if here is attached to item, probably is better to use a longer more descriptive name that reveals its intent?

What is this flag controlling in this case?

lerouxb: This is only to filter out the "write" options from the popover menu, right? So maybe a var name or set of var names that unambiguously reflects that?

paula-stacho: @lerouxb what do you mean by var name?

paula-stacho: does `areWriteActionsEnabled` sound better?

lerouxb (approved):

PR #5921, opened 2024-06-14
fix(ci): include releaser name in application artifacts COMPASS-7888

addaleax:
- Add our team's official email addresses and github usernames to `.mailmap` (we'll need to be using our mongodb.com email addresses soon anyway according to our DBX repo policy)
- Map the username for a newly triggered GH release to the name and email through the mailmap (and verify it's a mongodb.com address)
- Include it in the Compass package.json when we update it, ~~and include the package.json file in the generated app bundle~~ and use that to submit the information to papertrail
- Kick off CodeQL for the pushed, tagged commit from the release script as well

### DIFF
--- a/packages/compass/package.json
+++ b/packages/compass/package.json
@@ -123,7 +123,8 @@
           "**/macos-export-certificate-and-key/**",
           "**/system-ca/**",
           "**/node-forge/**",
-          "**/mongo_crypt_v1.*"
+          "**/mongo_crypt_v1.*",
+          "**/compass-application-package.json"
         ]
       },
       "rebuild": {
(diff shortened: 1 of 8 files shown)


packages/compass/package.json
```diff
@@ -149,6 +150,7 @@
     "test-renderer": "xvfb-maybe electron-mocha --no-sandbox --config ./.mocharc.renderer.js \"./src/app/**/*.spec.*\"",
     "check": "npm run typecheck && npm run lint && npm run depcheck",
     "webpack": "webpack-compass",
+    "postwebpack": "cp package.json build/assets/compass-application-package.json",
```
addaleax: Actually, I'll be checking with Dylan to see if we can integrate this with papertrail properly (just need to figure out how to manually set a submitter)

PR #5922, opened 2024-06-14
chore(e2e): use executeAsync rather than execute for async code

lerouxb:
Follow-up to https://github.com/mongodb-js/compass/pull/5915

I want to see if the timeout I added works for execute because the docs say it only works for executeAsync. Which might explain why I've seen execution hang indefinitely on the browser.execute inside browser.setFeature and then the afterEach hook it was in eventually hit the mocha timeout.

I'm waiting to see if that still happens and if it does we can try this. Still doesn't explain _why_ it would hang, but if it could at least time out and fail the relevant hook/test then that's already an improvement and will aid debugging.

(If this makes a difference then there are one or two other places we might also want to change.)


lerouxb:
I'm going to close this for now and can re-open it if it comes up again.

PR #5923, opened 2024-06-14
fix(compass-connections-navigation): fix for action button not being tabbable when tabbing through navigation tree

himanshusinghs:
## Description
This is a problem we identified while finalising the unified sidebar experience that when you tab through the navigation tree, the action buttons are not tabbable. This happens because the action buttons are never rendered and because of which when we tab, they never have a chance to capture the next focus.

Another issue that was only highlighted in the unified sidebar experience was that the active item was not getting highlighted even when navigation changed. Now this is not a problem in main because the the connections list changes all the time which leads to re-render.

Before:

https://github.com/mongodb-js/compass/assets/10037761/6fc035d5-e002-414c-b6cf-68a195086668

After:

https://github.com/mongodb-js/compass/assets/10037761/97a538bb-b788-4419-93f3-3cd809af7f1f

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-connections-navigation/src/base-navigation-item.tsx
+++ b/packages/compass-connections-navigation/src/base-navigation-item.tsx
@@ -26,6 +26,7 @@ type NavigationBaseItemProps = {
 
   canExpand: boolean;
   isExpanded: boolean;
+  isFocused: boolean;
   onExpand: (toggle: boolean) => void;
 
   actionProps: {
@@ -58,6 +59,7 @@ export const NavigationBaseItem = ({
   dataAttributes,
   canExpand,
   isExpanded,
+  isFocused,
   onExpand,
 }: NavigationBaseItemProps) => {
   const [hoverProps, isHovered] = useHoverState();
@@ -88,7 +90,7 @@ export const NavigationBaseItem = ({
           </ItemLabel>
         </ItemButtonWrapper>
         <ItemActionControls<Actions>
-          isVisible={isActive || isHovered}
+          isVisible={isActive || isHovered || isFocused}
           data-testid="sidebar-navigation-item-actions"
           iconSize="small"
           {...actionProps}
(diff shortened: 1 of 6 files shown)


packages/compass-connections-navigation/src/base-navigation-item.tsx
```diff
@@ -88,7 +90,7 @@ export const NavigationBaseItem = ({
           </ItemLabel>
         </ItemButtonWrapper>
         <ItemActionControls<Actions>
-          isVisible={isActive || isHovered}
+          isVisible={isActive || isHovered || isFocused}
```
himanshusinghs, line 93: So this is the main change that makes the action controls tabbable because we render the action controls also when the navigation item is focused.

himanshusinghs: [How we get isFocused](https://github.com/mongodb-js/compass/pull/5923/files#diff-f06442f5a7bdda153a9bab3f55177d9b4165e82f71e7ba13064dddf5ac63ad67R306-R308) is something you guys might wanna comment about.

alenakhineika: Do you have concerns about this approach?

packages/compass-connections-navigation/src/virtual-list/use-virtual-navigation-tree.tsx
```diff
@@ -130,7 +130,7 @@ export function useVirtualNavigationTree<T extends HTMLElement = HTMLElement>({
   activeItemId?: string;
   onExpandedChange(item: VirtualTreeItem, isExpanded: boolean): void;
   onFocusMove?: (item: VirtualTreeItem) => void;
-}): [React.HTMLProps<T>, string | undefined] {
+}): [React.HTMLProps<T>, string | undefined, boolean] {
   const rootRef = useRef<T | null>(null);
   const activeId = activeItemId || findFirstItem(items)?.id;
```
paula-stacho, line 135: I'm not sure why, but this findFirstItem bit is now broken

https://github.com/mongodb-js/compass/assets/5196720/c777bad3-1360-45e4-9c93-df9c0416f124

paula-stacho: correction: actually it's broken on main already and it's specifically after databases are selected, so not related to this PR

paula-stacho: https://github.com/mongodb-js/compass/pull/5928

lerouxb (approved):

PR #5944, opened 2024-06-18
(compass-sidebar): fixes disconnected item retaining its expanded state when connected again  COMPASS-7967

himanshusinghs:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-sidebar/src/components/use-filtered-connections.ts
+++ b/packages/compass-sidebar/src/components/use-filtered-connections.ts
@@ -1,3 +1,5 @@
+import _ from 'lodash';
+import toNS from 'mongodb-ns';
 import type {
   SidebarCollection,
   SidebarConnectedConnection,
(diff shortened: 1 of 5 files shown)


packages/compass-sidebar/src/components/use-filtered-connections.ts
```diff
@@ -341,20 +339,63 @@ type UseFilteredConnectionsHookResult = {
...
 
+  const activeConnections = useMemo(() => {
+    return connections
+      .filter(({ connectionStatus }) => {
+        return connectionStatus === ConnectionStatus.Connected;
+      })
+      .map(({ connectionInfo }) => connectionInfo);
+  }, [connections]);
+
+  // get rid of stale connection related metadata in the state
```
himanshusinghs, line 394: this is the change that fixes the bug - the rest of the changes are based on the tests that I added.

packages/compass-sidebar/src/components/use-filtered-connections.ts
```diff
@@ -341,20 +339,63 @@ type UseFilteredConnectionsHookResult = {
...
+  useEffect(() => {
+    dispatch({ type: CONNECTIONS_CHANGED, connections: activeConnections });
```
lerouxb, line 396: Shouldn't all the logic for determining what's active / expanded be in actions and reducers rather than in an effect inside a hook that triggers when the hook is used?

himanshusinghs: True that, infact it used to be in reducer but at some point we moved it to a react state because the filtering was not just for databases but also for connections which were not in redux. Moving this back to redux would require some refactoring in the slice to also include connections related data in there.

The idea of this hook is to serve as an intermediary for further refactoring to follow which is planned in https://jira.mongodb.org/browse/COMPASS-7982

lerouxb (approved):

PR #5949, opened 2024-06-19
chore(compass-sidebar): adds an entrypoint for shell in sidebar COMPASS-8015

himanshusinghs:
## Description
Adds an entrypoint in the sidebar navigation for opening shell. The action item on connection navigation is still retained.

### How it works?
- When you don't have any connection the navigation item is disabled
- When you have just one connection the navigation item will open shell for that connection
- When you have multiple connection it will always ask which connection to open the shell in.

### Open questions:
- Should we open the shell always in a new tab? (Answered - yes)
- Should we have something like "Remember my preference" in the select connection modal to remember which connection to open the shell in? If yes, we would be saving that as a preference so I guess this should also come up eventually in compass preferences modal?
(Answered - We should not have something like this because it might be confusing to users when they forget such setting exist. Instead it would be better to just always ask for which connection to open the shell in, unless there is just one connection (copied from Slack))

### Demo

https://github.com/mongodb-js/compass/assets/10037761/adc2a923-58cb-4b99-8fc3-0d234bf6f4f0

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


himanshusinghs:
Closing this as we are not going to do this.

PR #5950, opened 2024-06-20
chore(compass-sidebar, compass-connections-import-export): adds an entrypoint in the multiple connections sidebar for import and export of saved connections COMPASS-7981

himanshusinghs:
## Description
This PR adds a ConnectionImportExportProvider to support invoking the ImportExport modal from different places in the application (old connections list and the new sidebar). When the provider is not the render tree, we do not show the actions related to Import Export of connections (applicable for CompassWeb).

The import export logic has also been modified to consider that all the connections should be taken into account into when multiple connections flag is on because in multiple connections world, all the connections are saved (regardless of them being favorite or not).

### Demo

https://github.com/mongodb-js/compass/assets/10037761/cd420b43-15d5-4090-9bd0-765469a2a3eb

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-connection-import-export/src/components/import-modal.tsx
```diff
@@ -82,21 +86,22 @@ export function ImportConnectionsModal({
         displayName: (
           <>
             {conn.name}
-            {conn.isExistingFavorite && (
+            {conn.isExistingConnection && (
```
lerouxb, line 89: What happens here in single connections world? Is this only true for favourites?

himanshusinghs: This is just a rename to isExistingConnection (to not confuse anyhow that we're only working with favorites) but how we determine an existing connection is defined here - https://github.com/mongodb-js/compass/pull/5950/files#diff-ec1f998fbad63a79b9ef45caea37663b06897235f2af291a7612b8eea2b30744R103-R117

This changes between single and multi connection world.

packages/compass/src/app/components/home.tsx
```diff
@@ -135,64 +135,6 @@ async function reauthenticationHandler() {
...
-  }, []);
-
-  const ConnectionImportModal = useMemo(
-    () =>
-      function ConnectionImportModal() {
-        return (
-          <ImportConnectionsModal
-            open={importModalOpen}
-            setOpen={setImportModalOpen}
-            trackingProps={{ context: 'connectionsList' }}
```
himanshusinghs, line 157: I wanted to ask about this `trackingProps`. Since we now have a provider, we expect the trackingProps to be provided by location from where the modal will be opened.

For example [here](https://github.com/mongodb-js/compass/pull/5950/files#diff-dcb87fc0b13a0f3894f8e4ea2a0c1c2a92ca6da293b7afbb9be30b68f64236bfR156-R157) and [then also here](https://github.com/mongodb-js/compass/pull/5950/files#diff-dcb87fc0b13a0f3894f8e4ea2a0c1c2a92ca6da293b7afbb9be30b68f64236bfR156-R157).

As you can see - in both the places the context is connectionsList. Would we want to differentiate between the two maybe by saying "multipleConnectionsList" for the latter?

Also, we open the modals via ipc call ([example here](https://github.com/mongodb-js/compass/pull/5950/files#diff-bb7670b9b7b39521ac52e1e2596227d7bd4a1e77c34797f1902ab84812be2b45L56)). Should the tracking prop be different in that case, maybe "compassMenu"?

@lerouxb @mcasimir (tagging you guys, maybe you have an idea?)

lerouxb: My guess is yes. Maybe ask in slack to see what product/analytics would prefer?

himanshusinghs: To clear out this thread: we're keeping context: connectionsList for both the single and multi connections sidebar.
For when the modal is opened via menuBar the context is set to "menuBar"

lerouxb (approved):

PR #5959, opened 2024-06-21
fix(compass-sidebar): fixes the filtering behaviour for databases and collections COMPASS-8026

himanshusinghs:
## Description
This PR brings back the original filtering behaviour that was bugged in one of the recent changes. Ideally when you are filtering and the filter matches the database then all of the collections of the database must get listed (regardless of collections matching the filter or not).

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
+++ b/packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
@@ -45,14 +45,16 @@ const filterDatabases = (
 ): FilteredDatabase[] => {
   const results: FilteredDatabase[] = [];
   for (const db of databases) {
-    const isMatch = regex.test(db.name);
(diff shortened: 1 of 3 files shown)


lerouxb (approved):

paula-stacho:
wait, what does listed mean? we want them included, but not expanded

paula-stacho (changes requested):
This is not a change request, more of discussion request :) I removed the behaviour intentionally to improve the search for longer lists, where including "non-matches" in the results makes it harder to spot the actual results (they might even be out of sight below all the other items). The user can still expand the filter results, so we need to include the children. But they don't need to clutter the search results.
Sorry if I failed to make that clear before! 
I can see how in some scenarios the original approach might be better, so maybe we can discuss this with design again!

PR #5960, opened 2024-06-21
fix: ConnectionScopedTelemetry COMPASS-7968

paula-stacho:
### ConnectionScopedTelemetry
Hooks & locator combining `telemetry` and `connectionInfoAccess`, by currying the `connectionId` (from `connectionInfoAccess`) into `track`. They live in `compass-connections`. I'd rather they live in the telemetry, and maybe even be the telemetry default, but right now that would create a circular dependency.

- `useConnectionScopedTelemetry`
- `withConnectionScopedTelemetry`
- `useConnectionScopedTrackOnChange`
- `createConnectionScopedTelemetryLocator`

This covers most cases. Occasionally, the event is outside of `connectionInfoAccess` scope, and we get the `connection_id` in some other way.

### Telemetry strategy

From now on, events & payload have to be registered in `@mongodb-js/compass-telemetry/src/events`. This is a first small step towards having a clear telemetry strategy in the code. Small step because right now, the payload types are mostly placeholders, I am only requiring `connectionId` (where possible).
Adding new events, we should define the actual payload type. For the existing events, we can port the current payload types from the code gradually or via some script (or both 😅 ). // TODO: add a task

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-aggregations/src/modules/saved-pipeline.ts
```diff
@@ -209,17 +213,21 @@ export const saveCurrentPipeline =
       id: savedPipeline.id,
       num_stages: stagesLength,
       editor_view_type: mapPipelineModeToEditorViewType(getState()),
+      connectionId: connectionInfoAccess.getCurrentConnectionInfo().id,
```
mcasimir, line 216: `connection_id`

PR #5975, opened 2024-06-26
chore(e2e) take into account the connection name in e2e tests COMPASS-8002 COMPASS-7970

lerouxb:
Also adds the Refresh Databases action to the connection menu

Screenshot 2024-06-27 at 17 10 28


packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -212,6 +228,8 @@ async function assertCannotCreateDb(
 
 async function assertCannotCreateCollection(
   browser: CompassBrowser,
+  // TODO(COMPASS-8002): take into account connectionName
```
himanshusinghs, line 231: we should remove these todos now, right?

himanshusinghs (approved):
looking good, just not sure about those todos we still have there referencing COMPASS-8002

packages/compass-e2e-tests/tests/connection.test.ts
lerouxb: The problem is still that the assumption is there that this is the only connection with this database name and collection name. I've changed the interface but internally it isn't robust against there being multiple connections open.

PR #5976, opened 2024-06-26
chore(compass-sidebar): adds CSFLE / Non-Genuine MongoDB markers in the sidebar along with some alignments with the new sidebar design

himanshusinghs:
## Description
Addresses:
- [COMPASS-8000](https://jira.mongodb.org/browse/COMPASS-8000)
- [COMPASS-8024](https://jira.mongodb.org/browse/COMPASS-8024)
- [COMPASS-7957](https://jira.mongodb.org/browse/COMPASS-7957)
- [COMPASS-7877](https://jira.mongodb.org/browse/COMPASS-7877)

(All the above were pretty small tasks and were linked so I included all of them in the same PR)

Designs that were taken for reference are [here](https://www.figma.com/design/4IEA6SVacn9KZvSU1c3oIt/MC-Edge-Cases?node-id=94-19963&t=JmcMPRCImWCkIMfO-0).

Open Questions:
- Should we consider having a small size in the interface for the icon button, given that it [exists for ActionControls already](https://github.com/mongodb-js/compass/blob/main/packages/compass-components/src/components/item-action-controls.tsx#L88-L101) and now [also for the static markers](https://github.com/mongodb-js/compass/pull/5976/files#diff-eb88200ce008588a5505f6c176b5708b8618cfafa1281568746025c151469331R23-R29)?
(Answered: I removed the new usage of small sized IconButton and instead of that using ItemActionControls itself which removes the question altogether.)

Demo:

https://github.com/mongodb-js/compass/assets/10037761/e2982cd0-e55a-4ebc-a92d-6b543f0c92cb

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-sidebar/src/components/legacy/sidebar.tsx
```diff
@@ -220,8 +220,8 @@ export function Sidebar({
         />
         <CSFLEConnectionModal
           open={isCSFLEModalVisible}
-          setOpen={(open: boolean) => setIsCSFLEModalVisible(open)}
           csfleMode={csfleMode}
+          onClose={() => setIsCSFLEModalVisible(false)}
```
paula-stacho, line 224: thanks, this makes more sense!

paula-stacho:
Am I looking at the wrong designs? I see in the figma:  icons at 16x16, icon button at 28x28 (that means 6px padding)

packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
```diff
@@ -348,6 +348,8 @@ const mapStateToProps: MapStateToProps<
         connectionStatus: ConnectionStatus.Connected,
         isPerformanceTabSupported:
           !isDataLake && !!state.isPerformanceTabSupported[connectionId],
+        isGenuineMongoDB: instance?.genuineMongoDB.isGenuine !== false,
```
paula-stacho, line 351: this feels a tiny bit unsafe, as it really means "is not a confirmed non genuine". most likely not an issue, but I imagined that in the future someone would say "let's show a badge that it's genuine!" and we'd use this bool which is true for undefined.

paula-stacho:
Screenshot 2024-06-27 at 10 59 33

I wonder if we shouldn't align the top and the bottom menu

paula-stacho:
when the scrollbar is in, the lower menu is no longer aligned with the top 🤦 I'm not sure how to address that

Screenshot 2024-06-27 at 10 59 33

himanshusinghs:
> Am I looking at the wrong designs? I see in the figma: icons at 16x16, icon button at 28x28 (that means 6px padding)

I took [these designs](https://www.figma.com/design/4IEA6SVacn9KZvSU1c3oIt/MC-Edge-Cases?node-id=94-19963&t=JmcMPRCImWCkIMfO-0) as reference.
Icons should still be 16x16 both in implementation and in figma but the icon button in Figma is 28px but not in implementation because we have been using 20px icon button (ActionControl) and changing that to 28px made sidebar look pretty odd. So I kept using what was already there.

packages/compass-sidebar/src/components/legacy/sidebar-databases-navigation.tsx
himanshusinghs: I think it gets more clearer when we look at how this (genuineMongoDB) is [computed](https://github.com/mongodb-js/mongodb-build-info/blob/main/index.js#L90-L110) because it is set to false only for explicitly identified conditions. We might as well say that every connection is genuine unless it was explicitly identified not to be.

paula-stacho: ah I see. so looks like this can only be other than a boolean when there is no instance. which shouldn't even happen, feels like we're missing an if somewhere :D

paula-stacho (approved):
Looking fly! 😎

PR #6005, opened 2024-07-10
chore(e2e): port shell e2e usage to multiple connections COMPASS-8004

lerouxb:
This ports tests that use the shell and almost all shell tests. We'll need a small follow-up for doing the last bits once we hide the shell entry points if the shell is disabled. See the TODOs.


packages/compass-components/src/components/workspace-tabs/tab.tsx
```diff
@@ -254,6 +258,8 @@ function Tab({
       tabIndex={isSelected ? 0 : -1}
       aria-controls={tabContentId}
       data-testid="workspace-tab-button"
+      data-connectionName={connectionName}
```
lerouxb, line 261: I added these data attributes so we can be sure we're addressing the right tab.

`connectionName` because even with COMPASS-8038 I think it will be awkward to address. There's only title & subtitle and the breadcrumb is up to four levels deep for a view, not counting the fact that there can be different views for the same breadcrumb (Databases & Shell, for example.)

`type` because title & subtitle will always be ambiguous on their own because the title for the shell tab is the connection name and once we do COMPASS-8049 the Databases tab will also be named after the connection name. The visible difference is the icon.

packages/compass-e2e-tests/tests/global-preferences.test.ts
```diff
@@ -177,6 +177,9 @@ describe('Global preferences', function () {
     const compass = await init(this.test?.fullTitle());
     const browser = compass.browser;
     try {
+      // TODO: check that the shell is there before toggling the section so we
```
lerouxb, line 180: We need a follow-up ticket for once we hide the shell entry points when the shell is disabled.

packages/compass-e2e-tests/tests/global-preferences.test.ts
```diff
@@ -201,6 +204,7 @@ describe('Global preferences', function () {
           'This setting cannot be modified as it has been set in the global Compass configuration file.'
         );
       }
+      // TODO: This just passes for multiple connections because the shell section is never there.
```
lerouxb, line 207: Same ticket number here.

packages/compass-e2e-tests/tests/shell.test.ts
```diff
@@ -60,10 +54,17 @@ describe('Shell', function () {
     await browser.clickVisible(Selectors.ShellInfoModalCloseButton);
     await infoModalElement.waitForDisplayed({ reverse: true });
 
-    await browser.hideShell();
+    await browser.hideShell(connectionName);
   });
 
   it('shows and hides shell based on settings', async function () {
+    // TODO: Leaving this skipped until we decide what we're going to do. hide
```
lerouxb, line 61: And same ticket number here.

packages/compass-e2e-tests/helpers/commands/workspace-tabs.ts
```diff
@@ -0,0 +1,83 @@
...
+    .$(Selectors.CloseWorkspaceTab)
+    .click();
+
+  // wait until the tab goes away and if the confirmation modal opens, maybe confirm
+  await browser.waitUntil(async () => {
+    if (autoConfirmTabClose) {
+      // Tabs in "dirty" state can't be closed without confirmation
+      if (await browser.$(Selectors.ConfirmTabCloseModal).isExisting()) {
+        await browser.clickVisible(
+          browser.$(Selectors.ConfirmTabCloseModal).$('button=Close tab')
```
himanshusinghs, line 28: `$('button=Close tab')`
TIL: Never saw this way of writing a selector before. Question - shouldn't Close tab be quoted?

packages/compass-e2e-tests/tests/logging.test.ts
```diff
@@ -31,8 +29,17 @@ describe('Logging and Telemetry integration', function () {
 
       try {
         await browser.connectWithConnectionString();
-        await browser.shellEval('use test');
-        await browser.shellEval('db.runCommand({ connectionStatus: 1 })');
+
+        if (TEST_MULTIPLE_CONNECTIONS) {
+          // make sure we generate the screen event that the tests expect
+          await browser.navigateToMyQueries();
+        }
```
himanshusinghs, line 36: we probably need to update this, right? because in reality the screen event for 'My Queries' won't be emitted or different ones will be emitted upon connection which we will need to reflect here?
Should we ticket this?

himanshusinghs (commented):
This is looking good, left a few questions.

packages/compass-e2e-tests/tests/logging.test.ts
paula-stacho: for the purpose of testing`telemetry.screens()`, this seems fine to me, and actually probably makes the test more resilient since we don't just assume which screen is open by default

packages/compass-e2e-tests/helpers/commands/connect-with-connection-string.ts
```diff
@@ -71,7 +71,7 @@ export async function connectWithConnectionString(
 
     // some connection should be expanded (ie. connected) by now
     await browser
-      .$(`${Selectors.SidebarTreeItems} [aria-expanded=true]`)
-      .waitForExist({ reverse: true });
```
paula-stacho, line 75: how was this working until now?

packages/compass-e2e-tests/helpers/commands/hide-shell.ts
```diff
@@ -1,12 +1,23 @@
...
 import * as Selectors from '../selectors';
 
-export async function hideShell(browser: CompassBrowser): Promise<void> {
-  await retryWithBackoff(async function () {
-    const shellContentElement = await browser.$(Selectors.ShellContent);
-    if (await shellContentElement.isDisplayed()) {
-      await browser.clickVisible(Selectors.ShellExpandButton);
-    }
-  });
+export async function hideShell(
```
paula-stacho, line 6: maybe we can rename this to closeShell (also the file), so that it's easier to find for the future devs

packages/compass-e2e-tests/helpers/commands/show-shell.ts
```diff
@@ -1,17 +1,43 @@
...
-export async function showShell(browser: CompassBrowser): Promise<void> {
-  // Expand the shVdell
-  await retryWithBackoff(async function () {
-    const shellContentElement = await browser.$(Selectors.ShellContent);
-    if (!(await shellContentElement.isDisplayed())) {
-      // The toasts may be covering the shell, so we need to close them.
-      await browser.hideAllVisibleToasts();
-      await browser.clickVisible(Selectors.ShellExpandButton);
-    }
+export async function showShell(
```
paula-stacho, line 6: similarly here - openShell would be more fitting now

packages/compass-e2e-tests/helpers/commands/connect-with-connection-string.ts
lerouxb: If your selector matches nothing then waitForExist({ reverse: true }) cannot fail :)

PR #6009, opened 2024-07-10
feat(compass-crud): allows nested elements to also expand in steps

himanshusinghs:
## Description
This patch introduces "Show more" fields toggle also on the nested elements which we earlier used to render with all of their children when expanded. With the change, we will the take a similar approach as that of Document, showing only a default number of elements to begin with(set to 25) which user can later change by clicking the toggle.

https://github.com/user-attachments/assets/a8b43008-51e0-4acf-ac1a-4db2c672a614

Designs for the toggle have also changed a bit as per Slack discussion:

|Before|After|
|------|-----|
|image|image|
|image|image|

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/hadron-document/src/element.ts
```diff
@@ -736,7 +739,7 @@ export class Element extends EventEmitter {
         element.expand(expandChildren);
       }
     }
-    this.emit(ElementEvents.Expanded, this);
+    this._bubbleUp(ElementEvents.Expanded, this);
```
himanshusinghs, line 742: Note: This is intentional and used in the virtualised list.   I forgot to remove this from the commit when I was splitting but I also think its harmless if it goes in main with this patch.

gribnoysup (commented):
Have we talked with design about the button styles? The current one when added in the middle of the document looks very bulky compared to surrounding content, font size and family is not matching, feels out of place

packages/hadron-document/src/element.ts
```diff
@@ -836,6 +839,35 @@ export class Element extends EventEmitter {
...
+    }
+    return [...this.elements].slice(0, this.visibleElementsCount);
+  }
+
+  setVisibleElementsCount(newCount: number) {
+    if (!this._isExpandable(this.originalExpandableValue)) {
+      return;
+    }
+    this.visibleElementsCount = newCount;
+    this._bubbleUp(Events.VisibleElementsChanged, this, this.getRoot());
```
gribnoysup, line 854: This is not true if element is collapsed, right?

packages/hadron-document/src/element.ts
```diff
@@ -753,7 +756,7 @@ export class Element extends EventEmitter {
         element.collapse();
       }
     }
-    this.emit(ElementEvents.Collapsed, this);
+    this._bubbleUp(ElementEvents.Collapsed, this);
```
gribnoysup, line 759: We can bubble it, but then we are also create an interface that requires you to know the internal implementation. I feel like we should rather emit "visible changed" in addition to this even if appropriate

packages/hadron-document/src/element.ts
```diff
@@ -67,6 +69,7 @@ export class Element extends EventEmitter {
   invalidTypeMessage?: string;
   decrypted: boolean;
   expanded = false;
+  visibleElementsCount = DEFAULT_VISIBLE_ELEMENTS;
```
gribnoysup, line 72: I'd probably consider a slightly different approach and make this particular property private, and for public would write something that also takes expanded into account, because otherwise you can have a case where expanded is false and reading this value will tell you that there are visible elements, that is not true

packages/hadron-document/src/element.ts
himanshusinghs: ahh that's right - I missed that, thanks for spotting.

packages/hadron-document/src/element.ts
himanshusinghs: yea I was in a fix about wether to emit VisibleElementsChanged for all the actions that could possibly change the number of visible elements or wether to listen for those action events directly in the needed place. I opted [for latter](https://github.com/mongodb-js/compass/pull/6010/files#diff-dd307ccd5a00c28fdaa74fe18ee0fe7a3595235e5a1b11c534008d54881ecc83R82-R87), it probably was not the best idea. I will adapt the suggestion here as I have better understanding now. Thank you :)

packages/hadron-document/src/element.ts
himanshusinghs: ahh I think the name led to slight confusion. Maybe this property should be named - maxVisibleElements because that is what it is doing, setting a cap for maximum elements that can be made visible at a time and:
- to actually get the visible elements you use [getVisibleElements](https://github.com/mongodb-js/compass/pull/6009/files/d3418339ee6be5e0fedfe49c160281ecb134ca29#diff-6f9c0f59fcb29fb98792df56908d66e6b4a45b2f5187904e70afdf09053bb52dR842)
- to get a count of visible elements you use [getTotalVisibleElementsCount](https://github.com/mongodb-js/compass/pull/6009/files/d3418339ee6be5e0fedfe49c160281ecb134ca29#diff-6f9c0f59fcb29fb98792df56908d66e6b4a45b2f5187904e70afdf09053bb52dR857)

(I just realised that those methods should account for expanded or not in some way, I will add that.)

himanshusinghs: addressed this in - [dea3187](https://github.com/mongodb-js/compass/pull/6009/commits/dea318758d4165f6a588efbcb36f098181833735)

packages/hadron-document/src/element.ts
himanshusinghs: addressed this in - [628c98d](https://github.com/mongodb-js/compass/pull/6009/commits/628c98d58fd09bb664238fe9fe0989b4fc0992ab)

packages/hadron-document/src/element.ts
himanshusinghs: addressed this in recent changes

packages/compass-components/src/components/document-list/document.tsx
```diff
@@ -124,20 +140,20 @@ const HadronDocument: React.FunctionComponent<{
...
+        currentSize={document.maxVisibleElementsCount}
         totalSize={elements.length}
-        minSize={INITIAL_FIELD_LIMIT}
+        minSize={DEFAULT_VISIBLE_DOCUMENT_ELEMENTS}
         // In the editing mode we allow to show / hide less fields because
         // historically Compass was doing this for "performance" reasons
         step={editing ? 100 : 1000}
-        onSizeChange={setVisibleFieldsCount}
-      ></DocumentFieldsToggleGroup>
+        onSizeChange={handleVisibleFieldsChanged}
```
Anemy, line 155: This is existing functionality, but I figured I'd bring it up here, no action needed. A small UX improvement here for folks with lots of fields in their documents could be to adjust the scroll position when they click to hide a lot of fields to the relative scroll height after the fields are hidden. Currently it's easy to lose a document if it's not the last document and you hide a bunch of fields.

packages/hadron-document/src/element.ts
```diff
@@ -840,6 +850,47 @@ export class Element extends EventEmitter {
...
 
+  getVisibleElements() {
+    if (!this.elements || !this.expanded) {
+      return [];
+    }
+    return [...this.elements].slice(0, this.maxVisibleElementsCount);
+  }
+
+  setMaxVisibleElementsCount(newCount: number) {
+    if (!this._isExpandable(this.originalExpandableValue)) {
```
Anemy, line 861: We updated the `_isExpandable` usage in https://github.com/mongodb-js/compass/pull/6013
There is a thread on how it can be better used there. Now it's less of a static method.
```suggestion
    if (!this._isExpandable()) {
```
I'm thinking the recent changes are preventing a bug here where someone could be editing a sub document with a lot of fields in it and change the type.

Thanks for the code comment on the difference between the `25` when editing and `1000` when not editing.

packages/compass-components/src/components/document-list/visible-field-toggle.tsx
```diff
@@ -11,23 +12,39 @@ const container = css({
   paddingRight: spacing[3],
 });
 
-const button = css({
-  flex: 'none',
+const linkButtonStyles = css({
+  border: 'none',
+  background: 'none',
+  padding: 0,
```
Anemy, line 18: These styles look good in readonly, however for normal documents I found them maybe too far to the left and down. How about:
| before | after |
| - | - |
| Screenshot 2024-07-17 at 5 57 47 PM | Screenshot 2024-07-17 at 6 15 37 PM |
^ Looking at my example again, maybe I put it too far to the right 😅 - also have thought about bolding these actions?

gribnoysup: I like aligning the top level expand block with the document so that it matches other expand buttons (and yeah, does address the big gap in editable mode)! For bold we tried it out but specifically decided not to go with it to avoid making the doc card too visually noisy (see the discussion [here](https://mongodb.slack.com/archives/GDTJXPHD0/p1720618573269519))

packages/compass-components/src/components/document-list/visible-field-toggle.tsx
```diff
@@ -58,32 +75,43 @@ const DocumentFieldsToggleGroup: React.FunctionComponent<{
...
+          hideExternalIcon={true}
+          className={cx(linkButtonStyles, buttonClassName)}
           onClick={onHideClick}
-          className={button}
+          aria-label={hideButtonText}
           data-testid="hide-fields-button"
         >
-          Hide {hideSizeDiff} fields
-        </Button>
+          <Icon size="xsmall" glyph="ArrowUp"></Icon>
```
gribnoysup, line 109: I think in the designs that Ben shared it was CaretDown / CaretUp, not an arrow. Don't have a strong preference, but arrow seems a big big when compared to the text near it, so maybe worth trying

packages/compass-components/src/components/document-list/visible-field-toggle.tsx
himanshusinghs: Thank you for the suggestion here. I have gone ahead and applied some consistency as per suggestions above. So now:
- the top level toggle aligns with the expand icon of the top level fields
- the nested toggle aligns with the expand icon of the nested fields
- have removed the bottom padding from the toggle as it was causing a little disruption when two toggles were next to each other.

This is how it looks now:
|Description|Image|
|-----------|-----|
|Readonly Mode|image
|Readonly Mode expanded|image|
|Editable Mode|image|
|Editable Mode expanded|image|

packages/hadron-document/src/element.ts
himanshusinghs: Thanks for catching this Rhys. I took a pull from main but forgot to update this. Just did that.

packages/compass-components/src/components/document-list/document.tsx
himanshusinghs: Yea I noticed this after I implemented virtual list. I will see what I can do there to modify this behaviour a little.

Anemy (approved):

PR #6017, opened 2024-07-11
chore(e2e): Port My Queries tests, fix Refresh Databases COMPASS-8006

lerouxb:
* I realised that refresh databases wasn't actually doing anything. I just fixed it in this PR
* This un-skips the my queries tests for multiple connections
* I renamed the test file because it isn't actually per-instance
* I just flatted the nested context & multiple beforeEach hooks to fit with the style in this package. When you have a big test file and lots of nested describe/context each with their hooks I find it hard to follow what's actually in play. I also don't like putting too much work in a beforeEach especially in an e2e test because it is just as likely to fail as other test code and then it is the hook that fails and gets screenshotted rather than the test.
* I updated the test that checks that the sidebar's refresh databases button works because it looks like it couldn't fail at least in some cases.

### DIFF
--- a/packages/compass-e2e-tests/tests/my-queries-tab.test.ts
+++ b/packages/compass-e2e-tests/tests/my-queries-tab.test.ts
@@ -41,7 +41,7 @@ async function openMenuForQueryItem(
   await browser.$(Selectors.SavedItemMenu).waitForDisplayed();
 }
 
-describe('My Queries tab', function () {
+describe.only('My Queries tab', function () {
   let compass: Compass;
   let browser: CompassBrowser;
   const connectionName = connectionNameFromString(DEFAULT_CONNECTION_STRING);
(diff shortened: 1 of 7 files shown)


packages/compass-sidebar/src/components/multiple-connections/connections-navigation.tsx
```diff
@@ -497,6 +503,12 @@ const ConnectionsNavigation: React.FC<ConnectionsNavigationProps> = ({
   );
 };
 
+const onRefreshDatabases = (): SidebarThunkAction<void> => {
+  return (_dispatch, getState, { globalAppRegistry }) => {
+    globalAppRegistry.emit('refresh-databases');
```
lerouxb, line 508: As far as I can tell this refresh-databases and the one on the Databases tab where I copied it from will will refresh for all connections. Assuming I'm reading it correctly.

packages/compass-e2e-tests/tests/my-queries-tab.test.ts
```diff
@@ -267,121 +268,147 @@ describe('Instance my queries tab', function () {
...
+  const connectionName = connectionNameFromString(DEFAULT_CONNECTION_STRING);
+  const favoriteQueryName = 'list of numbers greater than 10 - query';
+  const newCollectionName = 'numbers-renamed';
+
+  before(async function () {
+    skipForWeb(this, 'saved queries not yet available in compass-web');
+
+    compass = await init(this.test?.fullTitle());
+    browser = compass.browser;
+  });
```
himanshusinghs, line 285: we probably moved it out of root describe block and that's why we need to skipForWeb here as well, should we just move  it back in, unless this was intentional?
If it was I recall seeing in compass somewhere a rule about only having one suite per test file.

packages/compass-sidebar/src/components/multiple-connections/connections-navigation.tsx
himanshusinghs: yea you will need to add connectionId to this most likely as second parameter.

himanshusinghs: and I think in the databases-collection, that's also a problem, we will also have to update the corresponding check in compass-app-stores which actually refreshes the database. Sorry I must have either missed this when I moved compass-app-stores events or accidentally skipped it.

packages/compass-e2e-tests/tests/my-queries-tab.test.ts
lerouxb: Like I said in the description I moved it deliberately.

lerouxb: Actually there's no real reason for that second beforeEach to exist and be any different. Fixing.

packages/compass-sidebar/src/components/multiple-connections/connections-navigation.tsx
lerouxb: Should be good now?

packages/compass-e2e-tests/tests/my-queries-tab.test.ts
```diff
@@ -274,8 +275,9 @@ describe('Instance my queries tab', function () {
       const favoriteQueryName = 'list of numbers greater than 10 - query';
       const newCollectionName = 'numbers-renamed';
 
-      /** saves a query and renames the collection associated with the query, so that the query must be opened with the "select namespace" modal */
-      async function setup() {
+      it('users can permanently associate a new namespace for an aggregation/query', async function () {
```
lerouxb, line 278: As described in the PR description I simplified some code here and brought it in line with this package's style.

himanshusinghs (approved):
Looks good 🚀 I will rebase mine after you merge this.

PR #6018, opened 2024-07-11
feat: new telemetry events for connections

paula-stacho:
### Existing events

**Connection Imported - when new connections are imported**
added
- connection_ids

### New events
**Connection Disconnected**
- connection_id

**Connection Created - when a new connection is saved**
- connection_id
- color (optional)

**Connection Removed**
- connection_id

** UPDATE ** Removed active_connection_count and inactive_connection_count upon further discussion. We might add these later if needed.

### Checklist
- [ ] Documentation is changed or added --> https://docs.google.com/spreadsheets/d/1Wi7np8CooM0b3xPHZNd-iYst_DnKEo7FBswbnSXKoys/edit?gid=953290655#gid=953290655 

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-connections/src/hooks/use-connection-repository.ts
```diff
@@ -89,18 +90,18 @@ export function useConnectionRepository(): ConnectionRepository {
...
         }
       });
 
       setNonFavoriteConnections((prevList) => {
-        if (areConnectionsEqual(prevList, nonFavoriteConnections)) {
+        if (areConnectionsEqual(prevList, newNonFavoriteConnections)) {
           return prevList;
         } else {
-          return nonFavoriteConnections;
+          return newNonFavoriteConnections;
```
paula-stacho, line 104: no change here, just touched this file while exploring some other options. there are higher scope consts with the name `favoriteConnections` and `nonFavoriteConnections`.

mcasimir:
> it also make sense to save before and not after connecting

💯 

The reason for this was to keep track of the history. Can we create a follow up ticket to clean this up? I'm not sure folks removing the FF will be aware of it.

mcasimir:
Did you consider splitting the shouldSave change from the new telemetry events? Seems to be quite loosely related.

paula-stacho:
> > it also make sense to save before and not after connecting
> 
> 💯
> 
> The reason for this was to keep track of the history. Can we create a follow up ticket to clean this up? I'm not sure folks removing the FF will be aware of it.

that's a valid concern. what needs to be cleaned up here is the `legacyShouldSaveConnection`, I changed the name to mark it for the cleanup. we do have a general cleanup ticket https://jira.mongodb.org/browse/COMPASS-7906, but I do wonder if we're doing good enough job of marking everything 🙈 I can add a todo with a link to that ticket.

paula-stacho:
> Did you consider splitting the shouldSave change from the new telemetry events? Seems to be quite loosely related.

I encountered several issues, one with the duplication and another with saving twice. the first was a problem on it's own so it got a ticket. this one was just unnecessary, it only became problem for the tracking (due to the asynchronous nature of repository updates we couldn't reliably tell that the connection already exists when it was saved second time). still, I can make it a separate ticket if that makes it easier to review

packages/compass-connections/src/connections-manager.ts
```diff
@@ -157,6 +157,12 @@ export class ConnectionsManager extends EventEmitter {
     );
   }
 
+  getConnectionIdsByStatus(status: ConnectionStatus): ConnectionInfoId[] {
+    return [...this.connectionStatuses.entries()]
+      .filter(([, connectionStatus]) => connectionStatus === status)
+      .map(([connectionId]) => connectionId);
+  }
```
himanshusinghs, line 164: This adds one more way to access this information besides our hooks - `useActiveConnections`, `useConnectionsWithStatus`. I think I understand the usecase here but we would want to be very deliberate about this because it might be fine but otherwise a tech debt (either this or the hooks?) that needs to be cleaned up. We don't need to block the PR on this but it would be a good idea to populate the compass-connections ticket with this information as well.

mcasimir:
> add a todo with a link to that ticket

that sounds great, should we update that ticket with a todo list too?

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -211,13 +236,53 @@ export function useConnections({
...
+    },
+    [favoriteConnections, recentConnections]
+  );
+
   const saveConnectionInfo = useCallback(
-    async (
-      connectionInfo: RecursivePartial<ConnectionInfo> &
-        Pick<ConnectionInfo, 'id'>
-    ) => {
+    async ({
```
mcasimir, line 257: can you elaborate on this change? what made us change how connections are saved here?

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -211,13 +236,53 @@ export function useConnections({
...
+      fullConnectionInfo,
```
mcasimir, line 258: what is `fullConnectionInfo` and what is `partialConnectionInfo`?

himanshusinghs (commented):
Looks promising. I left a small note but not a blocker.

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -211,13 +236,53 @@ export function useConnections({
...
+            const isActive =
+              connectionsManager.statusOf(fullConnectionInfo.id) ===
+              'connected'; // save normally happens first, but in case.
+            onConnectionCreated?.(fullConnectionInfo, {
+              active: isActive ? active + 1 : active,
+              inactive: isActive ? inactive : inactive + 1, // we can't use ConnectionRepository here, as it updates asynchronously
+            });
+          }
+        }
+        await repositorySaveConnection(connectionInfo || partialConnectionInfo);
```
mcasimir, line 285: `connectionInfo || partialConnectionInfo`

We already have
`const connectionInfo = fullConnectionInfo || partialConnectionInfo;` few lines above?

paula-stacho:
> > add a todo with a link to that ticket
> 
> that sounds great, should we update that ticket with a todo list too?

making a complete list is not going to be that simple at this point, and rather than making an incomplete list I'd leave it for when the ticket is picked up.

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -211,13 +236,53 @@ export function useConnections({
...
+        }
+      | {
+          fullConnectionInfo: ConnectionInfo;
+          partialConnectionInfo?: undefined;
+        }) => {
+      const connectionInfo = fullConnectionInfo || partialConnectionInfo;
       try {
-        await saveConnection(connectionInfo);
+        if (fullConnectionInfo) {
+          const isNewConnection = !findConnectionInfo(fullConnectionInfo.id);
```
mcasimir, line 273: can't you rely on the fact that `id` is set to know if is new, but also, isn't the callsite aware of this?

packages/compass-connections/src/stores/connections-store.tsx
paula-stacho: thanks, I forgot to update this line!

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -211,13 +236,53 @@ export function useConnections({
...
+          if (isNewConnection) {
+            const { active, inactive } = getActiveAndInactiveCount();
+            const isActive =
+              connectionsManager.statusOf(fullConnectionInfo.id) ===
+              'connected'; // save normally happens first, but in case.
+            onConnectionCreated?.(fullConnectionInfo, {
+              active: isActive ? active + 1 : active,
```
mcasimir, line 280: This is creating an implicit hidden coupling, personally i don't like particularly this style because it corrupts code over time in ways that aren't obvious.

When we add `active` and `inactive` (btw please let's at least call them "activeConnectionsCount"), we are admitting that this code knows what `onConnectionCreated` will do. That is the same than calling `track` here, in fact i would rather prefer that.

The 2 other alternatives are instead very clear:
1. we don't change `saveConnectionInfo`, which is a critical function, that is better if it stays simple and if it doesn't fail or gets more complex for unrelated reason. And instead we could emit an `onConnectionSaved` and have the entire telemetry related logic there (including counting active connections etc ..).

2. we very explicitly call track here, which would make deadly clear why all of this computation are needed, and would make it also easier to refactor code in a simpler way, just extracting it to functions  with a clear name.

packages/compass-connections/src/stores/connections-store.tsx
paula-stacho: we don't change how they are saved, we're adding tracking for connections that are new. this method is used for create & update both. in one case of update, it is called with partial connectionInfo, which doesn't fit the tracking.

separating create & update would be one option, it just doesn't fit with the current flow. I also didn't want to simply assume that the info is complete when the connection is new, so I opted to make the `partialConnectionInfo` an explicit overload.

packages/compass-connections/src/stores/connections-store.tsx
paula-stacho: see above

packages/compass-connections/src/connections-manager.ts
paula-stacho: I see your point, but the use cases are quite different - this is a simple one-time getter, similar to `statusOf`, while the hooks are maintaining a state and listening to updates

packages/compass-connections/src/stores/connections-store.tsx
paula-stacho: I don't follow the question - did you mean that the id is not set for a new connection? if so, this is actually incorrect, the id is created when the form for a new connection is open.

packages/compass/src/app/components/home.tsx
```diff
@@ -198,13 +201,65 @@ function Home({
...
+    },
+    [track]
+  );
+
+  const onConnectionRemoved = useCallback(
+    (
+      connectionInfo: ConnectionInfo | undefined,
+      activeAndInactiveConnectionsCount: { active: number; inactive: number }
+    ) => {
+      trackConnectionRemovedEvent(
```
mcasimir, line 254: Similar to my previous comment. I'm curious about what prompted this "callback" / "event-driven" style here.
 
It seems `onDisconnected`, `onConnectionCreated`, and `onConnectionRemoved` exclusively call the respective track functions and require adapting the calling point to surface additional information.

What is the benefit of this approach? From my perspective, this indirection might be obscuring the code's intention, making it harder to follow. I'm open to discuss this further of course, but i don't see much of an advantage to be doing this instead of just calling trackConnection<X>Event directly.

packages/compass-connections/src/stores/connections-store.tsx
mcasimir: Got it, this is creating new concepts though, `partialConnectionInfo` and `fullConnectionInfo` isn't that obvious in which way would be full and partial. If we want to keep this style, it may be better to add some jsdoc comment that clarifies how this function should be used, when you want to call what.

> I also didn't want to simply assume that the info is complete when the connection is new

This is a good example, i feel we are giving for granted that everyone understands what "complete" means here.

I also want to add that I do feel weird that we need to do an invasive change like this for tracking. I wonder what other options we considered

packages/compass-connections/src/stores/connections-store.tsx
mcasimir: Ah ok, i thought is something that is done on save

packages/compass/src/app/utils/telemetry.ts
```diff
@@ -200,6 +203,52 @@ export function trackNewConnectionEvent(
   }
 }
 
+export function trackConnectionDisconnectedEvent(
```
mcasimir, line 206: I did miss it when we moved connection attempt etc here, but i don't understand why we are moving these here. 

I would love if we have all of the domain specific `track` functions in one place, i don't think this is what is happening though. Why do we have this special "place" for connection events only? And if this is a road that we want to pursue, how do we make sure is consistent everywhere and that we don't have 5 different ways to do the same thing?

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -252,7 +331,35 @@ export function useConnections({
...
+      const saved = await saveConnectionInfo({
+        fullConnectionInfo: connectionInfo,
+      });
+      if (!saved) {
+        return;
+      }
+
+      if (activeConnectionId === connectionInfo.id) {
+        // Update the active connection if it's currently selected.
+        dispatch({
```
gribnoysup, line 355: Why other places that save info directly through saveConnectionInfo don't need to update this?

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -293,12 +402,12 @@ export function useConnections({
...
       connectionInfo: ConnectionInfo,
       dataService: DataService,
-      shouldSaveConnectionInfo: boolean
+      legacyShouldSaveConnectionInfo?: boolean // TODO: cleanup COMPASS-7906
     ) => {
       try {
         dispatch({ type: 'set-active-connection', connectionInfo });
-        onConnected?.(connectionInfo, dataService);
-        if (!shouldSaveConnectionInfo) return;
+        onConnected?.(connectionInfo, dataService, getActiveAndInactiveCount());
```
gribnoysup, line 409: Same as above: potentially stale

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -342,6 +458,10 @@ export function useConnections({
     );
     try {
       await connectionsManager.closeConnection(connectionId);
+      onDisconnected?.(
+        findConnectionInfo(connectionId),
+        getActiveAndInactiveCount()
```
gribnoysup, line 463: Same as above: both are potentially stale

packages/compass/src/app/components/home.tsx
gribnoysup: We initially had https://jira.mongodb.org/browse/COMPASS-7740 opened for this, but it's assigned to me and closed. I don't really remember why it got resolved like that, but I'm guessing we decided that it will be just resolved with some other refactoring that still needs to happen?

packages/compass/src/app/utils/telemetry.ts
gribnoysup: Answered in https://github.com/mongodb-js/compass/pull/6018#discussion_r1677948900 (I think those basically express the same concerns)

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -211,13 +236,53 @@ export function useConnections({
...
       try {
-        await saveConnection(connectionInfo);
+        if (fullConnectionInfo) {
+          const isNewConnection = !findConnectionInfo(fullConnectionInfo.id);
+          if (isNewConnection) {
+            const { active, inactive } = getActiveAndInactiveCount();
+            const isActive =
+              connectionsManager.statusOf(fullConnectionInfo.id) ===
+              'connected'; // save normally happens first, but in case.
+            onConnectionCreated?.(fullConnectionInfo, {
```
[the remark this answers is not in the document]
paula-stacho: nope, let me switch the order!

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -237,12 +302,26 @@ export function useConnections({
...
+      connectionsManager,
+      onConnectionCreated,
+    ]
   );
 
   const removeConnection = useCallback(
     async (connectionInfo: ConnectionInfo) => {
-      await deleteConnection(connectionInfo);
+      await repositoryRemoveConnection(connectionInfo);
+      const { active, inactive } = getActiveAndInactiveCount();
```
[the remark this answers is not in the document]
paula-stacho: you mean it's stale when we get them here or by the time the track is send?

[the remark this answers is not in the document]
paula-stacho: for the first I thought having `getActiveAndInactiveCount` as a dependency would do the trick, doesn't it? for the latter I'm not sure if we anything fresher than the state after the removal

packages/compass-connections/src/stores/connections-store.tsx
paula-stacho: good question. I actually only moved this function and I haven't questioned this bit. I'd say this is a legacy thing, because in multiple connections the activeConnection is only used by the connection form, and the form is closed when we save. perhaps @himanshusinghs can confirm? if I'm right we could mark this for a cleanup

packages/compass/src/app/components/home.tsx
mcasimir: yeah, that's probably a good guess. Though we probably thought you were doing this refactoring @gribnoysup :)

mcasimir: > desktop-only dependencies so had to stay away from the code shared with web

Isn't that addressed by providing a different implementation of services?

packages/compass-connections/src/stores/connections-store.tsx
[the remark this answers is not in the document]
gribnoysup: When you get them here. Function itself and all the values inside of it are going to be potentially stale at the moment of calling because it is called after the async operation, it doesn't really matter whether or not it's a React "callback" and what dependencies you provide here, there is no way to safely access values like that if you're not reading them from the state directly.

This is a common issue with closures, just harder to spot due to all the extra code we have here, if you'd simplify it to the very core, it would look something like this, hopefully it highlights better what is going wrong here:

```
// simulate react render call: when component renderes, everything defined
// inside the scope will be part of it and will not change
function render(recentConnections) {
  const getRecentConnectionsLength = () => recentConnections.length;
  const removeRecent = () => {
    setTimeout(() => {
      // pretend we did some async work here
      console.log(getRecentConnectionsLength())
    }, 0);
  }
  return { removeRecent };
}
var recentConnections = [];
var { removeRecent } = render(recentConnections);
removeRecent(); // doesn't log immediately, async operation is in progress
// we changed the value and triggered a re-render, but it doesn't matter because the
// function that we already triggered still has the old value in the scope
recentConnections = [1, 2, 3];
void render(recentConnections);
// async op is done and you'll see zero logged
```

This is also very similar to the common issue of accessing a stale state in redux thunks and why they recommend to always call `getState()` inside of a thunk action instead of doing it once at the beginning of async flow.

paula-stacho:
note: separated the legacy parameter change, as I need it for the Save&Connect update https://github.com/mongodb-js/compass/pull/6034

packages/compass-connections/src/stores/connections-store.tsx
paula-stacho: the change doesn't seem that invasive to me - it's a change in parameters of a function internal to `useConnections` hook

paula-stacho: I think I listed all the alternatives above. From those, separating create & update would definitely be more invasive, which leaves the alternative to "assume" that a new connection info is complete (with `as`). This is mildly less invasive (doesn't change the parameters of the internal function), but less secure.
perhaps you could check the change in the function again? now that we've removed the active/inactive count, the change is leaner.

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -360,7 +366,11 @@ export function useConnections({
         Pick<ConnectionInfo, 'id'>
     ) => {
       try {
-        return await saveConnection(connectionInfo);
+        const isNewConnection = !getConnectionInfoById(connectionInfo.id);
+        const savedConnectionInfo = await saveConnection(connectionInfo);
+        if (isNewConnection) {
+          onConnectionCreated?.(savedConnectionInfo);
```
gribnoysup, line 372: Similar to other callbacks, can you wrap these in `useCurrentRef` and use refs instead to make sure that the functions you are passing these to are stable and don't have direct dependency on these callbacks?

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -360,7 +366,11 @@ export function useConnections({
         Pick<ConnectionInfo, 'id'>
```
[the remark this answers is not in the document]
gribnoysup: Also looking closer at this, I'm pretty sure we never actually pass the partial here, so maybe this special case type is not even needed. I copied this from previous implementation without digging too deep

packages/compass-connections/src/connections-manager.ts
gribnoysup: Totally might be missing some code in this PR, but is it being used by something? I don't mind that we're adding this method here, but as we're going to change the manager too in the follow-up refactors, if we don't use it, I'd maybe skip adding more methods to it

packages/compass-connections/src/stores/connections-store.tsx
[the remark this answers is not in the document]
paula-stacho: actually we do pass partial in `oidcUpdateSecrets`

packages/compass-connections/src/stores/connections-store.tsx
```diff
@@ -360,7 +366,11 @@ export function useConnections({
...
     ) => {
       try {
-        return await saveConnection(connectionInfo);
```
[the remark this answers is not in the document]
paula-stacho: ah yes, I was trying to remember to add the return after the track and looks like I forgot, thank you!

packages/compass/src/app/utils/telemetry.ts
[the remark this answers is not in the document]
paula-stacho: sure we can! I followed the same flow for consistency, but if we'll get rid of the other callbacks soon, we might as well move those in. do we have a ticket number for the change?

packages/compass-connections/src/stores/connections-store.tsx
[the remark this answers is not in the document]
gribnoysup: Ahhh, true, missed that! Hmmm, that method is internal to this hook logic though and we do have the whole connection info object passed to it so we probably can do the merge there ourselves? Totally no strong preference though, mostly saying this because I saw your discussion above about this whole partial / full info type and maybe if we want to remove the ambiguity, we can just refactor away the use-case for this partial type

packages/compass/src/app/utils/telemetry.ts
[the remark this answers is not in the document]
gribnoysup: That'd be nice I think. Doing the callbacks here is totally valid approach and it makes sense you did it like that first, but just from the perspective this not being required and us planning to change this later it would be nice to already do this even though it breaks the pattern.

The ticket part is a bit confusing, if you want to add a `todo` somewhere here, you should probably just use the catch-all COMPASS-7397, we had another ticket more related to that, but it got closed when we were cleaning tickets up ([related discussion](https://github.com/mongodb-js/compass/pull/6018#discussion_r1678023926))

packages/compass-connections/src/stores/connections-store.spec.tsx
```diff
@@ -139,11 +152,11 @@ describe('useConnections', function () {
     renderHookWithContext({ onConnected });
 
     await waitFor(() => {
-      expect(onConnected).to.have.been.called;
+      expect(onConnected).to.have.been.calledOnce;
     });
 
     // autoconnect info should never be saved
-    expect(saveSpy).to.not.have.been.called;
```
gribnoysup, line 146: Not sure if this was auto-replaced by accident, but we kinda expect save to not be called at all in this case, not only not called once (this will not fail if it's called twice)

paula-stacho: yes that was an accident 🤦 thanks!

packages/compass-connections/src/stores/connections-store.spec.tsx
```diff
@@ -392,6 +406,98 @@ describe('useConnections', function () {
     });
   });
 
+  describe('#saveEditedConnection', function () {
+    it('new connection: should call save and onConnectionCreated', async function () {
```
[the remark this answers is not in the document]
paula-stacho: ah yes, thank you!

PR #6021, opened 2024-07-12
chore(e2e): Port usage of saving favourites including all of the in-use-encryption e2e tests to multiple connections COMPASS-8003

lerouxb:
Main change is that the in-use-encryption tests were ported. browser.shellEval() can now handle multiple commands and properly waits for them to be finished.

### DIFF
--- a/packages/compass-e2e-tests/helpers/commands/connect-with-connection-form.ts
+++ b/packages/compass-e2e-tests/helpers/commands/connect-with-connection-form.ts
@@ -1,6 +1,5 @@
 import type { CompassBrowser } from '../compass-browser';
 import type { ConnectFormState } from '../connect-form-state';
-import * as Selectors from '../selectors';
 
 export async function connectWithConnectionForm(
   browser: CompassBrowser,
@@ -8,9 +7,13 @@ export async function connectWithConnectionForm(
   connectionStatus: 'success' | 'failure' | 'either' = 'success',
   timeout?: number
 ): Promise<void> {
-  const sidebar = await browser.$(Selectors.SidebarTitle);
-  if (await sidebar.isDisplayed()) {
-    await browser.disconnect();
+  await browser.disconnect();
+
+  // If a connectionName is specified and a connection already exists with this
+  // name, make sure we don't add a duplicate so that tests can always address
+  // this new connection.
+  if (options.connectionName) {
+    await browser.removeConnection(options.connectionName);
   }
 
   await browser.setConnectFormState(options);
--- a/packages/compass-e2e-tests/helpers/commands/connect-with-connection-form.ts
+++ b/packages/compass-e2e-tests/helpers/commands/connect-with-connection-form.ts
@@ -7,7 +7,7 @@ export async function connectWithConnectionForm(
   connectionStatus: 'success' | 'failure' | 'either' = 'success',
   timeout?: number
 ): Promise<void> {
-  await browser.disconnect();
+  await browser.disconnectAll();
 
   // If a connectionName is specified and a connection already exists with this
   // name, make sure we don't add a duplicate so that tests can always address
--- a/packages/compass-e2e-tests/helpers/commands/shell-eval.ts
+++ b/packages/compass-e2e-tests/helpers/commands/shell-eval.ts
@@ -7,8 +7,6 @@ async function getOutputText(browser: CompassBrowser): Promise<string[]> {
   });
 }
 
-let screenshotNumber = 0;
-
 export async function shellEval(
   browser: CompassBrowser,
   connectionName: string,
@@ -49,8 +47,6 @@ export async function shellEval(
       return false;
     });
   }
-
-  await browser.screenshot(`shell-eval-${++screenshotNumber}.png`);
 
   const output = await getOutputText(browser);
 
--- a/packages/compass-e2e-tests/helpers/commands/shell-eval.ts
+++ b/packages/compass-e2e-tests/helpers/commands/shell-eval.ts
@@ -17,11 +17,11 @@ export async function shellEval(
   // focus it.
   await browser.openShell(connectionName);
 
-  const numLines = (await getOutputText(browser)).length;
-
   const strings = Array.isArray(input) ? input : [input];
 
   for (const str of strings) {
+    const numLines = (await getOutputText(browser)).length;
+
     const command = parse === true ? `JSON.stringify(${str})` : str;
 
     await browser.setCodemirrorEditorValue(Selectors.ShellInputEditor, command);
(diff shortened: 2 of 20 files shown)


packages/compass-e2e-tests/helpers/commands/connect-with-connection-form.ts
```diff
@@ -1,16 +1,19 @@
...
 export async function connectWithConnectionForm(
   browser: CompassBrowser,
   options: ConnectFormState,
   connectionStatus: 'success' | 'failure' | 'either' = 'success',
   timeout?: number
 ): Promise<void> {
-  const sidebar = await browser.$(Selectors.SidebarTitle);
-  if (await sidebar.isDisplayed()) {
-    await browser.disconnect();
+  await browser.disconnect();
```
lerouxb, line 10: Just made this in line with other places where we disconnect. No need to check anything. We'd have to branch in the checking.

packages/compass-e2e-tests/helpers/commands/connect-with-connection-form.ts
```diff
@@ -1,16 +1,19 @@
...
+
+  // If a connectionName is specified and a connection already exists with this
+  // name, make sure we don't add a duplicate so that tests can always address
+  // this new connection.
+  if (options.connectionName) {
+    await browser.removeConnection(options.connectionName);
```
lerouxb, line 16: Just making sure this mirrors connectWithConnectionString()

packages/compass-e2e-tests/helpers/commands/shell-eval.ts
```diff
@@ -7,10 +7,12 @@ async function getOutputText(browser: CompassBrowser): Promise<string[]> {
   });
 }
 
+let screenshotNumber = 0;
+
 export async function shellEval(
   browser: CompassBrowser,
   connectionName: string,
-  str: string,
+  input: string | string[],
```
lerouxb, line 13: shellEval() now supports multiple lines. We used it consecutively in some tests and the problem is that before, we just showed and hid it so it kept its state, but now we open a new one and close it so it loses state. So tests that would `use database` broke because database reset to `test` every time.

Now we can give it multiple lines.

Ideally we'd probably not be using the shell like this for fixture data anyway, but that's probably a larger change.

packages/compass-e2e-tests/helpers/commands/shell-eval.ts
```diff
@@ -19,23 +21,36 @@ export async function shellEval(
...
+        lines.length >
+        /**
+         * input line becomes an output line on enter press, so we are waiting
+         * for two new lines to appear, not just one
+         */
+        numLines + 1
+      ) {
+        // make sure the prompt shows up before entering another line or continueing on
+        await browser
+          .$(`${Selectors.ShellInput} [aria-label="Chevron Right Icon"]`)
```
lerouxb, line 43: My proudest achievement 😆  We were typing too fast for some of the commands and this waits for the input to give control back to the user.

packages/compass-e2e-tests/tests/connection-form.test.ts
```diff
@@ -65,8 +65,8 @@ describe('Connection form', function () {
     };
 
     if (TEST_MULTIPLE_CONNECTIONS) {
-      (expectedState.connectionColor = 'no-color'),
-        (expectedState.connectionFavorite = false);
+      expectedState.connectionColor = 'no-color';
```
lerouxb, line 68: No idea what prettier did here. I just cleaned it up.

packages/compass-e2e-tests/tests/instance-sidebar.test.ts
```diff
@@ -154,6 +154,7 @@ describe('Instance sidebar', function () {
     }
 
     // open the create database modal from the sidebar
+    // TODO: technically this should be scoped to the connection
```
lerouxb, line 157: I'm going go just deal with this in this PR. Hang tight..

packages/compass-e2e-tests/tests/in-use-encryption.test.ts
```diff
@@ -7,26 +7,32 @@ import {
...
 import { MongoClient } from 'mongodb';
 
 import delay from '../helpers/delay';
+import type { ConnectFormState } from '../helpers/connect-form-state';
 
 const CONNECTION_HOSTS = '127.0.0.1:27091';
 const CONNECTION_STRING = `mongodb://${CONNECTION_HOSTS}/`;
 
-async function refresh(browser: CompassBrowser) {
+async function refresh(browser: CompassBrowser, connectionName: string) {
```
lerouxb, line 22: This file is where most of the action happens, but github collapses it by default.

packages/compass-e2e-tests/tests/instance-sidebar.test.ts
```diff
@@ -154,9 +154,16 @@ describe('Instance sidebar', function () {
     }
 
     // open the create database modal from the sidebar
-    await browser.clickVisible(Sidebar.CreateDatabaseButton, {
-      screenshot: 'before-can-create-a-database-and-drop-it-click.png',
-    });
+    if (TEST_MULTIPLE_CONNECTIONS) {
```
lerouxb, line 157: These are drive-bys: I realised we weren't properly scoping the selector to the relevant connection.

packages/compass-e2e-tests/helpers/commands/connect-with-connection-form.ts
paula-stacho: I wonder if we could name the disconnect action differently then, something like ensureNoActiveConnections. sounds weird, but without your comment and checking the disconnect implementation I'd not think it works like it does.

lerouxb: I renamed it to disconnectAll, but I'm generally reserving those kinds of refactoring/distinctions for when we add multiple connections use cases and start differentiating actions like that.

packages/compass-e2e-tests/helpers/commands/shell-eval.ts
paula-stacho: nice!

paula-stacho (approved):
🎉

addaleax (approved):

packages/compass-e2e-tests/helpers/commands/shell-eval.ts
```diff
@@ -19,23 +19,34 @@ export async function shellEval(
...
+    // wait until more output appears
+    await browser.waitUntil(async () => {
+      const lines = await getOutputText(browser);
+      if (
+        lines.length >
+        /**
+         * input line becomes an output line on enter press, so we are waiting
+         * for two new lines to appear, not just one
+         */
+        numLines + 1
```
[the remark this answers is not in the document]
lerouxb: aah good catch. missed that when adding the loop. probably doesn't matter because the chevron goes away fast enough (in which case the line count check is probably not necessary anymore), but best to move it anyway.

PR #6028, opened 2024-07-16
chore(e2e): add test for add connection failures COMPASS-8009 COMPASS-8091

lerouxb:
Also when adding a connection and it fails to connect the toast now shows up.


packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
```diff
@@ -362,17 +369,17 @@ export function MultipleConnectionSidebar({
 
   const onNewConnectionConnect = useCallback(
     (connectionInfo: ConnectionInfo) => {
-      void connect({
-        ...cloneDeep(connectionInfo),
-      }).then(() => setIsConnectionFormOpen(false));
+      setIsConnectionFormOpen(false);
+      void _onConnect(connectionInfo).then(() => {
```
lerouxb, line 373: We weren't using all the same code for new connections as when connecting to existing saved connections which meant that errors weren't caught and the toast didn't appear.

packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -299,6 +300,29 @@ describe('Connection string', function () {
     }
   });
 
+  it('fails for authentication errors', async function () {
```
lerouxb, line 303: We didn't actually have any tests for failed connections except in auto-connect.test.ts for connecting from the CLI, so we didn't notice before that the toast never showed up.

packages/compass-e2e-tests/helpers/commands/connect-with-connection-string.ts
```diff
@@ -36,24 +36,26 @@ export async function connectWithConnectionString(
...
-    // There's a potential problem here: The list of connections is virtual, so
-    // the new connection might not be rendered. By searching for it if possible
-    // we can guarantee that it will be on screen. The search box won't exist
-    // for the first connection, but that's OK because once connected it will be
-    // the only connection so should be rendered.
-    if (await browser.$(Selectors.SidebarFilterInput).isExisting()) {
-      await browser.clickVisible(Selectors.SidebarFilterInput);
-      await browser.setValueVisible(Selectors.SidebarFilterInput, '');
-    }
+  if (connectionStatus === 'success') {
```
paula-stacho, line 39: could the part that is only for successful connections be separated into it's own helper? `connectWithConnectionString` is becoming somewhat fuzzy

packages/compass-e2e-tests/helpers/commands/wait-for-connection-result.ts
```diff
@@ -37,9 +40,11 @@ export async function waitForConnectionResult(
   }
 
   if (TEST_MULTIPLE_CONNECTIONS) {
-    await browser
-      .$(Selectors.ConnectionModal)
-      .waitForDisplayed({ reverse: true });
+    if (fromForm) {
```
paula-stacho, line 43: do we need to skip this? does it error without the form, or do we have cases when we want to keep the form open while connecting from the sidebar? asking because I find the `fromForm` parameter a bit confusing, so I'd rather avoid it if possible

paula-stacho (approved):
Looks good, thank you for the fix!

PR #6032, opened 2024-07-16
chore(e2e): cancel import/export when disconnecting, re-enable tests COMPASS-8008

lerouxb:
The e2e tests actually caught the regression. We weren't aborting the import/export so the toast would get all the errors from trying to work with a disconnected dataservice.

### DIFF
--- a/packages/compass-import-export/src/modules/export.ts
+++ b/packages/compass-import-export/src/modules/export.ts
@@ -153,6 +153,18 @@ type CloseExportAction = {
   type: ExportActionTypes.CloseExport;
 };
 
+export const connectionDisconnected = (
+  connectionId: string
+): ExportThunkAction<void> => {
+  return (dispatch, getState, { logger: { debug } }) => {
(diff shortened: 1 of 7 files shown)


packages/compass-import-export/src/modules/export.ts
```diff
@@ -153,6 +153,18 @@ type CloseExportAction = {
...
 };
 
+export const connectionDisconnected = (
+  connectionId: string
+): ExportThunkAction<void> => {
+  return (dispatch, getState, { logger: { debug } }) => {
+    const currentConnectionId = getState().export.connectionId;
+    debug('connectionDisconnected', { connectionId, currentConnectionId });
+    if (connectionId === currentConnectionId) {
+      dispatch(closeExport());
```
himanshusinghs, line 163: Should this be cancelExport?

himanshusinghs (changes requested):

himanshusinghs (approved):

PR #6033, opened 2024-07-16
chore(e2e): add connectionName support to more helpers COMPASS-8002

lerouxb:


packages/compass-e2e-tests/helpers/commands/index.ts
```diff
@@ -41,8 +41,7 @@ export * from './get-feature';
 export * from './set-feature';
 export * from './save-favorite';
 export * from './save-connection-string-as-favorite';
-export * from './select-connection';
-export * from './select-connection-menu-item';
+export * from './sidebar-connection';
```
lerouxb, line 44: I combined a bunch of sidebar connection related commands into one file to mirror what we do for workspace management already. Otherwise it is a bit of a proliferation of little files. It might also help if we do move these to page objects because they'd already be together.

packages/compass-e2e-tests/helpers/commands/workspace-tabs.ts
```diff
@@ -14,31 +16,36 @@ async function closeTab(
...
-      if (await browser.$(Selectors.ConfirmTabCloseModal).isExisting()) {
-        await browser.clickVisible(
-          browser.$(Selectors.ConfirmTabCloseModal).$('button=Close tab')
-        );
-        await browser
-          .$(Selectors.ConfirmTabCloseModal)
-          .waitForDisplayed({ reverse: true });
+  await browser.waitUntil(
+    async () => {
+      // keep retrying the click :(
```
lerouxb, line 22: Drive-by. This was something I moved out recently because clicking again won't work if the modal obscures it anyway, but it seems like it was necessary because it became flaky again.

packages/compass-e2e-tests/helpers/commands/workspace-tabs.ts
```diff
@@ -59,10 +67,15 @@ export async function closeWorkspaceTabs(
       // are multiple tabs then another tab will immediately become active and
       // trip up the logic that checks that the tab you closed went away.
       const id = await currentActiveTab.getAttribute('id');
+      debug('closing tab', { numTabsStart, id });
```
lerouxb, line 70: Just some debugging in case this still flakes.

packages/compass-e2e-tests/helpers/commands/sidebar-connection.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export async function getConnectionIdByName(
+  browser: CompassBrowser,
+  connectionName: string
+): Promise<string> {
+  // NOTE: this is only supported for multiple connections
+  return await browser
+    .$(Selectors.sidebarConnection(connectionName))
+    .getAttribute('data-connection-id');
+}
```
paula-stacho, line 13: given that we do not enforce unique names, this is a bit risky 🤔 . could we at least check if there are multiple and throw in that case? so that we don't end up with some weird test behaviour on accident.

packages/compass-e2e-tests/helpers/commands/collection-workspaces.ts
```diff
@@ -12,7 +13,12 @@ async function navigateToCollection(
   // state of Schema, and Validation tabs without re-connecting.
   closeExistingTabs = true
 ): Promise<void> {
+  const connectionId = TEST_MULTIPLE_CONNECTIONS
+    ? await browser.getConnectionIdByName(connectionName)
+    : undefined;
```
paula-stacho, line 18: are all these checks necessary because we don't have the connectionName in single-connection, or because the getter function doesn't work for it?

paula-stacho: just wondering if we could avoid all those ifs somehow

lerouxb: Actually I was confused and responded about the wrong thing at first. I thought of putting this check inside getConnectionIdByName, but then I have to make the return type be Promise<string|undefined> and then I have to check or assert the type of connectionId in a bunch of places.. this just felt cleaner overall.

We can't read out the connection id for single connection because it isn't somewhere persistent in the page.

lerouxb: Actually I just tried it again and it didn't cause any problems. I think I just happened to change other things in the meantime.

packages/compass-e2e-tests/helpers/commands/workspace-tabs.ts
paula-stacho: should we add todos to clean these up later?

paula-stacho (approved):
Neat! 🎉 Sounds like we could now also get rid of all the `delete <telemetryEvent>.connection_id;` that I've added for telemetry!

packages/compass-e2e-tests/helpers/commands/drop-collection-from-sidebar.ts
```diff
@@ -4,48 +4,24 @@ import * as Selectors from '../selectors';
...
-  await collectionElement.waitForDisplayed();
-
-  // open the drop collection modal from the sidebar
-  await browser.hover(collectionSelector);
-
-  // NOTE: if the menu was already open for another collection this could get
-  // confusing. Also this selector is just for the actions button and it is
-  // assumed that at this point it is the only one. But the drop confirmation
-  // usually catches that.
-  await browser.clickVisible(Selectors.SidebarNavigationItemShowActionsButton);
```
lerouxb, line 35: This wasn't scoped to the collection. If one collection was active it would show the three dot menu. If you then hover over a second collection it would show that three dot menu too..

PR #6037, opened 2024-07-17
fix: save&connect COMPASS-8068

paula-stacho:
## Description
Aside of renaming the button to save&connect, this PR is trying to make a clearer distinction between 'connect' and 'saveAndConnect', legacy and new code. For example making sure the submitForm is always explicitly one or the other - previously we had inconsistent behaviour between submit via button and submit via keystroke, likely because we didn't notice the other calls. So after I removed the extra "save", we were saving only once in the first case but zero times in the second! 🤦 I blame the complexity of the legacy form, and can't wait for the cleanup 😁  🔥 

Also I marked `ConnectFormActions` as legacy and replaced it's unit tests with tests for the new `ConnectionFormModalActions`. Today I learned these are still used for VSCode, but I don't think that's a reason to maintain the separate component - we just need some flags to prepare the new component for VSCode. I created [COMPASS-8098](https://jira.mongodb.org/browse/COMPASS-8098) to be done before we do the clean up and enable the new form in VSCode ([COMPASS-7762](https://jira.mongodb.org/browse/COMPASS-7762))

Scenarios tested:
- edit (submitted with btn or enter)
- duplicate (submitted with btn or enter)
- create (submitted with btn or enter)

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


lerouxb (approved):

packages/compass/src/app/utils/telemetry.ts
```diff
@@ -156,10 +156,11 @@ export function trackConnectionAttemptEvent(
...
 ): void {
   try {
-    const { favorite, lastUsed } = connectionInfo;
+    const { lastUsed, savedConnectionType } = connectionInfo;
+    const isFavorite = savedConnectionType === 'favorite';
     const trackEvent = {
-      is_favorite: Boolean(favorite),
-      is_recent: Boolean(lastUsed && !favorite),
+      is_favorite: isFavorite,
+      is_recent: Boolean(lastUsed && !isFavorite),
```
paula-stacho, line 163: this is a driveby - noticed via a broken e2e test. the favorite object contains the color and name, which can now exist on non-favorite connections 🙈 . we didn't rename because it would require a migration. the proper way to check for a favorite is via savedConnectionType.

lerouxb: had to update another unit test: https://github.com/mongodb-js/compass/pull/6037/commits/a19247034e840099f5eef3a4e7bfd64ae71eca11

PR #6040, opened 2024-07-17
feat(compass-editor, compass-query-bar): provide auto completed query history matching user input COMPASS-8018

VivianTNT:
## Description
COMPASS-8018

The query history autocompleter will still show up on focus but now users will also be able to type to match their input. The new autocompleter has 2 sections, with the original base completions for fields on the top and query history completions on the bottom (if any). 
On focus:
image
Both types of completions:
image
Just saved queries:
image

### Checklist
- [X] New tests and/or benchmarks are included

## Motivation and Context

- [ ] Bugfix
- [X] New feature
- [X] Dependency update
- [ ] Misc

## Open Questions

When a query is currently in an editor and the cursor is focused on it, pressing reset will trigger onFocus() for that particular editor and becomes unfocused after the {} and autocompleter are automatically triggered.

## Dependents

COMPASS-8017

## Types of changes

- [ ] *Backport Needed*


packages/compass-editor/src/codemirror/query-history-autocompleter.ts
```diff
@@ -92,3 +108,17 @@ function createInfo(query: SavedQuery): {
...
   };
 }
+
+// scales a number unscaledNum between [minAllowed, maxAllowed]
+export function scaleBetween(
+  unscaledNum: number,
+  minAllowed: number,
+  maxAllowed: number,
+  min: number,
+  max: number
```
paula-stacho, line 118: took me a bit to understand the min and max params, maybe we could name them differently to make things clearer - for example: 
```
newScaleMin,
newScaleMax,
originalScaleMin,
originalScaleMax,
```

packages/compass-editor/src/codemirror/query-history-autocompleter.ts
```diff
@@ -92,3 +108,17 @@ function createInfo(query: SavedQuery): {
...
+): number {
+  if (max === min) return 0;
```
paula-stacho, line 120: this looks like unexpected input. returning a number could make such input go unnoticed, and 0 might be out of the allowed scale - so it would be safer to throw

packages/compass-editor/src/codemirror/query-history-autocompleter.ts
VivianTNT: Sounds good, thanks for the suggestion!

packages/compass-editor/src/codemirror/query-history-autocompleter.ts
VivianTNT: I returned 0 because in the context of the autocompleter, originalScaleMin = originalScaleMax only when there is 1 lastExecuted time, in which case it doesn't matter how the autocompleter ranks the completion. The `boost` property in the options expects numbers between -99 and 99, so I could probably return any number between those. 

For the sake of making the input noticed, do you think it would be a better idea to throw an error in the function and then catch the error within the autocompleter options itself instead of handling it in the function?

packages/compass-query-bar/src/stores/query-bar-reducer.ts
```diff
@@ -440,6 +448,46 @@ const saveRecentQuery = (
...
+      } = getState();
+
+      const queryAttributes = getQueryAttributes(query);
+      // Ignore empty or default queries
+      if (isEmpty(queryAttributes)) {
+        return;
+      }
+
+      const existingFavoriteQuery = favoriteQueries.find((favoriteQuery) => {
+        return isQueryEqual(getQueryAttributes(favoriteQuery), queryAttributes);
```
Anemy, line 471: Since the favorite query is already being passed into this function and we're doing this same find there can we remove this check?
Then above we can pass `existingFavoriteQuery` to this function and make this function query argument type a bit stricter.

Anemy (commented):
Nice, left a comment or two on the recent changes. Looking good!

Anemy (commented):
Nice tests! Code looks good, left a comment on what we're autocompleting. We might want to match a bit less than we currently do.

Anemy (approved):

packages/compass-editor/src/codemirror/query-autocompleter-with-history.ts
```diff
@@ -21,10 +23,55 @@ export const createQueryWithHistoryAutocompleter = (
...
+
+  return async function fullCompletions(context: CompletionContext) {
+    if (context.state.doc.toString() === '{}') {
+      return queryHistoryAutocompleter(context);
+    }
+
+    const combinedOptions: Completion[] = [];
+    // completions assigned to a section appear below ones that are not assigned
+    const baseCompletions = await originalQueryAutocompleter(context);
+    const historyCompletions = await queryHistoryAutocompleter(context);
```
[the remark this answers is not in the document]
Anemy: Slack thread: https://mongodb.slack.com/archives/GDTJXPHD0/p1721663254763859

PR #6044, opened 2024-07-18
refactor(hadron-build): remove usage of "async" package and its tasks COMPASS-5051

alenakhineika:
## Description

Remove the usage of the `async` package by just using promises.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/hadron-build/commands/release.js
```diff
@@ -562,20 +557,18 @@ exports.run = (argv, done) => {
     task('store build configuration as json', writeConfigToJson)
   ].filter(Boolean));
 
-  return async.series(tasks, (_err) => {
-    try {
-      if (_err) {
-        return done(_err);
+  return tasks.reduce((promise, task) => {
```
mcasimir, line 560: Nit: here we can probably just go with a `for..of` with `async await`? it would simplify how it reads a lot.

packages/hadron-build/commands/release.js
```diff
@@ -562,20 +557,18 @@ exports.run = (argv, done) => {
...
+    return promise.then(task);
+  }, Promise.resolve()).then(() => {
+    // clean up copied npmrc
+    fs.rm(path.resolve(target.dir, '.npmrc'), (err) => {
```
mcasimir, line 564: this block was under `finally` and cleaning also in case of error.

mcasimir (approved):
🙌. Strange that this doens't cause a change also in `package-lock.json`? Left also comment on removing `.npmrc` in case of errors, which If i'm not wrong it may be worth addressing.

packages/hadron-build/lib/zip.js
```diff
@@ -1,67 +1,67 @@
...
-      '--symlinks',
-      opts.outPath,
-      './'
-    ];
-
-    execFileSync('zip', args, {
-      env: process.env,
-      cwd: path.join(opts.dir, '..'),
-      stdio: 'inherit'
+  const runZip = () => {
```
mcasimir, line 21: Nit: similar to above we can util.promisify `zipFolder` and use `async/await`

mcasimir (approved):

PR #6050, opened 2024-07-22
chore(compass-workspaces): update the window title in the workspace, match the workspace's connection COMPASS-8103

lerouxb:


Anemy (approved):
lgtm, might be nice to add a test but not required, didn't have one before either.

gribnoysup (approved):

PR #6056, opened 2024-07-23
feat(compass): display warning message to user when Compass cannot access cred…

alenakhineika:
…ential storage COMPASS-7819

## Description

Use Electron's `safeStorage.isEncryptionAvailable()` to check if the secret storage works. If not, display a toast with a warning message.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass/src/app/index.tsx
+++ b/packages/compass/src/app/index.tsx
@@ -265,6 +265,7 @@ const Application = View.extend({
         title:
           'Compass cannot access credential storage. You can still connect, but please note that passwords will not be saved.',
       });
+      track('Secret Storage Not Available');
     }
 
     document.querySelector('#loading-placeholder')?.remove();
--- a/packages/compass/src/app/index.tsx
+++ b/packages/compass/src/app/index.tsx
@@ -9,6 +9,7 @@ import { globalAppRegistry } from 'hadron-app-registry';
 import { defaultPreferencesInstance } from 'compass-preferences-model';
 import semver from 'semver';
 import { CompassElectron } from './components/entrypoint';
+import { openToast } from '@mongodb-js/compass-components';
 
 // https://github.com/nodejs/node/issues/40537
 dns.setDefaultResultOrder('ipv4first');
@@ -114,6 +115,10 @@ async function getWindowAutoConnectPreferences(): Promise<AutoConnectPreferences
   return await ipcRenderer?.call('compass:get-window-auto-connect-preferences');
 }
 
+async function checkSecretStorageIsAvailable(): Promise<boolean> {
+  return await ipcRenderer?.call('compass:check-secret-storage-is-available');
+}
+
 /**
  * The top-level application singleton that brings everything together!
  */
@@ -212,6 +217,7 @@ const Application = View.extend({
       (): Promise<AutoConnectPreferences> => {
         return Promise.resolve(initialAutoConnectPreferences);
       };
+    const isSecretStorageAvailable = await checkSecretStorageIsAvailable();
     const connectionStorage = new CompassRendererConnectionStorage(
       ipcRenderer,
       getInitialAutoConnectPreferences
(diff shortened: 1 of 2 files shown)


packages/compass/src/app/index.tsx
```diff
@@ -253,6 +259,14 @@ const Application = View.extend({
       this.queryByHook('layout-container')
     );
 
+    if (!isSecretStorageAvailable) {
+      openToast('secret-storage-not-available', {
+        variant: 'warning',
+        title:
+          'Compass cannot access credential storage. You can still connect, but please note that passwords will not be saved.',
```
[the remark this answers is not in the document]
alenakhineika: I was thinking about it, but still have doubts. On one hand, the toast might seem annoying, on the other hand, if they disabled this with the setting, they might forget about it over time. My initial idea with the warning pillow on the connect form came because of the extra reason, to always display this information to users, similarly to how we inform them in the case of CSFLE, that credentials will be stored on disk. With the current implementation, we show the toast on each Compass launch, not on disconnect.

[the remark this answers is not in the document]
Anemy: > My initial idea with the warning pillow on the connect form came because of the extra reason, to always display this information to users, similarly to how we inform them in the case of CSFLE, that credentials will be stored on disk.

I like that initial approach, it's less intrusive to someone who knows what they're doing on a custom OS setup. 
Looking at the slack thread: https://mongodb.slack.com/archives/GDTJXPHD0/p1721654142590129 
Sounds like Julia likes it. I'm cool with either approach, although I do think if we do the toast we should think about if how folks will want to hide it.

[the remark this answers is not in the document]
alenakhineika: Agreed in Slack do not introduce a settign for now, but add a telemetry event to track the number of occurrences.

packages/compass/src/main/application.ts
```diff
@@ -230,6 +230,9 @@ class CompassApplication {
         debug('Did not agree to license, quitting app.');
         app.quit();
       },
+      'compass:check-secret-storage-is-available': function () {
+        return safeStorage.isEncryptionAvailable();
```
Anemy, line 234: Looking at https://github.com/electron/electron/issues/33640 it sounds like if we haven't had the `whenReady` event this can return false, at least on windows. Since we're using it from the renderer I don't think this will be an issue, although we may want to add a promise to return this value when it's ready if it's not yet ready.
At the least I think it's worth calling out in a comment, although I do feel it could make it a bit prone to a future bug if we use it differently, so I think making sure it's ready first is good.
```suggestion
        // This will return `false` even if a credential store is available on windows if this is called
        // before electron's `whenReady`. https://github.com/electron/electron/issues/33640
        return safeStorage.isEncryptionAvailable();
```

Anemy (commented):
Left one mention of a possible issue for `isEncryptionAvailable` which we might want to at least call out in a comment.
Other than that lgtm!

packages/compass/src/main/application.ts
alenakhineika: It is safe to call `isEncryptionAvailable` here since the app is already ready and rendered, but I have left a comment for visibility. I also added `await app.whenReady()` before logging `SafeStorage initialized` so we have a proper value on Windows.

packages/compass/src/main/application.ts
```diff
@@ -230,6 +233,12 @@ class CompassApplication {
         debug('Did not agree to license, quitting app.');
         app.quit();
       },
+      'compass:check-secret-storage-is-available': function () {
+        // Accessing isEncryptionAvailable is not allowed when app is not ready on Windows
+        // https://github.com/electron/electron/issues/33640
+        // But here Compass is already rendered and it is safe to omit `await app.whenReady()`
+        return safeStorage.isEncryptionAvailable();
```
Anemy, line 240: Not a blocker, does it work if we instead make this method `async` and have the `await app.whenReady();` in here before we return? That way we aren't waiting for the `whenReady` till it's needed.
Looks like the process communication here already has support for async: https://github.com/mongodb-js/compass/blob/3868044de073e6db26b76b4fa4d716f29699606b/packages/hadron-ipc/src/main.ts#L64

Anemy (approved):
lgtm, left one suggestion, not a blocker.

packages/compass/src/main/application.ts
alenakhineika: We could make the callback of `compass:check-secret-storage-is-available` async and put `await app.whenReady()` inside before returning, but why would you want to do that, if we know that app got ready in main before the rendering process?

Anemy (approved):

PR #6057, opened 2024-07-24
feat(sidebar): Add the shell icon on hover to the connections in the sidebar COMPASS-8050 COMPASS-8071

lerouxb:
Nuance that might be missed is that the way the three dot menu is designed is that it is the "spillover" of the actions that don't fit in the sidebar. So the open shell action actually moved from the menu to the sidebar in between the add database and menu icons. ie. really the number of icons at which point the rest is collapsed into a menu has changed.

Screenshot 2024-07-24 at 11 44 35

This also brings back a test for multiple connections that checks that the icon goes away if the shell is disabled.


packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -210,11 +210,10 @@ async function assertCannotCreateDb(
 
   // open the create database modal from the sidebar
   if (TEST_MULTIPLE_CONNECTIONS) {
-    await browser.clickVisible(
-      Selectors.sidebarConnectionActionButton(
-        connectionName,
-        Sidebar.CreateDatabaseButton
-      )
+    await browser.selectConnectionMenuItem(
```
lerouxb, line 213: Cleaned this up because selectConnectionMenuItem() now works for the items that aren't collapsed too.

himanshusinghs (approved):

himanshusinghs:
Something odd happening with sidebar tests, might wanna check that.

PR #6060, opened 2024-07-25
feat(compass-editor, compass-query-bar): query history autocomplete formatting COMPASS-8063

VivianTNT:
## Description

COMPASS-8063

Using the existing query history popover formatting, added multiple css changes to the label, completion detail, and completion info panel. For the info panel, syntax highlighting was added (same styles as the option editor), as well as have the styling adapt to dark vs. light mode.

image
image

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [x] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-editor/src/codemirror/query-history-autocompleter.ts
```diff
@@ -92,7 +96,34 @@ function createInfo(query: SavedQuery): {
...
+    function emitBreak() {
+      code.appendChild(document.createTextNode('\n'));
+    }
+
+    if (formattedQuery) {
+      highlightCode(
+        formattedQuery,
+        languages['javascript-expression']().language.parser.parse(
+          formattedQuery
+        ),
```
Anemy, line 121: A slight performance improvement (I'm thinking it could factor in if someone is scrolling through a lot of query history items), could we define this parser once so we don't rebuild this language support object each time? It's probably not too costly, but I think it doesn't hurt to move it to a constant variable outside of this function.

packages/compass-editor/src/codemirror/query-history-autocompleter.ts
```diff
@@ -109,6 +140,13 @@ function createInfo(query: SavedQuery): {
   };
 }
 
+export type SavedQuery = {
+  lastExecuted: Date;
+  queryProperties: {
+    [propertyName: string]: any;
+  };
+};
```
Anemy, line 148: [nit] Can we move this type definition to higher up in the file where it used to be? Since the functions currently above this are using it.

Anemy (commented):
Nice! Changes look good. Left two small suggestions, let's wait till we hear back from Diancheng, then it should be good to go merge!

packages/compass-editor/src/codemirror/query-history-autocompleter.ts
VivianTNT: How much of it do I move out to a constant variable? Including `parser` or just up to` languages['javascript-expression']().language`?

Anemy (approved):

PR #6064, opened 2024-07-26
 feat(compass-editor, compass-query-bar): autocomplete icon COMPASS-8031

VivianTNT:
## Description

COMPASS-8031

Added icons next to each autocompletion item based on which type of completion it is; clock for query history, rectangular box for base completions.

image

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-editor/src/codemirror/ace-compat-autocompleter.ts
+++ b/packages/compass-editor/src/codemirror/ace-compat-autocompleter.ts
@@ -21,6 +21,7 @@ export function mapMongoDBCompletionToCodemirrorCompletion(
         ? completion.value
         : wrapField(completion.value, escape === 'always'),
     detail: completion.meta?.startsWith('field') ? 'field' : completion.meta,
+    type: 'base-autocompleter',
     info() {
       if (!completion.description) {
         return null;
--- a/packages/compass-editor/src/codemirror/ace-compat-autocompleter.ts
+++ b/packages/compass-editor/src/codemirror/ace-compat-autocompleter.ts
@@ -21,7 +21,7 @@ export function mapMongoDBCompletionToCodemirrorCompletion(
         ? completion.value
         : wrapField(completion.value, escape === 'always'),
     detail: completion.meta?.startsWith('field') ? 'field' : completion.meta,
-    type: 'base-autocompleter',
+    type: completion.meta?.startsWith('field') ? 'field' : 'method',
     info() {
       if (!completion.description) {
         return null;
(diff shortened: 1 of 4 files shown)


packages/compass-editor/src/codemirror/ace-compat-autocompleter.ts
```diff
@@ -21,6 +21,7 @@ export function mapMongoDBCompletionToCodemirrorCompletion(
         ? completion.value
         : wrapField(completion.value, escape === 'always'),
     detail: completion.meta?.startsWith('field') ? 'field' : completion.meta,
+    type: 'base-autocompleter',
```
Anemy, line 24: Adding this will start showing the rectangle box icon for all of our possible autocompletion types (`field`, `query`, `bson`, etc). This is shown in the screenshot. Should we instead show different icons for these different types? Maybe something similar to what VSCode does? https://code.visualstudio.com/docs/editor/intellisense#_types-of-completions 

Alternatively, we could hide icons for now and do that as a separate pr. It likely won't be too much so I think I'd prefer if we do it here. 

From a quick look I'm thinking right now we only need to add a new the regular box, which we can use for the functions, which would be `BSON` and `query`. Then we use the existing `base-autocompleter`/`field` rectangle box for the field autocompletions. Maybe there are more icons we can use for now? Not sure.
We get those non-field autocompletions from here: https://github.com/mongodb-js/devtools-shared/tree/main/packages/mongodb-constants 
Maybe some of the json-schema completions are `keyword`s.

packages/compass-editor/src/editor.tsx
```diff
@@ -334,14 +334,50 @@ function getStylesForTheme(theme: CodemirrorThemeType) {
...
+          backgroundRepeat: 'no-repeat',
+          backgroundPosition: 'center',
+        },
       },
       '& .cm-completionLabel': {
         flex: 1,
         overflow: 'hidden',
         textOverflow: 'ellipsis',
         whiteSpace: 'nowrap',
+        lineHeight: '110%',
```
Anemy, line 379: What is this used for?

VivianTNT: I noticed each completion wasn't centered vertically and for some reason couldn't get it to work using other properties (`display: flex` interferes with the ellipses). lineHeight is supposed to be the actual height of one completion box but I didn't know how to get that value. Can definitely look more into this!

Anemy: Cool, not a blocker or a big issue, was mostly curious. The css properties `alignItems`, `alignSelf`, `justifyContent` or `justifySelf` might be useful for the container or the classes here. They can help vertically align things in flex. It also may be that the icon is changing the height of the container. I can take more of a look later if you'd like.

VivianTNT: I think it might be because originally each of the completions were very slightly scrollable. I removed being able to scroll individual completions so that might have messed up the height, but not sure about this.

PR #6069, opened 2024-07-29
feat(compass-crud): adds ability to load more documents per page in Documents view COMPASS-6903

himanshusinghs:
## Description
This PR adds ability to view more documents per page in the Documents view using a select box in the CRUD header. To bypass the performance problem of rendering more documents we have added two more component for rendering virtual lists for both json and list views of compass-crud. We're utilising the existing package react-window for virtualisation. The other packages that we tried (virtua, react-virtuoso, @tanstack/virtual), all were a tad bit better in terms of performance but suffered problems with ResizeObserver's undelivered notifications error hence we decided to keep working with react-window here.

The PR also changes HadronDocument a little by preserving editing and deleting state on the HadronDocument itself. This is done to avoid the loss of state for documents that were being edited / deleted but pushed out of rendering window while scrolling.

So in total two important changes:
1. Virtualised lists for json and list view
2. New HadronDocumentEvents for notifying change in editing / deleting states

I would recommend starting with HadronDocument events as that is easier to go through and later catch up with the VirtualisedLists implementations.

Demo:

https://github.com/user-attachments/assets/91309898-13f5-4d34-8291-1825cbdfd8ad

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -706,19 +747,25 @@ class CrudStoreImpl
...
         const index = this.findDocumentIndex(doc);
-        this.state.docs![index] = new HadronDocument(d);
-        this.trigger(this.state);
+        const newDocs = this.state.docs
+          ? [...this.state.docs]
+          : this.state.docs;
+        newDocs?.splice(index, 1, new HadronDocument(d));
+        this.setState({
+          docs: newDocs,
+        });
```
himanshusinghs, line 760: I removed `this.trigger(this.state)` assuming that `this.setState` will do the same thing. Let me know if that's not the case.

gribnoysup (commented):
Nothing maybe to address in this particular PR, but this top block jumps so much when you change the value in the dropdown, is it just too many states changing sequentially? Maybe we can consolidate this somehow or do some sort of single in-progress state of them so that it's not so noticeable

Kapture 2024-07-31 at 14 21 14

packages/compass-crud/src/components/virtualized-document-list-view.tsx
```diff
@@ -0,0 +1,127 @@
...
+    ]
+  );
+
+  return (
+    <VirtualList
+      items={docs}
+      renderItem={renderItem}
+      estimateItemInitialHeight={estimateDocumentInitialHeight}
+      rowGap={spacing[200]}
+      className={className}
```
Anemy, line 118: Looks like for the `list` and `json` view we're now showing the padding in a way where it doesn't reach the whole page, and the scrollbar inside of the padding.
Can we update how we apply those styles, `listAndJsonStyles`, so that the padding is instead inside the scrollable container? Then the documents reach the whole way to the bottom of the screen as well.

| Compass | branch |
| - | - |
| Screenshot 2024-08-05 at 1 27 03 PM | Screenshot 2024-08-05 at 1 26 41 PM |

Anemy (commented):
Changes are looking good, left a couple comments on small things I noticed. Cool virtualization!

packages/compass-crud/src/components/virtualized-document-list-view.tsx
himanshusinghs: Just to confirm - is it that the documents don't reach all the way to the bottom or all the way to the right of Compass app (excluding padding-right)? Because I can see that the documents reach to the bottom up until the desired padding but yea the scrollbar getting in the way of padding-right is a real issue that troubles us also on the sidebar when the scroll is visible there. That's actually a problem with the library itself, with the way they render the scrollbar, its always inside the container and will render the scrollbar gutter as soon as it sees an overflow.

packages/compass-crud/src/components/editable-document.tsx
```diff
@@ -200,7 +221,7 @@ class EditableDocument extends React.Component<
     if (!this.state.editing && !this.state.deleting) {
       return (
         <DocumentList.DocumentActionsGroup
-          onEdit={this.handleEdit.bind(this)}
+          onEdit={this.handleStartEditing.bind(this)}
```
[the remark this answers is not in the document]
himanshusinghs: > Looks like there are two start editing functions here, is that intended? Looks like it used to set state and now calls it on the document.

Its now intended to call startEditing on HadronDocument because we shifted the document editing state over to HadronDocument itself. On call of that method, we update the state in the HadronDocument and also emit an event which gets picked up in this component and that updates the state of the component itself. This is done so we can preserve the state of editing / deleting when the list unmounts out of view components on scroll.

> Something we probably want to look at is when a user starts editing a document, then changes the documents shown per page. The document is still flagged for edits, but it gets back the non-edit actions:

This might be an issue, I might have missed writing a test for this. Will check it and cover that as well.

packages/compass-crud/src/components/virtualized-document-list-view.tsx
Anemy: > Just to confirm - is it that the documents don't reach all the way to the bottom or all the way to the right of Compass app (excluding padding-right)? 

I'd expect the documents to render up to the shell (when still scrolling). In main there isn't the space between the current document scrolled to and the shell (screenshot from this branch):
Screenshot 2024-08-05 at 5 00 24 PM

For the right, I expect the scrollbar to be on the farthest right, no space with the side of the app, with some spacing between the scrollbar and the document.

> That's actually a problem with the library itself, with the way they render the scrollbar, its always inside the container and will render the scrollbar gutter as soon as it sees an overflow.

Would removing the container padding and instead use margin on the documents work?

packages/compass-crud/src/components/editable-document.tsx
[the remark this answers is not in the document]
himanshusinghs: > This might be an issue, I might have missed writing a test for this. Will check it and cover that as well. Thank you for spotting <3

Ok so it turned out that component at the index retains its editing state but the underlying HadronDocument has obviously changed because of the refresh due to the change in page size. Fixed this in [fbb5ab7](https://github.com/mongodb-js/compass/pull/6069/commits/fbb5ab7238ac2e4dd4b5b261317ef2b5f22e36c1) and added a test for this as well.

packages/compass-crud/src/components/virtualized-document-list-view.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+const estimateDocumentInitialHeight = (doc: HadronDocument) => {
+  const DEFAULT_VISIBLE_FIELDS = 25;
+  const keylineCardBorder = spacing[25] + spacing[25];
+  // top and bottom padding taken up by the document card
+  const occupiedStaticSizePerItem = spacing[400] + spacing[400];
+  // total size occupied by the "Show more fields" button whenever visible
+  const showMoreButtonSize =
+    DEFAULT_VISIBLE_FIELDS < doc.elements.size
+      ? 38 //spacing[400] + spacing[400]
```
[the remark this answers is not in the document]
himanshusinghs: I updated this as per the new size of the "Show more" ([9b99950](https://github.com/mongodb-js/compass/pull/6069/commits/9b9995011d6741b0ce6b5725f5a2a596b0021279))

packages/compass-crud/src/components/virtualized-document-list-view.tsx
himanshusinghs: Thank you for the suggestion Rhys 🙌  I could not get margins to work for us (likely because everything in the list container is calculated and margins / paddings invalidates those calculations) but there were [suggestions from the library](https://github.com/bvaughn/react-window#can-i-add-padding-to-the-top-and-bottom-of-a-list) which I followed here in this [1ecac93](https://github.com/mongodb-js/compass/pull/6069/commits/1ecac93e8bd0af2a8fd870f551904e0a07970c64). This fixes the problem you mentioned but as a side effect the scroll goes a little below the actual last item itself.

image


So far my feeling is that it is unavoidable because of how we mimick adding padding to the scroll container which basically increasing the height of the container by the amount of padding.

himanshusinghs: And this is how it looks for the right view of the scroll. As you can notice I purposely reduced the spacing between the scrollbar and the KeylineCard but that is to align the KeylineCard with the end of the top toolbar header contents. Let me know what you think?

image

packages/compass-crud/src/components/virtualized-document-json-view.tsx
```diff
@@ -109,7 +107,12 @@ const VirtualizedDocumentJsonView: React.FC<
       renderItem={renderItem}
       estimateItemInitialHeight={estimateDocumentInitialHeight}
       rowGap={spacing[200]}
-      className={className}
+      useStableScrollbarGutter={true}
+      paddingLeft={spacing[400]}
+      // paddingRight is only 6px because we are using stable scrollbar gutter
+      // which takes 10px already
```
Anemy, line 113: We might want to double check how it looks on windows, the scroll bars there work a bit differently from mac, probably not an issue, I have a mac if you'd like me to check (not a blocker we can do a follow up for any small style things).

Anemy (approved):
lgtm ! Nice work on this!

packages/compass-crud/src/components/virtualized-document-list-view.tsx
Anemy: Loots good !

packages/compass-crud/src/components/virtualized-document-json-view.tsx
Anemy: Tried it out - looks Good!

Anemy (approved):

PR #6075, opened 2024-07-30
feat(compass-editor, compass-query-bar): added star icon to autocomplete favorite queries

VivianTNT:
## Description

Added star icon to autocomplete favorite query completions.
image

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-query-bar/src/components/option-editor.tsx
+++ b/packages/compass-query-bar/src/components/option-editor.tsx
@@ -24,7 +24,7 @@ import { useAutocompleteFields } from '@mongodb-js/compass-field-store';
 import { applyFromHistory } from '../stores/query-bar-reducer';
 import { getQueryAttributes } from '../utils';
 import type { BaseQuery } from '../constants/query-properties';
-import { SavedQuery } from '@mongodb-js/compass-editor/dist/codemirror/query-history-autocompleter';
+import type { SavedQuery } from '@mongodb-js/compass-editor';
 
 const editorContainerStyles = css({
   position: 'relative',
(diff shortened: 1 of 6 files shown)


packages/compass-query-bar/src/components/option-editor.tsx
```diff
@@ -21,10 +21,10 @@ import { usePreference } from 'compass-preferences-model/provider';
 import { lenientlyFixQuery } from '../query/leniently-fix-query';
 import type { RootState } from '../stores/query-bar-store';
 import { useAutocompleteFields } from '@mongodb-js/compass-field-store';
-import type { RecentQuery } from '@mongodb-js/my-queries-storage';
 import { applyFromHistory } from '../stores/query-bar-reducer';
 import { getQueryAttributes } from '../utils';
 import type { BaseQuery } from '../constants/query-properties';
+import { SavedQuery } from '@mongodb-js/compass-editor/dist/codemirror/query-history-autocompleter';
```
Anemy, line 27: Looks like this import is using the `dist` folder. Let's use the generic package import and make sure we export it there. It also looks like we're only using this as a type so we can use the `type` import syntax. 
```suggestion
import type { SavedQuery } from '@mongodb-js/compass-editor';
```

Since the `SavedQuery` type in `compass-editor` is only used in auto complete how do you feel about a more descriptive name so down the line folks don't get it mixed up? Maybe `SavedQueryAutocompletion`?

Anemy (commented):
Nice! Looks like there's a failing unit test, left one comment on type import with a suggestion. Also pinged you on article on slack on how `as` in typescript can introduce some type errors overtime, we should be able to avoid it here.

Anemy (approved):
lgtm! Let's merge once tests pass

PR #6079, opened 2024-07-31
feat(compass-query-bar): selectively autocomplete query history in additional options COMPASS-8118

VivianTNT:
## Description

Filter bar is unchanged and will display all saved queries. The additional option editors (i.e. project, sort, collation, etc.) will only display saved queries that contain that specific property. 

image

image

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-query-bar/src/components/option-editor.tsx
+++ b/packages/compass-query-bar/src/components/option-editor.tsx
@@ -161,7 +161,8 @@ export const OptionEditor: React.FunctionComponent<OptionEditorProps> = ({
       ? createQueryWithHistoryAutocompleter(
           savedQueries
             .filter((query) => {
-              const isOptionNameInQuery = optionName === 'filter' || optionName in query.queryProperties;
+              const isOptionNameInQuery =
+                optionName === 'filter' || optionName in query.queryProperties;
               const isUpdateNotInQuery = !('update' in query.queryProperties);
               return isOptionNameInQuery && isUpdateNotInQuery;
             })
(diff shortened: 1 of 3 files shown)


packages/compass-query-bar/src/components/option-editor.tsx
```diff
@@ -83,6 +83,7 @@ const insightsBadgeStyles = css({
 });
 
 type OptionEditorProps = {
+  barName: string;
```
Anemy, line 86: How do you feel about renaming this property to `optionName`? We don't refer to these as editors as bars (only the whole query bar) so I'm thinking we can make it a bit more descriptive.

VivianTNT: Agreed, just changed it

Anemy (approved):

packages/compass-query-bar/src/components/option-editor.tsx
```diff
@@ -158,7 +160,14 @@ export const OptionEditor: React.FunctionComponent<OptionEditorProps> = ({
     return isQueryHistoryAutocompleteEnabled
       ? createQueryWithHistoryAutocompleter(
           savedQueries
-            .filter((query) => !('update' in query))
+            .filter((query) => {
+              const isOptionNameInQuery =
+                optionName !== 'filter'
+                  ? optionName in query.queryProperties
+                  : true;
```
Anemy, line 167: Another way to write this is
```suggestion
              const isOptionNameInQuery = optionName === 'filter' || optionName in query.queryProperties;
```

Anemy: Ah oops, this probably makes the reformat check fail, need to run reformat on it.

Anemy (approved):

PR #6083, opened 2024-08-01
fix(sidebar): Make the hover/active background effect take up the full element. ie. flush with the sides

lerouxb:
I still have to get buy-in for this, but this is the PR so long.

### DIFF
--- a/packages/compass-connections-navigation/src/base-navigation-item.tsx
+++ b/packages/compass-connections-navigation/src/base-navigation-item.tsx
@@ -37,8 +37,6 @@ const menuStyles = css({
 });
 
 const itemContainerStyles = css({
-  paddingLeft: spacing[300],
-  paddingRight: spacing[300],
   cursor: 'pointer',
   color: 'var(--item-color)',
   backgroundColor: 'var(--item-bg-color)',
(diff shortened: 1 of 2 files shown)


packages/compass-connections-navigation/src/base-navigation-item.tsx
```diff
@@ -59,10 +57,8 @@ const itemWrapperStyles = css({
   display: 'flex',
   height: ROW_HEIGHT,
   alignItems: 'center',
-  paddingLeft: spacing[100],
-  paddingRight: spacing[100],
+  paddingRight: spacing[400],
```
lerouxb, line 60: I actually think we can push this further right. Usually there's a scrollbar which brings it out of alignment with the three dot menu next to CONNECTIONS anyway.

Anemy: Further right sounds good.

lerouxb:
> A bit unrelated, looking at it I feel the items have too much padding/spacing on their left.

The items are lined up with the search box and connections heading plus its icons and I didn't change that here yet. I think it is OK - we can change it later if people want to

PR #6094, opened 2024-08-07
fix: do not replace tabs of different connections COMPASS-8113

paula-stacho:
## Description
[COMPASS-8113](https://jira.mongodb.org/browse/COMPASS-8113)
Forces a new tab if both the active & the next workspace are of a different connection.

https://github.com/user-attachments/assets/f1bae7cb-2f74-45ad-aa50-e92f78218af8

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-workspaces/src/stores/workspaces.ts
+++ b/packages/compass-workspaces/src/stores/workspaces.ts
@@ -318,10 +318,23 @@ const reducer: Reducer<WorkspacesState> = (
           : state;
       }
 
-      // ... otherwise check if we can replace the current tab based on its
-      // replace handlers and force new tab opening if we can't
       if (currentActiveTab) {
-        forceNewTab = canReplaceTab(currentActiveTab) === false;
+        // if both the new workspace and the existing one are connection scoped,
+        // make sure we do not replace tabs between different connections
+        if (
+          action.workspace.type !== 'Welcome' &&
+          action.workspace.type !== 'My Queries' &&
+          currentActiveTab.type !== 'Welcome' &&
+          currentActiveTab.type !== 'My Queries'
+        ) {
+          forceNewTab =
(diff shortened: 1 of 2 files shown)


packages/compass-workspaces/src/stores/workspaces.ts
```diff
@@ -318,10 +318,23 @@ const reducer: Reducer<WorkspacesState> = (
...
       }
 
-      // ... otherwise check if we can replace the current tab based on its
-      // replace handlers and force new tab opening if we can't
       if (currentActiveTab) {
-        forceNewTab = canReplaceTab(currentActiveTab) === false;
+        // if both the new workspace and the existing one are connection scoped,
+        // make sure we do not replace tabs between different connections
+        if (
+          action.workspace.type !== 'Welcome' &&
```
mcasimir, line 325: Do we need to enumerate the types or this can be checked with `currentActiveTab.connectionId && action.workspace.connectionId`?

mcasimir (approved):

packages/compass-workspaces/src/stores/workspaces.ts
paula-stacho: Typescript does not allow to reference the property, because some of the workspaces don't have it. If we really wanted to avoid this, we could add `connectionId?: never` to the global workspaces, but I'd rather enumerate like this

this is not valid:
```
type MyType = { type: 'A' } | { type: 'B', prop: string };

const hasProp = (obj: MyType) => !!obj.prop;
```

this is valid:
```
type MyType = { type: 'A', prop?: never } | { type: 'B', prop: string };

const hasProp = (obj: MyType) => !!obj.prop;
```

PR #6096, opened 2024-08-07
chore(e2e): Add E2E tests for multiple connections flows

lerouxb:
Still in progress. Mostly some preliminary refactoring.


packages/compass-connections-navigation/src/navigation-item.tsx
```diff
@@ -184,20 +184,21 @@ export function NavigationItem({
     }
     if (item.type === 'connection') {
       return {
-        'data-is-active': `${isActive}`,
+        'data-is-active': isActive.toString(),
```
lerouxb, line 187: drive-by: fixing some warnings

packages/compass-connections-navigation/src/navigation-item.tsx
```diff
@@ -184,20 +184,21 @@ export function NavigationItem({
...
         'data-connection-id': item.connectionInfo.id,
         'data-connection-name': item.name,
+        'data-is-connected': (item.connectionStatus === 'connected').toString(),
```
lerouxb, line 190: There is unfortunately no easy way for an e2e test to tell if a connection is connected/disconnected. I was using expanded true/false, but it is unreliable because I now also use collapse all to make sure all the connections are rendered to work around the virtual list.

Visually there's a little green dot and the fonts are different and if you hover over it you get different options, but that's tricky to select on in tests.

packages/compass-e2e-tests/helpers/insert-data.ts
```diff
@@ -1,168 +1,217 @@
...
-  const database = typeof db === 'string' ? client.db(db) : db;
-  try {
-    await database.dropDatabase();
-  } catch (err) {
-    const codeName = (err as MongoServerError).codeName;
-    if (codeName !== 'NamespaceNotFound') {
-      throw err;
-    }
-  }
+import {
```
lerouxb, line 4: Fair amount of cleanup, refactoring and changes in this file. Probably easiest to review just the file as if it is new.

packages/compass-e2e-tests/helpers/insert-data.ts
```diff
@@ -1,168 +1,217 @@
...
+          location: { type: 'Point', coordinates: [lon(), lat()] },
+        }))
+      );
+    })
+  );
+}
 
-  const lon = () => Math.random() * 360 - 180;
-  const lat = () => Math.random() * 180 - 90;
+// WARNING: please don't export _dropDatabase because this file is written to
```
lerouxb, line 183: Added this warning and renamed the function to have an underscore in front of it because every once in a while someone starts using it on the outside. Also started writing documentation on how to write/debug e2e tests, what our standards and best practises are and so on.

packages/compass-e2e-tests/helpers/insert-data.ts
```diff
@@ -1,168 +1,217 @@
...
+    }
+  }
+}
 
-  await db.collection('geospatial').insertMany(
-    [...Array(1000).keys()].map(() => ({
-      location: { type: 'Point', coordinates: [lon(), lat()] },
-    }))
-  );
+// WARNING: please don't export _createBlankCollection because this file is
```
lerouxb, line 199: Same as above.

packages/compass-e2e-tests/helpers/insert-data.ts
lerouxb: Also some more comments at the top of the file around how to use test databases and collections and where to add them.

packages/compass-e2e-tests/package.json
```diff
@@ -17,8 +17,12 @@
     "test-packaged": "npm run test -- -- --test-packaged-app",
     "test-packaged-ci": "npm run test-ci -- -- --test-packaged-app",
     "reformat": "npm run eslint . -- --fix && npm run prettier -- --write .",
-    "start-server": "mongodb-runner start --id=e2e --topology=replset --secondaries=0 -- --port 27091",
-    "stop-server": "mongodb-runner stop --id=e2e",
+    "start-server-1": "mongodb-runner start --id=e2e-1 --topology=replset --secondaries=0 -- --port 27091",
+    "stop-server-1": "mongodb-runner stop --id=e2e-1",
+    "start-server-2": "mongodb-runner start --id=e2e-2 --topology=replset --secondaries=0 -- --port 27092",
+    "stop-server-2": "mongodb-runner stop --id=e2e-2",
+    "start-servers": "npm run start-server-1 && npm run start-server-2",
```
lerouxb, line 24: There are two now. Yes we could have just had two connections to the same server, but this way we can hopefully pick up if something unexpectedly appears in the wrong server, for example.

packages/compass-e2e-tests/tests/collection-rename.test.ts
```diff
@@ -6,16 +6,15 @@ import {
...
 import type { CompassBrowser } from '../helpers/compass-browser';
-import { createBlankCollection, dropDatabase } from '../helpers/insert-data';
+import { createDummyCollections } from '../helpers/insert-data';
 import * as Selectors from '../helpers/selectors';
 
-const initialName = 'numbers';
-const newName = 'renamed';
-
-const databaseName = 'rename-collection';
+const databaseName = 'test';
```
lerouxb, line 15: I changed this file to just use the same test database we use everywhere so it will be cleaned up between tests in the standard way along with everything else. No reason for a new database to test renaming a collection.

packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -342,6 +346,16 @@ describe('Connection string', function () {
         .getText();
       expect(errorMessage).to.equal('Authentication failed.');
     }
+
+    // for multiple connections click the review button in the toast
```
lerouxb, line 350: There was a bullet point in the ticket to test this and I realised I could just tack it onto an existing test.

packages/compass-e2e-tests/tests/tabs.test.ts
```diff
@@ -129,4 +134,96 @@ describe('Global Tabs', function () {
     // When confirmed, should remove the tab
     expect(await browser.$$(Selectors.workspaceTab())).to.have.lengthOf(0);
   });
+
+  it("should close a connection's tabs when disconnecting", async function () {
```
lerouxb, line 138: New test.

packages/compass-connections-navigation/src/navigation-item.tsx
paula-stacho: this actually points to an a11y problem, we should probably add an aria-label to the green dot. not sure if that'd be convenient enough for all the e2e tests, but it's for sure not good that we don't have anything beside little visual changes.

packages/compass-e2e-tests/helpers/commands/close-settings-modal.ts
```diff
@@ -11,7 +11,6 @@ export async function closeSettingsModal(
   const settingsModalElement = await browser.$(Selectors.SettingsModal);
 
   await settingsModalElement.waitForDisplayed();
-  await browser.screenshot('settings-modal.png');
```
paula-stacho, line 14: what were these screenshots used for?

lerouxb: Anything named modal: I worked on the modals 22 months ago and just screenshotted them all using CI so we can look at before vs after. Everything else: debugging flakes that are long gone.

lerouxb: From the PR description:

> Drive-by: I removed a bunch of often two-year-old debug screenshot code because there are just so many screenshots cluttering the output when you're looking for an error screenshot.

packages/compass-connections-navigation/src/navigation-item.tsx
lerouxb: the green dot is a different element to the connection so would still be a pain because then we'd need separate iterative code to go back up the tree. You can do that in addition to this.

packages/compass-e2e-tests/helpers/commands/connect-form.ts
```diff
@@ -0,0 +1,1013 @@
...
+    : Selectors.ConnectionTitle;
+
+  const connectionTitle = await browser.$(connectionTitleSelector);
+  await connectionTitle.waitUntil(async () => {
+    return (await connectionTitle.getText()) === 'New Connection';
+  });
+
+  await browser.waitUntil(async () => {
+    return (
+      (await browser.getConnectFormConnectionString(true)) ===
```
paula-stacho, line 44: detail, but why do we explicitly focus the input here?

packages/compass-connections-navigation/src/navigation-item.tsx
paula-stacho: that's fair!

packages/compass-e2e-tests/helpers/commands/close-settings-modal.ts
paula-stacho: ah, sorry, somehow I missed that comment

packages/compass-e2e-tests/helpers/commands/connect-form.ts
```diff
@@ -0,0 +1,1013 @@
...
+
+export async function setConnectFormState(
+  browser: CompassBrowser,
+  state: ConnectFormState
+): Promise<void> {
+  await browser.resetConnectForm();
+
+  // Something to keep in mind is that if you specify both connectionString AND
+  // other options, then the other options are going to override the
+  // connectionString. You probably want just one or the other.
```
paula-stacho, line 471: is it a hard rule that could even be validated against and throw, or more of a rule of thumb?

lerouxb: Rule of thumb, I think.

packages/compass-e2e-tests/helpers/commands/connect.ts
```diff
@@ -0,0 +1,233 @@
...
+  { connectionStatus = 'success', timeout }: ConnectionResultOptions = {}
+) {
+  const waitOptions = typeof timeout !== 'undefined' ? { timeout } : undefined;
+
+  if (connectionStatus === 'either') {
+    // TODO(COMPASS-7600): this doesn't support compass-web yet, but also isn't
+    // encountered yet For the rare cases where we don't care whether it fails
+    // or succeeds
+    await browser
+      .$(`${Selectors.DatabasesTable},${Selectors.ConnectionFormErrorMessage}`)
```
paula-stacho, line 144: the error part will not work for multiple connections because it's missing the toast click

packages/compass-e2e-tests/helpers/commands/connect.ts
```diff
@@ -0,0 +1,233 @@
...
+    if (TEST_COMPASS_WEB) {
+      // In compass-web, for now, we land on the Databases tab after connecting
+      await browser
+        .$('[data-testid="workspace-tab-button"][data-type=Databases]')
+        .waitForDisplayed();
+    } else if (TEST_MULTIPLE_CONNECTIONS) {
+      // For multiple connections, make sure the exact named connection is
+      // expanded. We're assuming that expanded means connected, although it is
+      // technically possible to collapse a connected connection. Something we
+      // might want to improve on later.
```
paula-stacho, line 159: looks like we can remove this comment

packages/compass-e2e-tests/helpers/selectors.ts
```diff
@@ -303,6 +307,13 @@ export const Multiple = {
     '[data-testid="connections-list-title-actions-import-saved-connections-action"]',
 
   InUseEncryptionMarker: '[data-action="open-csfle-modal"]',
+
+  ConnectedConnectionItems:
+    '[role="treeitem"][aria-level="1"] [data-is-connected=true]',
+
+  connectionItemByName: (connectionName: string, connected: boolean) => {
```
paula-stacho, line 314: `connectionItemByNameAndStatus`? to prevent confusion in case we'll want one without the status

packages/compass-e2e-tests/helpers/commands/connect.ts
lerouxb: This is also old code I moved. It has probably just never been encountered. The next subtask is to test the multiple connection errors and then I'll have to change this. There's only one test that uses this and it is currently skipped for multiple connections.

packages/compass-e2e-tests/helpers/selectors.ts
lerouxb: We can also just change it once we need it. This is all internal detail to this one package.

lerouxb: Another design might be to turn the second parameter into an options hash and have connected be optional defaulting to undefined. If it is undefined it is igored.

packages/compass-e2e-tests/helpers/commands/connect.ts
lerouxb: I updated the TODO comment.

packages/compass-e2e-tests/helpers/commands/disconnect.ts
```diff
@@ -15,76 +77,65 @@ export async function disconnectAll(
...
-        } catch (error) {
-          debug('ignoring', error);
-        }
-      }
-    }
+export async function disconnectByName(
+  browser: CompassBrowser,
+  connectionName: string
+) {
+  await resetForDisconnect(browser, { closeToasts: false });
```
paula-stacho, line 127: if I get it right, the reset now happens 1 + N times for `disconnectAll`, maybe we could make it skippable in this case?

packages/compass-e2e-tests/helpers/insert-data.ts
```diff
@@ -1,168 +1,217 @@
...
+import {
+  DEFAULT_CONNECTION_STRING_1,
+  DEFAULT_CONNECTION_STRING_2,
+  TEST_MULTIPLE_CONNECTIONS,
+} from './compass';
+
+// This is a list of all the known database names that get created by tests so
+// that we can know what to drop when we clean up before every test. If a new
+// test starts to create another database, add it here so that it will also be
+// cleaned up.
```
paula-stacho, line 13: I just know I will forget to do that 😅 Can't we get it with listDatabases?

packages/compass-e2e-tests/helpers/insert-data.ts
```diff
@@ -1,168 +1,217 @@
...
+let clients: MongoClient[];
+let test_dbs: Db[];
+
+before(async () => {
+  // Insert data on both connections so that the same databases and collections
+  // will exist on both servers and then anything that's not properly scoped to
+  // the correct connection has a chance to operate on the wrong one and
+  // hopefully fail a test.
+  // This should also mean that the database or collection name that we try and
+  // use is always ambiguous, so we're forced to deal with it early in tests.
```
paula-stacho, line 55: 👍

packages/compass-e2e-tests/helpers/insert-data.ts
lerouxb: Sure. File a ticket :)

packages/compass-e2e-tests/helpers/commands/disconnect.ts
paula-stacho: doesn't it take some time?

packages/compass-e2e-tests/helpers/insert-data.ts
lerouxb: (It worked this way before and I'm hesitant to do that kind of thing as a drive-by because at minimum it is going to have to filter out some system/builtin databases)

packages/compass-e2e-tests/helpers/commands/disconnect.ts
lerouxb: milliseconds. nothing compared to the general overhead of creating connections and connecting to them and browsing to where we need to go and waiting for toasts and for collections to load, etc, etc.

PR #6101, opened 2024-08-08
fix(telemetry): better hostname parsing for telemetry COMPASS-8148

alenakhineika:
## Description

- Return hostname without port
- Fall back to original hostname from uri when resolving srv failed

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass/src/app/utils/telemetry.ts
+++ b/packages/compass/src/app/utils/telemetry.ts
@@ -115,15 +115,15 @@ async function getHostnameForConnection(
         return null;
       }
     );
-    if (!uri) {
-      return null;
+    if (uri) {
+      connectionStringData = new ConnectionString(uri, {
+        looseValidation: true,
+      });
     }
-    connectionStringData = new ConnectionString(uri, {
-      looseValidation: true,
-    });
   }
 
-  return connectionStringData.hosts[0];
+  const [hostname] = (connectionStringData.hosts[0] ?? '').split(':');
+  return hostname;
 }
 
 async function getConnectionData(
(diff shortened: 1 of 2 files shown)


packages/compass/src/app/utils/telemetry.ts
```diff
@@ -115,15 +115,15 @@ async function getHostnameForConnection(
...
+        looseValidation: true,
+      });
     }
-    connectionStringData = new ConnectionString(uri, {
-      looseValidation: true,
-    });
   }
 
-  return connectionStringData.hosts[0];
+  const [hostname] = (connectionStringData.hosts[0] ?? '').split(':');
```
addaleax, line 125: I feel like this is code I've written elsewhere as well, but keep in mind that the hostname itself can also contain `:` if it is an IPv6 address

addaleax (approved):

packages/compass/src/app/utils/telemetry.ts
alenakhineika: Can it be Atlas host as well with an IPv6 address?

alenakhineika: And another question, did we solve this in this other project or do we keep doing the same? I am wondering if we want to address it here or this is fine leave as is.

alenakhineika: Note: As discussed in Slack we want to handle IPv6.

addaleax (approved):

PR #6109, opened 2024-08-13
fix(connection-form): move Red from the start of the color wheel to the end COMPASS-8146

lerouxb:
Now people can start complaining that Pink looks red enough to be an error and Orange and Yellow could both be warnings 😆 

Screenshot 2024-08-13 at 13 49 16
Screenshot 2024-08-13 at 13 49 20


lerouxb:
Apparently not what we want. Closing.

PR #6110, opened 2024-08-13
fix(connect-form): rotate the colour wheel so that red lines up with red COMPASS-8146

lerouxb:
If the user previously selected 'color1' in the Single Connection world they now get Green in Multiple Connections, not Red. If the user selected 'color6' (which used to be Red) they now still get Red. The color order still follows the color wheel, it just has a different starting point.

Screenshot 2024-08-13 at 15 03 21


mcasimir (approved):

packages/compass-e2e-tests/helpers/commands/connect-form.ts
```diff
@@ -442,15 +442,15 @@ async function waitForElementAnimations(browser: CompassBrowser, element: any) {
...
-  color1: 'Red',
-  color2: 'Pink',
-  color3: 'Orange',
-  color4: 'Yellow',
-  color5: 'Green',
-  color6: 'Teal',
-  color7: 'Blue',
-  color8: 'Iris',
-  color9: 'Purple',
+  color1: 'Green',
```
lerouxb, line 445: I can probably import this from usec-connection-color.ts or a common place rather than maintain another copy here, but I suppose it really isn't the end of the world if the E2E tests expect the colours to be what they are.

paula-stacho: it's actually not a bad thing to check for 👍

paula-stacho (approved):

PR #6117, opened 2024-08-14
chore(e2e): add E2E test for multiple connection errors COMPASS-8153

lerouxb:
Connects twice with different errors, then clicks review on each error toast and makes sure that the modal populates with that connection's info and shows the error.

### DIFF
--- a/packages/compass-e2e-tests/tests/connection.test.ts
+++ b/packages/compass-e2e-tests/tests/connection.test.ts
@@ -281,6 +281,12 @@ async function assertCannotCreateCollection(
   await createModalElement.waitForDisplayed({ reverse: true });
 }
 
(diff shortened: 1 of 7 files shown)


packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -961,12 +1077,6 @@ describe('SRV connectivity', function () {
   });
 
   it('resolves SRV connection string using OS DNS APIs', async function () {
-    if (TEST_MULTIPLE_CONNECTIONS) {
```
lerouxb, line 964: In case it isn't obvious, I also re-enabled this test for multiple connections.

addaleax (approved):

PR #6125, opened 2024-08-15
chore(e2e): firstRun: true by default, remove connections

lerouxb:
The connections test file creates a ton of connections and we're just about at the point where virtual scrolling kicks in. browser.removeAllConnections() allows us to remove them all between tests.

Furthermore firstRun now defaults to true so that each test file now gets its own user directory and therefore things like connections, saved queries and so on get reset every time. Only the tests that care about starting compass twice in a row and keeping the user directory have it set to false. This should isolate tests better with the downside being that we're almost always testing the first run use case.

### DIFF
--- a/packages/compass-e2e-tests/helpers/commands/index.ts
+++ b/packages/compass-e2e-tests/helpers/commands/index.ts
@@ -7,6 +7,7 @@ export * from './close-welcome-modal';
 export * from './connect';
 export * from './connect-form';
 export * from './disconnect';
+export * from './remove-connections';
 export * from './shell-eval';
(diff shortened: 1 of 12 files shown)


gribnoysup (approved):
Looks like a few failing tests actually expect the session to be preserved and so are failing right now due to missing connections in the navigation? Otherwise looks good, thanks for looking into it!

lerouxb:
> Looks like a few failing tests actually expect the session to be preserved and so are failing right now due to missing connections in the navigation?

yeah I was too lazy to find them manually, so waited for CI to tell me 😆

PR #6132, opened 2024-08-16
chore: extract event types COMPASS-8058

mcasimir:
Extracts types for events send via the track function to one central file.

TODO (in a follow up so new events aren't added in the meanwhile):
- use stricter, more precise types for properties
- document properties
- generate tracking plan in markdown


packages/compass-shell/src/modules/worker-runtime.ts
```diff
@@ -59,13 +59,13 @@ export function createWorkerRuntime(
...
       },
-      // Prefix Segment events with `Shell ` to avoid event name collisions.
+      // Forward Segment events from the shell.
       // We always enable telemetry here, since the track call will
       // already check whether Compass telemetry is enabled or not.
       track: ({ event, properties }) => {
         return track(
-          `Shell ${event}`,
-          properties,
+          'Shell Event',
```
addaleax, line 67: Is this still going to be consumable in Segment/Amplitude?

packages/compass-telemetry/src/telemetry-events.ts
```diff
@@ -0,0 +1,889 @@
...
+  name: 'Database Created';
+  payload: {
+    is_capped: boolean;
+    has_collation: boolean;
+    is_timeseries: boolean;
+    is_clustered: boolean;
+    is_fle2: boolean;
+    expires: boolean;
+  };
+}>;
```
addaleax, line 780: Honest Q: Who's the intended target audience for the JSDoc output, and would they maybe also be fine with this file? This is quite clean and readable already imo

mcasimir: I would imagine that being ok yeah. Is us, analytics folks and product. I think that the comments would be enough to read. The main effort remaining would really be adding the missing documentation. I've grabbed what i could from the existing tracking plan, but it lacks lots of events :(.

packages/compass-shell/src/modules/worker-runtime.ts
mcasimir: asked analytics, this is my speculation, i believe it may require an additional index, after that it would actually be a little more comfortable since they don't need to strip "Shell " from the event (nothing impossible) for cross-tool analysis .. but i don't actually know.

Let's see, in case we can find an alternative, i guess just skipping type checking this one track call here may be ok.

addaleax: Yeah, I'd imagine that `as any` is fine for this one. Plus, if we ever do something similar for the shell, we could go back to having proper typings here pretty easily again

mcasimir: Reverted and used a template literal type, thanks @paula-stacho

packages/compass-telemetry/src/types.ts
```diff
@@ -1,22 +1,19 @@
...
-export type AsyncTrackFunction = (
-  event: string,
-  parameters?: TrackParameters,
-  connectionInfo?: TelemetryConnectionInfo
-) => Promise<void>;
+export interface TrackFunction {
+  (
+    eventName: TelemetryEvent['name'],
+    payload: Omit<TelemetryEventPayload, 'connection_id'>,
+    connectionInfo?: TelemetryConnectionInfo | undefined
```
paula-stacho, line 17:
```suggestion
    connectionInfo?: TelemetryConnectionInfo
```

packages/compass-telemetry/src/telemetry-events.ts
```diff
@@ -0,0 +1,1435 @@
...
+/**
+ * This event is fired when user adds/remove a stage or changes the stage name in the stage editor view
+ *
+ * @category Aggregation Builder
+ */
+type AggregationEditedEvent = ConnectionScoped<{
+  name: 'Aggregation Edited';
+  payload: {
+    num_stages?: any | number;
+    editor_view_type?: string | 'stage' | 'text' | 'focus';
```
paula-stacho, line 54: so just string?

paula-stacho:
@mcasimir I am curious what did you use to extract the types?

packages/compass-telemetry/src/generic-track.ts
```diff
@@ -33,9 +37,11 @@ export const createTrack = ({
       return;
     }
 
-    if (typeof parameters === 'function') {
+    let parameters: Record<string, unknown> = parametersOrFn;
```
paula-stacho, line 40: this should not work, because at this point we don't know that the parameters are not a function

packages/compass-telemetry/src/types.ts
```diff
@@ -1,22 +1,19 @@
...
+  ): void;
+}
```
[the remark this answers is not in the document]
paula-stacho: But I actually tried the code above and it doesn't work, I think the problem is with the `{}` payloads - I understand why they were added, but it looks like the warning is not for nothing and the type really works in an unexpected way

packages/compass-telemetry/src/telemetry-events.ts
lerouxb: similarly if num_stages is any | number it might as well be any?

packages/compass-telemetry/src/types.ts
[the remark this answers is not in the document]
paula-stacho: TIL: Extract 😊

packages/compass-telemetry/src/telemetry-events.ts
```diff
@@ -0,0 +1,1435 @@
...
+/**
+ * This event is fired when a user cancel a running aggregation
+ *
+ * @category Aggregation Builder
+ */
+type AggregationCanceledEvent = ConnectionScoped<{
+  name: 'Aggregation Canceled';
+  payload: {
+    //
+  };
```
paula-stacho, line 80:
```suggestion
  payload: Record<string, never>;
```

mcasimir:
> I am curious what did you use to extract the types

typescript

packages/compass-telemetry/src/telemetry-events.ts
mcasimir: These types are extracted with a script, that's why i got `string` in some places. In my opinion, where the values are meaningful for analysis we should keep the stricter type and document them, my intention is to address any manual tweak in a follow up:

> TODO (in a follow up so new events aren't added in the meanwhile):
>
> use stricter, more precise types for properties
> document properties
> generate tracking plan in markdown

packages/compass-telemetry/src/generic-track.ts
mcasimir: That's true! thanks for catching it! I was using `Omit` on that function type, which just changes the type

packages/compass-telemetry/src/types.ts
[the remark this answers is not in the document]
mcasimir: Yup that's the goal, let me try that out!

packages/compass-telemetry/src/telemetry-events.ts
mcasimir: Not that it makes a big difference, but i find the empty payload a lot more immediate to read, is there a benefit to use the  `Record<string, never>`?

addaleax: eslint should yell at you if you use `{}` because `{}` just means any non-nullish value 🙂

packages/compass-telemetry/src/types.ts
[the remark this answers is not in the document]
mcasimir: Works wonderfully. And now it does catch a lot of "undefined" that the extraction script completely ignored.

PR #6151, opened 2024-08-22
fix(compass-import-export): stream import errors to file with back pressure COMPASS-7820

lerouxb:
Draft because I still have to update the tests.

We were keeping all import errors in memory and only writing it to a file at the end. This now writes them to the log file during the import and waits for the write to continue so we get proper back pressure. This also means that I finally got to replace collection-stream.ts which is the only bit that survived the rewrite largely intact. Lots of unnecessary code got removed in the process. There is also a lot more code reused between import csv and import json now.

The easiest way to test the most command / worst case scenario is to export a large collection then import the file to a new collection twice. The second time every id will trigger a duplicate key error. If you import a large file, that's a large amount of stuff temporarily kept in memory.

It would have been tricky to update the existing stream based code to do that and node's been supporting async iterators and generators for some time now plus that "just works" with streams, so I took the opportunity to rewrite most of import to use that. I think this nicely simplifies matters.

As a drive-by I renamed the UI state var from `errors` to `firstErrors` because we only keep up to 5 errors in the UI. There was a bug in the telemetry where it took the length of that array rather than using the numErrors var.

Probably easiest to just read through `packages/compass-import-export/src/import/import-csv.ts` or `packages/compass-import-export/src/import/import-json.ts` (they use most of the same code now, so very little code left in each) and then `doImport()` in `packages/compass-import-export/src/import/import-utils.ts`. Also obviously packages/compass-import-export/src/import/import-writer.ts which replaced collection-stream.


packages/compass-import-export/src/import/import-csv.spec.ts
```diff
@@ -778,43 +738,35 @@ describe('importCSV', function () {
       errorCallback,
     });
 
-    expect(result.dbStats.insertedCount).to.equal(0);
+    expect(result.docsWritten).to.equal(0);
 
     expect(progressCallback.callCount).to.equal(3);
-    expect(errorCallback.callCount).to.equal(1); // once for the batch
+    expect(errorCallback.callCount).to.equal(3);
```
lerouxb, line 744: There was this inconsistency where it was one error callback per batch of errors in some cases. It is always one per error now.

packages/compass-import-export/src/import/import-csv.ts
```diff
@@ -1,47 +1,91 @@
...
+        const fixupTo = fixupFrom.replace(/\r$/, '');
+        this.headerFields[this.headerFields.length - 1] = fixupTo;
+      }
+
+      this.parsedHeader = {};
+      for (const name of this.headerFields) {
+        this.parsedHeader[name] = parseCSVHeaderName(name);
+      }
+
+      // TODO(COMPASS-7158): make sure array indexes start at 0 and have no
```
lerouxb, line 67: This was an existing TODO I just moved with the code.

gribnoysup (approved):

PR #6166, opened 2024-08-29
feat(connection-form): support for multiple kms options from same provider COMPASS-8082

mabaasit:
<details>
<summary>Preview</summary>

image

</details>

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-connections/src/stores/connections-store-redux.ts
addaleax, line 1: Note quite sure where to put this, so just leaving it here on the first file:

Screenshot 2024-08-29 at 19 01 49

(For a connection configured with 2 `local` providers, `local` and `local:1`)

Clicking the empty button crashes the app

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/kms-provider-content.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+function getNextKmsProviderName<T extends KMSProviderType>(
+  kmsProviderType: T,
+  currentProviders: string[]
+): KMSProviderName<T> {
+  const currentNums = currentProviders
+    .map((name) => parseInt(name.split(':')[1], 10))
+    .filter((num) => !isNaN(num));
+  const name = currentNums.length === 0 ? 1 : Math.max(...currentNums) + 1;
+  return `${kmsProviderType}:${name}`;
```
addaleax, line 44: I don't know if we can select the name for the user here. The name _is_ persisted in the key vault collection, and while I'm not entirely sure, I would guess that it has to match between different usages of the same key (i.e. the name used by the application `kmsProviders` config should match the one used here).

I wanted to verify this myself but ran into some issues with my local FLE setup that I'll probably need to look into first

addaleax (commented):
Left two comments before diving into a bigger review, hopefully tomorrow 🙂

packages/compass-connections/src/stores/connections-store-redux.ts
mabaasit: actually i did not know that user can choose kms provider here. thanks for catching this.

packages/databases-collections/src/modules/create-namespace.ts
```diff
@@ -168,13 +168,18 @@ const reducer: Reducer<CreateNamespaceState, Action> = (
...
     return { ...INITIAL_STATE };
   }
 
+  if (isAction<CloseAction>(action, CreateNamespaceActionTypes.Close)) {
+    // When a modal is closed, we should not clear the connectionMetaData
+    return {
+      ...INITIAL_STATE,
+      connectionMetaData: state.connectionMetaData,
+    };
+  }
```
mabaasit, line 181: Drive-by: when opening modal to create a collection after closing it first time, the UI did not show encryption options, because we set `connectionMetaData` to initial state (`{}`).

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/kms-provider-content.tsx
mabaasit: I updated the PR to add support for renaming kms. Currently it allows users to edit the name while warning them about:
1. Empty name
2. Duplicate name

In these cases, if a user proceeds to connect, currently it considers the last valid name (as we do not rename). 

It hides the full name of the provider (the prefix part) - e.g if the user types local123, the actual name being `local:local123`, we hide the `local:`. I am waiting design input on how to communicate this to the user.

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/kms-provider-content.tsx
```diff
@@ -0,0 +1,101 @@
...
+  currentProviders: string[] = []
+): KMSProviderName<T> {
+  // For name, we are prefixing it with the type of the provider separated by a colon.
+  // This is because the name of the kms provider should always have the type of the provider
+  // and we are not showing it to the user and it is used internally.
+  if (currentProviders.length === 0) {
+    return `${kmsProviderType}:${kmsProviderType}1`;
+  }
+  const currentNums = currentProviders // local:local1
+    .map((name) => name.split(':')[1]?.replace(kmsProviderType, '')) // '1'
```
addaleax, line 33: The `[1]` here works because KMS provider names need to follow a `^<type>:[a-zA-Z0-9_]+$` scheme – should we validate that the name part only consists of `[a-zA-Z0-9_]+` to ensure that it doesn't contain a `:`?

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/kms-provider-content.tsx
```diff
@@ -0,0 +1,101 @@
...
+    .map((x) => parseInt(x, 10)) // 1
+    .filter((x) => !isNaN(x));
+  const nextNum = currentNums.length === 0 ? 1 : Math.max(...currentNums) + 1;
```
addaleax, line 36:
```suggestion
  const nextNum = Math.max(0, ...currentNums) + 1;
```

packages/connection-form/src/utils/csfle-handler.ts
```diff
@@ -152,29 +178,25 @@ export function handleUpdateCsfleKmsTlsParam({
...
   } else {
     tls[action.key] = action.value;
   }
   const tlsOptions = autoEncryption.tlsOptions ?? {};
-  if (Object.keys(tls).length === 0) {
-    delete tlsOptions[action.kmsProvider];
-  } else {
-    tlsOptions[action.kmsProvider] = tls;
-  }
+  tlsOptions[action.kmsProviderName] = tls;
```
addaleax, line 189: Our of curiousity, what's bringing on this change? If the KMS provider name changes, does this "leak" TLS options that do not correspond to any of the KMS providers?

packages/connection-info/src/connection-secrets.ts
```diff
@@ -190,10 +181,75 @@ export function extractSecrets(connectionInfo: Readonly<ConnectionInfo>): {
...
 
-function omitPropertiesWhoseValuesAreEmptyObjects<
-  T extends Record<string, Record<string, unknown>>
->(obj: T): Partial<T> {
+function omitPropertiesWhoseValuesAreEmptyObjects<T extends Document>(obj: T) {
   return Object.fromEntries(
     Object.entries(obj).filter(([, value]) => Object.keys(value).length > 0)
-  ) as Partial<T>;
+  );
+}
```
addaleax, line 188: This just removes the return type annotation and reduces type safety overall – any particular reason for that?

packages/connection-info/src/connection-secrets.ts
```diff
@@ -190,10 +181,75 @@ export function extractSecrets(connectionInfo: Readonly<ConnectionInfo>): {
...
+
+function extractAutoEncryptionSecrets(data: Document) {
```
addaleax, line 190:
```suggestion
function extractAutoEncryptionSecrets(data: AutoEncryptionOptions) {
```

and this should have a return type annotation as well

packages/connection-info/src/connection-secrets.ts
```diff
@@ -190,10 +181,75 @@ export function extractSecrets(connectionInfo: Readonly<ConnectionInfo>): {
...
+  function flattenObject(
+    obj: Document,
+    path = '',
+    result: Document = {}
+  ): Record<string, unknown> {
+    _.forEach(obj, (value, key) => {
+      key = key.replaceAll('.', '\\.');
```
addaleax, line 197: This probably works out okay because our keys do not contain literal `\.` sequences, but this code overall isn't the easiest to read – I don't think there's anything requiring us to stick to a paradigm where we specify the set of paths that we want to redact as a list of strings, manually redacting these objects the way we do for other sub-properties of `ConnectionOptions` is probably just fine too and might end up being both more readable and more type-safe

packages/connection-form/src/utils/csfle-handler.ts
mabaasit: ah, this i missed completely when renaming. i'll ensure in rename, tlsoption is also renamed. 
and this should remain as original code. will revert this change.

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/kms-provider-card.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const [hoverProps, isHovered] = useHoverState();
+  const [cardName, setCardName] = useState(kmsProviderName);
+  const [nameError, setNameError] = useState<string | undefined>();
+
+  const onRenameKmsProvider = useCallback(
+    (name: string) => {
+      const newName = `${kmsProviderType}:${name}` as KMSProviderName<T>;
+      setCardName(newName);
+      if (name === '') {
+        setNameError('Name cannot be empty');
```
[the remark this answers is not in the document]
mabaasit: this was enforcing all the kms provider names to be prefixed by the type. i'll remove this check and allow users to have kms type as name.

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/kms-provider-content.tsx
mabaasit: good point, should we actually force this pattern (`^<type>:[a-zA-Z0-9_]+$`) on the name of the kms? it will also simply extracting secrets?

addaleax: I think it's okay to do that at least through TextInput validation if we can, yeah

packages/connection-form/src/components/advanced-options-tabs/csfle-tab/kms-provider-card.tsx
```diff
@@ -0,0 +1,244 @@
...
+        connectionOptions={connectionOptions}
+        updateConnectionFormField={updateConnectionFormField}
+        kmsProviderType={kmsProviderType}
+        kmsProviderName={kmsProviderName}
+        fields={fields}
+        clientCertIsOptional={clientCertIsOptional}
+        noTLS={noTLS}
+      />
+    </Card>
+  );
```
addaleax, line 241: Just wanted to say, I tried this out locally and it all feels very smooth and intuitive! This is amazing work ✨

packages/connection-form/src/utils/csfle-handler.ts
```diff
@@ -184,10 +206,55 @@ export function handleUpdateCsfleKmsTlsParam({
...
+    };
+  }
+
+  function filterEmptyValues(obj: Record<string, any> | undefined) {
+    const values = Object.fromEntries(
+      Object.entries(obj ?? {})
+        .filter(([, v]) => Object.keys(v ?? {}).length > 0)
+        .map(([k, v]) => [k, v])
+    );
+    return Object.keys(values).length > 0 ? values : undefined;
```
addaleax, line 232: Nit, we can use generics here to improve type safety:

```suggestion
  function filterEmptyValues<T extends object>(obj: T | undefined): Partial<T> | undefined {
    const values = Object.fromEntries(
      Object.entries(obj ?? {})
        .filter(([, v]) => Object.keys(v ?? {}).length > 0)
    );
    return Object.keys(values).length > 0 ? values as Partial<T> : undefined;
```

packages/connection-info/src/connection-secrets.ts
```diff
@@ -190,10 +181,62 @@ export function extractSecrets(connectionInfo: Readonly<ConnectionInfo>): {
...
+function extractAutoEncryptionSecrets(data: AutoEncryptionOptions): {
+  data: AutoEncryptionOptions & AutoEncryptionKMSAndTLSOptions;
+  secrets: AutoEncryptionKMSAndTLSOptions;
+} {
+  const secrets: AutoEncryptionKMSAndTLSOptions = {};
+  // Secrets are stored in a kmsProviders and tlsOptions
+  const { kmsProviders, tlsOptions, ...result } = data;
+
+  for (const kmsProviderName in kmsProviders ?? {}) {
+    const key = kmsProviderName.match(/\b(local|aws|gcp|azure)\b(?:)?/)?.[1] as
```
addaleax, line 210: Just a question – Does

```suggestion
    const key = kmsProviderName.split(':')[0] as
```

work here?

Currently, this would end up matching something like `kmip:local` as `local`, which hopefully won't be a common case, but maybe this can just be avoided this way?

packages/connection-info/src/connection-secrets.ts
```diff
@@ -190,10 +181,62 @@ export function extractSecrets(connectionInfo: Readonly<ConnectionInfo>): {
...
+
+function extractAutoEncryptionSecrets(data: AutoEncryptionOptions): {
+  data: AutoEncryptionOptions & AutoEncryptionKMSAndTLSOptions;
+  secrets: AutoEncryptionKMSAndTLSOptions;
+} {
+  const secrets: AutoEncryptionKMSAndTLSOptions = {};
+  // Secrets are stored in a kmsProviders and tlsOptions
+  const { kmsProviders, tlsOptions, ...result } = data;
+
+  for (const kmsProviderName in kmsProviders ?? {}) {
```
addaleax, line 209:
```suggestion
  for (const kmsProviderName of Object.keys(kmsProviders ?? {})) {
```

Not a huge deal, but `for ... in` is generally considered a bit less reliable because it walks the prototype chain, so while that doesn't make a big difference here, it's something we shouldn't start making common use of in the codebase

addaleax (commented):
🚀

packages/connection-info/src/connection-secrets.ts
mabaasit: fixed in [3f954e6](https://github.com/mongodb-js/compass/pull/6166/commits/3f954e6b6a3fb85f4943a929b5c551ff8901bb29)

addaleax (approved):
Looks good, feel free to consider the remaining unresolved comments optional 👍

packages/connection-form/src/utils/csfle-handler.ts
mabaasit: reverted in 1b179f689066e3d4948304a2345e0fad9335c7fa

packages/connection-info/src/connection-secrets.ts
addaleax: I mean the return type annotation is still missing but that's not a big deal

mabaasit: 🤦🏽

PR #6176, opened 2024-08-30
chore: improve types and documentation for telemetry

mcasimir:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-aggregations/src/components/stage-editor/stage-editor.tsx
+++ b/packages/compass-aggregations/src/components/stage-editor/stage-editor.tsx
@@ -143,7 +143,7 @@ export const StageEditor = ({
         {
           num_stages: num_stages,
           stage_index: index + 1,
-          stage_action: 'stage_content_changed' as const,
+          stage_action: 'stage_content_changed',
           stage_name: stageOperator,
           editor_view_type: editor_view_type,
         },
--- a/packages/compass-aggregations/src/components/stage-editor/stage-editor.tsx
+++ b/packages/compass-aggregations/src/components/stage-editor/stage-editor.tsx
@@ -77,7 +77,7 @@ type StageEditorProps = {
   syntaxError: PipelineParserError | null;
   serverError: MongoServerError | null;
   num_stages: number;
-  editor_view_type: string;
+  editor_view_type: 'text' | 'stage' | 'focus';
   className?: string;
   onChange: (index: number, value: string) => void;
   editorRef?: React.Ref<EditorRef>;
@@ -143,7 +143,7 @@ export const StageEditor = ({
         {
           num_stages: num_stages,
           stage_index: index + 1,
-          stage_action: 'stage_content_changed',
+          stage_action: 'stage_content_changed' as const,
           stage_name: stageOperator,
           editor_view_type: editor_view_type,
         },
--- a/packages/compass-connections/src/stores/connections-store-redux.ts
+++ b/packages/compass-connections/src/stores/connections-store-redux.ts
@@ -1509,6 +1509,7 @@ export const connect = (
       track(
         'Connection Attempt',
         {
+          is_favorite: connectionInfo.savedConnectionType === 'favorite',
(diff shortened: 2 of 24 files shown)


packages/compass-connections/src/stores/connections-store-redux.ts
```diff
@@ -1509,10 +1509,6 @@ export const connect = (
       track(
         'Connection Attempt',
         {
-          is_favorite: connectionInfo.savedConnectionType === 'favorite',
```
mcasimir, line 1512: drive by, these attributes aren't meaningful anymore with multiple connections

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -489,7 +489,7 @@ class CrudStoreImpl
    * to update if the labels change at some point.
    */
   modeForTelemetry() {
-    return this.state.view.toLowerCase();
+    return this.state.view.toLowerCase() as 'list' | 'json' | 'table';
```
mcasimir, line 492: is there a better way?

packages/compass-connections/src/stores/connections-store-redux.ts
lerouxb: Are they, though? We still have the concept of a favourite that we haven't removed it. Can't ever remove it if we don't keep stats on whether it gets used.

himanshusinghs: [Relevant thread](https://mongodb.slack.com/archives/G2L10JAV7/p1724951574301919)

I suggested removing them because the significance of is_favorite is not the same as it was before. Earlier is_favorite was the only way to differentiate wether a connection is saved or recent but now all the connections are saved and is_favorite is more UI related, allowing us to group connections together. So for the telemetry point that it used to represent earlier, it does not make sense to keep it anymore. However if we would like to track connections being marked as favorite maybe we can introduce a different attribute?

mcasimir: ahh ok, so is favorite is still a thing, then i will keep it

mcasimir: and is recent doesn't make sense because everything saved now right?

packages/compass-crud/src/stores/crud-store.ts
mabaasit: `as Lowercase<DocumentView>`?

packages/compass-connections/src/stores/connections-store-redux.ts
lerouxb: I think is_recent can be dropped, yes.

mcasimir: ok re-added is_favorite then! thanks!

packages/compass-aggregations/src/components/stage-editor/stage-editor.tsx
```diff
@@ -143,7 +143,7 @@ export const StageEditor = ({
         {
           num_stages: num_stages,
           stage_index: index + 1,
-          stage_action: 'stage_content_changed',
+          stage_action: 'stage_content_changed' as const,
```
lerouxb, line 146: I'm not understanding what these `as const` bits are doing and why they would be necessary.

mcasimir: as in move `as const` at the end? They are there so `stage_action` is not considered a string

mcasimir: ah wait, you are right this should not be needed here though. I'll take another look and remove them where is redundant

lerouxb (approved):

PR #6196, opened 2024-09-04
fix(indexes): check for vector search support when showing edit templates COMPASS-8235

Anemy:
COMPASS-8235

A bit of a note, this was likely missed/caused by the `isVectorSearchSupported` property being optional on the component. I'm thinking allowing undefined for boolean properties like this is a pattern we should avoid. Probably not worth a lint rule, more of something to look out for.

### DIFF
--- a/packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
+++ b/packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
@@ -14,18 +14,18 @@ import {
 const containerStyles = css({
   display: 'flex',
   flexDirection: 'column',
-  gap: spacing[1],
+  gap: spacing[100],
 });
 
 const dropdownLabelStyles = css({
   display: 'flex',
-  gap: spacing[1],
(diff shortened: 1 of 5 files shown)


packages/compass-indexes/src/components/search-index-template-dropdown/index.tsx
```diff
@@ -14,18 +14,18 @@ import {
...
   display: 'flex',
-  gap: spacing[1],
+  gap: spacing[100],
   alignItems: 'center',
 });
 
 type SearchIndexTemplateDropdownProps = {
   tooltip: string;
-  isVectorSearchSupported?: boolean;
+  isVectorSearchSupported: boolean;
```
lerouxb, line 28: `?: boolean` is actually a pretty common pattern in our code where we make opt-in things optional for convenience. Unless that's really a required field. So it would be hard to come up with a linting rule that works in all cases.

lerouxb (approved):

PR #6220, opened 2024-09-10
chore(ci): build and test packaged app on the newer macos version COMPASS-8090

kraenhansen:
## Description

A friendly takeover of https://github.com/mongodb-js/compass/pull/5990 incorporating @mcasimir's feedback: https://github.com/mongodb-js/compass/pull/5990#discussion_r1662392749

Updates the build variants used for packaging the app on MacOS to 14 and adds two new test variants for the MacOS 14. We might want to simply remove `test-packaged-app-macos-11-arm` and `test-packaged-app-macos-11-x64` in the future.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/.evergreen/buildvariants-and-tasks.yml
+++ b/.evergreen/buildvariants-and-tasks.yml
@@ -62,7 +62,7 @@ buildvariants:
     expansions:
       silk_asset_group: compass-macos
     display_name: Package MacOS Intel
-    run_on: macos-1100
+    run_on: macos-14
     tasks:
       - name: package-compass
       - name: package-compass-isolated
@@ -71,7 +71,7 @@ buildvariants:
     expansions:
       silk_asset_group: compass-macos-arm
     display_name: Package MacOS Arm64
-    run_on: macos-1100-arm64
+    run_on: macos-14-arm64
     tasks:
       - name: package-compass
       - name: package-compass-isolated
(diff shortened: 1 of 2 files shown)


gribnoysup (approved):
Thanks! And seems like the flake I was seeing in my branch gone away, neat!

addaleax (approved):

.evergreen/buildvariants-and-tasks.yml
```diff
@@ -71,7 +71,7 @@ buildvariants:
     expansions:
       silk_asset_group: compass-macos-arm
     display_name: Package MacOS Arm64
-    run_on: macos-1100-arm64
+    run_on: macos-14-arm64
```
addaleax, line 74: Just for context, we've been building on a fairly low macOS version because we've felt that it gives us a better sense of guarantee that Compass will actually run on the macOS versions we claim to support. Currently, in our docs, that's 10.12 – which is definitely wrong at this point, Electron 27 raised the minimum to 10.15, so I'd recommend closing this ticket with "Documentation Changes Needed".

Now, I've chatted with some people in devprod before, but haven't really found anyone who is intimately familiar with what macOS's actual forward/backward compatibility story looks like in practice – i.e. answers to questions like "if we package on macOS 14, can we still relatively safely claim that it will run on 10.15 (or on 11, etc.)? Does it even really matter?".

All of which is to say: I think this change is fine, but in the long run we'll probably want to figure out what the actual story here is.

PR #6234, opened 2024-09-12
chore(telemetry): generate tracking plan from types COMPASS-8058

mcasimir:
Generates a tracking plan in `docs/tracking-plan.md` from `telemetry-events.ts` along the other auto-generated files.


packages/compass-telemetry/src/telemetry-events.ts
```diff
@@ -79,7 +82,7 @@ export type ConnectionScopedProperties = {
   /**
    * The id of the connection associated to this event.
    */
-  connection_id: string;
+  connection_id?: string;
```
mcasimir, line 85: looks like `connection_id` is not guaranteed for all the connection scoped events

scripts/generate-tracking-plan.ts
```diff
@@ -0,0 +1,326 @@
...
+  const program = ts.createProgram(['inMemoryFile.ts'], compilerOptions, host);
+  const checker = program.getTypeChecker();
+  return { sourceFile, checker };
+}
+
+function generateMarkdownPlan(
+  events: TelemetryEventInfo[],
+  identifyTraits: TelemetryEventInfo
+) {
+  const byCategory = _.groupBy(events, 'category');
```
Anemy, line 266: Totaly nit, since we're only using lodash for this can we do it in regular ts and avoid the dependency?

Anemy (approved):

scripts/generate-tracking-plan.ts
mcasimir: True, removed and also sorted everything

packages/compass-telemetry/src/telemetry-events.ts
lerouxb: Any idea which ones? Sounds dodgy.

mcasimir: Screen is one that is hard to fix, and then we found some others when connecting from CLI without an existing id

lerouxb (approved):

PR #6264, opened 2024-09-19
feat(compass-indexes): poll regular indexes COMPASS-8214

lerouxb:
Still a draft because this needs some more tests.

### DIFF
--- a/packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/search/text-search.tsx
+++ b/packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/search/text-search.tsx
@@ -17,7 +17,7 @@ import type { RootState } from '../../../../modules';
 import type { WizardComponentProps } from '..';
 import { FieldCombobox } from '../field-combobox';
 import {
-  type SearchIndexesStatus,
+  type FetchStatus,
   fetchIndexes,
 } from '../../../../modules/search-indexes';
 
@@ -100,7 +100,7 @@ export const TextSearch = ({
   onFetchIndexes,
 }: WizardComponentProps & {
   indexes: SearchIndex[];
-  indexesStatus: SearchIndexesStatus;
+  indexesStatus: FetchStatus;
   onFetchIndexes: () => void;
 }) => {
   const [formData, setFormData] = useState<TextSearchState>({
--- a/packages/compass-aggregations/src/modules/search-indexes.ts
+++ b/packages/compass-aggregations/src/modules/search-indexes.ts
@@ -3,14 +3,14 @@ import type { PipelineBuilderThunkAction } from '.';
 import type { SearchIndex } from 'mongodb-data-service';
 import { isAction } from '../utils/is-action';
 
-enum SearchIndexesStatuses {
+enum FetchStatuses {
   INITIAL = 'INITIAL',
   LOADING = 'LOADING',
   READY = 'READY',
   ERROR = 'ERROR',
 }
 
-export type SearchIndexesStatus = keyof typeof SearchIndexesStatuses;
+export type FetchStatus = keyof typeof FetchStatuses;
 
 export enum ActionTypes {
   FetchIndexesStarted = 'compass-aggregations/search-indexes/FetchIndexesStarted',
@@ -39,13 +39,13 @@ export type SearchIndexesAction =
 type State = {
   isSearchIndexesSupported: boolean;
   indexes: SearchIndex[];
-  status: SearchIndexesStatus;
+  status: FetchStatus;
 };
 
 export const INITIAL_STATE: State = {
   isSearchIndexesSupported: false,
   indexes: [],
-  status: SearchIndexesStatuses.INITIAL,
+  status: FetchStatuses.INITIAL,
 };
 
 const reducer: Reducer<State, Action> = (state = INITIAL_STATE, action) => {
@@ -54,7 +54,7 @@ const reducer: Reducer<State, Action> = (state = INITIAL_STATE, action) => {
   ) {
     return {
       ...state,
-      status: SearchIndexesStatuses.LOADING,
+      status: FetchStatuses.LOADING,
     };
   }
   if (
@@ -66,7 +66,7 @@ const reducer: Reducer<State, Action> = (state = INITIAL_STATE, action) => {
     return {
       ...state,
       indexes: action.indexes,
-      status: SearchIndexesStatuses.READY,
+      status: FetchStatuses.READY,
     };
   }
   if (
@@ -74,7 +74,7 @@ const reducer: Reducer<State, Action> = (state = INITIAL_STATE, action) => {
   ) {
     return {
       ...state,
-      status: SearchIndexesStatuses.ERROR,
+      status: FetchStatuses.ERROR,
     };
   }
   return state;
@@ -90,8 +90,8 @@ export const fetchIndexes = (): PipelineBuilderThunkAction<Promise<void>> => {
 
     if (
       !dataService ||
-      status === SearchIndexesStatuses.LOADING ||
-      status === SearchIndexesStatuses.READY
+      status === FetchStatuses.LOADING ||
+      status === FetchStatuses.READY
     ) {
       return;
     }
--- a/packages/compass-indexes/src/modules/regular-indexes.ts
+++ b/packages/compass-indexes/src/modules/regular-indexes.ts
@@ -1,12 +1,14 @@
+import { cloneDeep } from 'lodash';
 import type { IndexDefinition } from 'mongodb-data-service';
 import type { AnyAction } from 'redux';
 import {
   openToast,
   showConfirmation as showConfirmationModal,
 } from '@mongodb-js/compass-components';
-import { cloneDeep } from 'lodash';
 
-import { isAction } from './../utils/is-action';
+import { FetchStatuses, NOT_FETCHABLE_STATUSES } from '../utils/fetch-status';
+import type { FetchStatus, FetchingStatus } from '../utils/fetch-status';
+import { isAction } from '../utils/is-action';
 import { ActionTypes as CreateIndexActionTypes } from './create-index';
 import type {
   CreateIndexSpec,
@@ -42,6 +44,9 @@ export type InProgressIndex = {
 };
 
 export enum ActionTypes {
+  IndexesOpened = 'compass-indexes/regular-indexes/indexes-opened',
+  IndexesClosed = 'compass-indexes/regular-indexes/indexes-closed',
+
   FetchIndexesStarted = 'compass-indexes/regular-indexes/fetch-indexes-started',
   FetchIndexesSucceeded = 'compass-indexes/regular-indexes/fetch-indexes-succeeded',
   FetchIndexesFailed = 'compass-indexes/regular-indexes/fetch-indexes-failed',
@@ -54,7 +59,7 @@ export enum ActionTypes {
 
 type FetchIndexesStartedAction = {
   type: ActionTypes.FetchIndexesStarted;
-  isRefreshing: boolean;
+  status: FetchingStatus;
 };
 
 type FetchIndexesSucceededAction = {
@@ -74,16 +79,16 @@ type FailedIndexRemovedAction = {
 
 export type State = {
   indexes: RegularIndex[];
-  isRefreshing: boolean;
+  status: FetchStatus;
   inProgressIndexes: InProgressIndex[];
-  error: string | null;
+  error?: string;
 };
 
 export const INITIAL_STATE: State = {
+  status: FetchStatuses.NOT_READY,
   indexes: [],
   inProgressIndexes: [],
-  isRefreshing: false,
-  error: null,
+  error: undefined,
 };
 
 export default function reducer(
@@ -95,8 +100,7 @@ export default function reducer(
   ) {
     return {
       ...state,
-      error: null,
-      isRefreshing: action.isRefreshing,
+      status: action.status,
     };
   }
 
@@ -116,7 +120,7 @@ export default function reducer(
     return {
       ...state,
       indexes: allIndexes,
-      isRefreshing: false,
+      status: FetchStatuses.READY,
     };
   }
 
@@ -125,8 +129,15 @@ export default function reducer(
   ) {
     return {
       ...state,
-      error: action.error,
-      isRefreshing: false,
+      // We do no set any error on poll or refresh and the
+      // previous list of indexes is shown to the user.
+      // If fetch fails for refresh or polling, set the status to READY again.
+      error:
+        state.status === FetchStatuses.FETCHING ? action.error : state.error,
+      status:
+        state.status === FetchStatuses.FETCHING
+          ? FetchStatuses.ERROR
+          : FetchStatuses.READY,
     };
   }
 
@@ -210,10 +221,10 @@ export default function reducer(
 }
 
 const fetchIndexesStarted = (
-  isRefreshing: boolean
+  status: FetchingStatus
 ): FetchIndexesStartedAction => ({
   type: ActionTypes.FetchIndexesStarted,
-  isRefreshing,
+  status,
 });
 
 const fetchIndexesSucceeded = (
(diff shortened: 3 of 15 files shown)


packages/compass-indexes/src/modules/regular-indexes.ts
```diff
@@ -73,30 +86,45 @@ type FailedIndexRemovedAction = {
 };
 
 export type State = {
+  isVisible: boolean;
   indexes: RegularIndex[];
-  isRefreshing: boolean;
+  status: FetchStatus;
   inProgressIndexes: InProgressIndex[];
-  error: string | null;
+  error?: string;
```
lerouxb, line 93: This was string or null here and string or undefined in search indexes. I just picked one and made them the same.

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -375,12 +367,11 @@ export default function reducer(
...
       error:
-        state.status === SearchIndexesStatuses.FETCHING
-          ? action.error
-          : undefined,
-      status: SearchIndexesStatuses.FETCHING
-        ? SearchIndexesStatuses.ERROR
-        : SearchIndexesStatuses.READY,
+        state.status === FetchStatuses.FETCHING ? action.error : state.error,
+      status:
+        state.status === FetchStatuses.FETCHING
```
lerouxb, line 372: This was actually a nasty bug I introduced in the refactor. FetchStatuses.FETCHING is always true 😄

packages/compass-components/src/components/item-action-controls.tsx
```diff
@@ -567,7 +569,9 @@ export function DropdownMenuButton<Action extends string>({
             rightGlyph={<Icon glyph={'CaretDown'} />}
             title={buttonText}
           >
-            <span className={hiddenOnNarrowStyles}>{buttonText}</span>
+            <span className={hideOnNarrow ? hiddenOnNarrowStyles : undefined}>
```
lerouxb, line 572: This is a drive-by. Recently I made the dropdown button collapse when the view gets narrow, but for the create index button you get when search indexes is available that's a bit much. So I added a way to opt out of it.

packages/compass-indexes/src/modules/index-view.ts
```diff
@@ -18,6 +24,27 @@ export default function reducer(
   state = INITIAL_STATE,
   action: AnyAction
 ): IndexView {
+  // The create index button has a dropdown where you can select regular or
```
lerouxb, line 27: This is another drive-by. This is functionality we had before I refactored it, then accidentally dropped during the recent refactor.

packages/compass-indexes/src/modules/search-indexes.spec.ts
```diff
@@ -269,6 +272,96 @@ describe('search-indexes module', function () {
...
+          This looks like a pretty tight loop, but the await at least gives the
+          stub time to resolve. If we wait using a timeout it will never fire
+          because this test takes control of the clock. In fact this while
+          condition is never false when the test passes so it never enters the
+          loop because the await on this function itself gives the stub time to
+          execute.
+          */
+          await Promise.resolve();
+        }
+      };
```
gribnoysup, line 312: Don't have a strong preference for or against writing a dedicated helper, but we usually just use `waitFor` from the testing-library for stuff like this instead of writing our own

packages/compass-indexes/src/modules/regular-indexes.ts
```diff
@@ -257,13 +301,75 @@ export const fetchIndexes = (
...
+> => {
+  return async (dispatch, getState) => {
+    const { status } = getState().regularIndexes;
+
+    // If we are in a READY state, then we have already fetched the data
+    // and are refreshing the list.
+    const newStatus: FetchStatus =
+      status === FetchStatuses.READY
+        ? FetchStatuses.REFRESHING
+        : FetchStatuses.FETCHING;
```
gribnoysup, line 316: Kinda feels like it mixes responsibilites of the action too much: the "state machine" job that's usually should be the responsibility of the reducer is done by the action now, additionally from the reducer perspective if would make one action to look like a different one: like imagine you're debugging an issue with the indexes feching and inspecing the log of actions passing through the store, you now can't track the action to the exact location it was dispatched from because refresh action creator will also emit a fetch event that is usually coming from a different place in the code

packages/compass-aggregations/src/components/aggregation-side-panel/stage-wizard-use-cases/search/text-search.tsx
```diff
@@ -17,7 +17,7 @@ import type { RootState } from '../../../../modules';
 import type { WizardComponentProps } from '..';
 import { FieldCombobox } from '../field-combobox';
 import {
-  type SearchIndexesStatus,
+  type FetchStatus,
```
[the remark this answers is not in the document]
lerouxb: I just realised I accidentally changed the aggregations module with find/replace. I'll just undo the changes in this module.

packages/compass-aggregations/src/modules/search-indexes.ts
```diff
@@ -3,14 +3,14 @@ import type { PipelineBuilderThunkAction } from '.';
 import type { SearchIndex } from 'mongodb-data-service';
 import { isAction } from '../utils/is-action';
 
-enum SearchIndexesStatuses {
+enum FetchStatuses {
```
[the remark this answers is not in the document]
lerouxb: Same as above. I'll just undo this.

packages/compass-indexes/src/modules/regular-indexes.ts
```diff
@@ -73,30 +86,45 @@ type FailedIndexRemovedAction = {
 };
 
 export type State = {
+  isVisible: boolean;
```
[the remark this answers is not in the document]
lerouxb: Hmm I suppose. I'll clean it up.

packages/compass-indexes/src/modules/regular-indexes.ts
lerouxb: I kinda just copied that from search indexes to make it work the same and didn't think too much of it. Should I just move this into fetchIndexes() or are you suggesting a bigger refactor?

packages/compass-indexes/src/modules/search-indexes.spec.ts
lerouxb: I'll see if that works. If it does any kind of timeout or interval internally it will hang because of the comment above.

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -457,6 +448,54 @@ const updateSearchIndexSucceeded = (): UpdateSearchIndexSucceededAction => ({
...
+    });
+
+    pollInterval = setInterval(() => {
+      void dispatch(pollSearchIndexes());
+    }, POLLING_INTERVAL);
+
+    // if this is the first time, also refresh the list
+    if (getState().searchIndexes.status === 'NOT_READY') {
+      await dispatch(refreshSearchIndexes());
+    }
```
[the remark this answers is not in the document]
lerouxb: I can move it. For some reason I just had it in my mind that I shouldn't fire off actions in there.

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -607,8 +646,7 @@ const fetchIndexes = (
       return;
```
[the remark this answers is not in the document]
lerouxb: you mean just remove NOT_AVAILABLE and get that from elsewhere?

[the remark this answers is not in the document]
gribnoysup: Either removing NOT_AVAILABLE for search indexes and just storing and syncing this piece of data with other ampersand model props, or using NOT_AVAILABLE fetching status as a proxy for those props both for search and regular indexes

packages/compass-indexes/src/modules/search-indexes.ts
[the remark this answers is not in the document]
gribnoysup: If they are state setters we should avoid in favor of just using initial state, but otherwise activate is part of the lifecycle similar to react lifecycles, so similarly actions can generally happen in those lifecycles

packages/compass-indexes/src/modules/search-indexes.spec.ts
```diff
@@ -269,6 +273,88 @@ describe('search-indexes module', function () {
...
+          isSearchIndexesSupported: true,
+        },
+        {
+          getSearchIndexes: getSearchIndexesStub,
+        }
+      );
+
+      const waitForStatus = async (status: FetchStatus) => {
+        await waitFor(() => {
+          return store.getState().regularIndexes.status === status;
```
gribnoysup, line 303: `waitFor` resolves when the function inside stops throwing errors, so it should be something like this instead

```suggestion
          expect(store.getState().regularIndexes.status).to.eq(status)
```

packages/compass-indexes/src/modules/regular-indexes.ts
gribnoysup: Moving it to fetch indexes would kinda have the same issue: action doing the work of the reducer, it is better in terms of at least it's sort of just one place now, but you're still loosing information about what actually happened in the application. The easiest way to align with style guide here would be to pass what type of action triggered the fetch and then reducer to make a decision about the resulting state based on that: so like the action type would have something like `trigger: 'poll' | 'refresh' | 'initial-fetch'` instead of a new status as prop. That way we're making it clear what was the trigger and remove the "avoid using actions as setters" issue

packages/compass-indexes/src/modules/search-indexes.spec.ts
lerouxb: Oof.  And it didn't fail because any `await` causes next tick :)

packages/compass-indexes/src/modules/search-indexes.ts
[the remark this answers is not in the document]
gribnoysup: (but also yeah, I'm not sure if we want to untangle this too, we'll have to stop rewriting this at some point)

packages/compass-aggregations/src/modules/search-indexes.ts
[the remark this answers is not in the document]
gribnoysup: Thanks! And yeah, totally missed that this was a different plugin even, sorry!

packages/compass-indexes/src/stores/store.ts
```diff
@@ -80,8 +80,8 @@ export function activateIndexesPlugin(
       searchIndexes: {
         ...SEARCH_INDEXES_INITIAL_STATE,
         status: options.isSearchIndexesSupported
-          ? SearchIndexesStatuses.NOT_READY
-          : SearchIndexesStatuses.NOT_AVAILABLE,
+          ? FetchStatuses.NOT_READY
+          : FetchStatuses.NOT_AVAILABLE,
```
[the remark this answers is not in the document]
lerouxb: Since we're getting rid of NOT_AVAILABLE, I guess this would just be NOT_READY whenever there aren't indexes yet for any reason.

packages/compass-indexes/src/modules/search-indexes.ts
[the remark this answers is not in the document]
lerouxb: I elected to remove NOT_AVAILABLE. I just added it to the root like isReadonlyView and isWritable as isSearchIndexesSupported.

packages/compass-indexes/src/modules/regular-indexes.ts
lerouxb: I got the concept of a "fetch reason" now which then gets mapped to status in the reducer.

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -457,6 +448,54 @@ const updateSearchIndexSucceeded = (): UpdateSearchIndexSucceededAction => ({
   type: ActionTypes.UpdateSearchIndexSucceeded,
 });
 
+export const POLLING_INTERVAL = 5000;
+
+let pollInterval: ReturnType<typeof setInterval> | undefined = undefined;
+
+export const openSearchIndexes = (): IndexesThunkAction<
```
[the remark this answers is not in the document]
lerouxb: Yeah I changed to startPollingX and stopPollingX.

packages/compass-indexes/src/modules/regular-indexes.ts
```diff
@@ -257,13 +304,59 @@ export const fetchIndexes = (
...
 > => {
   return async (dispatch) => {
-    return dispatch(fetchIndexes(true));
+    return await dispatch(fetchIndexes(FetchReasons.POLL));
+  };
+};
+
+export const POLLING_INTERVAL = 5000;
+
+let pollInterval: ReturnType<typeof setInterval> | undefined = undefined;
```
gribnoysup, line 335: Hmmm, I know that we're technically stop polling if the tab goes off screen, but I wonder if it would maybe make sense to already take the tab id into account somehow here. We have this clunky way of passing tab id around as a prop for the components, potentially we'd have to change that too if we'd decide to change that. Just thinking that if we ever allow to splitting tabs in the same window, this will cause an issue that's hard to spot

packages/compass-indexes/src/utils/fetch-reason.ts
```diff
@@ -0,0 +1,16 @@
+export enum FetchReasons {
+  /**
+   *
```
gribnoysup, line 3: I think those are pretty self explanatory, but highlighting just in case you wanted to add comments but forgot 🙂

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -632,11 +629,11 @@ export const refreshSearchIndexes = (): IndexesThunkAction<
...
     // and are refreshing the list.
-    const newStatus: SearchIndexesStatus =
-      status === SearchIndexesStatuses.READY
-        ? SearchIndexesStatuses.REFRESHING
-        : SearchIndexesStatuses.FETCHING;
-    await dispatch(fetchIndexes(newStatus));
+    const reason: FetchReason =
+      status === FetchStatuses.READY
+        ? FetchReasons.REFRESH
+        : FetchReasons.INITIAL_FETCH;
```
gribnoysup, line 635: I think one thing that I was missing so far is that we're using the same action for user triggering through the refresh button, initial plugin activation, and re-fetching the indexes after we created one through the modal. I was hoping that introducing this `reason` to the action we would be able to avoid deriving the state in the action from the previous state, that is generally not recommended, but seems like specifically because this particular action is used for this initial fetching triggered from the activate, we still have to keep this around.

Using this for both user triggered and programmatic refreshes kinda makes sense to me, but do you think we can remove the initial fetch from the list of things that this action creator is used for and either have a dedicated action creator for that, or just use `fetchIndexes` with INITIAL_FETCH passed there as a reason directly in the `activate` method? I'd really want us to remove the deriving state from a previous one like that in these actions

packages/compass-indexes/src/modules/search-indexes.ts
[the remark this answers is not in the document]
gribnoysup: Neat! This seems to fit the "no derived state in store" rule nicely, especially because then we're remapping it back to `isSupported` in the view

packages/compass-indexes/src/modules/search-indexes.ts
```diff
@@ -83,7 +47,7 @@ export enum ActionTypes {
 
 type FetchSearchIndexesStartedAction = {
   type: ActionTypes.FetchSearchIndexesStarted;
-  status: 'REFRESHING' | 'POLLING' | 'FETCHING';
+  reason: FetchReason;
```
gribnoysup, line 50: Nice!

gribnoysup (approved):
Looks pretty good! I'd still maybe try to remove the whole switching the reason dynamically part in the refresh action creator (see the comment there), but even without this it's very neat!

packages/compass-indexes/src/utils/fetch-reason.ts
lerouxb: Forgot :)

packages/compass-indexes/src/modules/search-indexes.ts
lerouxb: I made separate functions for the initial fetch, refresh and poll.

packages/compass-indexes/src/modules/regular-indexes.ts
lerouxb: Like `const pollIntervalByTabId = {};` ?

gribnoysup: We usually stick with `Map` for similar cases in other stores, but just an object like that would also work, yeah 👍

lerouxb: Aah only saw this comment now. Changed to a Map.

PR #6280, opened 2024-09-25
chore: add missing `prettier` configs

kraenhansen:
## Description

A couple of packages that didn't have a .prettierrc.json config:
- ./packages/bson-transpilers
- ./packages/hadron-build
- ./packages/hadron-type-checker
- ./packages/mongodb-explain-compat

This adds the missing configs and changes to source-files reflecting that.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/bson-transpilers/lib/symbol-table/shelltocsharp.js
addaleax, line 1: These are generated files, so I imagine we should add them to the prettierignore instead?

packages/compass/.prettierignore
```diff
@@ -3,3 +3,6 @@ dist
 coverage
 build
 src/deps
+
+# Generated
+THIRD-PARTY-NOTICES.md
```
kraenhansen, line 8: This got copied on my system while building the app and kept the `check` from passing locally.

packages/compass-e2e-tests/.prettierignore
```diff
@@ -3,3 +3,7 @@
 fixtures
 .nyc_output
 coverage
+
+# Generated files
+electron-versions.json
+write-electron-versions.js
```
kraenhansen, line 9: These got created on my system while building the app and kept the `check` from passing locally.

Anemy (approved):
lgtm, changes look good, makes me happy we're aligning more things. I am a bit uncertain on some of the yaml changes (there are a lot of changes). I wonder if we want those in the prettier ignore also? Fine as is.

PR #6289, opened 2024-09-26
feat(global-writes): add unsharded state COMPASS-8276

mabaasit:
This PR includes:
1. Warning icon on Global Writes tab (along with a tooltip);
2. Unsharded state which allows users to create a shard key for a collection;
3. Initial sharding state, which is shown once server starts to create a shard key

<details>
<summary>Preview</summary>

Tab Warning
image

Unsharded State
image

Sharding State (more work in next ticket)
image

</details>

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-global-writes/src/store/reducer.ts
```diff
@@ -1,26 +1,251 @@
...
+export type ShardingStatus = keyof typeof ShardingStatuses;
+
 export type RootState = {
   namespace: string;
   isNamespaceSharded: boolean;
-  status: keyof typeof ShardingStatuses;
+  status: ShardingStatus;
+  createShardkey: {
+    isLoading: boolean;
+  };
```
mabaasit, line 71: This was not part of the tech doc, but while working on this, I realzied we need to disable the create-shard-button and added this. as an alternative, we can also await on the dispatch action on the react side and while its waiting we disable the button.

packages/compass-global-writes/src/store/reducer.ts
```diff
@@ -1,26 +1,251 @@
...
+        });
+
+      if (!isNamespaceManaged) {
+        dispatch({
+          type: GlobalWritesActionTypes.SetIsManagedNamespace,
+          isNamespaceManaged,
+        });
+        return;
+      }
+      // TODO (COMPASS-8277): Now fetch the sharding key and possible process error.
```
mabaasit, line 177: Here in the follow up tickets, we want to fetch the shard keys and also the agent process errors and based on that information, transition to other states. Currently this leaves user in a loading state, but i think that's fine with the current state of the project.

packages/compass-global-writes/src/store/reducer.ts
```diff
@@ -1,26 +1,251 @@
...
+      );
+      openToast('global-writes-fetch-shard-info-error', {
+        title: `Failed to fetch sharding information: ${
+          (error as Error).message
+        }`,
+        dismissible: true,
+        timeout: 5000,
+        variant: 'important',
+      });
+    }
```
mabaasit, line 193: we also want to have another state like `ERROR` or something, where user does not see loading and can see the error and possibly action to refetch.

packages/compass-global-writes/src/services/atlas-global-writes-service.ts
```diff
@@ -1,5 +1,107 @@
+import toNS from 'mongodb-ns';
 import type { AtlasService } from '@mongodb-js/atlas-service/provider';
+import type { CreateShardKeyData } from '../store/reducer';
+
+type ZoneMapping = unknown;
```
mabaasit, line 5: leaving it unknow for now, will implement it when needed

packages/compass-global-writes/src/components/index.tsx
```diff
@@ -1,34 +1,66 @@
...
+type GlobalWritesProps = {
+  shardingStatus: ShardingStatus;
+};
+
+function getStateViewBasedOnShardingStatus(shardingStatus: ShardingStatus) {
+  switch (shardingStatus) {
+    case ShardingStatuses.NOT_READY:
+      return (
+        <div className={centeredContent}>
+          <SpinLoaderWithLabel progressText="Loading ..." />
```
addaleax, line 43:
```suggestion
          <SpinLoaderWithLabel progressText="Loading …" />
```

packages/compass-global-writes/src/components/states/sharding.tsx
```diff
@@ -0,0 +1,42 @@
...
+  display: 'flex',
+  flexDirection: 'column',
+  gap: spacing[400],
+});
+
+export function ShardingState() {
+  return (
+    <div className={containerStyles}>
+      <Banner variant={BannerVariant.Info}>
+        <strong>Sharding your collection ...</strong>
```
addaleax, line 24:
```suggestion
        <strong>Sharding your collection …</strong>
```

(or I guess you can &hellip; like the &apos; below – but just fyi, &apos; is just a literal `'` that you don't really need to escape here anyway. And even more pedantically, you'd want to use `‘ ’` instead of `'` 😄)

packages/compass-global-writes/src/store/reducer.ts
```diff
@@ -1,26 +1,251 @@
...
+      dispatch({
+        type: GlobalWritesActionTypes.ShardingInProgressFinished,
+      });
+    } catch (error) {
+      logger.log.error(
+        logger.mongoLogId(1_001_000_331),
+        'AtlasFetchError',
+        'Error creating cluster shard key',
+        {
+          error: error as Error,
```
addaleax, line 235: I don't think we can log literal `Error` objects?

mabaasit: removed to log the message only

packages/compass-global-writes/src/store/reducer.ts
gribnoysup: Can you clarify why is this not just part of sharding status? I'm maybe missing the purpose, but this seems like just part of the status transition for creation flow, so why are we moving this state into its own field?

mabaasit: this is a state between `UNSHARDED` and `SHARDING`, time duration for which user clicks the create shard key and server acknowledges the change - keeps the button disabled, and for error shows the toast.

packages/compass-global-writes/src/store/reducer.ts
```diff
@@ -1,26 +1,251 @@
...
+  ManagedNamespace,
+  | 'customShardKey'
+  | 'isCustomShardKeyHashed'
+  | 'isShardKeyUnique'
+  | 'numInitialChunks'
+  | 'presplitHashedZones'
+>;
+
+enum GlobalWritesActionTypes {
+  SetIsManagedNamespace = 'global-writes/SetIsManagedNamespace',
```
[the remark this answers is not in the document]
mabaasit: fixed in [a0eb2be](https://github.com/mongodb-js/compass/pull/6289/commits/a0eb2be8bf1eb53c218238c39110dc3457c3fa2c). 

this action is dispatched after we call the api to check if the namespace is managed or not, in `fetchClusterShardingData` action.

packages/compass-global-writes/src/components/index.tsx
```diff
@@ -1,34 +1,66 @@
...
+        </div>
+      );
+    case ShardingStatuses.UNSHARDED:
+      return <UnshardedState />;
+    case ShardingStatuses.SHARDING:
+      return <ShardingState />;
+    default:
+      return null;
+  }
+}
```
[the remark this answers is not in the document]
mabaasit: fixed in [5f03a49](https://github.com/mongodb-js/compass/pull/6289/commits/5f03a497acf04177ae48a213796c80f6553affc2)

packages/compass-global-writes/src/store/reducer.ts
gribnoysup: So this loading state is mutually exclusive with all the other states, is that correct? If yes, in that case we should just add it as another state in list of available state transitions for the component, using multiple states to declare mutually exclusive, state-machine-like, states of the application can lead to potentially weird cases and hard to reason about logic as it basically allows application to be in multiple states at once.

mabaasit: Yes that's correct. It does make sense to add these to the state as well, its just that I think the list will grow a little and in such case, i find it a bit hard to maintain the transitions.

But I will try to do that. Thanks for the suggestion.

mabaasit: added in [23c1361](https://github.com/mongodb-js/compass/pull/6289/commits/23c136137c70829d9a7ca778edd358114de24a75)

packages/compass-global-writes/src/plugin-title.tsx
```diff
@@ -1,5 +1,78 @@
...
+          </Body>
+        </Tooltip>
+      )}
+    </div>
+  );
+};
+
+export const GlobalWritesTabTitle = connect(
+  ({ isNamespaceSharded, status }: RootState) => ({
+    showWarning: !isNamespaceSharded && status !== ShardingStatuses.NOT_READY,
```
paula-stacho, line 76: why don't we check for `ShardingStatuses.UNSHARDED` directly here? and how come this doesn't show the warning by `ShardingStatuses.SUBMITTING_FOR_SHARDING` - does that fall under `isNamespaceSharded`? 🤔

packages/compass-global-writes/src/index.ts
```diff
@@ -28,6 +28,6 @@ const CompassGlobalWritesHadronPlugin = registerHadronPlugin(
 export const CompassGlobalWritesPlugin = {
   name: 'GlobalWrites' as const,
   provider: CompassGlobalWritesHadronPlugin,
-  content: GlobalWrites,
-  header: GlobalWritesTabTitle,
+  content: GlobalWrites as React.FunctionComponent,
+  header: GlobalWritesTabTitle as React.FunctionComponent,
```
paula-stacho, line 32: do we need these assertions?

packages/compass-global-writes/src/components/index.spec.tsx
```diff
@@ -1,12 +1,22 @@
...
-      .exist;
+  it('renders plugin in NOT_READY state', function () {
+    renderWithStore(<GlobalWrites shardingStatus={'NOT_READY'} />);
+    expect(screen.getByText(/loading/i)).to.exist;
+  });
+
+  it('renders plugin in UNSHARDED state', function () {
+    renderWithStore(<GlobalWrites shardingStatus={'UNSHARDED'} />);
+    expect(screen.getByTestId('shard-collection-button')).to.exist;
+  });
```
paula-stacho, line 16: can we add SUBMITTING_FOR_SHARDING?

packages/compass-global-writes/src/plugin-title.tsx
mabaasit: actually, it does make sense to use `ShardingStatuses.UNSHARDED` here which means for any state change we have to ensure we update the `isNamespaceSharded`. it should and it does show the warning its in  `SUBMITTING_FOR_SHARDING` state (the state when Submit For sharding button is disabled). Once the server accepts the request, we transition to `SHARDING` state, where this warning is not shown.

mabaasit: ~actually, it does make sense to use `ShardingStatuses.UNSHARDED` here which means for any state change we have to ensure we update the `isNamespaceSharded`.~ it should and it does show the warning its in  `SUBMITTING_FOR_SHARDING` state (the state when Submit For sharding button is disabled). Once the server accepts the request, we transition to `SHARDING` state, where this warning is not shown.

packages/compass-global-writes/src/index.ts
mabaasit: hm, good point. so the current setup is such that all the plugins registered with `registerHadronPlugin` get same props defined by `CollectionTabPluginMetadata`. For this plugin, as i am not adding those and when connecting with redux, the [ts check fails with this error](https://spruce.mongodb.com/task/10gen_compass_main_check_check_patch_df47e20728978e1544dccaef9251273df0081782_66f55f9463a74c0007777c7b_24_09_26_13_20_21/logs?execution=0).

packages/compass-global-writes/src/components/index.spec.tsx
mabaasit: added in [59787bb](https://github.com/mongodb-js/compass/pull/6289/commits/59787bb7983cd9b6b8a72c7bf8a010b88824d66d)

paula-stacho:
Perhaps we could add some left padding, so that the LG hover isn't cut off:
Screenshot 2024-09-27 at 13 41 04
I'm thinking either 16px on the checkboxes (so that they are aligned with the text, skipping the toggle icon), or 4px on the form and description. What do you think?

packages/compass-global-writes/src/store/reducer.ts
```diff
@@ -19,8 +80,165 @@ const initialState: RootState = {
...
+
+  if (
+    isAction<SubmittingForShardingErroredAction>(
+      action,
+      GlobalWritesActionTypes.SubmittingForShardingErrored
+    )
+  ) {
+    return {
+      ...state,
+      status: ShardingStatuses.UNSHARDED,
```
paula-stacho, line 132: do we have another ticket for the error handling?

mabaasit:
> Perhaps we could add some left padding, so that the LG hover visual isn't cut off: Screenshot 2024-09-27 at 13 41 04 I'm thinking either 16px on the checkboxes (so that they are aligned with the text, skipping the toggle icon), or 4px on the form and description. What do you think?

I add some spacing to avoid this in [eca4068](https://github.com/mongodb-js/compass/pull/6289/commits/eca4068d73e71e966d3075bd8953ea49f2ad607e) (but only for the radio options and aligned checkbox accordingly)

paula-stacho (approved):
Looks good!

PR #6294, opened 2024-09-27
fix: remove compass-preferences from connection-form completely COMPASS-8098

paula-stacho:
## Description

Removing last reference to compass-preferences from connection-form, because it's causing some issues in the VSCode integration.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/connection-form/src/hooks/use-connection-color.ts
```diff
@@ -188,22 +171,10 @@ export function useConnectionColor(): {
...
-      return CONNECTION_COLOR_CODES;
-    }
-  };
+  const connectionColorCodes = () => CONNECTION_COLOR_CODES.slice(0, 9);
 
   return {
-    connectionColorToHex: isMultiConnectionEnabled
-      ? newColorCodeToHex
-      : colorCodeToHex,
+    connectionColorToHex: newColorCodeToHex,
```
Anemy, line 177: [nit] Does it make sense to drop the `new` now?

packages/connection-form/src/hooks/use-connection-color.ts
```diff
@@ -161,22 +160,6 @@ export function useConnectionColor(): {
     [isDarkMode]
   );
 
-  const colorCodeToHex = useCallback(
-    (colorCode: string | undefined): string | undefined => {
-      if (!colorCode) {
-        return;
-      }
-
-      const migratedColor = legacyColorsToColorCode(colorCode);
```
Anemy, line 170: We only have one other use of `legacyColorsToColorCode`, and it's in the color picker. Maybe we can remove `legacyColorsToColorCode` and `LEGACY_COLORS_TO_COLOR_CODE_MAP` now? Someone would be using a color from ~3 years ago, and we'd only slightly be helping them select a color inside of the color picker UI.

Anemy (approved):
lgtm! One suggestion on some old code we can remove, not a blocker, feel free to merge as is

packages/connection-form/src/hooks/use-connection-color.ts
paula-stacho: that must be part of the legacy bits. I'll try to remove it + the parts that use it, hopefully it's not too deep of a rabbit hole 🤞

PR #6304, opened 2024-09-30
feat: add sharded state COMPASS-8279

paula-stacho:
## Description

- show sharded state to the user
- shows zone table (without search & collapse - follow up in https://jira.mongodb.org/browse/COMPASS-8336)
- it lets user to unmanage the namespace
- also sets states like `SHARD_KEY_INVALID`, `SHARD_KEY_MISMATCH` or `SHARDING_ERROR`, but doesn't add the UI for them (will be added in https://jira.mongodb.org/browse/COMPASS-8278 and https://jira.mongodb.org/browse/COMPASS-8349)

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/atlas-service/src/make-automation-agent-op-request.ts
gribnoysup, line 1: Just FYI, you'd need to adjust this after https://github.com/mongodb-js/compass/pull/6305 lands, all this code is still moving around and we made a decision to not keep domain logic inside AtlasService

paula-stacho: thank you!

packages/compass-global-writes/src/components/states/shard-key-correct.tsx
```diff
@@ -0,0 +1,200 @@
...
+    return shardKey.fields[1].name;
+  }, [shardKey]);
+
+  const { atlasMetadata } = useConnectionInfo();
+
+  const sampleCodes = useMemo(() => {
+    const { collection, database } = toNS(namespace);
+    return {
+      findingDocuments: `use ${database}\ndb.${collection}.find({"location": "US-NY", "${customShardKeyField}": "<id_value>"})`,
+      insertingDocuments: `use ${database}\ndb.${collection}.insertOne({"location": "US-NY", "${customShardKeyField}": "<id_value>",...<other fields>})`,
```
gribnoysup, line 73: You can have symbols in the collection name that make dot notation property access break, we usually use the `["<escaped coll name>"]` instead because it's more robust

packages/compass-global-writes/src/components/states/shard-key-correct.tsx
```diff
@@ -0,0 +1,200 @@
...
+        </Body>
+        <Body>
+          Replace the text to perform operations on different documents. US-NY
+          is an ISO 3166 location code referring to New York, United States. You
+          can look up other ISO 3166 location codes below.
+        </Body>
+      </div>
+
+      <div className={codeBlockContainerStyles}>
+        <Label htmlFor="finding-documents">Finding documents</Label>
```
gribnoysup, line 118: There is no corresponding id attr anywhere in this file, what is this pointing to?

packages/compass-global-writes/src/components/states/shard-key-correct.tsx
```diff
@@ -0,0 +1,200 @@
...
+          Each document’s first field should include an ISO 3166-1 Alpha-2 code
+          for the location it belongs to.
+        </Body>
+        <Body>
+          We also support ISO 3166-2 subdivision codes for countries containing
+          a cloud provider data center (both ISO 3166-1 and ISO 3166-2 codes may
+          be used for these countries). All valid country codes and the zones to
+          which they map are listed in the table below. Additionally, you can
+          view a list of all location codes{' '}
+          <Link href="/static/atlas/country_iso_codes.txt">here</Link>.
```
gribnoysup, line 143: Where is this pointing to?

packages/compass-global-writes/src/components/states/shard-key-correct.tsx
```diff
@@ -0,0 +1,200 @@
...
+      </div>
+    </div>
+  );
+}
+
+export default connect(
+  (state: RootState) => ({
+    namespace: state.namespace,
+    // For this view, sharKey is always defined
+    shardKey: state.shardKey as ShardKey,
```
gribnoysup, line 192: Can we be stricter with this instead of asserting the types? Make sure that types are explicit when this property is present and throw if we got into an unexpected state for example. Weird that we're allowing to render something that is potentially broken, especially because inside the component we are accessing properties on an object that can be actually undefined

packages/compass-global-writes/src/components/states/shard-key-correct.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+export type ShardKeyCorrectProps = {
+  namespace: string;
+  shardKey: ShardKey;
+  shardZones: ShardZoneData[];
+  isUnmanagingNamespace: boolean;
+  onUnmanageNamespace: () => void;
+};
+
+export function ShardKeyCorrect({
```
gribnoysup, line 56: Would be nice to have a bit of padding at the bottom here
image

packages/compass-global-writes/src/components/states/shard-key-correct.tsx
```diff
@@ -0,0 +1,200 @@
...
+      </Body>
+      <div>
+        <Button
+          data-testid="shard-collection-button"
+          onClick={onUnmanageNamespace}
+          disabled={isUnmanagingNamespace}
+          variant="primary"
+          leftGlyph={
+            isUnmanagingNamespace ? (
+              <SpinLoader title="Unmanaging collection" />
```
gribnoysup, line 177: Pretty sure leafygreen buttons have the loading state support now, maybe we can use that instead so that we're mor aligned with design system?

paula-stacho: thanks!

gribnoysup:
GitHub collapsed the `reducer` file and I only looked at the changes there today while revieweing again, sorry for not including this in my initial review

packages/compass-global-writes/src/components/states/shard-key-correct.tsx
```diff
@@ -0,0 +1,206 @@
...
+    return shardKey.fields[1].name;
+  }, [shardKey]);
+
+  const { atlasMetadata } = useConnectionInfo();
+
+  const sampleCodes = useMemo(() => {
+    const { collection, database } = toNS(namespace);
+    return {
+      findingDocuments: `use ${database}\ndb.["${collection}"].find({"location": "US-NY", "${customShardKeyField}": "<id_value>"})`,
+      insertingDocuments: `use ${database}\ndb.["${collection}"].insertOne({"location": "US-NY", "${customShardKeyField}": "<id_value>",...<other fields>})`,
```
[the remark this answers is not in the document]
paula-stacho: ah of course, thanks!

packages/compass-global-writes/src/store/reducer.ts
```diff
@@ -240,5 +375,154 @@ export const createShardKey =
...
+      type: GlobalWritesActionTypes.SubmittingForShardingFinished,
+      managedNamespace,
+    });
+    openToast(toastId, {
+      title: 'Sharding your collection …',
+      description: 'This should not take too long.',
+      dismissible: true,
+      variant: 'progress',
+      // TODO: Add actionElement to cancel sharding.
+    });
```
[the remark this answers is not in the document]
paula-stacho: the "sharding in progress" is another ticket, this one was meant to focus on "shard key correct", so I haven't paid too much attention to this bit. you're right, I will remove it!

packages/compass/src/main/application.ts
```diff
@@ -316,7 +316,7 @@ class CompassApplication {
           );
 
           log.warn(
-            mongoLogId(1_001_000_332),
+            mongoLogId(1_001_000_333),
```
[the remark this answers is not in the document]
paula-stacho: good point!

packages/compass-global-writes/src/components/states/shard-key-correct.tsx
```diff
@@ -0,0 +1,210 @@
...
+          view a list of all location codes{' '}
+          <Link href="/static/atlas/country_iso_codes.txt">here</Link>.
+        </Body>
+        <Body>
+          {atlasMetadata?.projectId && atlasMetadata?.clusterName && (
+            <>
+              Locations’ zone mapping can be changed by navigating to this
+              clusters{' '}
+              <Link
+                href={`/v2/${atlasMetadata?.projectId}/clusters/edit/${atlasMetadata?.clusterName}`}
```
gribnoysup, line 167: Easy to miss, mms client uses hash router inside the project, so you need a hash here (which makes me think that we might want to put all the url logic like that in one place maybe so that it's easier to stay consistent, not an action for now, just a thought)

```suggestion
                href={`/v2/${atlasMetadata?.projectId}#/clusters/edit/${atlasMetadata?.clusterName}`}
```

packages/compass-global-writes/src/store/index.spec.ts
```diff
@@ -36,108 +74,139 @@ describe('GlobalWritesStore Store', function () {
...
+                  },
+                  unique: false,
+                },
+              ],
+            };
+          }
+        },
+      });
+      await store.dispatch(fetchClusterShardingData());
+      // expect(store.getState().status).to.equal('SHARD_KEY_CORRECT'); //  no idea why this does not work
```
gribnoysup, line 156: Is the state not matching or the check? What state do you see here? (we probably want to either remove this comment altogether or resolve, a bit weird to just leave this comment 🙂)

packages/compass-global-writes/src/store/reducer.ts
```diff
@@ -64,43 +101,163 @@ export enum ShardingStatuses {
...
+  zoneName: string;
+  zoneLocations: string[];
+};
 export type RootState = {
   namespace: string;
-  isNamespaceSharded: boolean;
-  status: ShardingStatus;
-};
+  managedNamespace?: ManagedNamespace;
+  // status: ShardingStatus;
```
gribnoysup, line 154:
```suggestion
```

gribnoysup (approved):

packages/compass-global-writes/src/store/index.spec.ts
paula-stacho: I was seeing NOT_READY. Putting the check behind waitFor helped, but I'm not sure why it was needed 🤔

gribnoysup: I guess it's because this fetch action is not the one that actually will update the state, and the one that will is not awaited inside this one:

image

Not sure if it's this intentional or not

paula-stacho: ah right! I was looking at this code and read it wrong 🤦  outside of tests it currently doesn't make difference if we wait for this one, since the `fetchClusterShardingData` is not awaited.

PR #6317, opened 2024-10-02
chore(ci): log electron app messages

kraenhansen:
## Description

This adds runtime options when starting the Electron app in e2e tests to write the Electron log into a file which will be attached to the task on Evergreen.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-e2e-tests/helpers/compass.ts
+++ b/packages/compass-e2e-tests/helpers/compass.ts
@@ -143,8 +143,8 @@ export const serverSatisfies = (
   );
 };
 
-// For the user data dirs
-let i = 0;
+// For the user data dirs and logs
+let runCounter = 0;
 
 interface Coverage {
   main?: string;
@@ -569,7 +569,7 @@ async function processCommonOpts({
   if (!defaultUserDataDir) {
     defaultUserDataDir = path.join(
       os.tmpdir(),
-      `user-data-dir-${Date.now().toString(32)}-${++i}`
+      `user-data-dir-${Date.now().toString(32)}-${runCounter}`
     );
   }
   const chromedriverLogPath = path.join(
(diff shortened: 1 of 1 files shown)


lerouxb (approved):

packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -644,6 +645,15 @@ async function startCompassElectron(
     chromeArgs.push('--showed-network-opt-in=true');
   }
 
+  // Logging output from Electron, even before the app loads any JavaScript
+  const electronLogFile = path.join(LOG_PATH, `electron-${runCounter}.log`);
+  chromeArgs.push(
+    // See https://www.electronjs.org/docs/latest/api/command-line-switches#--enable-loggingfile
+    '--enable-logging=file',
+    // See https://www.electronjs.org/docs/latest/api/command-line-switches#--log-filepath
+    `--log-file=${electronLogFile}`
```
lerouxb, line 654: Hmm. I wonder if we should add --log-level while at it.

kraenhansen: It's info by default and I've found that it's already pretty chatty.
Here's the list of available levels: https://source.chromium.org/chromium/chromium/src/+/main:base/logging.h;l=370-378?q=logging::LogSeverity&ss=chromium

Reading [the `--log-level` documentation](https://www.electronjs.org/docs/latest/api/command-line-switches#--log-leveln) it seems to me that one would need to play around with `--v` and `--vmodule` to get specific verbose logging and it's probably best to adjust on a case-by-case basis:
> Note that two complimentary logging mechanisms in Chromium -- LOG() and VLOG() -- are controlled by different switches. --log-level controls LOG() messages, while --v and --vmodule control VLOG() messages. So you may want to use a combination of these three switches depending on the granularity you want and what logging calls are made by the code you're trying to watch.
> 
> See [Chromium Logging source](https://source.chromium.org/chromium/chromium/src/+/main:base/logging.h) for more information on how LOG() and VLOG() interact. Loosely speaking, VLOG() can be thought of as sub-levels / per-module levels inside LOG(INFO) to control the firehose of LOG(INFO) data.
> 
> See also --enable-logging, --log-level, --v, and --vmodule.

PR #6321, opened 2024-10-03
fix(proxy): show friendly error when proxy config is not supported COMPASS-8345

kmruiz:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass/src/main/application.ts
+++ b/packages/compass/src/main/application.ts
@@ -1,7 +1,8 @@
 import './disable-node-deprecations'; // Separate module so it runs first
 import path from 'path';
 import { EventEmitter } from 'events';
-import { BrowserWindow, Event, ProxyConfig, dialog } from 'electron';
+import type { BrowserWindow, Event, ProxyConfig } from 'electron';
+import { dialog } from 'electron';
 import { app, safeStorage, session } from 'electron';
 import { ipcMain } from 'hadron-ipc';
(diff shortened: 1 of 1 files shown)


packages/compass/src/main/application.ts
```diff
@@ -301,7 +301,41 @@ class CompassApplication {
...
+            'userData'
+          )}${sep}AppPreferences${sep}General.json`;
+
+          dialog.showErrorBox(
+            'Unsupported proxy configuration',
+            `${headline}\n\n
+            To reset the proxy configuration, remove the "proxy" key in ${configPath} and restart Compass.`
+          );
+
+          app.quit();
```
Anemy, line 337: Why do we quit here? Does the error popup also get closed when the app quits? Haven't tried.

kmruiz: The application will continue running and crash because of the proxy missconfig, showing another error and another stack trace. Early quitting avoid this and gives the user all the context to action on the error.

Do you have any suggestion on how to improve it?

Anemy: Gotcha, just checking, no suggestion on improvement, I'm not too familiar with the issue. I was worried the app.quit would make the popup `showErrorBox` disappear as well.

Anemy (approved):

PR #6336, opened 2024-10-09
feat(connection-form): allow editing connected connections, but only the name, colour and favorite checkbox COMPASS-8160

lerouxb:
See [figma design](https://www.figma.com/proto/JmNVeRAx398lbu3kNQiuZc/Multiple-connection-UX-tweaks?page-id=0%3A1&node-id=5-1995&node-type=canvas&viewport=25%2C421%2C0.62&t=FN4NntbaFlOMRK7Q-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=5%3A1995&show-proto-sidebar=1) and [slack thread](https://mongodb.slack.com/archives/C05MLSFMUJC/p1728388841356229).

Screenshot 2024-10-09 at 12 33 01

### DIFF
--- a/packages/connection-form/src/components/connection-form-actions.tsx
+++ b/packages/connection-form/src/components/connection-form-actions.tsx
@@ -44,7 +44,7 @@ export type ConnectionFormModalActionsProps = {
 
   onCancel?(): void;
   onSave?(): void;
-  onSaveAndConnect(): void;
+  onSaveAndConnect?(): void;
 };
 
 export function ConnectionFormModalActions({
@@ -89,7 +89,11 @@ export function ConnectionFormModalActions({
           <div className={saveAndConnectStyles}>
             <Button
               data-testid="save-button"
-              variant={ButtonVariant.PrimaryOutline}
+              variant={
+                onSaveAndConnect
+                  ? ButtonVariant.PrimaryOutline
+                  : ButtonVariant.Primary
+              }
               disabled={false}
               onClick={onSave}
             >
@@ -98,13 +102,15 @@ export function ConnectionFormModalActions({
           </div>
         )}
 
-        <Button
-          data-testid="connect-button"
-          variant={ButtonVariant.Primary}
-          onClick={onSaveAndConnect}
-        >
-          {saveAndConnectLabel}
-        </Button>
+        {onSaveAndConnect && (
+          <Button
+            data-testid="connect-button"
+            variant={ButtonVariant.Primary}
+            onClick={onSaveAndConnect}
+          >
+            {saveAndConnectLabel}
+          </Button>
+        )}
       </div>
     </div>
   );
--- a/packages/connection-form/src/components/connection-form.tsx
+++ b/packages/connection-form/src/components/connection-form.tsx
@@ -19,6 +19,8 @@ import {
   cx,
   palette,
   useDarkMode,
+  Button,
+  Icon,
 } from '@mongodb-js/compass-components';
 import { cloneDeep } from 'lodash';
 import ConnectionStringInput from './connection-string-input';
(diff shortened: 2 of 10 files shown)


packages/connection-form/src/components/connection-form-actions.tsx
```diff
@@ -89,7 +89,11 @@ export function ConnectionFormModalActions({
           <div className={saveAndConnectStyles}>
             <Button
               data-testid="save-button"
-              variant={ButtonVariant.PrimaryOutline}
+              variant={
+                onSaveAndConnect
```
lerouxb, line 93: Basically: If there is no save&connect button (the Connect button), make the Save button the "default" (ie. solid green) one. type for both are still implicitly "button", though. there is no default. Would be ambiguous anyway because we have a custom onSubmit handler that would have to be kept in sync.

packages/connection-form/src/components/connection-form.tsx
```diff
@@ -429,16 +422,58 @@ function ConnectionForm({
...
   );
 
   const showHelpCardsInForm = useConnectionFormSetting('showHelpCardsInForm');
 
+  /*
+  To test onSubmit on a form (ie. what happens when you press {enter} and not
+  what happens when you click a submit button) you have to use submit() which only
+  works with role="form", but eslint doesn't like that because role="form"
+  should be redundant.
+  https://stackoverflow.com/questions/59362804/pressing-enter-to-submit-form-in-react-testing-library-does-not-work
```
lerouxb, line 470: If there's a better way to simulate submitting a form by pressing enter rather than pressing a button so that the callback gets called in tests, please let me know.

packages/connection-form/src/components/connection-form.tsx
```diff
@@ -451,7 +486,8 @@ function ConnectionForm({
         <div className={formHeaderStyles}>
           <div className={formHeaderContentStyles}>
             <H3 className={headingWithHiddenButtonStyles}>
-              {initialConnectionInfo.favorite?.name ?? 'New Connection'}
+              {(initialConnectionInfo.favorite?.name ?? 'New Connection') ||
+                'Edit Connection'}
```
lerouxb, line 490: Drive-by to fix the case where you edit a connection that has no favorite name.

Screenshot 2024-10-08 at 14 14 53

packages/connection-form/src/components/form-help/form-help.tsx
```diff
@@ -29,13 +30,13 @@ const descriptionStyles = css({
 });
 
 const sectionDarkModeStyles = css({
-  '--theme-background-color': palette.blue.dark3,
-  '--theme-border-color': palette.blue.dark2,
+  '--theme-background-color': palette.gray.dark3,
```
lerouxb, line 33: Note that according to the design, the blue help text boxes on the right are now grey.

packages/connection-form/src/components/connection-form.tsx
```diff
@@ -303,10 +315,12 @@ type ConnectionFormPropsWithoutSettings = {
   darkMode?: boolean;
   initialConnectionInfo: ConnectionInfo;
   connectionErrorMessage?: string | null;
+  disableEditingConnectedConnection?: boolean;
```
lerouxb, line 318: Optional new prop with undefined acting like false, so vscode should be unaffected.

packages/connection-form/src/components/connection-form.tsx
```diff
@@ -461,9 +497,36 @@ function ConnectionForm({
...
         <div className={formContentContainerStyles}>
           <div className={formContentStyles}>
+            {disableEditingConnectedConnection && onDisconnectClicked && (
+              <Banner
+                data-testid="disabled-connected-connection-banner"
+                className={disabledConnectedConnectionBannerStyles}
+              >
+                <div className={disabledConnectedConnectionContentStyles}>
+                  <div>
+                    While connected, you may only personalize your
```
addaleax, line 507: This isn't really what "personalize" means, unless we consider connections to be people

addaleax (approved):

packages/connection-form/src/components/connection-form.tsx
lerouxb: Fits the dictionary definition.

PR #6344, opened 2024-10-10
feat(connection-form): show a banner on the connection form explaining why you can't edit the connection COMPASS-8264

lerouxb:
Screenshot 2024-10-10 at 09 29 10

### DIFF
--- a/packages/connection-form/src/components/connection-form.tsx
+++ b/packages/connection-form/src/components/connection-form.tsx
@@ -521,7 +521,7 @@ function ConnectionForm({
                 </div>
               </Banner>
             )}
-            {protectConnectionStrings && (
+            {protectConnectionStrings && !disableEditingConnectedConnection && (
               <Banner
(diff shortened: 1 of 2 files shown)


packages/connection-form/src/components/connection-form.tsx
```diff
@@ -521,6 +521,17 @@ function ConnectionForm({
                 </div>
               </Banner>
             )}
+            {protectConnectionStrings && (
```
gribnoysup, line 524: I know it's a corner case, but would it make more sense to make this banner show up only when you can actually edit the connection string taking connection status into account? Otherwise it's getting pretty packed in the when editing a connected one:

image

gribnoysup (approved):

packages/connection-form/src/components/connection-form.tsx
lerouxb: Aah well spotted. Fixed.

PR #6351, opened 2024-10-11
chore(e2e): small screen fixes

lerouxb:


packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -1020,19 +1020,28 @@ export async function init(
   // For browser.executeAsync(). Trying to see if it will work for browser.execute() too.
   await browser.setTimeout({ script: 5_000 });
 
-  // larger window for more consistent results
-  const [width, height] = await browser.execute(() => {
-    // in case setWindowSize() below doesn't work
-    window.resizeTo(window.screen.availWidth, window.screen.availHeight);
+  if (TEST_COMPASS_WEB) {
```
lerouxb, line 1023: Turns out window.resizeTo is only working in the browser. Sometimes. Also window.screen.availWidth, window.screen.availHeight was returning nonsense in electron world so even if it did size the window with webdriverio's browser.setWindowSize() it wasn't being helpful.

So I added a new path for electron land and kept the old one for compass web.

gribnoysup (approved):

PR #6361, opened 2024-10-14
WIP: order latest first COMPASS-6706

lerouxb:
## TODO:
* [ ] make sure we send the default order through to all relevant places. ie. explain

---

Super hacky, will need lots of cleanup but at least we can try it out.

`{ _id: -1 }` for normal collections, undefined whenever the `_id_` index does not exist. ie. timeseries and views.

### DIFF
--- a/.evergreen/print-compass-env.sh
+++ b/.evergreen/print-compass-env.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-export MONGODB_DEFAULT_VERSION=6.0.x
+export MONGODB_DEFAULT_VERSION=7.0.x
 
 if [[ $OSTYPE == "cygwin" ]]; then
     export PLATFORM='win32'
--- a/THIRD-PARTY-NOTICES.md
+++ b/THIRD-PARTY-NOTICES.md
@@ -1,5 +1,5 @@
 The following third-party software is used by and included in **Mongodb Compass**.
-This document was automatically generated on Wed Oct 16 2024.
+This document was automatically generated on Thu Oct 17 2024.
 
 ## List of dependencies
 
--- a/docs/tracking-plan.md
+++ b/docs/tracking-plan.md
@@ -1,7 +1,7 @@
 
 # Compass Tracking Plan
 
-Generated on Wed, Oct 16, 2024 at 01:37 PM
+Generated on Thu, Oct 17, 2024 at 03:33 PM
 
 ## Table of Contents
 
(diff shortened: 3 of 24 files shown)


packages/compass-query-bar/src/constants/query-option-definition.ts
```diff
@@ -30,7 +30,7 @@ export const OPTION_DEFINITION: {
   sort: {
     name: 'sort',
     type: 'document',
-    placeholder: "{ field: -1 } or [['field', -1]]",
+    placeholder: '{ $_id: -1}',
```
lerouxb, line 33: This would obviously have to change based on whatever rules we end up with too.

packages/compass-crud/src/utils/cancellable-queries.ts
```diff
@@ -7,7 +7,7 @@ export async function countDocuments(
   dataService: DataService,
   preferences: PreferencesAccess,
   ns: string,
-  filter: BSONObject,
+  filter: BSONObject | undefined,
```
lerouxb, line 10: drive-by

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -109,20 +110,34 @@ const INITIAL_BULK_UPDATE_TEXT = `{
...
+  }
+
+  return { _id: -1 };
+}
+
 export const fetchDocuments: (
   dataService: DataService,
-  serverVersion: string,
-  isDataLake: boolean,
+  fetchDocumentsOptions: FetchDocumentsOptions,
```
lerouxb, line 129: This is getting ugly 😆

packages/compass-crud/src/stores/crud-store.ts
```diff
@@ -183,9 +195,20 @@ const extractCollectionStats = (collection: Collection): CollectionStats => {
...
     avg_document_size: coll.avg_document_size,
+    index_details: coll.index_details,
   };
 };
 
+function getDefaultSort(
+  collectionStats: CollectionStats | null
+): Sort | undefined {
+  if (collectionStats?.index_details._id_) {
+    return { _id: -1 };
```
lerouxb, line 206: I'm now not sure. It seems like `{ $natural: -1 }` is actually just as fast as `{ $natural: 1 }` *(the default) at least in most cases. If it is in MOST cases we'd have to just change this logic and return `{ $natural: -1 }`. If it works and is fine in ALL cases, then most of this PR is probably not needed. We could just always use that as a default and we don't even need this function.

lerouxb: But then.. I don't think we want $natural at all because it's just how documents are sorted in wiredtiger, which means any order really. For example, an update can bump an old document to the top of the natural order.

PR #6364, opened 2024-10-14
show the search warning in all three places

lerouxb:
These are the 6 cases for when there are no output documents returned.

If an Atlas-only stage is involved:

stage-editor_atlas
text_atlas
focus_atlas

If no atlas-only stage is involved:

stage-editor_no-documents
text_no-documents
focus_no-documents

### DIFF
--- a/packages/compass-aggregations/src/components/atlas-no-results.tsx
+++ b/packages/compass-aggregations/src/components/atlas-no-results.tsx
@@ -1,0 +1,48 @@
+import React from 'react';
+
+import {
+  css,
+  palette,
+  spacing,
+  Body,
+  useDarkMode,
+  Subtitle,
+} from '@mongodb-js/compass-components';
+
+const centeredContent = css({
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  height: '100%',
+  padding: spacing[3],
+  flexDirection: 'column',
+  textAlign: 'center',
+});
+
+const missingAtlasIndexLightStyles = css({
+  color: palette.green.dark2,
+});
+
+const missingAtlasIndexDarkStyles = css({
+  color: palette.green.base,
+});
+
+export default function AtlasNoResults() {
+  const darkMode = useDarkMode();
+
+  return (
+    <div className={centeredContent}>
+      <Subtitle
+        className={css(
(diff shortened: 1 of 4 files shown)


packages/compass-aggregations/src/components/atlas-no-results.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
```
mcasimir, line 1: NIT: is this AtlasSearchNoResults, missingAtlasSearchIndex ...?

lerouxb: Any atlas-only stage. I didn't fix that bug. I'm not even sure if it is a bug since more stages than just $search relate to atlas search.

lerouxb: isAtlasOnly() seems to map to $search, $searchMeta, $vectorSearch now. Which actually sounds about right.

mcasimir: then this was probably messed up before, i think we should change it. 

This behaviour is not intended for any atlas stage. I can totally see us adding a new atlas only stage inside devtools-shared that is not a search feature, and get this weird error which is only meant for search.

lerouxb: Updated now. Let me know what you think.

mcasimir (approved):

packages/compass-aggregations/src/components/focus-mode/focus-mode-stage-preview.spec.tsx
```diff
@@ -86,6 +86,18 @@ describe('FocusModeStagePreview', function () {
       const preview = screen.getByTestId('focus-mode-stage-preview');
       expect(within(preview).getByText(/no preview documents/i)).to.exist;
     });
+    it('renders missing search index text for $search', async function () {
```
mcasimir, line 89: super NIT feel free to ignore, any reason not to `for searchStage of [$search, $vectorSearch, ...]` this one?

lerouxb: good idea. I'll change.

PR #6370, opened 2024-10-15
fix(compass-generative-ai): Increase the character size restriction for query and aggregation genai COMPASS-8369

ruchitharajaghatta:
## Description
Currently if folks use the generative ai feature in Compass with a collection with documents with a lot of fields, we'll show an error:
"Sorry, your collections have too many fields to process. Please try 
using this feature on a collection with smaller documents."
PROMPT_TOO_LONG error from mms
This is because we don't want to exceed the token limit of open ai. We recently changed from gpt-3.5-turbo to gpt-4o. The new limit is 128,000 tokens and OpenAI approximates 1 token as ~= 4 english chars, so we updated this field to 4*128,000 = 512000 in the server. Since we want the final validation to be on the server lets do 10x what the server limit is at. The server does some further changes to schema and sample documents so we should let it be the decider. The frontend should only prevent uploading massive schemas and documents in the request. So, in compass the limit is set to 10x this (5,120,000).

Also updated the PROMPT_TOO_LONG message to be more transparent and clear that this is the limit causing the error. 
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-generative-ai/src/atlas-ai-service.spec.ts
+++ b/packages/compass-generative-ai/src/atlas-ai-service.spec.ts
@@ -204,7 +204,7 @@ describe('AtlasAiService', function () {
               userInput: 'test',
               collectionName: 'test',
               databaseName: 'peanut',
-              sampleDocuments: [{ test: '4'.repeat(600000) }],
+              sampleDocuments: [{ test: '4'.repeat(5120001) }],
               requestId: 'abc',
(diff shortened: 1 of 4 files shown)


Anemy (approved):
lgtm ! Nice

PR #6374, opened 2024-10-17
feat(compass-global-writes): sharding error state COMPASS-8349

paula-stacho:
## Description
Notes: 
- the `CreateShardKeyForm` is simply moved out of unsharded state, as it is now reused for both `Unsharded` and `ShardingError` states - changes pointed out in a comment
- the `ShardingError` also displays the error and a button to cancel sharding. this means that from this state the user can either go back (via the cancel button), or forward (by submitting the form)
- again, the implementation is based on mms. for example:
  - the error parsing https://github.com/10gen/mms/blob/fdad01af9558be4f9e940e0aaa50d5a3a1f0c6cf/client/packages/common/utils/deploymentStatusError.ts#L26
  - the banner being a warning level 🤷 (these levels are one of the things I want to check with design) https://github.com/10gen/mms/blob/74cf2c73b098b977cfbed907c1990b13a11b93c5/client/packages/project/metrics/components/DataExplorerShardKey.tsx#L903

I have thought about whether this state needs to be separate, or if it would better be the unsharded state with optional error. In the end I continued the approach with separate states and the store behaving as a state machine. It still makes sense to me and I find it better than having a ton of conditional bits in the components. But, I've been working on this for a while, so I'm definitely biased and I wouldn't be surprised if the solution isn't that clear to someone else - let me know if that's the case and we can talk about how to make it clearer!

Preview:
if you happen to read the error message - this one actually should be handled in a different way (key mismatch screen), but this was the easiest for me to reproduce
Screenshot 2024-10-24 at 14 40 09

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-global-writes/src/components/create-shard-key-form.tsx
```diff
@@ -0,0 +1,312 @@
...
+            </div>
+          )}
+        </Accordion>
+        <div>
+          <Button
+            data-testid="shard-collection-button"
+            onClick={onSubmit}
+            disabled={!secondShardKey || isSubmittingForSharding}
+            variant="primary"
+            isLoading={isSubmittingForSharding}
```
paula-stacho, line 302: I know I said no changes, but this is one small change. It used to be
```
disabled={!secondShardKey || isSubmittingForSharding}
variant="primary"
leftGlyph={
  isSubmittingForSharding ? (
    <SpinLoader title="Creating shard key" />
  ) : undefined
}
```
I changed it because we use the isLoading in other buttons by now.

packages/compass-global-writes/src/components/create-shard-key-form.tsx
```diff
@@ -0,0 +1,339 @@
...
+      isCancellingSharding: [
+        ShardingStatuses.CANCELLING_SHARDING,
+        ShardingStatuses.CANCELLING_SHARDING_ERROR,
+      ].includes(state.status),
+    };
+  },
+  {
+    onCreateShardKey: createShardKey,
+  }
+)(CreateShardKeyForm);
```
paula-stacho, line 339: Another small change - this is now connected directly, and it works with error/no error statuses.

packages/compass-global-writes/src/components/common-styles.ts
```diff
@@ -0,0 +1,19 @@
+import { css, spacing } from '@mongodb-js/compass-components';
+
+export const bannerStyles = css({
+  textAlign: 'justify',
```
mabaasit, line 4: this adds extra space in the banner and the message looks a bit weird (in the screenshot attached in the PR description). also, would be nice if we can split the error message by new line (server error has it).

paula-stacho: changed this setting for the error specifically. otherwise most messages I tried looked better with justify. also added pre-wrap on the error.

mabaasit (approved):

PR #6378, opened 2024-10-18
chore: bump the major version we run unit tests on in ci

lerouxb:
We were running unit tests on 6.0.x in CI and whatever mongodb-runner uses by default locally.

### DIFF
--- a/.evergreen/print-compass-env.sh
+++ b/.evergreen/print-compass-env.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-export MONGODB_DEFAULT_VERSION=6.0.x
+export MONGODB_DEFAULT_VERSION=7.0.x
 
 if [[ $OSTYPE == "cygwin" ]]; then
     export PLATFORM='win32'
--- a/packages/data-service/src/connect-mongo-client.spec.ts
+++ b/packages/data-service/src/connect-mongo-client.spec.ts
@@ -252,8 +252,10 @@ describe('connectMongoClient', function () {
         expect(error).to.be.instanceOf(Error);
 
         // propagates the tunnel error
-        expect(error.message).to.match(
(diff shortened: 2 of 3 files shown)


.evergreen/print-compass-env.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-export MONGODB_DEFAULT_VERSION=6.0.x
+export MONGODB_DEFAULT_VERSION=7.0.x
```
lerouxb, line 5: I'm thinking we should probably not set this here and use it print-compass-env.js and then somehow rather fill in a default if nothing is set in the compass tests. That would be the only way to have a consistent default locally and in CI because we don't use this file when just doing `npm test` locally.

packages/data-service/src/connect-mongo-client.spec.ts
```diff
@@ -252,8 +252,10 @@ describe('connectMongoClient', function () {
         expect(error).to.be.instanceOf(Error);
 
         // propagates the tunnel error
-        expect(error.message).to.match(
-          /(All configured authentication methods failed|ENOTFOUND compass-tests\.fakehost\.localhost)/
+        // NOTE: this heavily depends on which server version we're running on
```
paula-stacho, line 255: do we have to test for a specific message? can't we just confirm an error?

lerouxb: I don't know. I didn't write the original test and I don't want to unnecessarily change the intention.

paula-stacho: or, if we want to test the propagation, is there a way to mock the error?

lerouxb: Yes, but this file is an integration test, not a unit test.

gribnoysup (approved):

PR #6389, opened 2024-10-22
chore(compass-schema): "Schema Exported" event COMPASS-8332

kraenhansen:
## Description

Adds a "Schema Exported" telemetry-event firing when clicking the "Share Schema as JSON" menu item:

Screenshot 2024-10-22 at 19 29 10

I've verified the implementation running

```
node ./scripts/logserver.js
```

and

```
HADRON_METRICS_SEGMENT_API_KEY='ignore' HADRON_METRICS_SEGMENT_HOST='http://localhost:8000' npm start
```

### Checklist

## Motivation and Context
- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Open Questions

- None

## Types of changes

- [ ] *Backport Needed*


packages/compass-schema/src/stores/store.ts
```diff
@@ -181,6 +184,21 @@ export function activateSchemaPlugin(
...
+      const { schema } = this.state;
+      // Use a function here to a) ensure that the calculations here
+      // are only made when telemetry is enabled and b) that errors from
+      // those calculations are caught and logged rather than displayed to
+      // users as errors from the core schema sharing logic.
+      const trackEvent = () => ({
+        has_schema: hasSchema,
+        schema_width: schema?.fields?.length ?? 0,
+        schema_depth: schema ? calculateSchemaDepth(schema) : 0,
+        geo_data: schema ? schemaContainsGeoData(schema) : false,
```
Anemy, line 197: We do these same schema telemetry metadata calculations in the `_trackSchemaAnalyzed` function currently. How do you feel about using a shared function for them?

Anemy (approved):
lgtm, one suggestion on reducing some code duplication, not a blocker.

packages/compass-schema/src/stores/store.ts
paula-stacho: might be worth noting - we're planning to get rid of the menu action altogether in oncoming schema-export project, so this is most likely a temporary tracking event, just to see how many people are actually using the action we're about to remove.

paula-stacho: @kraenhansen please confirm with Julia, but I think we don't even need any metadata

kraenhansen: Clarification from Julia:

> It would also be great if we include the same properties that we include in the "Schema Analyzed" event. 
> For the naming–I think it will be best to call it "Schema Exported."

kraenhansen: > we're planning to get rid of the menu action altogether

I believe that speaks even stronger in favour of not introducing a shared function 👍

Anemy (approved):

PR #6399, opened 2024-10-25
feat(compass-global-writes): incomplete sharding setup COMPASS-8372

paula-stacho:
## Description
When the collection has already been geosharded with one key, it cannot be sharded with another one. Until now, we allowed this because we went from SHARD_KEY_CORRECT -> UNSHARDED. This PR handles the state when collection has already been sharded (shardKey exists), but it is not managed now. The new state is INCOMPLETE_SHARDING_SETUP, and from this state we can only resume sharding. The request is the same, but the key description is taken from the existing key, not user input.

The change looks bigger than it is - `example-commands-markup.tsx` and `shard-zones-description.tsx` are just moved out of `shard-key-correct.tsx`. The bulk of changes is in the reducer, which had to be rearranged a bit to incorporate the new scenario (we haven't realised this possibility exists until later on).

Screenshot 2024-10-31 at 13 13 03

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-global-writes/src/components/states/incomplete-sharding-setup.tsx
+++ b/packages/compass-global-writes/src/components/states/incomplete-sharding-setup.tsx
@@ -27,7 +27,7 @@ const containerStyles = css({
   marginBottom: spacing[400],
 });
 
-const unmanageBtnStyles = css({
+const manageBtnStyles = css({
   marginTop: spacing[100],
 });
 
@@ -68,7 +68,7 @@ export function IncompleteShardingSetup({
             onClick={onResume}
             variant={ButtonVariant.Default}
             isLoading={isSubmittingForSharding}
-            className={unmanageBtnStyles}
+            className={manageBtnStyles}
           >
             Enable Global Writes
           </Button>
--- a/packages/compass-global-writes/src/components/states/incomplete-sharding-setup.tsx
+++ b/packages/compass-global-writes/src/components/states/incomplete-sharding-setup.tsx
@@ -1,0 +1,100 @@
+import {
+  Banner,
+  BannerVariant,
+  Button,
+  spacing,
+  css,
+  ButtonVariant,
+  Link,
+} from '@mongodb-js/compass-components';
+import React from 'react';
+import ShardKeyMarkup from '../shard-key-markup';
+import {
+  resumeManagedNamespace,
+  ShardingStatuses,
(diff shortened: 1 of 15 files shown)


packages/compass-global-writes/src/components/states/incomplete-sharding-setup.tsx
```diff
@@ -0,0 +1,100 @@
...
+          rel="noreferrer"
+        >
+          Read more about Global Writes
+        </Link>
+        <div>
+          <Button
+            data-testid="manage-collection-button"
+            onClick={onResume}
+            variant={ButtonVariant.Default}
+            isLoading={isSubmittingForSharding}
```
mabaasit, line 70: nit: should we add spinner here as well while its loading?

packages/compass-global-writes/src/store/reducer.ts
```diff
@@ -561,6 +598,23 @@ export const fetchClusterShardingData =
...
+      throw new Error('Cannot resume managed namespace without a shardKey');
+    }
+    const data: CreateShardKeyData = {
+      customShardKey: shardKey.fields[1].name,
+      isShardKeyUnique: shardKey.isUnique,
+      isCustomShardKeyHashed: shardKey.fields[1].type === 'HASHED',
+      numInitialChunks: null, // default
+      presplitHashedZones: false, // default
+    };
+    await dispatch(createShardKey(data));
```
mabaasit, line 614: When I resume sharding, it eventually transitions to `SHARDING` status and if I cancel that process, it takes me back to `UNSHARDED` status but it should transition to `INCOMPLETE_SHARDING_SETUP` status, right?

packages/compass-global-writes/src/components/states/incomplete-sharding-setup.tsx
paula-stacho: ah right, I wrote this before finding out about the loadIndicator, thanks for noticing!

packages/compass-global-writes/src/store/reducer.ts
paula-stacho: you're right, I missed that one! thanks a lot!

PR #6427, opened 2024-10-31
fix(user-data): use semaphore to limit reads COMPASS-7256

mabaasit:
### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-user-data/src/user-data.spec.ts
+++ b/packages/compass-user-data/src/user-data.spec.ts
@@ -158,6 +158,20 @@ describe('user-data', function () {
         expect(mongoshData?.[1]).to.be.instanceOf(Stats);
       }
     });
+
+    it('reads many number of files', async function () {
+      const files = Array.from({ length: 10000 }, (_, i) => [
+        `data${i}.json`,
+        JSON.stringify({ name: `VSCode${i}` }),
+      ]);
+
+      await Promise.all(
+        files.map(([filepath, data]) => writeFileToStorage(filepath, data))
+      );
+
+      const result = await getUserData().readAll();
+      expect(result.data).to.have.lengthOf(10000);
+    });
   });
 
   context('UserData.readOne', function () {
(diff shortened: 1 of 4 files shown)


packages/compass-user-data/src/user-data.ts
```diff
@@ -68,6 +69,7 @@ export class UserData<T extends z.Schema> {
   private readonly serialize: SerializeContent<z.input<T>>;
   private readonly deserialize: DeserializeContent;
   private readonly getFileName: GetFileName;
+  private readonly semaphore = new Semaphore(1000);
```
addaleax, line 72: The Node.js thread pool for fs operations (`UV_THREADPOOL_SIZE`) has a default size of 4 or so. I think a lower number, say 50 or 100, would also be fine here and result in similar performance while at the same time still being significantly better at reducing the risk of EMFILE/ENFILE errors

packages/compass-user-data/src/user-data.spec.ts
```diff
@@ -158,6 +158,20 @@ describe('user-data', function () {
...
+
+    it('reads many number of files', async function () {
+      const files = Array.from({ length: 10000 }, (_, i) => [
+        `data${i}.json`,
+        JSON.stringify({ name: `VSCode${i}` }),
+      ]);
+
+      await Promise.all(
+        files.map(([filepath, data]) => writeFileToStorage(filepath, data))
+      );
```
addaleax, line 170: Wouldn't this test also run the risk of running into file descriptor limitations?

addaleax (approved):

packages/compass-user-data/src/user-data.spec.ts
mabaasit: yeah that's a good point. i'll remove this test.

PR #6439, opened 2024-11-04
chore(compass-components): less document whitespace

kraenhansen:
## Description

As a preparation for COMPASS-4635 this solves the first part of the designed proposed in EXPO-2098: Making the expand-all button more compact and trimming padding.

### Compact expand-all button

Screenshot 2024-11-04 at 12 41 31

### Less padding on the left side of the trash button and between the "+" and the "▶"

Screenshot 2024-11-04 at 12 41 55

### Checklist
- [ ] New tests and/or benchmarks are included
- [ ] Documentation is changed or added
- [ ] I have signed the MongoDB Contributor License Agreement (https://www.mongodb.com/legal/contributor-agreement)

## Motivation and Context
- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*
- [x] Patch (non-breaking change which fixes an issue)
- [ ] Minor (non-breaking change which adds functionality)
- [ ] Major (fix or feature that would cause existing functionality to change)

PR #6446, opened 2024-11-05
feat(editor): expandable json editor

kraenhansen:
## Description

A sub-task of the design proposed in COMPASS-4635 is to add the "expand all" / "collapse all" button to the JSON editor view of documents.

Merging this PR will:
- Update the spacing of the actions container and styles injected by the `JSONEditor` component to use the non-deprecated variants, as well as pull the actions container to span the entire width.
- Add `pointer-events` css properties to avoid capturing clicks on whitespace of the actions container.
- Add a "compact" prop to the editors `ActionButton` component, using [the same hack introduced in #6439](https://github.com/mongodb-js/compass/pull/6439/files/3bc54710e5b07b8a7771973a509af83ef92f1eb6#diff-08b32be8f1f7995a0af322d352a6f5390db3c1c14bcf897bc6c8e595f113ccb2R47-R52). I hope to replace this by [an upstream change to leafygreen-ui](https://github.com/mongodb/leafygreen-ui/pull/2535).
- Add the "Expand all" / "Collapse all" button to the `ActionsContainer` conditionally rendered on the presence of an `onExpend` prop.
- Pass a callback for the `onExpend` from the `JSONEditor` into the editor, which calls the doc's collapse / expand methods.
- Adds 20px of spacing to the gutters of the editor when the "expand all" button is visible to prevent collisions with the fold / unfold icons in the gutter.

Notice how the "dismissal" behaviour is a bit different in the JSON editor than the regular document editor: As the focus is brought to the action button when clicking it, it will prevent the action group from disappearing. I.e. it is not enough to simply move the mouse pointer outside of the element to get the action buttons to hide. This is not new functionality but the effect of the existing styling here:

https://github.com/mongodb-js/compass/blob/05a557b9b16a2fc18ebfc264c2a10c25be92525b/packages/compass-editor/src/editor.tsx#L1382

https://github.com/user-attachments/assets/4ccbe12a-3327-49c6-afda-ec97b21f367f

### Checklist

## Motivation and Context
- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-crud/src/components/json-editor.tsx
+++ b/packages/compass-crud/src/components/json-editor.tsx
@@ -45,8 +45,7 @@ const editorDarkModeStyles = css({
 });
 
 const actionsGroupStyles = css({
-  paddingTop: spacing[2],
-  paddingRight: spacing[2],
+  padding: spacing[200],
 });
 
 export type JSONEditorProps = {
--- a/packages/compass-crud/src/components/json-editor.tsx
+++ b/packages/compass-crud/src/components/json-editor.tsx
@@ -257,6 +257,14 @@ const JSONEditor: React.FunctionComponent<JSONEditorProps> = ({
     onDeletionFinished,
   ]);
 
+  const toggleExpandCollapse = useCallback(() => {
+    if (doc.expanded) {
+      doc.collapse();
+    } else {
+      doc.expand();
+    }
+  }, [doc]);
+
   // Trying to change CodeMirror editor state when an update "effect" is in
   // progress results in an error which is why we timeout the code mirror update
   // itself.
@@ -296,6 +304,8 @@ const JSONEditor: React.FunctionComponent<JSONEditorProps> = ({
         className={cx(editorStyles, darkMode && editorDarkModeStyles)}
         actionsClassName={actionsGroupStyles}
         completer={completer}
+        onExpand={editing ? undefined : toggleExpandCollapse}
+        expanded={expanded}
       />
       <DocumentList.DocumentEditActionsFooter
         doc={doc}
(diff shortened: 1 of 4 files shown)


gribnoysup:
I'm a bit confused, so just double-checking: in the designs the expand / collapse button is added to the "Document" read-only view (which IIRC was the main user complaint, because JSON editor already has a top level expand button, so this one is just duplicating it more or less), but this PR adds it to the JSON view, there will be a follow-up?

Also in the JSON view in this implementation it aligns a bit weird with the document itself:

image

In the designs the button is aligne by the middle of the first line of the editor:

image

Was there a change in the design that's just not documented?

kraenhansen:
> JSON editor already has a top level expand button

The button on the top level doesn't expand all the child fields.

> should we somehow hide the button in the editor itself in these cases?

> it aligns a bit weird with the document itself

> Was there a change in the design that's just not documented?

I'd defer those decisions to Diancheng, I'll ask for her opinion.

gribnoysup:
> The button on the top level doesn't expand all the child fields.

Sorry, I didn't communicate what's confusing here well, my bad. To clarify: we have two buttons with the same icon on the same line / level in UI doing slightly different things, how do user understands the difference and do they need to? Is there a point in having two?

gribnoysup:
> Here's the user request for what this PR is solving. https://feedback.mongodb.com/forums/924283-compass/suggestions/46586038-expand-all-in-json-view-should-be-brought-back

Yeah, so it seems like we broke the old behavior by switching editors and user wants the existing top level button to expand / collapse all and not just one level (instead of an extra button). It's just a handful of people who upvoted, so having two buttons is probably okay if it looks okay to you from design perspective

kraenhansen:
I've adjusted the offset of the expand/collapse button to match the design.

Screenshot 2024-11-05 at 14 36 47

gribnoysup (approved):

PR #6447, opened 2024-11-05
fix(crud): expandable readonly document COMPASS-4635

kraenhansen:
## Description

This is the final sub-task of COMPASS-4635 adding an "expand all" / "collapse all" button to the read-only view.

Merging this PR will:
- Make the `ReadonlyDocument` expandable, by registering relevant event listeners on the `doc` updating a local state.
- Pass a callback into the `DocumentActionsGroup` component to render the "expand button".
- Add extra gutter width in the `Element` component when used in a `ReadonlyDocument` to avoid conflicts with the field-level expand/collapse icons.

https://github.com/user-attachments/assets/3cbaa1d1-36b2-40d4-8139-e0bbf74dd320

### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-components/src/components/document-list/element.tsx
+++ b/packages/compass-components/src/components/document-list/element.tsx
@@ -301,6 +301,10 @@ const elementSpacer = css({
   flex: 'none',
 });
 
+const readOnlySpacer = css({
+  width: spacing[900],
+});
+
 const elementExpand = css({
   width: spacing[3],
   flex: 'none',
@@ -371,13 +375,15 @@ export const calculateShowMoreToggleOffset = ({
   editable,
   level,
   alignWithNestedExpandIcon,
+  extraGutterWidth = 0,
 }: {
   editable: boolean;
   level: number;
   alignWithNestedExpandIcon: boolean;
+  extraGutterWidth: number | undefined;
 }) => {
   // the base padding that we have on all elements rendered in the document
-  const BASE_PADDING_LEFT = spacing[50];
+  const BASE_PADDING_LEFT = spacing[50] + extraGutterWidth;
   const OFFSET_WHEN_EDITABLE = editable
     ? // space taken by element actions
       spacing[300] +
@@ -402,6 +408,7 @@ export const HadronElement: React.FunctionComponent<{
   onEditStart?: (id: string, field: 'key' | 'value' | 'type') => void;
   lineNumberSize: number;
   onAddElement(el: HadronElementType): void;
+  extraGutterWidth?: number;
 }> = ({
   value: element,
   editable,
@@ -409,6 +416,7 @@ export const HadronElement: React.FunctionComponent<{
   onEditStart,
   lineNumberSize,
   onAddElement,
+  extraGutterWidth,
 }) => {
   const darkMode = useDarkMode();
   const autoFocus = useAutoFocusContext();
@@ -457,8 +465,9 @@ export const HadronElement: React.FunctionComponent<{
         editable,
         level,
         alignWithNestedExpandIcon: true,
+        extraGutterWidth,
       }),
-    [editable, level]
+    [editable, level, extraGutterWidth]
   );
 
   const isValid = key.valid && value.valid;
@@ -566,6 +575,9 @@ export const HadronElement: React.FunctionComponent<{
               ></AddFieldActions>
             </div>
           </div>
+        )}
+        {typeof extraGutterWidth === 'number' && (
+          <div style={{ width: extraGutterWidth }} />
         )}
         <div className={elementSpacer} style={{ width: elementSpacerWidth }}>
           {/* spacer for nested documents */}
--- a/packages/compass-components/src/components/document-list/element.tsx
+++ b/packages/compass-components/src/components/document-list/element.tsx
@@ -363,8 +363,12 @@ const elementKeyDarkMode = css({
   color: palette.gray.light2,
 });
 
-const calculateElementSpacerWidth = (editable: boolean, level: number) => {
-  return (editable ? spacing[100] : 0) + spacing[400] * level;
+const calculateElementSpacerWidth = (
+  editable: boolean,
+  level: number,
+  extra: number
+) => {
+  return (editable ? spacing[100] : 0) + extra + spacing[400] * level;
 };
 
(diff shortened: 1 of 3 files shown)


packages/compass-components/src/components/document-list/element.tsx
```diff
@@ -371,13 +375,15 @@ export const calculateShowMoreToggleOffset = ({
...
+  extraGutterWidth = 0,
 }: {
   editable: boolean;
   level: number;
   alignWithNestedExpandIcon: boolean;
+  extraGutterWidth: number | undefined;
 }) => {
   // the base padding that we have on all elements rendered in the document
-  const BASE_PADDING_LEFT = spacing[50];
+  const BASE_PADDING_LEFT = spacing[50] + extraGutterWidth;
```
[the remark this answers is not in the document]
kraenhansen: The line you're referencing is used only to calculate the offset of the "Show more" link when displaying long lists of values. But you're right something was broken: I wasn't [passing the `extraGutterWidth`](https://github.com/mongodb-js/compass/pull/6447/commits/82a5c8756de2cd2cdfbdd2df73a63eafdcc0d369) through to the child elements of an element.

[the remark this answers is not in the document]
kraenhansen: I've updated the video in the description.

packages/compass-crud/src/components/readonly-document.tsx
```diff
@@ -48,13 +112,32 @@ class ReadonlyDocument extends React.Component<ReadonlyDocumentProps> {
...
    *
    * @returns {Array} The elements.
    */
   renderElements() {
-    return <DocumentList.Document value={this.props.doc} />;
+    return (
+      <DocumentList.Document
+        value={this.props.doc}
+        // Provide extra whitespace for the expand button
+        extraGutterWidth={spacing[900]}
```
gribnoysup, line 138: If we want this to be very flexible in the component interface we probably should note that this value should match exactly the space left by hidden editing action buttons in the editable card, so that in case the editable card gets updated we can find this and update the value

gribnoysup (commented):
Seems like in readonly the extra fields button is now misaligned

image

Still thinking that maybe if we're trying to add a uniform gutter to the whole Document here to everything, there might be a better place for that than manually passing it to all the nested elements like Element, there is probably a spot that doesn't require us to track down the value in multiple places and only do it in one 🤔

kraenhansen:
> Seems like in readonly the extra fields button is now misaligned

My bad again. It seems I haven't got sufficient test data to trigger these edge-cases. I forgot to pass the `extraGutterWidth` to the `calculateShowMoreToggleOffset` call in the `HadronDocument` component: https://github.com/mongodb-js/compass/pull/6447/commits/5278c0ce15fd21e5ab12c41e1ddd000bdac8b441

packages/compass-crud/src/components/readonly-document.tsx
kraenhansen: I don't think there a requirement that this offset match any other state as such. Just that it's enough for the expand button to not overlay the icons of the individual elements.

packages/compass-components/src/components/document-list/element.tsx
```diff
@@ -567,6 +572,9 @@ export const HadronElement: React.FunctionComponent<{
             </div>
           </div>
         )}
+        {typeof extraGutterWidth === 'number' && (
+          <div style={{ width: extraGutterWidth }} />
+        )}
         <div className={elementSpacer} style={{ width: elementSpacerWidth }}>
```
gribnoysup, line 578: Nit: this element is known to be a performance bottleneck due to the sheer amount of dom elements it might end up rendering on the page (yeah, even if they are just empty divs), so one thing I'd suggest here is to combine your extra width into the existing element:

```suggestion
        <div className={elementSpacer} style={{ width: elementSpacerWidth + (extraGutterWidth ?? 0) }}>
```

(or you can include it in the existing calculate extra width function above)

It's less of an issue now that we're virtualizing the list of these document cards in most cases, but doesn't hurt to be a bit more vigilant about that in this particular element.

Also if you want to keep the extra element, you might want to add the same elementSpacer class to it to make sure that it won't get the default flex shrink value that would allow it to be resized inside of this flex parent

gribnoysup (approved):

kraenhansen:
I've updated the code to use the existing element spacer 👍

Here's a screenshot of the "show more" links on fields and array items:

Screenshot 2024-11-06 at 13 35 09
Screenshot 2024-11-06 at 13 45 35

packages/compass-components/src/components/document-list/element.tsx
```diff
@@ -378,23 +382,20 @@ export const calculateShowMoreToggleOffset = ({
...
+    extraGutterWidth
+  );
+  const editableOffset = editable
     ? // space taken by element actions
       spacing[300] +
       // space and margin taken by line number element
       spacing[400] +
-      spacing[100] +
-      // element spacer width that we render
-      calculateElementSpacerWidth(editable, level)
```
kraenhansen, line 388: Since this takes the `editable` as an argument and displaying the spacer isn't conditional on the value of `editable` I've moved it out of the `OFFSET_WHEN_EDITABLE` (now `editableOffset`).

packages/compass-components/src/components/document-list/element.tsx
```diff
@@ -378,23 +382,20 @@ export const calculateShowMoreToggleOffset = ({
   alignWithNestedExpandIcon: boolean;
   extraGutterWidth: number | undefined;
 }) => {
-  // the base padding that we have on all elements rendered in the document
-  const BASE_PADDING_LEFT = spacing[50] + extraGutterWidth;
```
kraenhansen: I've updated the casing on these local variables, since they are not global constants.

packages/compass-components/src/components/document-list/element.tsx
kraenhansen: I've updated the code to use the existing spacer and fixed what I believe was a bug in the calculation of the spacer width 🤞

gribnoysup (approved):

PR #6449, opened 2024-11-06
chore(connections-navigation): add connect button to connections in sidebar COMPASS-8381

kraenhansen:
## Description

Merging this PR will:
- Add a connect action, represented as a `Button` to every disconnected connection item in the sidebar.
- Disable the default action (which was connecting) when clicking the connection in the sidebar as well as removing the visual cue for the interaction (the `cursor: 'pointer'`).
- Update E2E tests to connect by clicking the new button.
- Invert a test which expected clicking an inactive connection to connect.

https://github.com/user-attachments/assets/4743b209-1aed-463e-be3e-c3fe478fe2fa

### Checklist
- [x] Update e2e tests to click the connect button
- [ ] Fix the breaking `@mongodb-js/compass-sidebar` tests

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Open Questions

- Given the intent to eventually extend this "Connect" button with a drop-down, I wonder if the existing `ItemAction` abstraction is the right for this feature or I should introduce another prop to inject a custom component into the `NavigationBaseItem`.

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-connections-navigation/src/base-navigation-item.tsx
+++ b/packages/compass-connections-navigation/src/base-navigation-item.tsx
@@ -17,6 +17,7 @@ type NavigationBaseItemProps = {
   isExpandDisabled: boolean;
   isExpanded: boolean;
   isFocused: boolean;
+  hasDefaultAction: boolean;
   icon: React.ReactNode;
   style: React.CSSProperties;
 
@@ -37,7 +38,6 @@ const menuStyles = css({
 });
 
 const itemContainerStyles = css({
-  cursor: 'pointer',
   color: 'var(--item-color)',
   backgroundColor: 'var(--item-bg-color)',
   '&[data-is-active="true"] .item-wrapper': {
@@ -51,6 +51,10 @@ const itemContainerStyles = css({
   svg: {
     flexShrink: 0,
   },
+});
+
+const itemContainerWithActionStyles = css({
+  cursor: 'pointer',
 });
 
 const itemWrapperStyles = css({
@@ -93,6 +97,7 @@ export const NavigationBaseItem: React.FC<NavigationBaseItemProps> = ({
   isExpandDisabled,
   isExpanded,
   isFocused,
+  hasDefaultAction,
   onExpand,
   children,
 }) => {
@@ -100,7 +105,9 @@ export const NavigationBaseItem: React.FC<NavigationBaseItemProps> = ({
   return (
     <div
       data-testid="base-navigation-item"
-      className={itemContainerStyles}
+      className={cx(itemContainerStyles, {
+        [itemContainerWithActionStyles]: hasDefaultAction,
+      })}
       {...hoverProps}
       {...dataAttributes}
     >
--- a/packages/compass-connections-navigation/src/navigation-item.tsx
+++ b/packages/compass-connections-navigation/src/navigation-item.tsx
@@ -251,6 +251,9 @@ export function NavigationItem({
           isActive={isActive}
           isFocused={isFocused}
           isExpanded={!!item.isExpanded}
+          hasDefaultAction={
+            item.type !== 'connection' || item.connectionStatus === 'connected'
+          }
           icon={itemIcon}
           name={item.name}
           style={style}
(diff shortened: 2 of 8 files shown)


packages/compass-connections-navigation/src/base-navigation-item.tsx
```diff
@@ -93,14 +97,17 @@ export const NavigationBaseItem: React.FC<NavigationBaseItemProps> = ({
...
   onExpand,
   children,
 }) => {
   const [hoverProps, isHovered] = useHoverState();
   return (
     <div
       data-testid="base-navigation-item"
-      className={itemContainerStyles}
+      className={cx(itemContainerStyles, {
+        [itemContainerWithActionStyles]: hasDefaultAction,
```
kraenhansen, line 109: Ideally this would be co-located with or derived from the implementation in `onDefaultAction` in `connections-navigation-tree.tsx`, but that is not trivial because of the virtualization.

packages/compass-connections-navigation/src/navigation-item.tsx
```diff
@@ -251,6 +251,9 @@ export function NavigationItem({
           isActive={isActive}
           isFocused={isFocused}
           isExpanded={!!item.isExpanded}
+          hasDefaultAction={
+            item.type !== 'connection' || item.connectionStatus === 'connected'
```
kraenhansen, line 255: Ideally this would be co-located with or derived from the implementation in `onDefaultAction` in `connections-navigation-tree.tsx`, but that is not trivial because of the virtualization.

himanshusinghs (approved):
LGTM 🚀 

Something else I noticed, definitely not urgent for now. I feel like the ItemActionControls component interface need a little revisit (maybe from design POV) because one combination will render the UI un-usable, having collapseAfter set to anything >1 and one or more of the rendered actions having expandedPresentation of button. I think what I am trying to suggest is having some typecheck (if possible) to only allow combination that make sense from rendering POV but yea that depends on what design is thinks is feasible from rendering POV.

himanshusinghs:
I ran the branch locally and have a two clarifying questions:
1. I suppose we don't even render the connections in the disconnected mode in MMS so Betsy's  concern about not showing the button on compass-web should be addressed by that but can you please confirm this?
2. Do we want to cross-confirm with design whether the right-arrow icon still make sense when disconnected? It definitely made sense earlier as it signified that the rows could be expanded (even when disconnected), maybe not so much anymore?

gribnoysup:
> I suppose we don't even render the connections in the disconnected mode

We do, we don't automatically connect to every cluster in the list

lerouxb (approved):

betsybutton (approved):
LGTM from a UX perspective!

kraenhansen:
> It definitely made sense earlier as it signified that the rows could be expanded (even when disconnected), maybe not so much anymore?

I think that's a valid point.
- if removed completely, it will introduce some "jumping" of elements as the connection gets established
- if left blank, it will introduce some strange whitespace

Perhaps a "Disconnect" icon would be more appropriate? As this is out of scope for the original design, I suggest leaving that up for another (quick and fast) PR to avoid holding back this one.

bsradcliffe:
I would like to preserve the chevrons. They indicate that a node can be expanded, but it's simply disabled in the collapsed state. This is a fairly standard convention in explorer trees in GUIs. I wouldn't want to introduce any inconsistent leading edges/jagged jumping between the root level nodes.

Disconnect is something that is needs to be "buried" for lack of a better word. It shouldn't be immediately accessible at the top level since a potential misclick can terminate the connection and close all of the associated connection's tabs.

kraenhansen:
> Disconnect is something that is needs to be "buried" for lack of a better word. It shouldn't be immediately accessible at the top level since a potential misclick can terminate the connection and close all of the associated connection's tabs.

If this is referring to my last suggestion, I fear I didn't make myself clear:
To clarify, my suggestion was to simply replace the "CaretRight" with the "Disconnect" icon from the Leafygreen icons, for the already disconnected connections in the sidebar. I.e. I'm not suggesting adding any new interaction (which could be misclicked).

bsradcliffe:
Gotcha. Even so, as we only utilize that left glyph for chevrons, I would still stay no to the proposal. Every left-most glyph should be a caret of some sort where child nodes can exist beneath it.

bsradcliffe:
Question here: in the video posted, why is it a single connect CTA and not the split button as the designs show?

CleanShot 2024-11-13 at 09 18 07@2x

PR #6472, opened 2024-11-11
fix(compass-global-writes): show error on all invalid statuses COMPASS-8451

djechlin-mongodb:
### Checklist

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-global-writes/src/plugin-title.tsx
```diff
@@ -49,7 +55,7 @@ export const PluginTitle = ({ showWarning }: { showWarning: boolean }) => {
               }}
             >
               <Icon
-                glyph="ImportantWithCircle"
+                glyph={showWarning ? 'Warning' : 'ImportantWithCircle'}
```
paula-stacho, line 58: The other way around :) I know it's confusing, but we want it to match the banner icon
Screenshot 2024-11-12 at 18 10 07
Screenshot 2024-11-12 at 18 15 34

packages/compass-global-writes/src/plugin-title.tsx
```diff
@@ -49,7 +55,7 @@ export const PluginTitle = ({ showWarning }: { showWarning: boolean }) => {
...
                 aria-label="warning"
```
paula-stacho, line 59: Could we make this message and the color of the icon variable as well?

djechlin-mongodb: Is this what you mean?

Error: 
image

Warning:
image

djechlin-mongodb: Code quality question, it seems weird that the properties of the Icon are so coupled to each other, should we just pull out a WarningIcon and ErrorIcon or something pre-parameterized? Does this pattern come up a lot in compass?

paula-stacho: You're right, with this many differences it makes sense to move the icons into their own function components so that the main render is more readable. And yes that's a common pattern, not only in compass. If your question is more towards reusable components in Compass, check the `compass-components` package. But I wouldn't put these in there as the warning = ImportantWithcircle and error = Warning is a bit odd. I'd just keep them in the same file.

djechlin-mongodb:
Reviewed with Paula this morning - we think this is LGTM once Paula's refactor PR is in. And that PR makes this one easier to  merge so it's better to wait.

paula-stacho (approved):
The last detail - the colors are the other way around 😅 Otherwise looks great!

PR #6495, opened 2024-11-14
chore(components): refactor `InteractivePopover` to take a type argument for the trigger element

kraenhansen:
## Description

As a follow-up to this https://github.com/mongodb-js/compass/pull/6486#discussion_r1841958522, I suggest adding a type argument to the `InteractivePopover` to specify the concrete `HTMLElement` used when declaring the type of the `ref` passed through the `trigger` callback.

We can't just pass `React.Ref<HTMLElement>` as that (as an example) isn't possible to pass via `ref` to a `<button />`:

```tsx
const ref = React.useRef<HTMLElement>(null);
return <button ref={ref} />;
```

☝️ yields this TS error related to the `ref` prop:

```
Type 'RefObject<HTMLElement>' is not assignable to type 'LegacyRef<HTMLButtonElement> | undefined'.
  Type 'RefObject<HTMLElement>' is not assignable to type 'RefObject<HTMLButtonElement>'.
    Type 'HTMLElement' is missing the following properties from type 'HTMLButtonElement': disabled, form, formAction, formEnctype, and 15 more.ts(2322)
```

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-components/src/components/interactive-popover.tsx
+++ b/packages/compass-components/src/components/interactive-popover.tsx
@@ -41,12 +41,12 @@ const closeButtonStyles = css({
   right: spacing[2],
 });
 
-type InteractivePopoverProps = {
+type InteractivePopoverProps<TriggerElement extends HTMLElement> = {
   className?: string;
   children: React.ReactNode;
   trigger: (triggerProps: {
-    onClick: React.MouseEventHandler<HTMLButtonElement>;
-    ref: React.LegacyRef<HTMLButtonElement>;
+    onClick: React.MouseEventHandler<TriggerElement>;
+    ref: React.Ref<TriggerElement>;
     children: React.ReactNode;
   }) => React.ReactElement;
   hideCloseButton?: boolean;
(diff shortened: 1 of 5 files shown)


packages/compass-components/src/components/interactive-popover.tsx
```diff
@@ -42,12 +42,12 @@ const closeButtonStyles = css({
   right: spacing[2],
 });
 
-type InteractivePopoverProps = {
+type InteractivePopoverProps<TriggerElement extends HTMLElement> = {
```
kraenhansen, line 45: This could be named `T` but it felt a bit too vague 🤔

packages/compass-sidebar/src/components/connections-filter-popover.tsx
```diff
@@ -102,7 +102,7 @@ export default function ConnectionsFilterPopover({
               onMouseLeave={handleButtonMouseLeave}
               active={open}
               aria-label="Filter connections"
-              ref={ref as React.Ref<unknown>}
+              ref={ref}
```
kraenhansen, line 105: I believe it's possible to pass the `ref` here without having to provide an explicit type parameter to the `InteractivePopover` above, is possible because the `IconButton` use the `forwardRef` utility internally.

gribnoysup (approved):
Neat, thanks for the follow up to clean this up!

Anemy (approved):

PR #6496, opened 2024-11-14
chore(connection-form): move connection-form-modal, rename connection-form-modal-actions COMPASS-8471

lerouxb:
This moves connection-form-modal to compass-connections and renames connection-form-modal-actions to just connection-form-actions since it is used direcly in connection-form, not connection-form-modal.


packages/connection-form/src/components/connection-form.tsx
```diff
@@ -622,7 +622,7 @@ function ConnectionForm({
               : formFooterBorderLightModeStyles
           )}
         >
-          <ConnectionFormModalActions
+          <ConnectionFormActions
```
lerouxb, line 625: ☝️ See this is the actions for connection-form, but really for connection-form-modal specifically.

lerouxb:
apparently not this, then.

PR #6528, opened 2024-11-25
chore(preferences-model): enable updating the cloudFeatureRolloutAccess preference for web COMPASS-8572

Anemy:
COMPASS-8572

This updates the runtime memory preferences for Compass web, `CompassWebPreferencesAccess`, to allow updating of the `cloudFeatureRolloutAccess` preference. This is updated when the hello endpoint returns that the feature is enabled.


ruchitharajaghatta (approved):
makes sense and looks good to me!

PR #6531, opened 2024-11-26
chore(data-service): disable utf8 validation when retrieving metadata compass needs in order to function COMPASS-8517

lerouxb:
There's an old server bug where it could truncate content in the middle of a UTF-8 codepoint: https://jira.mongodb.org/browse/SERVER-24007

This can make compass unusable if it happens for one of the commands that compass fires off when you connect or browse to a collection. It is reasonably safe to just turn off utf8 validation in those cases. We still leave it on when reading/writing documents and therefore users aren't at risk of writing back broken data to their database.

### DIFF
--- a/packages/data-service/src/data-service.ts
+++ b/packages/data-service/src/data-service.ts
@@ -1168,7 +1168,11 @@ class DataServiceImpl extends WithLogContext implements DataService {
   @op(mongoLogId(1_001_000_031))
   async killOp(id: number, comment?: string): Promise<Document> {
     const db = this._database('admin', 'META');
-    return runCommand(db, { killOp: 1, id, comment });
+    return runCommand(
+      db,
+      { killOp: 1, id, comment },
+      { enableUtf8Validation: false }
+    );
   }
 
   isWritable(): boolean {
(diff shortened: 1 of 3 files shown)


packages/data-service/src/run-command.ts
```diff
@@ -334,7 +334,7 @@ export const runCommand: RunCommand = (
 
   return db.command(
     { ...spec },
-    { readPreference, ...options }
+    { readPreference, enableUtf8Validation: false, ...options }
```
lerouxb, line 337: This disables utf8 validation for runCommand by default. You can still re-enabe that by specifying it as an option. I had a look through all the places where we use runCommand and it seems to overlap perfectly with all the places where we want to disable the validation. ie. all the places where we get metadata, none of the places where we read/update documents.

addaleax: Yeah ... are you sure this is a good idea? It seems very easy to accidentally enableUtf8Validation when that is _not_ desirable this way

packages/data-service/src/data-service.ts
```diff
@@ -1226,7 +1229,7 @@ class DataServiceImpl extends WithLogContext implements DataService {
     try {
       const cursor = this._database(databaseName, 'CRUD').listCollections(
         filter,
-        { nameOnly }
+        { nameOnly, enableUtf8Validation: false }
```
addaleax, line 1232: I don't think we can do this for collection names, we would end up accessing a collection name that doesn't exist this way

packages/data-service/src/run-command.ts
lerouxb: OK I removed it from here and then added it to every runCommand individually. Then we can argue about them on a case by case basis :)

packages/data-service/src/data-service.ts
```diff
@@ -1083,49 +1083,52 @@ class DataServiceImpl extends WithLogContext implements DataService {
...
+                    else: 0,
                   },
-                  else: 0,
                 },
               },
             },
-          },
-        ])
+          ],
+          { enableUtf8Validation: false }
```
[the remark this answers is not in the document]
lerouxb: Added this one here because the user's error log file contained a lot of aggregate failed commands and we have a hunch it might be this one.

addaleax (approved):

PR #6541, opened 2024-11-29
fix(compass-components): don't cut off list view value editor tooltips COMPASS-8581

lerouxb:
I don't know why the portal was turned off there. The expand/collapse buttons are very similar in that they have tooltips and they use portals, so it should be safe to do that here too?

I noticed that the key editor's error tooltip doesn't work at all. Can't immediately tell why. I filed a [ticket](https://jira.mongodb.org/browse/COMPASS-8586). The only possible error there seems to be a duplicate key. It does go red to show that it is an error, so the validation does work. The tooltip just doesn't trigger.


Anemy (approved):
These were without portal as they would teleport to the top left of the screen when a user had one open and then navigated to another tab. It may have been an old tooltip issue that was fixed along the way.
Related pr https://github.com/mongodb-js/compass/pull/3574

PR #6548, opened 2024-12-03
fix(components): fix element key editor tooltip and revert not updating validity COMPASS-8586

Anemy:
[COMPASS-8586](https://jira.mongodb.org/browse/COMPASS-8586)

- Fixes the duplicate key error tooltip not showing.
- Fixes the `revert` action not reverting the validity of a field if it updates the field name making it not a duplicate.
- Shows a red error around values that are invalid similar to what we do with keys.

**Error tooltip that is now shown**
Screenshot 2024-12-03 at 2 03 29 PM

**Error validity state of field**
| before | after |
| - | - |
| Screenshot 2024-12-03 at 2 26 17 PM | Screenshot 2024-12-03 at 2 02 04 PM |

### DIFF
--- a/packages/compass-components/src/components/document-list/element-editors.tsx
+++ b/packages/compass-components/src/components/document-list/element-editors.tsx
@@ -83,6 +83,7 @@ export const KeyEditor: React.FunctionComponent<{
           usePortal={false}
           trigger={({
             className,
+            children,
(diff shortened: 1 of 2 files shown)


lerouxb (approved):
nice! and thanks

PR #6552, opened 2024-12-06
chore(compass-components): refactor `ItemActionControls`

kraenhansen:
## Description

As preparations for future changes to this code, I took some time to refactor the `item-action-controls.tsx` into a directory of multiple files.

### Checklist

## Motivation and Context
- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-components/src/components/actions/action-glyph.tsx
```diff
@@ -0,0 +1,26 @@
+// As we are using this component to render icon in MenuItem,
+// and it does cloneElement on glyph, here we are accepting all the
```
lerouxb, line 2: all the? :)

packages/compass-components/src/components/actions/types.ts
```diff
@@ -0,0 +1,37 @@
+import type { glyphs } from '@leafygreen-ui/icon';
+
+import type { ItemActionButtonSize } from './constants';
+
+export type ItemBase<Action extends string> = {
```
lerouxb, line 5: nit: I don't think as a rule we usually have the types separated into their own file.

lerouxb (approved):

packages/compass-components/src/components/actions/types.ts
kraenhansen: The main reason for this file is because two mutually independent files (`item-action-menu.tsx` and `item-action-group.tsx`) have declarations which union this type. I don't want one to be considered more important by the other importing from it and I don't want to hoist it as that'll introduce a circular import.

PR #6553, opened 2024-12-06
fix(serverstats): update top chart write lock percentage of load visual COMPASS-8604

Anemy:
COMPASS-8604
Small visual bug on how we show the % of write on the load in the top / hottest collections chart. Javascript related (typescript would catch this sort of thing).
Not easily seen which is probably why it's never been fixed. This bug has probably been in there since it was written 5+ years ago lol.

This is the chart for reference:
Screenshot 2024-12-06 at 12 58 58 PM


paula-stacho (approved):
wow, good catch! this really needs a rewrite 🙈 I'd maybe even rename the variables to say 'Write' and 'Read' so that they're not that easy to confuse

PR #6558, opened 2024-12-10
fix(e2e): actually test the packaged app when running npm run test-packaged

lerouxb:
Edit: This seems to only affect npm run test-packaged, not npm run test-packaged-ci which is what we run in CI. So it was fine in CI, just not locally. Depending on which command you run locally.

Since the refactor a couple of months ago we've been passing too many `--` to the app and the arg parser was missing `--test-packaged-app` so it would re-build compass every time.

You can test this by running the E2E tests against the locally installed beta, dev or GA like this:

```
DEBUG=compass-e2e-tests*,electron*,hadron*,mongo* COMPASS_APP_NAME="MongoDB Compass" COMPASS_APP_PATH="/Applications/MongoDB Compass.app"  npm run test-packaged
```


packages/compass-e2e-tests/helpers/test-runner-context.ts
```diff
@@ -279,6 +279,8 @@ if ('then' in parsedArgs && typeof parsedArgs.then === 'function') {
 export const context = parsedArgs as CommonParsedArgs &
   Partial<DesktopParsedArgs & WebParsedArgs>;
 
+debug('context:', context);
```
lerouxb, line 282: Handy for seeing if it actually picked up all the command-line args it should have.

Anemy: Screenshot 2024-12-10 at 12 23 46 PM

Nice add

gribnoysup: It's printed already a few lines down with secrets removed from the list:

```
const contextForPrinting = Object.fromEntries(
  Object.entries(context).map(([k, v]) => {
    return [k, /password/i.test(k) ? '<secret>' : v];
  })
);
debug('Running tests with the following arguments:', contextForPrinting);
```

Is this print doing something different?

lerouxb: Sorry didn't see it. We can remove it again.

lerouxb: I just merged it https://github.com/mongodb-js/compass/pull/6561

PR #6559, opened 2024-12-11
chore(e2e): webdriverio 9

lerouxb:
Basically to take in https://github.com/puppeteer/puppeteer/issues/13390 so that the firefox download will work.


packages/compass-e2e-tests/helpers/commands/add-collection.ts
```diff
@@ -37,7 +37,7 @@ export async function addCollection(
   collectionOptions?: AddCollectionOptions,
   screenshotPath?: string
 ): Promise<void> {
-  const createModalElement = await browser.$(Selectors.CreateCollectionModal);
+  const createModalElement = browser.$(Selectors.CreateCollectionModal);
```
lerouxb, line 40: browser.$() now returns a chainable promise element which isn't a promise itself anymore.

packages/compass-e2e-tests/helpers/commands/connect-form.ts
```diff
@@ -319,8 +317,8 @@ async function getCheckedRadioValue(
   browser: CompassBrowser,
   selector: string
 ): Promise<string | null> {
-  const elements = await browser.$$(selector);
-  for (const element of elements) {
+  const elements = browser.$$(selector);
+  for await (const element of elements) {
```
lerouxb, line 321: browser.$$ returns a chainable element promise array so a) it isn't a promise itself anymore and b) we have to async iterate over the result.

packages/compass-e2e-tests/tests/auto-connect.test.ts
```diff
@@ -287,22 +287,23 @@ describe('Automatically connecting from the command line', function () {
...
         const type = await currentActiveTab.getAttribute('data-type');
         return type === 'Welcome';
       });
 
       // no toasts to signify that anything is connecting
-      const toasts = await browser.$(Selectors.LGToastContainer).$$('div');
-      expect(toasts).to.have.lengthOf(0);
+      const numToasts = await browser.$(Selectors.LGToastContainer).$$('div')
+        .length;
+      expect(numToasts).to.equal(0);
```
lerouxb, line 300: The change to the result of $$ kinda breaks the nice to.have.lengthOf. Checking the length (which we have to await) is basically the same until the test fails. The error is now 1 != 0 rather than expected [a, b, c, d] to have length of 0. Which is a bit less nice but I just found this fix easiest at scale.

packages/compass-e2e-tests/helpers/commands/click-visible.ts
```diff
@@ -9,7 +9,7 @@ interface ClickOptions {
 
 export async function clickVisible(
   browser: CompassBrowser,
-  selector: string | ChainablePromiseElement<WebdriverIO.Element>,
```
lerouxb, line 12: ChainablePromiseElement is not generic anymore

packages/compass-e2e-tests/helpers/commands/wait-for-aria-disabled.ts
```diff
@@ -3,20 +3,18 @@ import type { CompassBrowser } from '../compass-browser';
 
 export async function waitForAriaDisabled(
   browser: CompassBrowser,
-  selector:
-    | string
-    | ChainablePromiseElement<WebdriverIO.Element>
-    | WebdriverIO.Element,
```
lerouxb, line 9: No need to also allow WebdriverIO.Element for convenience because you can't ordinarily get that type using .$() lookups.

packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -178,7 +176,7 @@ export class Compass {
     // TS infers the type of `message` correctly here, which would conflict with
     // what we get from `import type { ConsoleMessage } from 'puppeteer'`, so we
     // leave out an explicit type annotation.
-    page.on('console', (message) => {
+    page.on('console', (message: any) => {
```
lerouxb, line 179: Same here.

packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -646,6 +644,11 @@ async function startCompassElectron(
         binary: maybeWrappedBinary,
         args: chromeArgs,
       },
+      // from https://github.com/webdriverio-community/wdio-electron-service/blob/32457f60382cb4970c37c7f0a19f2907aaa32443/packages/wdio-electron-service/src/launcher.ts#L102
+      'wdio:enforceWebDriverClassic': true,
+    },
+    'wdio:chromedriverOptions': {
+      // TODO: enable logging so we don't have to debug things blindly
```
lerouxb, line 651: Maybe I'm just really bad at finding undocumented options, but I would be eternally grateful to whoever gets us more logs because I just lost hours of the app silently and cryptically stopping on startup before ever showing a window.

packages/compass-e2e-tests/helpers/commands/wait-for-aria-disabled.ts
```diff
@@ -3,24 +3,22 @@ import type { CompassBrowser } from '../compass-browser';
...
   const expectedValue = isDisabled ? 'true' : 'false';
 
   await browser.waitUntil(
     async () => {
+      const element = getElement();
       return (await element.getAttribute('aria-disabled')) === expectedValue;
     },
     {
-      timeoutMsg: `Expected ${selector} to have attribute "aria-disabled" equal to "${expectedValue}"`,
+      timeoutMsg: `Expected element to have attribute "aria-disabled" equal to "${expectedValue}"`,
```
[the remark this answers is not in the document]
lerouxb: Apparently we're never using it with a string 😆 It is always already a chainable promise element in all the places we use this. (I went to go look to see if it is worth branching on string type.)

PR #6571, opened 2024-12-17
feat(e2e): Smoke tests task & initial entrypoint COMPASS-8531

lerouxb:
This is also documented by yargs and in comments, but for convenience I'm adding a summary here.

For running locally and testing with compass you packaged yourself:

```shell
HADRON_DISTRIBUTION=compass IS_OSX=true npm run smoketest -w compass-e2e-tests -- --skip-download
```

For running locally and testing against compass that was packaged by evergreen for a pull request:

```shell
# DEV_VERSION_IDENTIFIER should be printed out by the package task and in the
# format yy-mm-dd-dev.hms (leave it out for testing with the version specified
# in packages/compass/package.json, ie. for locally packaged or beta or stable)
export DEV_VERSION_IDENTIFIER=24.12.14-dev.5046

# EVERGREEN_BUCKET_NAME is currently always the same
export EVERGREEN_BUCKET_NAME=mciuploads

# EVERGREEN_BUCKET_KEY_PREFIX should be printed out by the package task and is
# in the format ${project}/${revision}_${revision_order_id}
export EVERGREEN_BUCKET_KEY_PREFIX=10gen-compass-main/4754c3d04086f2473560d1d8b14f7d83f68e1017_nikola.irinchev_586

# HADRON_DISTRIBUTION can be compass, compass-readonly or compass-isolated
export HADRON_DISTRIBUTION=compass

# ARCH will pick up platform.arch if not set. Useful for setting it to something
# else if you want to download/test different packages
export ARCH=arm64

# IS_OSX could also be IS_UBUNTU, IS_RHEL or IS_WINDOWS. That's to mirror our
# existing variables used in evergreen. You have to specify one of them
export IS_OSX=true 

npm run smoketest -w compass-e2e-tests
```


.evergreen/buildvariants-and-tasks.in.yml
```diff
@@ -60,6 +60,40 @@ const PACKAGE_BUILD_VARIANTS = [
...
 
+const SMOKETEST_BUILD_VARIANTS = [
+  {
+    name: 'smoketest-ubuntu',
+    display_name: 'Smoketest Ubuntu',
+    run_on: 'ubuntu2004-large',
+    depends_on: 'package-ubuntu',
+  },
+  
+//  {
```
lerouxb, line 71: For now I have most platforms commented out because this is a lot of new tasks that will do nothing for the moment.

.evergreen/buildvariants-and-tasks.in.yml
```diff
@@ -424,6 +471,22 @@ tasks:
       - func: save-all-artifacts
         vars:
           compass_distribution: <%= distribution %>
+  <% } %>
+
+<% for (const distribution of ['compass']) { %>
```
lerouxb, line 476: Also not looping through COMPASS_DISTRIBUTIONS here yet.

.evergreen/preinstall.sh
```diff
@@ -12,6 +12,11 @@ echo "NPM_VERSION: $NPM_VERSION"
 echo "APPDATA: $APPDATA"
 echo "PATH: $PATH"
 
+# these are super useful if you want to run the smoke tests locally
```
lerouxb, line 15: We're going to need these when we want to test with the files produced by a PR or beta or ga build, so I'm just adding it next to other useful/essential variables being output. They aren't secrets so that's fine.

packages/compass-e2e-tests/smoke-test.ts
```diff
@@ -0,0 +1,338 @@
...
+  .check((argv) => {
+    if (!argv.skipDownload) {
+      if (!(argv.bucketName && argv.bucketKeyPrefix)) {
+        throw new Error(
+          'Either supply EVERGREEN_BUCKET_NAME and EVERGREEN_BUCKET_KEY_PREFIX or specify --skip-download'
+        );
+      }
+    }
+
+    if (!(argv.isWindows || argv.isOSX || argv.isUbuntu || argv.isRHEL)) {
```
lerouxb, line 76: We could detect these as a default, but I'm not sure we want to duplicate [the logic](https://github.com/mongodb-js/compass/blob/c8f0a6b3869fba35ab8ad80e2f3095e2eab09224/.evergreen/print-compass-env.sh#L7-L28)?

I spent way too much time thinking about this and going back and forth.

packages/compass-e2e-tests/tsconfig.json
```diff
@@ -1,7 +1,4 @@
 {
   "extends": "@mongodb-js/tsconfig-compass/tsconfig.common.json",
-  "compilerOptions": {
```
lerouxb, line 3: Driveby from uncommitted changes.

packages/hadron-build/commands/download.js
```diff
@@ -43,12 +42,18 @@ const handler = function handler(argv) {
...
   });
 
-  Promise.all(downloads).catch(abortIfError);
-};
+  return Promise.all(downloads)
+}
+
+const handler = function handler(argv) {
+  cli.argv = argv;
+  run(argv).catch(abortIfError);
```
lerouxb, line 50: In the end I'm not using download because as is it would download all the packages and therefore mean we have to depend on every package task rather than just the relevant one for the platform.

I could modify download more to filter but we already have the info we need from hadron-build info anyway so I just download straight in the script.

But I think the change here is harmless and a very minor improvement anyway.

packages/compass-e2e-tests/smoke-test.ts
```diff
@@ -0,0 +1,338 @@
...
+        await fs.mkdir(path.dirname(filepath), { recursive: true });
+        const url = `https://${context.bucketName}.s3.amazonaws.com/${context.bucketKeyPrefix}/${name}`;
+        console.log(url);
+        return downloadFile(url, filepath);
+      })
+    );
+    verifyPackagesExist(packages);
+  }
+
+  // TODO(COMPASS-8533): extract or install each package and then test the Compass binary
```
lerouxb, line 163: The rest of the epic goes here :)

packages/compass-e2e-tests/smoke-test.ts
```diff
@@ -0,0 +1,338 @@
...
+    if (!buildInfoIsUbuntu(buildInfo)) {
+      throw new Error('missing ubuntu package keys');
+    }
+    names = ubuntuFilenameKeys.map((key) => buildInfo[key]);
+  }
+
+  const extension = config.extension;
+
+  return names
+    .filter((name) => !extension || name.endsWith(extension))
```
lerouxb, line 284: By optionally filtering the files we'll test we can easily run this on just one file at a time during development and we can also change the evergreen config to add one task per file if we wanted to at some point.

packages/compass-e2e-tests/smoke-test.ts
```diff
@@ -0,0 +1,338 @@
...
+
+        // handle redirects
+        if (code > 300 && code < 400 && !!response.headers.location) {
+          return resolve(downloadFile(response.headers.location, targetFile));
+        }
+
+        // save the file to disk
+        const fileWriter = createWriteStream(targetFile).on('finish', () => {
+          resolve();
+        });
```
kraenhansen, line 311: Perhaps you'd want to register an 'error' listener here as well (calling `reject`), in case the write stream fails?

packages/compass-e2e-tests/smoke-test.ts
```diff
@@ -0,0 +1,338 @@
...
+  .option('arch', {
+    type: 'string',
+    choices: ['x64', 'arm64'],
+    demandOption: true,
+    default: () => {
+      if (process.env.ARCH) {
+        return process.env.ARCH;
+      }
+      return process.arch;
+    },
```
kraenhansen, line 56:
```suggestion
    default: () => process.env.ARCH ?? process.arch,
```

packages/compass-e2e-tests/smoke-test.ts
kraenhansen: Should we also check to ensure these are mutually exclusive?

packages/compass-e2e-tests/smoke-test.ts
```diff
@@ -0,0 +1,338 @@
...
+    await Promise.all(
+      packages.map(async ({ name, filepath }) => {
+        await fs.mkdir(path.dirname(filepath), { recursive: true });
+        const url = `https://${context.bucketName}.s3.amazonaws.com/${context.bucketKeyPrefix}/${name}`;
+        console.log(url);
+        return downloadFile(url, filepath);
+      })
+    );
+    verifyPackagesExist(packages);
+  }
```
kraenhansen, line 161:
```suggestion
  if (!context.skipDownload) {
    await Promise.all(
      packages.map(async ({ name, filepath }) => {
        await fs.mkdir(path.dirname(filepath), { recursive: true });
        const url = `https://${context.bucketName}.s3.amazonaws.com/${context.bucketKeyPrefix}/${name}`;
        console.log(url);
        return downloadFile(url, filepath);
      })
    );
  }
  verifyPackagesExist(packages);
```

kraenhansen (approved):
I've got a few minor suggestions, but I like the approach.

The only bigger thing is: I have a feeling "smoke test" is too generic for what we're doing with this project. I mean, this will specifically test "installs" and "updating", right?

packages/compass-e2e-tests/smoke-test.ts
lerouxb: I've been thinking about that. Either make them mutually exclusive or allow you to "OR" them together. Then you can use this to download everything if you want to or at least test multiple .zip files at a time. Sure it will break later, but much of the code could run on mac for testing.

packages/compass-e2e-tests/smoke-test.ts
lerouxb: I suppose that won't hurt. I find streams such a footgun. There really should be builtin utilities somewhere for these cases. We use the `download` npm package in a few other places but it hasn't been published to in 5 years, supports no ESM, no TypeScript.. with no obvious perfect alternative I could quickly find.

kraenhansen: How about using the `finished` function on the builtin `node:stream/promises` module?

https://nodejs.org/api/stream.html#streamfinishedstream-options

kraenhansen: Or perhaps [this answer](https://stackoverflow.com/a/77210219) on stack-overflow:

```ts
import { writeFile } from 'node:fs/promises';
import { Readable } from 'node:stream';

const response = await fetch(url);
const stream = Readable.fromWeb(response.body);
await writeFile(targetFile, stream);
```

addaleax: Late to the party, but the best-practice generic way here (i.e. not relying on this particularity of the `writeFile()` API) would be to use `stream.promises.pipeline(response, createWriteStream(targetFile));` – that already takes care of error handling and waiting for the operation to finish for you.

PR #6569, opened 2024-12-16
chore(deps): update LG packages

mabaasit:
### Checklist
- [ ] New tests and/or benchmarks are included
- [ ] Documentation is changed or added
- [ ] I have signed the MongoDB Contributor License Agreement (https://www.mongodb.com/legal/contributor-agreement)

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*
- [ ] Patch (non-breaking change which fixes an issue)
- [ ] Minor (non-breaking change which adds functionality)
- [ ] Major (fix or feature that would cause existing functionality to change)

PR #6572, opened 2024-12-17
chore(aggregations): Refactor aggregations stage toolbar

kraenhansen:
## Description

An alternative to the refactor which happened in https://github.com/mongodb-js/compass/pull/6569#discussion_r1887231014 favoring composition over arrays of props.

I included a change to re-export the `GlyphName` from LG instead of deriving and exporting our own from the "components" package.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-aggregations/src/components/stage-toolbar/option-menu.tsx
+++ b/packages/compass-aggregations/src/components/stage-toolbar/option-menu.tsx
@@ -1,14 +1,6 @@
-import React, { useMemo, useState } from 'react';
+import React, { useState } from 'react';
 import { connect } from 'react-redux';
-import {
-  Menu,
-  MenuItem,
-  css,
-  IconButton,
-  Icon,
-  palette,
-  spacing,
-} from '@mongodb-js/compass-components';
+import { Menu, IconButton, Icon } from '@mongodb-js/compass-components';
 import {
   addStage,
   removeStage,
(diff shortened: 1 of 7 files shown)


packages/compass-aggregations/src/components/stage-toolbar/option-menu.tsx
```diff
@@ -105,18 +47,36 @@ export const OptionMenu = ({
...
-            <Icon color={palette.gray.dark2} glyph={item.icon} size="small" />
-            {item.label}
-          </div>
-        </MenuItem>
-      ))}
+      <OptionMenuItem
+        label="Add stage after"
+        icon="PlusWithCircle"
+        onClick={() => onAddStageClick(index)}
+        setMenuOpen={setMenuOpen}
```
mabaasit, line 54: What benefit does it have to pass these two here instead of just `onClick` and then something like:

```:js
	onClick={() => {
        onAddStageClick(index);
        setMenuOpen(false);
    }}
```

mabaasit (approved):

packages/compass-aggregations/src/components/stage-toolbar/option-menu.tsx
mabaasit: not a blocker though.

kraenhansen: This just ensures that `setMenuOpen(false)` is always called when `onClick` is called, not making that a choice of the caller.

PR #6577, opened 2024-12-19
feat(connections-navigation): connect in new window via connect split-button COMPASS-8473

kraenhansen:
## Description

Merging this PR will:
- Adjust the "connect button" in the sidebar to display a `SplitButton` with a dropdown menu, containing options to "Connect Here" and "Connect in a New Window".
- Adding the concept of the `ItemActionControls` being "hideable" (through the `setHideable` callback) allowing children (the `ConnectButton` in particular) to prevent the controls from being hidden once the cursor leaves the connection row. I use this to prevent controls (hence `SplitButton` and it's `Menu`) to be removed from the tree as the cursor is moved out of the `ConnectButton`, across and into the menu: If the menu is opened, the controls stay visible.
- Propagate events through "connection actions" → "connections Redux store" → "globalAppRegistry" → Main IPC → Window Manager to finally create a new window and ultimately storing a "connection id" in main-process state associated with this window.
- Auto-connect the newly created window from the associated state.
- For consistency, I added auto-connect when invoked as a CLI and a connection ID is passed.

https://github.com/user-attachments/assets/eaad1a6d-f03a-472e-9d13-7b02f027c595

### TODOs
- [x] Rebase once https://github.com/mongodb-js/compass/pull/6560 merge.
- [x] Ensure the SplitButton is only shown on Compass Desktop (and add a test for this)
- [x] Focus / activate the connection when auto-connected
- [x] Update the LeafyGreen `SplitButton` component once we're able to pass `renderDarkMenu` https://github.com/mongodb/leafygreen-ui/pull/2601 and to get https://github.com/mongodb/leafygreen-ui/pull/2596. I'm currently using a local install of the `@leafygreen-ui/split-button`.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions
- I did add support for passing a connection id as positional argument when invoked via the CLI, but is this desirable?

## Dependents
- https://github.com/mongodb/leafygreen-ui/pull/2596
- https://github.com/mongodb/leafygreen-ui/pull/2601
- https://github.com/mongodb-js/compass/pull/6560

## Types of changes

- [ ] *Backport Needed*


packages/compass-components/package.json
```diff
@@ -63,6 +63,7 @@
     "@leafygreen-ui/search-input": "^4.0.0",
     "@leafygreen-ui/segmented-control": "^9.0.0",
     "@leafygreen-ui/select": "^13.0.0",
+    "@leafygreen-ui/split-button": "../../../leafygreen-ui/packages/split-button/leafygreen-ui-split-button-v3.0.0.tgz",
```
kraenhansen, line 66: This is a temporary workaround and needs to be updated once my proposed changes merge and release.

packages/compass-components/src/components/actions/item-action-controls.tsx
```diff
@@ -39,7 +33,7 @@ export type ItemActionControlsProps<Action extends string> = {
   'data-testid'?: string;
 };
 
-export function ItemActionControls<Action extends string>({
+export function ItemActionControls<Action extends string = string>({
```
lerouxb, line 36: TIL that you can have a default

addaleax: Yeah – it usually just reduces type safety, though, because then consumers are likely to omit the type (and the default is typically the most generic type available).

This also seems to be one of those cases where type safety goes away – I don't seen an obvious reason for that in the rest of the code, though, so maybe in this particular case we could reconsider?

kraenhansen: 🙂 TBH, it's likely not strictly needed here, since we'll always be using this [in a way that can infer the type of this type parameter from the `props`](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAYg9nACgJzmAzgHgEIENkB8UAvFAN5QBG+AXFHslAL4DcAUACYQDGANvtABmAVwB23YAEs4o2AkwAVKBAAewCKI7oo6YMkmiA5gQAUYVBjrwkFrAoIBKOqEhxBUc2nQBtAETVkXwBddm4ZXShkCHRhXmASOTgTCgC6XxA4XyhcbTDRCKYHdgB6YqhyqAA9AH42IA).

packages/compass-sidebar/src/components/multiple-connections/sidebar.tsx
```diff
@@ -182,6 +183,9 @@ export function MultipleConnectionSidebar({
           onConnect={(connectionInfo) => {
             void connect(connectionInfo);
           }}
+          onConnectInNewWindow={(connectionInfo) => {
+            connectInNewWindow(connectionInfo);
+          }}
```
gribnoysup, line 188:
```suggestion
          onConnectInNewWindow={connectInNewWindow}
```

packages/compass/src/main/auto-connect.ts
```diff
@@ -5,6 +5,7 @@ import { hasDisallowedConnectionStringOptions } from './validate-connection-stri
 import COMPASS_ICON from './icon';
 import { createLogger, mongoLogId } from '@mongodb-js/compass-logging';
 import { redactConnectionString } from 'mongodb-connection-string-url';
+import { UUID } from 'mongodb';
```
gribnoysup, line 8:
```suggestion
import { UUID } from 'bson';
```

packages/compass-e2e-tests/helpers/selectors.ts
```diff
@@ -282,6 +282,8 @@ export const Single = {
 export const Multiple = {
   ConnectionsTitle: '[data-testid="connections-header"]',
   ConnectButton: '[data-action="connection-connect"]',
+  ConnectInNewWindowButton: '[data-action="connection-connect-in-new-window"]',
+  ConnectDropdownButton: '[aria-label="see more connection options"]',
```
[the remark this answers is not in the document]
kraenhansen: What in particular don't you like about this? That it will break if the button label gets updated?

This is targeting a "trigger" button within the `SplitButton`, but most props are forwarded to the "left most" button: https://github.com/mongodb/leafygreen-ui/blob/main/packages/split-button/src/SplitButton/SplitButton.tsx#L88

I might be able to select this as a button sibling of some `data-testid` 🤔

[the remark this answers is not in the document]
kraenhansen: This is the trigger I'm targeting 🤔

```html
<button
  data-lgid="lg-button"
  type="button"
  class="lg-ui-button-0000 leafygreen-ui-lt73jo"
  aria-disabled="false"
  aria-label="see more connection options"
  aria-controls="lg-split-button-menu-56"
  aria-expanded="true"
  aria-haspopup="true">
  ...
</button>
```

Alternatively I can select this relative to its sibling:

```
[data-testid="sidebar-navigation-item-actions-connection-connect-action"] + button
```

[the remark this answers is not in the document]
gribnoysup: We also usually use testid selectors in e2e tests, which are also mentioned as recommented by the documentation, so if we can pass a testid here, that's also an option and consistent with most of the other selectors. If we can't, it might be worth asking leafygreen team to add the option to pass it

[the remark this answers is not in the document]
kraenhansen: I'll do a PR for LG adding a constant `data-testid` attribute to the trigger element.

Related, I noticed that LG is also selecting the trigger in their test, relative to the other button: https://github.com/mongodb/leafygreen-ui/blob/f5d842764faf5b498f799d1710288ca964a93141/packages/split-button/src/SplitButton/SplitButton.spec.tsx#L52

packages/compass-connections-navigation/src/connect-button-with-menu.tsx
```diff
@@ -0,0 +1,109 @@
...
+        >
+          Connect Here
+        </ConnectMenuItem>,
+        <ConnectMenuItem
+          key="connection-connect-in-new-window"
+          action="connection-connect-in-new-window"
+          glyph="OpenNewTab"
+          onClick={onClick}
+        >
+          Connect in a New Window
```
betsybutton, line 102: Can we make this simply "In New Window"

packages/compass-connections-navigation/src/connect-button-with-menu.tsx
```diff
@@ -0,0 +1,109 @@
...
+      setOpen={setOpen}
+      triggerAriaLabel="see more connection options"
+      menuItems={[
+        <ConnectMenuItem
+          key="connection-connect"
+          action="connection-connect"
+          glyph="Connect"
+          onClick={onClick}
+        >
+          Connect Here
```
betsybutton, line 94: Can we simplify this to just "Here"

packages/compass/src/app/index.tsx
```diff
@@ -368,6 +368,11 @@ const app = {
         });
       }
     );
+    // Propagate events from global app registry to the main process
+    globalAppRegistry.on('connect-in-new-window', (connectionId: string) => {
+      void ipcRenderer?.call('app:connect-in-new-window', connectionId);
+    });
```
gribnoysup, line 374: FWIW adding support for web is literally just adding same three lines that call `window.open` instead of ipcRenderer in compass-web entrypoint

packages/compass-components/package.json
```diff
@@ -46,7 +46,7 @@
     "@leafygreen-ui/guide-cue": "^7.0.2",
     "@leafygreen-ui/hooks": "^8.3.4",
     "@leafygreen-ui/icon": "^13.1.2",
-    "@leafygreen-ui/icon-button": "^16.0.2",
+    "@leafygreen-ui/icon-button": "16.0.2",
```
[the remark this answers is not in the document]
kraenhansen: Yep. I meant to capture an issue on upgrading this as well, the v16.0.3 release is updating the types of the component, heres a comment on the original LG PR introducing the breaking change: https://github.com/mongodb/leafygreen-ui/pull/2703#issuecomment-2652021481

packages/compass/package.json
```diff
@@ -180,6 +180,7 @@
   },
   "dependencies": {
     "@mongosh/node-runtime-worker-thread": "^3.0.2",
+    "bson": "^6.10.1",
```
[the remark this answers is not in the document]
kraenhansen: How is that? It's being used here:

https://github.com/mongodb-js/compass/blob/9959e8da804926e7961c252dd4589d7194f43a87/packages/compass/src/main/auto-connect.ts#L8

Is this because you expect this to be included in the Webpack bundle?

packages/compass/src/app/index.tsx
kraenhansen: > three lines that call `window.open`

Ideally, I don't want the compass codebase to make assumptions on what URL it is being rendered at, which is why I wanted to delegate the opening of a separate browser tab to MMS.

packages/compass/package.json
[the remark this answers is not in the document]
gribnoysup: This code is bundled with webpack before being packaged, bson ends up bundled within the distribution, this dependency that will be installed with node_modules in "production" will never be required, you're basically just adding some dead code to the packaged application. Only dependencies that are externalized out of the bundle are part of the dependencies list here. This is why most of the dependencies are dev here

packages/compass-components/package.json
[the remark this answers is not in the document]
kraenhansen: > Do we have a ticket to follow up with a fix for that?

Not yet. I'll create that as part of merging this PR - this is what I meant by:

> I meant to capture an issue on upgrading this as well

packages/compass/src/app/index.tsx
kraenhansen: Makes sense. I'ved linked to this thread from the [COMPASS-8970](https://jira.mongodb.org/browse/COMPASS-8970) ticket.

kraenhansen:
@betsybutton this is what it looks like now 👍 
Screenshot 2025-02-13 at 11 32 51

packages/compass-components/package.json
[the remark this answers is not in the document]
kraenhansen: I've created [COMPASS-8973](https://jira.mongodb.org/browse/COMPASS-8973).

packages/compass/package.json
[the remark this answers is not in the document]
kraenhansen: Makes sense. I've moved it to `devDependencies`

packages/compass-e2e-tests/tests/connection.test.ts
```diff
@@ -641,6 +645,73 @@ describe('Connection string', function () {
...
+    const [mainWindow, ...otherWindows] = await browser.getWindowHandles();
+    for (const window of otherWindows.reverse()) {
+      await browser.switchToWindow(window);
+      await browser.closeWindow();
+    }
+    await browser.switchToWindow(mainWindow);
+  });
+
+  it('can connect in new window', async function (this) {
+    skipForWeb(this, 'connecting in new window is not supported on web');
```
gribnoysup, line 675: Leave a TODO here with a ticket numer so that it's easier to find?

packages/compass-preferences-model/src/preferences-schema.ts
```diff
@@ -404,6 +407,17 @@ export const storedUserPreferencesProps: Required<{
     validator: z.number().default(Date.now()),
     type: 'number',
   },
+  /**
+   * Enables a dropdown in the connections sidebar to connect in a new window.
+   */
+  enableConnectInNewWindow: {
```
gribnoysup, line 413: Ditto about TODO

packages/compass/src/app/index.tsx
kraenhansen: > let's discuss excluding features like that before jumping on implementation

I definitely don't share your sentiment that I've been "jumping on implementation". I have had extensive discussions with @bsradcliffe in Figma on the design and I've implemented this against the design detailed in [COMPASS-8381](https://jira.mongodb.org/browse/COMPASS-8381), which has clearly stated the dropdown as a non-goal:
> A few clarifications on the design, since it doesn't have all the details flushed out: 
> - Clicking the primary trigger surface connects you immediately in the same window.
> - Button appears on hover of a connection node.
> - Definitely disable dropdown for Compass Web.

> we also just added more extra work than the actual implementation would require to ourselves

Please clarify that statement. What extra work? To whom? And compared to what?

gribnoysup (commented):
Hmmm, when running this locally, I'm getting an error in the main console every time I open a connection in the new window:

```
Error sending from webFrameMain:  Error: Failed to serialize arguments
    at s.send (node:electron/js2c/browser_init:2:91812)
    at _.send (node:electron/js2c/browser_init:2:75243)
    at resolve (webpack://mongodb-compass/../hadron-ipc/src/main.ts?:95:20)
    at IpcMainImpl.eval (webpack://mongodb-compass/../hadron-ipc/src/main.ts?:111:7)
    at IpcMainImpl.emit (node:events:518:28)
    at WebContents.<anonymous> (node:electron/js2c/browser_init:2:85784)
    at WebContents.emit (node:events:518:28)
Error sending from webFrameMain:  Error: Failed to serialize arguments
    at s.send (node:electron/js2c/browser_init:2:91812)
    at _.send (node:electron/js2c/browser_init:2:75243)
    at resolve (webpack://mongodb-compass/../hadron-ipc/src/main.ts?:95:20)
    at IpcMainImpl.eval (webpack://mongodb-compass/../hadron-ipc/src/main.ts?:111:7)
    at IpcMainImpl.emit (node:events:518:28)
    at WebContents.<anonymous> (node:electron/js2c/browser_init:2:85784)
    at WebContents.emit (node:events:518:28)
```

Have you seen that?

gribnoysup:
Added some logging, a whole browser window is getting passed to the resolve there at some point:

```
BrowserWindow {
  setBounds: [Function (anonymous)],
  _events: [Object: null prototype] {
    blur: [Function (anonymous)],
    focus: [ [Function (anonymous)], [Function: onFocus] ],
    show: [Function: visibilityChanged],
    hide: [Function: visibilityChanged],
    minimize: [Function: visibilityChanged],
    maximize: [Function: visibilityChanged],
    restore: [Function: visibilityChanged],
    closed: [ [Function (anonymous)], [Function], [Function] ],
    'ready-to-show': [Function: bound onceWrapper] { listener: [Function: resolver] },
    error: [Function: bound onceWrapper] {
      listener: [Function: errorListener]
    }
  },
  _eventsCount: 10,
  _browserViews: [],
  devToolsWebContents: [Getter]
}
Error sending from webFrameMain:  Error: Failed to serialize arguments
```

packages/compass/src/main/window-manager.ts
```diff
@@ -265,6 +272,8 @@ class CompassWindowManager {
         return getWindowAutoConnectPreferences(bw, compassApp.preferences);
       },
       'test:show-connect-window': () => showConnectWindow(compassApp),
+      'app:connect-in-new-window': (event, connectionId: string) =>
+        showConnectWindow(compassApp, { connectionId }),
```
gribnoysup, line 276: These listeners will try to return value back to renderer, if value is not serialized, you'll get an error. Seems like an existing issue that you just copied, we probably want to fix it for both cases
```suggestion
      'test:show-connect-window': () => void showConnectWindow(compassApp),
      'app:connect-in-new-window': (event, connectionId: string) =>
        void showConnectWindow(compassApp, { connectionId }),
```

packages/compass/src/app/index.tsx
kraenhansen: We had a zoom to talk through this ☝️

packages/compass-e2e-tests/tests/connection.test.ts
kraenhansen: Added a todo 👍

packages/compass-preferences-model/src/preferences-schema.ts
kraenhansen: Added a todo here as well 👍

gribnoysup (approved):
Awesome work! Happy to see this landing in main!

PR #6579, opened 2024-12-20
feat(e2e): smoketest macOS .dmg files COMPASS-8709

lerouxb:
[COMPASS-8709](https://jira.mongodb.org/browse/COMPASS-8709)

Installs the .app found in the .dmg to /Applications then runs the time to first query test on the app in /Applications.

To run locally:

```
m -rf /Applications/MongoDB\ Compass\ Dev.app
HADRON_DISTRIBUTION=compass ARCH=arm64 IS_OSX=true ./smoke-test.ts --extension=dmg --skip-download
```


packages/compass-e2e-tests/installers/mac-dmg.ts
```diff
@@ -0,0 +1,39 @@
...
+  appName: string,
+  { filepath }: Package
+): Promise<InstalledAppInfo> {
+  const fullDestinationPath = `/Applications/${appName}.app`;
+
+  if (existsSync(fullDestinationPath)) {
+    throw new Error(`${fullDestinationPath} already exists`);
+  }
+
+  exec('hdiutil', ['attach', filepath]);
```
lerouxb, line 28: I'm in two minds about whether this should just be a shell script at this point, but I suppose it is easy to execute a shell script in its place at a later stage. I suppose we'll see what happens once we get to the windows setup .exe and .msi what works best there.

For .tar.gz and .zip files it would likely just be one line anyway.

packages/compass-e2e-tests/smoke-test.ts
```diff
@@ -333,6 +366,32 @@ function verifyPackagesExist(packages: Package[]): void {
...
+  const result = spawnSync(
+    'npm',
+    [
+      'run',
+      '--unsafe-perm',
+      'test-packaged',
+      '--workspace',
+      'compass-e2e-tests',
+      '--',
+      '--test-filter=time-to-first-query',
```
lerouxb, line 379: The filter will probably change once we add tests that test auto-update since it is probably easiest to test that by running the app inside an e2e test so we can use webdriverio. But this is good enough for now because it tests that the app starts up and can execute a query.

packages/compass-e2e-tests/installers/helpers.ts
```diff
@@ -0,0 +1,19 @@
+import type { SpawnSyncReturns } from 'child_process';
+
+export function assertSpawnSyncResult(
```
addaleax, line 3: Non-blocking, and no need to change anything in this PR, but:

The typical best practice in Node.js in general is to use `execFile` (and usually promisified so it's not completely unworkable). `spawnSync` is obviously also _fine_ to use out of convenience in tests, since there's no need to care about performance or limitations of the synchronous method as much, but I'd be wary of building an entire test helper system around it – if you're at the point where you're introducing helpers already, you might as well just stick to the async method. (Again, this all matters only to a limited extent, but "follow best practices for production code even in tests" just makes it easier to also follow best practices in actual production code)

addaleax (approved):

packages/compass-e2e-tests/installers/helpers.ts
lerouxb: Regarding sync vs async: I was on the fence and I'll just change it to async.

Regarding exec vs execFile vs spawn vs spawn: I have no idea when to use which. I suppose exec() at least has a callback for when it finishes, but I got stuck on the fact that in a general sense I don't want all the stdout and stderr in a variable. In these cases there's little output, but in a more general sense you can never be too sure. To me the ideal seems to pass it through to stdout and stderr and just resolve a promise with an error. None of them seem to do that or close to it out of the box.

lerouxb: ie. I basically want to happen what happens when you call a program in a shell script 😆

lerouxb: I made a promisified async spawn helper rather and removed assertSpawnSyncResult. This way I can error on anything except exit code 0 and stdio gets passed through rather than accumulate all the e2e test output in the result.

.evergreen/functions.yml
```diff
@@ -674,7 +674,7 @@ functions:
             export COMPASS_E2E_DISABLE_CLIPBOARD_USAGE="true"
           fi
 
-          npm run --workspace compass-e2e-tests smoketest
+          npm run --unsafe-perm --workspace compass-e2e-tests smoketest
```
lerouxb, line 677: "Let's see if we really need this" and then _forget about it for a week_ 🤦

lerouxb:
Closing in favor of https://github.com/mongodb-js/compass/pull/6582 because I made a real mess here..

PR #6586, opened 2025-01-02
feat(mocked-compass): build mocked compass for auto-update tests

lerouxb:
Built on https://github.com/mongodb-js/compass/pull/6584. Just testing to see if this works in CI.


packages/compass-e2e-tests/smoke-test.ts
```diff
@@ -168,6 +182,8 @@ async function run() {
 
   verifyPackagesExist(packages);
 
+  await buildMockedCompass();
```
lerouxb, line 185: I think the plan now is to split this out into its own evergreen task that builds the mocked compass app and signs/notarises it then just uploads it to S3. Probably somewhere in or near `https://mciuploads.s3.amazonaws.com/${EVERGREEN_PROJECT}/compass/update-builds` because that's where we store the dev update next version id.

Then this smoketest task will depend on that task in addition to package-compass and it will just download it from s3.

PR #6598, opened 2025-01-08
chore(connections): disconnect when we encounter a non-retryable error code on an atlas connection CLOUDP-286331

Anemy:
[CLOUDP-286331](https://jira.mongodb.org/browse/CLOUDP-286331)

When we're on cloud we listen for non-retry-able errors on failed server heartbeats. These can happen when:
- A user's session has ended.
- The user's roles have changed (so they lack permissions).
- The cluster / group they are trying to connect to has since been deleted.

When we encounter one we disconnect. This is to avoid polluting logs/metrics and to avoid constantly retrying to connect when we know it'll fail.

Screenshot 2025-01-08 at 6 17 27 PM

When a user runs a command after we've disconnected they end up with errors like this:

Screenshot 2025-01-08 at 6 18 23 PM

Which [we surface ourselves](https://github.com/mongodb-js/compass/blob/dce7b3b9c45b6572ce628f27d4c493c795b210bf/packages/data-service/src/data-service.ts#L2455). While they do have a message from the toast I'm thinking we probably want to give something less cryptic there. A bit more context, we also chatted about fully disconnecting from the connection, which would close all of their open tabs, but that may end up in them losing some work which would be a frustrating ux. Discussion: https://mongodb.slack.com/archives/C069YM25L8N/p1735662529296989 

Not an easily testable flow as these are mostly internals of data service and mongo client, curious if folks think it's worth some mocks to facilitate tests for it. We'd need to simulate the `serverHeartbeatFailed` events on a mongo client and have the custom mongo client passed to the data service.

Custom close codes in mms: https://github.com/10gen/mms/blob/24f5af9c5318a5da746ad328547591f376449dc0/server/src/main/com/xgen/cloud/services/clusterconnection/runtime/res/CustomCloseCodes.java#L5
Where they are returned on pings which form the `serverHeartbeatFailed` events: https://github.com/10gen/mms/blob/24f5af9c5318a5da746ad328547591f376449dc0/server/src/main/com/xgen/cloud/services/clusterconnection/runtime/res/ClusterConnectionEndpoint.java#L92


packages/compass-connections/src/stores/connections-store-redux.ts
```diff
@@ -1507,6 +1508,46 @@ function isAtlasStreamsInstance(
...
+      ] ?? 'Unknown';
+}
+
+const openConnectionClosedWithNonRetryableErrorToast = (
+  connectionInfo: ConnectionInfo,
+  error: Error
+) => {
+  openToast(`non-retryable-error-encountered--${connectionInfo.id}`, {
+    title: `Unable to connect to ${getConnectionTitle(connectionInfo)}`,
+    description: `Reason: ${getDescriptionForNonRetryableError(error)}`,
```
Anemy, line 1546: We can mention here that a user will have to refresh or disconnect and reconnect before they can use the connection.

syn-zhu (commented):
Is there another PR on the MMS side you're currently working on?

packages/compass-connections/src/stores/connections-store-redux.ts
```diff
@@ -1701,6 +1742,36 @@ const connectWithOptions = (
           return;
         }
 
+        let showedNonRetryableErrorToast = false;
+        if (connectionInfo.atlasMetadata) {
```
gribnoysup, line 1746: Do we need to gate this for atlas connections? Not that it makes a big difference, but adding logic branching hen not required can maybe be avoided

Not doing this and maybe updating our ws-proxy to return those in some cases might allow you to add e2e tests for that

gribnoysup:
> Not an easily testable flow as these are mostly internals of data service and mongo client

I think you definitely should be able to add unit tests in the compass-connections for this by emitting the event manually on a mocked dataService, we have all the tooling to setup a test like that and at least it would cover the error parsing logic. Should be possible to add some e2e tests too if we want to by modifying our ws-proxy code, we can chat more about how to set that up

packages/compass-connections/src/stores/connections-store-redux.ts
```diff
@@ -1507,6 +1508,46 @@ function isAtlasStreamsInstance(
   }
 }
 
+const NonRetryableErrorCodes = [3000, 3003, 4004, 1008] as const;
```
gribnoysup, line 1511: Maybe worth linking to the code in mms / expanding the description here a bit so that the context for where those codes are coming from is not lost and easy to link to the source

Anemy:
@syn-zhu 
> Is there another PR on the MMS side you're currently working on?

No, this is the only branch I've been working on. Is there work you foresee us needing on the mms side? I see there's already support for the ping/pongs checking the rolls and throwing these errors https://github.com/10gen/mms/blob/de2a9c463cfe530efb8e2a0941033e8207b6cb11/server/src/main/com/xgen/cloud/services/clusterconnection/runtime/res/ClusterConnectionEndpoint.java#L521 
Maybe I'm missing some work we need to do there?

packages/compass-connections/src/stores/connections-store-redux.ts
Anemy: Nice suggestion, sounds good, removed the branching logic.

Added 2 unit tests, I'll work on an e2e test now.

packages/compass-connections/src/stores/connections-store-redux.ts
Anemy: Screenshot 2025-01-13 at 5 14 31 PM

packages/compass-connections/src/stores/connections-store-redux.ts
```diff
@@ -1465,6 +1466,57 @@ function isAtlasStreamsInstance(
...
 
+// We listen for non-retry-able errors on failed server heartbeats.
+// These can happen when:
+// - A user's session has ended.
+// - The user's roles have changed.
+// - The cluster / group they are trying to connect to has since been deleted.
+// When we encounter one we disconnect. This is to avoid polluting logs/metrics
+// and to avoid constantly retrying to connect when we know it'll fail.
+// These error codes can be found at
+// https://github.com/10gen/mms/blob/de2a9c463cfe530efb8e2a0941033e8207b6cb11/server/src/main/com/xgen/cloud/services/clusterconnection/runtime/res/CustomCloseCodes.java
```
syn-zhu, line 1477: This seems obvious now, but might it make sense to add a comment somewhere explicitly clarifying that this only applies to compass-web?

Since that's still technically the special case

syn-zhu (approved):
one comment, everything else looks good

Anemy:
Chatted a bit with Sergey in slack, not going to go for an e2e for this at least for now. I'm going to wait on merging until I can reliably get this working with our main branch of mms. It might take some backend changes, having trouble returning the correct errors at the moment, it may go away once we merge https://github.com/10gen/mms/pull/116288

PR #6600, opened 2025-01-09
chore(e2e): Lint for restricted globals in e2e tests

kraenhansen:
## Description

The WebdriverIO API allows for execution of serialized JavaScript functions in the browser via the `browser.execute` method. We're currently using this. To get typescript support for this, our tsconfig includes `"dom"` as a `lib`, which has the unfortunate drawback that the types for DOM globals become available to all the code, which is running on Node.js and doesn't have values for these available at runtime.

This is an issue first and foremost because the usefulness of TypeScript degrades fast when you cannot trust it and most recently when I wanted to use the `fetch` implementation built into Node.js, as the DOM's `fetch` declaration overlap and differ slightly.

<details><summary>TLDR; It's not trivial to restrict access to these types</summary>
In situations like this, I've reached for a multi-project setup using TypeScript references in the past: They can specify subsets of files which are allowed access to types based on specific TypeScript configs.

This won't work in this particular situation however, because:
1. The WebdriverIO's `browser.execute` is inlined with other code which is not allowed to access DOM APIs and the tsconfigs can't control types at the granularity of blocks of code. This could be worked around by extracting these functions into separate files with a known extension (such as `whatever.dom.ts`) which would indicate code which would be unrestricted from accessing DOM globals, but this does come with a drawback of decreased cohesion between the function and the code calling ti.
2. TypeScript project references need project configs to use "composite", which entails either `noEmit: false` or `emitOnlyDeclaration: true`, meaning they will emit code, which is not ideal in our scenario, as we're using `ts-node` to strip types and interpret the code without an explicit build step.
</details>

My proposal for an interim solution is inspired by [this comment](https://github.com/microsoft/TypeScript/issues/14306#issuecomment-2527855530), where the `"globals"` package is used in conjunction with the `"no-restricted-globals"` eslint rule to produce errors and warnings when accessing DOM-only globals.

Merging this PR will:
- Enable eslint errors for code accessing DOM globals in general.
- De-escalate to eslint warnings for files which we know have `browser.execute` calls.
- Add eslint-ignore comments for existing use of these DOM globals, where I've verified the code is executed through `browser.execute` calls.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43285,6 +43285,7 @@
         "electron-to-chromium": "^1.5.76",
         "eslint": "^7.25.0",
         "glob": "^10.2.5",
+        "globals": "^15.14.0",
         "hadron-build": "^25.6.1",
         "lodash": "^4.17.21",
         "mocha": "^10.2.0",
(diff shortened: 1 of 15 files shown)


lerouxb (approved):
Makes sense to me!

PR #6612, opened 2025-01-14
feat(compass): show the autoupdate installed toast even when we downgrade COMPASS-8808

lerouxb:
This also adds logging that will be super helpful in debugging auto-update. Especially during testing. And some data-testids we can use in the tests to address things in the relevant toasts.

### DIFF
--- a/packages/compass/src/app/components/update-toasts.tsx
+++ b/packages/compass/src/app/components/update-toasts.tsx
@@ -53,6 +53,7 @@ const RestartCompassToastContent = ({
       <button
         className={cx(buttonStyles, darkmode && buttonDarkStyles)}
         onClick={onUpdateClicked}
+        data-testid="auto-update-restart-button"
       >
         Restart
       </button>
@@ -83,6 +84,7 @@ export function onAutoupdateExternally({
           Compass features.
         </Body>
         <Link
+          data-testid="auto-update-download-link"
           as="a"
           target="_blank"
           href={'https://www.mongodb.com/try/download/compass'}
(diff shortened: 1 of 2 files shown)


packages/compass/src/app/index.tsx
```diff
@@ -379,9 +379,15 @@ const app = {
...
+    log.info(mongoLogId(1_001_000_338), 'Main Window', 'Recent version info', {
+      previousVersion: state.previousVersion,
+      highestInstalledVersion: state.highestInstalledVersion,
+      APP_VERSION,
+    });
+
     if (
-      semver.gt(APP_VERSION, state.previousVersion) &&
-      state.previousVersion !== DEFAULT_APP_VERSION
+      state.previousVersion !== DEFAULT_APP_VERSION &&
```
lerouxb, line 389: Same justification and comment as https://github.com/mongodb-js/compass/pull/6608/files#r1913799566

packages/compass/src/app/components/update-toasts.tsx
```diff
@@ -53,6 +53,7 @@ const RestartCompassToastContent = ({
       <button
         className={cx(buttonStyles, darkmode && buttonDarkStyles)}
         onClick={onUpdateClicked}
+        data-testid="auto-update-restart-button"
```
lerouxb, line 56: As a drive-by I'm adding these test ids so the latest nightly dev release will end up with them and then the auto-update tests can use them.

gribnoysup (approved):

PR #6614, opened 2025-01-14
chore(e2e): remove the retry

lerouxb:
I have a suspicion that retry isn't helping and is actually making things worse. Let's see what happens in CI for this PR. We don't have to merge it and we can easily revert it too.


mabaasit (approved):

lerouxb:
This CI run was green (which is pretty rare at the moment, but I suspect that's still just a coincidence) so I'm merging it.

If it turns out that the flake we're seeing on mac arm64 really is just confined to the first time we start electron and we can't figure out why, then I think we can try to work around it by starting the app and immediately closing it before we start the tests. Then just catch any errors in that step.

PR #6618, opened 2025-01-15
chore(e2e): throw if we're unexpectedly building Compass COMPASS-8615

lerouxb:
This is here to make very sure that we're definitely testing  the packaged app when we intend to. Mostly to not trip me up every time I read over that code.

### DIFF
--- a/packages/compass-e2e-tests/helpers/compass.ts
+++ b/packages/compass-e2e-tests/helpers/compass.ts
@@ -949,18 +949,20 @@ async function getCompassBuildMetadata(): Promise<BinPathOptions> {
 }
 
 export async function buildCompass(
-  force = false,
   compassPath = COMPASS_DESKTOP_PATH
 ): Promise<void> {
-  if (!force) {
-    try {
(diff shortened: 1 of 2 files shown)


packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -949,18 +949,20 @@ async function getCompassBuildMetadata(): Promise<BinPathOptions> {
 }
 
 export async function buildCompass(
-  force = false,
```
lerouxb, line 952: drive-by, but we never use this

packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -949,18 +949,20 @@ async function getCompassBuildMetadata(): Promise<BinPathOptions> {
...
   compassPath = COMPASS_DESKTOP_PATH
```
lerouxb, line 952: yes the second parameter became the first, but we never pass either of these parameters, so that didn't require other changes.

PR #6621, opened 2025-01-16
Add sandboxing and refactored smoke test harness

kraenhansen:
## Description

Merging this PR will:

1. Add sandboxes: The creation of temporary directories for installers and app to store files into.
1. Simplify the `downloadFile` to use `fetch` and avoid having to implement following redirection manually and adopt to [Anna's suggestion](https://github.com/mongodb-js/compass/pull/6571#discussion_r1892661445) of using `stream.promises.pipeline`.
1. I've replaced `skipDownload` (which had two purposes - testing locally built / packaged compass as well as spending less time downloading) with a download cache based on the `ETag` header of the server response and a new `--forceDownload` flag to clear the cache before download.
1. When building and packaging Compass locally, a `dist/target.json` with "build info" is produced, so to test a locally built and packaged Compass, we don't need to invoke Hadron build again. I've added a `--localPackage` to skip hadron-build and download entirely and instead use the `compass/dist` directory (and its artifacts).
1. I wanted to make the parsing / assertion of the build-info bind types, such that the extraction of the filename property could be generalized and we could have code like: `if (kind === 'osx_dmg') { console.log(buildInfo.osx_dmg_filename)` (i.e. a conditional on the kind of package narrows buildInfo). I've added the extraction of the filename from the "build info" based on the kind of package as well.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [x] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-e2e-tests/smoke-test.ts
```diff
@@ -339,5 +226,5 @@ run()
   })
   .catch(function (err) {
     console.error(err.stack);
-    process.exit(1);
+    process.exitCode = 1;
```
kraenhansen, line 229: From https://nodejs.org/api/process.html#processexitcode:

> Rather than calling process.exit() directly, the code should set the process.exitCode and allow the process to exit naturally by avoiding scheduling any additional work for the event loop:

lerouxb: It is like I know this and keep independently rediscovering it yet still forget.

packages/compass-e2e-tests/helpers/smoke-test/build-info.ts
```diff
@@ -0,0 +1,189 @@
...
+      'HADRON_APP_VERSION',
+      'HADRON_PRODUCT',
+      'HADRON_PRODUCT_NAME',
+      'HADRON_READONLY',
+      'HADRON_ISOLATED',
+      'DEV_VERSION_IDENTIFIER',
+      'IS_RHEL',
+    ])
+  );
+  writeBuildInfo(infoArgs);
```
lerouxb, line 187: Part of me wants to change hadron-build so we can get this info as a variable without an intermediate file, but I've been a bit weary of changing that code because it is used in so many ways and often only at release time. Although I suppose nowadays we release dev too, so it should be safer.

packages/compass-e2e-tests/smoke-test.ts
```diff
@@ -156,163 +157,49 @@ async function run() {
...
+
+  try {
+    const appName = buildInfo.productName;
+    if (kind === 'osx_dmg') {
+      const { appPath, uninstall } = installMacDMG({
+        appName,
+        filepath,
+        destinationPath: context.sandboxPath,
+      });
+      cleanups.push(uninstall);
```
lerouxb, line 176: We should probably uninstall before clearing out the sandbox because if it runs an uninstaller that uninstaller will be inside the sandbox.

lerouxb (approved):
Great work!

packages/compass-e2e-tests/helpers/smoke-test/downloads.ts
```diff
@@ -0,0 +1,58 @@
+import assert from 'node:assert';
+import fs from 'node:fs';
+import path from 'node:path';
+import stream from 'node:stream';
+import { type fetch as undiciFetch } from 'undici-types';
+
+// Hacking types here because the DOM types doesn't match the fetch implemented by Node.js
+const fetch = globalThis.fetch as unknown as typeof undiciFetch;
```
kraenhansen, line 8: This could be avoided if we moved the code into its own package with types properly locked down.

packages/compass-e2e-tests/smoke-test.ts
kraenhansen: Good point - deleting the sandbox should be the last cleanup task 👍

PR #6629, opened 2025-01-17
feat(smoke-tests): test auto-updating from the version of Compass that was just packaged COMPASS-8536

lerouxb:
Replaces https://github.com/mongodb-js/compass/pull/6608

This changes the smoketest script's interface slightly. There is now a `--tests` parameter so you can tell it which smoketest you want to run. Options are currently `'time-to-first-query'` and  `'auto-update-from'`. Soon we'll add `'auto-update-to'`. An `'install-only'` might be useful too for when you're just working on an installer.

The benefit here is that we can control the exception where mac doesn't auto-update (just run a different test) from CI while still testing it locally.

So you can run it like:

```
npm run start -- --package=osx_dmg --tests auto-update-from
```

We now also use the real compass-mongodb-com update server. In CI it gets checked out next to `compass`. This is likely where you'd have it during development anyway. We `npm link` it into the codebase (it is an optional peer dependency) and then dynamically import it if you run the auto-update server. The import error is caught, it logs a helpful hint that you need to npm link it and rethrows the error. Since it is a dynamic import you only need this if you're running the `'auto-update-from'` test.


packages/compass-smoke-tests/src/cli.ts
```diff
@@ -201,12 +220,55 @@ async function run() {
...
+}: AutoUpdateServerOptions) {
+  const env: Record<string, string> = {
+    ...process.env,
+    PORT: port.toString(),
+  };
+  if (allowDowngrades) {
+    env.UPDATE_CHECKER_ALLOW_DOWNGRADES = 'true';
+  }
+
+  // a git repo that is not published to npm that evergreen clones for us in CI
```
lerouxb, line 240: I don't know how best to cleanly start this, especially in dev. This assumes that `compass-mongodb-com` is cloned right next to `compass` which is hacky.

My initial plan was to start the server using docker in evergreen so it is already running before we execute the smoke tests, but we don't have docker in every platform where we want to run the smoke tests, so I think we have to start the server manually.

Furthermore we'll also have two of these servers with different configuration that we'll need back-to-back because updating from the most recent release TO this test subject will require telling the auto-update server about this new unreleased copy of compass.

packages/compass-e2e-tests/tests/auto-update.test.ts
```diff
@@ -0,0 +1,67 @@
...
+import {
+  init,
+  cleanup,
+  Selectors,
+  screenshotPathName,
+} from '../helpers/compass';
+
+describe('Auto-update', function () {
+  it('auto-update from', async function () {
+    if (process.env.TEST_NAME !== 'AUTO_UPDATE_FROM') {
```
lerouxb, line 11: Open to ideas for a better name for this env var. I think we'll also have AUTO_UPDATE_TO once we add that test next to this one. These tests kinda have to be the only test you run at a time because they replace the test subject.

.evergreen/functions.yml
```diff
@@ -648,6 +648,36 @@ functions:
 
           npm run --unsafe-perm --workspace compass-e2e-tests test-packaged-ci
 
+  get-compass-mongodb-com:
```
lerouxb, line 651: change to get-update-server

packages/compass-smoke-tests/src/cli.ts
```diff
@@ -200,12 +218,53 @@ async function run() {
...
+  });
+}
+
+type RunE2ETestOptions = {
   appName: string;
   appPath: string;
 };
 
-function runTest({ appName, appPath }: RunTestOptions) {
+function runE2ETest({ appName, appPath }: RunE2ETestOptions) {
```
lerouxb, line 267: runTimeToFirstQuery

packages/compass-smoke-tests/src/installers/mac-dmg.ts
```diff
@@ -15,6 +15,8 @@ export function installMacDMG({
 
   execute('hdiutil', ['attach', filepath]);
 
+  execute('umask', []);
```
lerouxb, line 18: cleanup

packages/compass-e2e-tests/tests/auto-update.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        await linkElement.waitForDisplayed();
+        expect(await linkElement.getAttribute('href')).to.equal(
+          'https://www.mongodb.com/try/download/compass?utm_source=compass&utm_medium=product'
+        );
+      }
+    } finally {
+      await browser.screenshot(screenshotPathName('auto-update-from'));
+      await cleanup(compass);
+
+      if (process.platform === 'darwin' && process.env.HOME) {
```
lerouxb, line 55: cleanup

packages/compass-smoke-tests/src/installers/mac-dmg.ts
```diff
@@ -33,13 +35,23 @@ export function installMacDMG({
       'Application Support',
       appName
     );
+    const shipitDir = path.resolve(
```
lerouxb, line 38: cleanup

packages/compass-smoke-tests/src/installers/mac-dmg.ts
```diff
@@ -33,13 +35,23 @@ export function installMacDMG({
...
-      fs.rmSync(settingsDir, { recursive: true });
+    for (const dir of [settingsDir, shipitDir]) {
+      if (fs.existsSync(dir)) {
+        console.log(`${dir} already exists. Removing.`);
+        fs.rmSync(dir, { recursive: true });
+      }
     }
   }
 
+  execute('xattr', ['-dr', 'com.apple.quarantine', appPath]);
```
lerouxb, line 53: It isn't quarantined, just remove this line.

lerouxb: Replace with an assertion.

packages/compass-smoke-tests/src/cli.ts
```diff
@@ -229,6 +288,58 @@ function runTest({ appName, appPath }: RunTestOptions) {
...
+
+function runUpdateTest({
+  appName,
+  appPath,
+  autoUpdatable,
+  testName,
+}: RunUpdateTestOptions) {
+  const server = startAutoUpdateServer({
+    allowDowngrades: true,
+    port: 8080,
```
lerouxb, line 306: change to 0, read the port out of standard output

packages/compass-smoke-tests/package.json
```diff
@@ -34,11 +34,13 @@
     "@mongodb-js/eslint-config-compass": "^1.2.5",
     "@mongodb-js/prettier-config-compass": "^1.1.5",
     "@mongodb-js/tsconfig-compass": "^1.1.5",
+    "cross-spawn": "^7.0.5",
     "depcheck": "^1.4.1",
     "eslint": "^7.25.0",
     "hadron-build": "^25.6.5",
     "lodash": "^4.17.21",
     "prettier": "^2.7.1",
+    "tree-kill": "^1.2.2",
```
lerouxb, line 43: process.kill rather

packages/compass-smoke-tests/src/cli.ts
```diff
@@ -200,12 +218,53 @@ async function run() {
...
+    '..',
+    'compass-mongodb-com'
+  );
+
+  if (!fs.existsSync(cwd)) {
+    throw new Error(`compass-mongodb-com does not exist: ${cwd}`);
+  }
+
+  console.log('Starting auto-update server', cwd);
+  return crossSpawn.spawn('npm', ['run', 'start'], {
```
lerouxb, line 254: Rather just try and run this programmatically.

packages/compass-smoke-tests/src/cli.ts
```diff
@@ -200,12 +218,53 @@ async function run() {
...
+    ...process.env,
+    PORT: port.toString(),
+  };
+  if (allowDowngrades) {
+    env.UPDATE_CHECKER_ALLOW_DOWNGRADES = 'true';
+  }
+
+  // a git repo that is not published to npm that evergreen clones for us in CI
+  // next to the Compass code
+  const cwd = path.join(
```
lerouxb, line 240: npm link it on the outside, use require(), catch it.

packages/compass-smoke-tests/tsconfig.json
```diff
@@ -3,7 +3,11 @@
   "compilerOptions": {
     "noEmit": true,
     "types": ["node"],
-    "lib": ["ES2021"]
+    "lib": ["ES2021"],
+    "module": "NodeNext",
+    "paths": {
+      "compass-mongodb-com": ["./src/compass-mongodb-com.d.ts"]
```
lerouxb, line 9: Needed for ts-node https://github.com/TypeStrong/ts-node?tab=readme-ov-file#missing-types

packages/compass-smoke-tests/tsconfig.json
```diff
@@ -3,7 +3,11 @@
   "compilerOptions": {
     "noEmit": true,
     "types": ["node"],
-    "lib": ["ES2021"]
+    "lib": ["ES2021"],
+    "module": "NodeNext",
```
lerouxb, line 7: Needed so that typescript doesn't compile the dynamic import to a require and then complain that we're require()-ing an ES module.

packages/compass-e2e-tests/tests/auto-update.test.ts
```diff
@@ -0,0 +1,78 @@
...
+  it('auto-update from', async function () {
+    if (process.env.TEST_NAME !== 'auto-update-from') {
+      // we don't want this test to execute along with all the others under
+      // normal circumstances because it is destructive - it overwrites Compass
+      // itself
+      this.skip();
+    }
+
+    // run the app and wait for it to auto-update
+    console.log('starting compass the first time');
```
lerouxb, line 25: I have another branch open where I'm switching from console.log to debug. I don't want to conflict with that now, so will fix all these together in a future PR.

packages/compass-smoke-tests/package.json
```diff
@@ -24,7 +24,7 @@
     "eslint": "eslint",
     "prettier": "prettier",
     "lint": "npm run eslint . && npm run prettier -- --check .",
-    "depcheck": "compass-scripts check-peer-deps && depcheck",
+    "depcheck": "depcheck",
```
kraenhansen, line 27: 😬 If this doesn't handle `optionalDependencies`, we should probably add a ticket tracking a fix for that.

packages/compass-smoke-tests/src/cli.ts
```diff
@@ -104,7 +107,14 @@ const argv = yargs(hideBin(process.argv))
...
     default: false,
-  });
+  })
+  .option('tests', {
+    type: 'string',
+    choices: SUPPORTED_TESTS,
+    description: 'Which tests to run',
+  })
+  .array('tests')
+  .default('tests', []);
```
kraenhansen, line 117: I _think_ this is doable in a single call of the `option` builder function 🤔

```suggestion
  .option('tests', {
    type: 'array',
    string: true,
    choices: SUPPORTED_TESTS,
    description: 'Which tests to run',
    default: [],
  });
```

I think th

packages/compass-smoke-tests/src/cli.ts
```diff
@@ -184,23 +194,47 @@ async function run() {
...
   const install = getInstaller(kind);
 
   try {
-    const { appPath, uninstall } = install({
-      appName,
-      filepath,
-      destinationPath: context.sandboxPath,
-    });
+    if (context.tests.length === 0) {
+      console.log('Warning: not performing any tests. Pass --tests.');
```
kraenhansen, line 208: Or, perhaps we should just default to all tests? 🤔

packages/compass-smoke-tests/package.json
lerouxb: I think that's for compass plugins, though. Which is probably why it is a separate script and not part of depcheck.

lerouxb: I'll ask @gribnoysup  what he thinks.

lerouxb: Checking this rather than npm link, removing it as a peer dep.
https://github.com/mongodb-js/compass/pull/6629/commits/5926b93f63987dc61a7d3aff98a8ca0879e1a7f3

PR #6634, opened 2025-01-20
chore(deps): bump mongodb-log-writer MONGOSH-1992

gagik:
Includes a version bump and adjustments for the breaking change from the latest major release.

Related to https://github.com/mongodb-js/devtools-shared/pull/498

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8703,7 +8703,7 @@
         "@mongosh/errors": "2.3.7",
         "@mongosh/history": "2.3.7",
         "@mongosh/types": "2.3.7",
-        "mongodb-log-writer": "^1.4.2",
+        "mongodb-log-writer": "^2.1.0",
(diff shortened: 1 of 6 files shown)


addaleax (approved):
Does it matter whether this is bundled with the corresponding shell updates?

gagik:
@addaleax I imagine it wouldn't? We may end up having two versions as the dependencies (this is the case because of the oidc-plugin right now looking at package-lock, though I don't know if this means we're fully bundling both) but don't think it'd be anything problematic functionality wise. Hopefully caught by the tests if I'm wrong.

addaleax (approved):
@gagik Yeah, probably just a bit of unnecessary paranoia on my end around making sure our dependencies are on the same version 🙂

lerouxb (approved):

gagik:
Redundant, has been updated already.

PR #6638, opened 2025-01-21
feat: download and test with the latest release

lerouxb:
Still a work in progress because this should test that it will upgrade to the new package. Right now it is just running an e2e test against the latest release for the relevant channel.

I'm just opening this draft early to get some feedback.


packages/compass-smoke-tests/src/build-info.ts
```diff
@@ -25,13 +27,21 @@ function assertObjectHasKeys(
...
+export type CommonBuildInfo = Record<typeof commonKeys[number], string> & {
+  channel: 'dev' | 'beta' | 'stable';
+};
 
 export function assertCommonBuildInfo(
   buildInfo: unknown
 ): asserts buildInfo is CommonBuildInfo {
   assertObjectHasKeys(buildInfo, 'buildInfo', commonKeys);
+  assert(
+    SUPPORTED_CHANNELS.includes((buildInfo as any).channel),
```
lerouxb, line 40: I don't know what would be the best way to do any of the changes in this file. But basically: I need the `channel` variable in common build info.

packages/compass-smoke-tests/src/cli.ts
```diff
@@ -145,6 +145,60 @@ async function getTestSubject(
...
+type Arch = 'arm64' | 'x64';
+
+type PlatformShortName =
+  | 'darwin-arm64'
+  | 'darwin-x64'
+  | 'windows'
+  | 'linux_deb'
+  | 'linux_rpm';
+
+function getPlatformShortName(
```
lerouxb, line 157: The URL for downloading the latest release ends in things like osx, darwin-arm64, darwin-x64, windows, linux_dev or linux_rpm. This just calculates that bit. I'm calling it a "platform short name" until we think of something better.

packages/compass-smoke-tests/src/cli.ts
```diff
@@ -196,6 +252,33 @@ async function run() {
...
+    const appName = buildInfo.productName;
+
+    const { appPath, uninstall } = install({
+      appName,
+      filepath: releasepath,
+      destinationPath: context.sandboxPath,
+    });
+
+    try {
+      runTest({ appName, appPath });
```
lerouxb, line 274: Again: This should actually see if you can update from this version. I'm just including it as a kind of sanity check that releasepath works.

packages/compass-smoke-tests/src/downloads.ts
```diff
@@ -18,6 +18,14 @@ export async function downloadFile({
 }: DownloadFileOptions): Promise<string> {
   const response = await fetch(url);
 
+  if (!targetFilename) {
+    // if no filename was specified, work out the filename based on the final
+    // URL the way a browser would
+    targetFilename = path.basename(new URL(response.url).pathname);
```
lerouxb, line 24: We don't actually know the filename in advance because we don't know the latest version number in advance. I'm open to suggestions for what to do here.

packages/compass-smoke-tests/src/downloads.ts
```diff
@@ -50,5 +58,7 @@ export async function downloadFile({
     fs.createWriteStream(outputPath)
   );
 
+  console.log(outputPath);
```
lerouxb, line 61: temporary ;)

packages/compass-smoke-tests/src/build-info.ts
```diff
@@ -8,6 +8,8 @@ import { type PackageKind } from './packages';
 import { type SmokeTestsContext } from './context';
 import { pick } from 'lodash';
 
+const SUPPORTED_CHANNELS = ['dev', 'beta', 'stable'] as const;
```
kraenhansen, line 11:
```suggestion
const SUPPORTED_CHANNELS = ['dev', 'beta', 'stable'] as const;
type Channel = typeof SUPPORTED_CHANNELS[number];
```

packages/compass-smoke-tests/src/build-info.ts
```diff
@@ -25,13 +27,21 @@ function assertObjectHasKeys(
 
 // subsets of the hadron-build info result
 
-export const commonKeys = ['productName'] as const;
-export type CommonBuildInfo = Record<typeof commonKeys[number], string>;
+export const commonKeys = ['productName', 'channel'] as const;
+export type CommonBuildInfo = Record<typeof commonKeys[number], string> & {
+  channel: 'dev' | 'beta' | 'stable';
```
kraenhansen, line 32: This could more easily be derived from `SUPPORTED_CHANNELS` (see the comment declaring `Channel` above).

```suggestion
  channel: Channel;
```

packages/compass-smoke-tests/src/downloads.ts
kraenhansen: Does it have to be stored with the "correct" filename? Alternatively, I would think there's a header with the filename 🤔

packages/compass-smoke-tests/src/downloads.ts
kraenhansen: We could introduce use of the `debug` package, that we're using elsewhere?

packages/compass-smoke-tests/src/downloads.ts
lerouxb: I looked for Content-Disposition, but alas:

```
URL https://mciuploads.s3.amazonaws.com/10gen-compass-main/e20763a4901d6f8dbea880c350b1e78d14afec4a_7792/mongodb-compass-25.1.21-dev.81257-darwin-arm64.dmg
Headers Headers {
  'x-amz-id-2': 'YiYH6YWPzeZUuCou36dhk8Xpf2Ndk1hl9pGVOZAHviyuam4zXCBNUYn/INzd6++vaL64VpWVLeE=',
  'x-amz-request-id': '1DN9Z06M06AR1RF2',
  date: 'Tue, 21 Jan 2025 16:08:38 GMT',
  'last-modified': 'Tue, 21 Jan 2025 08:31:07 GMT',
  'x-amz-expiration': 'expiry-date="Thu, 22 Jan 2026 00:00:00 GMT", rule-id="lifecycle-expiration"',
  etag: '"f1c5d713ab89bfdda5ea120dd007ecda-28"',
  'x-amz-server-side-encryption': 'AES256',
  'accept-ranges': 'bytes',
  'content-type': 'application/x-apple-diskimage',
  'content-length': '146989989',
  server: 'AmazonS3'
}
```

packages/compass-smoke-tests/src/downloads.ts
lerouxb: yeah. I suppose that's indicative of just how lazy I am 😆 I'll get to it ASAP..

packages/compass-smoke-tests/src/build-info.ts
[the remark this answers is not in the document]
lerouxb: buildInfo isn't a channel, though. it _has_ a channel. And it is already unknown.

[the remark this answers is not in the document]
lerouxb: Something like:

```typescript
export function assertCommonBuildInfo(
  buildInfo: unknown
): asserts buildInfo is CommonBuildInfo {
  assertObjectHasKeys(buildInfo, 'buildInfo', commonKeys);
  assert(
    SUPPORTED_CHANNELS.includes((buildInfo as { channel: Channel }).channel),
    `Expected ${
      (buildInfo as { channel: Channel }).channel
    } to be in ${SUPPORTED_CHANNELS.join(',')}`
  );
}
```

packages/compass-smoke-tests/src/downloads.ts
lerouxb: If we don't give it a filename it will probably be a bit confusing with the cache? It is also one way to try and parse out the version. Although I suppose we can also make a separate request to get the latest versions from somewhere.

lerouxb: And I suppose by the time we make a request to get the version we probably have everything we need to be able to calculate a filename up front 🤔

kraenhansen: > Does it have to be stored with the "correct" filename?

Could you just provide it some filename? I mean, does it need to be a specific filename with a particular version in the filename?

packages/compass-smoke-tests/src/build-info.ts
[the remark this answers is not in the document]
lerouxb: Already did. I added channel to commonKeys.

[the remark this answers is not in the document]
kraenhansen: Then why the `as { channel: Channel }`? 🤔

lerouxb:
Closing in favour of https://github.com/mongodb-js/compass/pull/6669

PR #6669, opened 2025-01-30
feat(smoke-tests): test updating from the latest release to the newly packaged app COMPASS-8532 COMPASS-8535

lerouxb:
Unfortunately because we skip the update tests on mac in CI and we don't have the windows .exe installer done yet (needed when installing the latest release windows), this won't actually run in CI yet.


packages/compass-e2e-tests/tests/auto-update.test.ts
```diff
@@ -13,66 +13,71 @@ function wait(ms: number) {
...
+        const compass = await init(`${testName} restart`, {
+          firstRun: false,
+        });
+        const { browser } = compass;
+        try {
+          await browser.$(Selectors.AutoUpdateToast).waitForDisplayed();
+          await browser
+            .$(Selectors.AutoUpdateReleaseNotesLink)
+            .waitForDisplayed();
+          // TODO: when updating to a known version we know the version, so
```
lerouxb, line 74: Same as above.

packages/compass-smoke-tests/src/build-info.ts
```diff
@@ -25,13 +29,21 @@ function assertObjectHasKeys(
...
+export type CommonBuildInfo = Record<typeof commonKeys[number], string> & {
+  channel: Channel;
+};
 
 export function assertCommonBuildInfo(
   buildInfo: unknown
 ): asserts buildInfo is CommonBuildInfo {
   assertObjectHasKeys(buildInfo, 'buildInfo', commonKeys);
+  assert(
+    SUPPORTED_CHANNELS.includes((buildInfo as { channel: Channel }).channel),
```
lerouxb, line 42: I still can't get TypeScript to infer that buildInfo is an object at this point. It thinks buildInfo is still any and not an object, so I can't just tell it that buildInfo.channel is a channel.

kraenhansen: One way would be to update the signature of `assertObjectHasKeys` to something which explicitly `asserts`:

https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXygGdCQYMB5AIwCtwMAJIgaRAE9CAeF9+EADwwhUwQvEIYYWVAHMA2gF0AfAAocNAFzw0Aa1Q4A7qgA08VFAC2ITeMkyT2toU3dCASk1ESZUWurwsogBK4DgwwBwYrAAOIDiIphYgJpExcfAO7IoA3ABQoJCwCGB44vC+HqisuZ6kFDR0jIQuqjQmAESUyFgQwACSqIg4bSayAOSDOKMmo5SwU-AzUABeo-IEosWo4q65vgB0nd19Azg5APRn8PAAegD8QA

```typescript
function assertObjectHasKeys<Keys extends string[]>(
  obj: unknown,
  name: string,
  keys: Keys
): asserts obj is Record<typeof name, typeof keys> {
  // ...
}
```

lerouxb: That's not gonna work. Name is the name of the object. We want Keys on the left of the Record<>. And all we can really say is that their values are `any` at this point.

kraenhansen: Sorry for the confusion, for some reason I though the name passed was the name of a property on the object 🙈

https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXygGdCQYMB5AIwCtwMAJIgaRAE9CAeAKHnhffggAHhhCpgheHCjA8EVvEIYYWVAHMA2gF0AND3hVqgkWInwASuBwxgHfoQ2pkAW0qld8NAGtUOAO6oAHx6gQAUODQAXJ6oPv6oOvCoUM4g0Uoq6olebITR9gCU0UQkZJIRRliShgDcXFwA9A2KyJQkGOWI8BgAFgg9MjB4ALSUyFgQwPCqiDhSIITIEBj1GKwADggAwgOoqCAQ8AC88ABEfgOiAG6kp3VgeErwAMoAqgAK7+TmACoAogARAD6W3oAEEAHIQv4AGWex3gGnOlxANxgpy0dS4wnW1gw8AeqCeD2czjw9gRGgA5OshsBkJgISkQFTElS0YRcKhWfAqWBdvsIFStARJISlHUcXjuhttjhSXgAELjSYASVQswRlgeNg4a02OC6JLJqHsjhcbhgHgyqjUgXgADJ4ABvfT8qB7A7RHYewV1AC+WKlZCQaEwXNFpQwW3lJuVE2A6tmoX0YwTSZw0W8vgCXCKkdIHXgabVGrmVXgMYVqHjpc1rt4JULhjojEI9lCJcTZbZXYzPON5NyBTqvDen2+-2BoMh0LhADpVJBkKBCJ2Vd3ZvP3Z7DkRKwKDgU9P6gA

packages/compass-e2e-tests/tests/auto-update.test.ts
```diff
@@ -13,66 +13,94 @@ function wait(ms: number) {
...
         await cleanup(compass);
       }
-    }
-  });
+
+      if (process.env.AUTO_UPDATE_UPDATABLE === 'true') {
+        console.log(
+          'pause to make sure the app properly exited before starting again'
+        );
+        await wait(10_000);
```
kraenhansen, line 74: Could we poll the `ps` with some `pid` instead?  🤔 Ideally from within the `cleanup` function.

packages/compass-e2e-tests/tests/auto-update.test.ts
```diff
@@ -13,66 +13,94 @@ function wait(ms: number) {
...
+          firstRun: false,
+        });
+        const { browser } = compass;
+        try {
+          await browser.$(Selectors.AutoUpdateToast).waitForDisplayed();
+          const releaseNotesLink = browser.$(
+            Selectors.AutoUpdateReleaseNotesLink
+          );
+          await releaseNotesLink.waitForDisplayed();
+          // for now we only know the new version in the auto-update-to case
```
kraenhansen, line 88: Would this be easier to get to now that we're running the update server from within the CLI process?

packages/compass-smoke-tests/src/build-info.ts
```diff
@@ -25,13 +28,21 @@ function assertObjectHasKeys(
...
+export type CommonBuildInfo = Record<typeof commonKeys[number], string> & {
+  channel: Channel;
+};
 
 export function assertCommonBuildInfo(
   buildInfo: unknown
 ): asserts buildInfo is CommonBuildInfo {
   assertObjectHasKeys(buildInfo, 'buildInfo', commonKeys);
+  assert(
+    SUPPORTED_CHANNELS.includes(buildInfo.channel as Channel),
```
kraenhansen, line 41: 🎉

packages/compass-smoke-tests/src/cli.ts
```diff
@@ -198,38 +200,69 @@ async function run() {
...
     for (const testName of context.tests) {
+      const installerPath =
+        testName === 'auto-update-to'
+          ? await getLatestRelease(
+              channel,
+              context.arch,
+              kind,
+              context.forceDownload
+            )
+          : filepath;
```
kraenhansen, line 225: Won't this depend on the order if multiple `tests` are passed to the CLI?
I suggest we make this a bit less powerful and run only one test per invocation of the CLI (effectively changing `context.tests` to `context.test`) to make it easier to reason about what installer needs to run first.

kraenhansen: Looking at this again, it does feel a bit misplaced here, instead of being a part of the actual test function, which would increase the cohesion 🤔

packages/compass-smoke-tests/src/releases.ts
```diff
@@ -0,0 +1,115 @@
...
+type Arch = 'arm64' | 'x64';
+
+type PlatformShortName =
+  | 'darwin-arm64'
+  | 'darwin-x64'
+  | 'windows'
+  | 'linux_deb'
+  | 'linux_rpm';
+
+function getPlatformShortName(
```
kraenhansen, line 13: I'd like a doc string on this function - especially the return value of the kinds confuses me.

packages/compass-smoke-tests/src/tests/auto-update-to.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export async function testAutoUpdateTo({
+  appName,
+  appPath,
+  autoUpdatable,
+  channel,
+  version,
+  bucketKeyPrefix,
+}: RunUpdateToTestOptions) {
+  process.env.PORT = '0'; // dynamic port
```
kraenhansen, line 23: Ideally, we would be able to pass this through the default function exported by the update server package, which would then be passed to the `UpdateChecker` constructor:

https://github.com/10gen/compass-mongodb-com/blob/a1fb99908815d3c3ab0efb5960430cc3faf99a15/src/update-checker.js#L40

kraenhansen (approved):
Great stuff overall - I have a few requests, non of which are necessarily blocking.

packages/compass-smoke-tests/src/tests/auto-update-to.ts
lerouxb: yeah it is a bit janky at the moment. That's not where the port is even used, though. update-checker is not the server itself. I think we can make a PR to compass-mongodb-com to allow us to clean it up, then another one here to use that. I'd do it as a follow-up, though, so we can have the tests running in the meantime.

packages/compass-e2e-tests/tests/auto-update.test.ts
lerouxb: Yeah, but where do we get the pid? I think we can clean it up but in an effort to get somewhere let's do that as a follow-up.

packages/compass-smoke-tests/src/releases.ts
lerouxb: What do you want the doc string to say?

packages/compass-smoke-tests/src/cli.ts
lerouxb: OK what I've done now is move all this stuff into the tests. I also changed it so that auto-update-to does not unnecessarily download the newly packaged app when it is going to be installing the latest release anyway.

Let me know what you think.

packages/compass-e2e-tests/tests/auto-update.test.ts
lerouxb: Maybe. Haven't thought of it. But if we do I'd just do it in a follow-up.

PR #6672, opened 2025-01-31
chore(telemetry): log when getting public cloud info fails

lerouxb:
This is with me hardcoding a `throw new Error('This is a test')`:

```
{"t":{"$date":"2025-01-31T10:48:37.186Z"},"s":"E","c":"COMPASS-TELEMETRY","id":1001000339,"ctx":"getPublicCloudInfo","msg":"Error fetching cloud info","attr":"This is a test"}
```

### DIFF
--- a/packages/compass/src/app/utils/telemetry.ts
+++ b/packages/compass/src/app/utils/telemetry.ts
@@ -5,6 +5,8 @@ import { getCloudInfo } from 'mongodb-cloud-info';
 import ConnectionString from 'mongodb-connection-string-url';
 import resolveMongodbSrv from 'resolve-mongodb-srv';
 import type { KMSProviders, MongoClientOptions } from 'mongodb';
+import { createLogger } from '@mongodb-js/compass-logging';
+const { log, mongoLogId } = createLogger('COMPASS-TELEMETRY');
 
 type HostInformation = {
   is_localhost: boolean;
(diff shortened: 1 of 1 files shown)


packages/compass/src/app/utils/telemetry.ts
```diff
@@ -38,6 +40,14 @@ async function getPublicCloudInfo(host: string): Promise<{
...
     };
   } catch (err) {
+    debug(`getCloudInfo failed for "${host}": ${(err as Error).message}`);
+
+    log.error(
+      mongoLogId(1_001_000_339),
+      'getPublicCloudInfo',
+      (err as Error).message,
+      { host }
+    );
```
addaleax, line 50: `log.error()` is intended for errors that should stand out in the application logs as they affect the ability of the application to run, which this isn't. We also generally only provide the error message as an attribute, and instead described what failed in the message text

```suggestion
    log.warn(
      mongoLogId(1_001_000_339),
      'getPublicCloudInfo',
      'Failed to look up host cloud information for telemetry',
      { host, error: (err as Error).message }
    );
```

addaleax (approved):

lerouxb:
weird:
https://parsley.mongodb.com/evergreen/10gen_compass_main_unit_tests_macos_test_electron_patch_139fa5df87a126d917014fe32c1f3da2c0bd5a6a_679cfe8dd2868d000729d0e4_25_01_31_16_47_10/0/task?bookmarks=0,91766

The hostname is test-000-shard-00-00.test.mongodb.net, but that's only used in a different test 🤔

addaleax (approved):

PR #6684, opened 2025-02-07
feat(smoke-tests): test tar installer for on Ubuntu COMPASS-8711

kraenhansen:
## Description

Merging this PR will:
1. Implement the Linux .tar installer (using the `tar` command and verifying compass `--version` via `xvfb-run`).
2. Arm CI with the Ubuntu build variant to execute the smoke tests for the Ubuntu tar installer (only running `time-to-first-query` for now).

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/.evergreen/functions.yml
+++ b/.evergreen/functions.yml
@@ -692,15 +692,15 @@ functions:
             npm run --unsafe-perm --workspace @mongodb-js/compass-smoke-tests start -- --package=osx_dmg --tests=time-to-first-query
           fi
 
-          #if [[ "$IS_UBUNTU" == "true" ]]; then
+          if [[ "$IS_UBUNTU" == "true" ]]; then
             # TODO: linux_deb
-            # TODO: linux_tar
-          #fi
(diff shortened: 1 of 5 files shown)


.evergreen/functions.yml
```diff
@@ -694,7 +694,7 @@ functions:
 
           #if [[ "$IS_UBUNTU" == "true" ]]; then
             # TODO: linux_deb
-            # TODO: linux_tar
+            npm run --unsafe-perm --workspace @mongodb-js/compass-smoke-tests start -- --package=linux_tar --tests=time-to-first-query
```
lerouxb, line 697: In this case it should be safe to run autoupdate-from because it won't auto-update, but it will tell the user that there is an update. And there's test code for that.

But we can also enable that in a separate PR.

lerouxb (approved):

PR #6702, opened 2025-02-10
feat(smoketests): run tests on readonly and isolated too COMPASS-8128

lerouxb:


.evergreen/functions.yml
```diff
@@ -678,8 +678,14 @@ functions:
 
           if [[ "$IS_WINDOWS" == "true" ]]; then
             npm run --unsafe-perm --workspace @mongodb-js/compass-smoke-tests start -- --package=windows_setup --tests time-to-first-query
-            npm run --unsafe-perm --workspace @mongodb-js/compass-smoke-tests start -- --package=windows_zip --tests auto-update-from
-            npm run --unsafe-perm --workspace @mongodb-js/compass-smoke-tests start -- --package=windows_msi --tests auto-update-from
+            if [[ "$HADRON_DISTRIBUTION" == "compass-isolated"]]; then
```
lerouxb, line 681: Part of me wants to keep this functions.yml clean so it specifies the full set for the default compass distribution case and then rather just override the tests parameter inside the cli to exclude the variants that won't work with the distribution.

packages/compass-smoke-tests/src/tests/read-only.ts
```diff
@@ -0,0 +1,57 @@
+import fs from 'node:fs';
+import { type SmokeTestsContext } from '../context';
+import { execute } from '../execute';
+import { getInstaller } from '../installers';
+import { createSandbox } from '../directories';
+import { getTestSubject } from '../test-subject';
+
+export async function testReadOnly(context: SmokeTestsContext) {
```
lerouxb, line 8: This whole file is basically a copy/paste of time-to-first-query.ts. We can probably refactor it.

PR #6708, opened 2025-02-11
fix(ci): fail the package job if the file is not signed on windows COMPASS-8950

lerouxb:


addaleax (approved):

kraenhansen (approved):

lerouxb:
I've cherry-picked this commit into https://github.com/mongodb-js/compass/pull/6709 because then I can use it to prove that that fix worked whereas if I merge it alone it will just make CI red.

PR #6709, opened 2025-02-11
fix: sign the windows setup .exe COMPASS-8945 COMPASS-8950

lerouxb:
* assert on required env vars in signtool.go
* log errors in signtool.go (the stdout and stderr redirect wasn't working)
* have @mongodb-js/signing-utils where signtool.go expects it
* actually assert that the setup .exe is signed when verifying artifacts


addaleax (approved):

packages/compass/package.json
```diff
@@ -231,6 +231,7 @@
     "@mongodb-js/my-queries-storage": "^0.22.0",
     "@mongodb-js/prettier-config-compass": "^1.2.0",
     "@mongodb-js/sbom-tools": "^0.7.0",
+    "@mongodb-js/signing-utils": "^0.3.7",
```
lerouxb, line 234: So the bug was that signtool.go is doing `node -e 'require("@mongodb-js/signing-utils")'` from the packages/compass folder and that node module doesn't exist there.

In the past it probably worked because the dep was hoisted and then it broke because now it is not.

lerouxb: I think doing something like `npx run sign -w @mongodb-js/hadron-build -- path-to-file` from signtool.go would be a cleaner fix. I'll try that in a follow-up.

packages/hadron-build/signtool/signtool.go
```diff
@@ -55,12 +59,15 @@ func main() {
...
 
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
+	stdoutStderr, err := cmd.CombinedOutput()
 
-	err := cmd.Run()
 	if err != nil {
 		fmt.Println("Error signing the file")
+		fmt.Printf("%s\n", stdoutStderr)
+		log.Println(err)
```
lerouxb, line 67: Ideally this would be `log.fatal(err)` but then we don't see any of the logging and electron winstaller throws a huge error, none of which is helpful. Maybe it doesn't flush its subprocess's output? I'm not sure.

gribnoysup (approved):

PR #6717, opened 2025-02-12
chore(schema): move non serializable services off of state to services

Anemy:
Follow up cleanup from https://github.com/mongodb-js/compass/pull/6668


paula-stacho (approved):
Neat! 🎉

PR #6723, opened 2025-02-14
chore(schema, telemetry): add and update schema telemetry events COMPASS-8818

Anemy:
COMPASS-8818

Adds a new telemetry events: `Schema Analysis Cancelled`.
Updates the existing `Schema Exported` and `Schema Analyzed` to have new events.
Updated the way we gather schema metadata for the events to make it so that we only traverse the tree once instead of twice, also makes it so that we unblock the thread occasionally while traversing this tree in case of a very large tree.

We're only tracking the export when folks click the copy button, or click the export button, see discussion in https://mongodb.slack.com/archives/C07JL273J06/p1739389537801429

We only track the `optional_field_count` and `variable_type_count` properties for top level fields re https://mongodb.slack.com/archives/C07JL273J06/p1739398926141009

### DIFF
--- a/packages/compass-schema/src/modules/schema-analysis.spec.ts
+++ b/packages/compass-schema/src/modules/schema-analysis.spec.ts
@@ -418,5 +418,228 @@ describe('schema-analysis', function () {
         });
       });
     });
+
+    describe('variable_type_count', function () {
+      describe('with fields having multiple types', function () {
+        let schema: Schema;
+        before(async function () {
+          schema = await mongoDBSchemaAnalyzeSchema([
+            { pineapple: 123 },
+            { pineapple: 'string' },
+            { pineapple: true },
+            { singleType: 'onlyString' },
+          ]);
+        });
+
+        it('counts fields with more than one type', async function () {
+          const { variable_type_count } = await calculateSchemaMetadata(schema);
+          expect(variable_type_count).to.equal(1);
+        });
+      });
+
+      describe('with no fields', function () {
+        let schema: Schema;
+        before(async function () {
+          schema = await mongoDBSchemaAnalyzeSchema([{}]);
+        });
+
+        it('returns zero', async function () {
+          const { variable_type_count } = await calculateSchemaMetadata(schema);
+          expect(variable_type_count).to.equal(0);
+        });
+      });
+
+      describe('with all single-type fields', function () {
+        let schema: Schema;
+        before(async function () {
+          schema = await mongoDBSchemaAnalyzeSchema([
+            { a: 1 },
+            { a: 2, b: 2 },
+            { a: 2, c: 3 },
+          ]);
+        });
+
+        it('returns zero', async function () {
+          const { variable_type_count } = await calculateSchemaMetadata(schema);
+          expect(variable_type_count).to.equal(0);
+        });
+      });
+
+      describe('with differing nested fields fields', function () {
+        let schema: Schema;
+        before(async function () {
+          schema = await mongoDBSchemaAnalyzeSchema([
+            { required: { notAlwaysPresent: 'yes' } },
+            { required: { notAlwaysPresent: 123456 } },
+            { required: { sometimesHere: true } },
+          ]);
+        });
+
+        it('returns does not count the optionals', async function () {
+          const { variable_type_count } = await calculateSchemaMetadata(schema);
+          expect(variable_type_count).to.equal(0);
+        });
+      });
+    });
+
+    describe('optional_field_count', function () {
+      describe('with optional fields', function () {
+        let schema: Schema;
+        before(async function () {
+          schema = await mongoDBSchemaAnalyzeSchema([
+            { pineapple: 'yes', optional: 123 },
+            { pineapple: 'yes' },
+            { pineapple: 'yes', optional: 'maybe' },
+          ]);
+        });
+
+        it('counts fields missing from some documents', async function () {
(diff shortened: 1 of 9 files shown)


packages/compass-schema/src/stores/schema-analysis-reducer.ts
```diff
@@ -164,10 +164,58 @@ export const geoLayersDeleted = (
...
+  userCancelled = false
+): SchemaThunkAction<void> => {
   return (dispatch, getState, { analysisAbortControllerRef }) => {
     if (!analysisAbortControllerRef.current) return;
-    analysisAbortControllerRef.current?.abort();
+    analysisAbortControllerRef.current?.abort(
+      userCancelled ? userCancelledAnalysisAbortReason : undefined
+    );
+  };
+};
```
gribnoysup, line 181: I would suggest not to overload actions with meaning like that, it always makes colocating logic and managing them harder in the long run when you're trying to make them too generic (it's like not exactly the "don't use setters" point from the style guide, but it's similar in what effects it has). You have a pretty clear user action here, clicking the cancel button, it should abort the analysis and track the interaction, if you align this action with the actual event in the app (user clicked a cancel button), then you don't need to do this text-based check in the error handler of the analysis method, all the logic related to user-triggered cancelation is stored in one place. There is a question of storing the start time somewhere, and as it becomes state shared between actions, you can add it to the store itself, reducer is well positioned to preserve it on the "start" action.

For the "cleanup", I'd suggest to either create a special action for it, or, because it just need to call abort on a few controllers and controllers are created in the scope of the activate method, you can call abort there directly without the need for it to be a redux store action.

paula-stacho:
should we distinguish between copy and download?

packages/compass-schema/src/modules/schema-analysis.ts
```diff
@@ -91,76 +81,148 @@ export const analyzeSchema = async (
...
+    [bsonType: string]: number;
+  } = {};
+  let variableTypeCount = 0;
+  let optionalFieldCount = 0;
+  let unblockThreadCounter = 0;
+
+  async function traverseSchemaTree(
+    fieldsOrTypes: SchemaField[] | SchemaType[],
+    isRoot = false
+  ): Promise<number> {
```
paula-stacho, line 146: I find it a bit odd that the function calculates the 4 things and returns one of them. Can we make it return all? `{ schemaDepth, variableTypeCount, ... }`

packages/compass-schema/src/modules/schema-analysis.spec.ts
```diff
@@ -250,164 +246,399 @@ describe('schema-analysis', function () {
...
-              type: 'Point',
-              coordinates: [-73.856077, 40.848447],
+    describe('field_types', function () {
+      describe('with mixed bson types', function () {
+        let schema: Schema;
+        before(async function () {
+          schema = await mongoDBSchemaAnalyzeSchema([
+            { number: 42, value: true, string: 'test' },
+            {
+              number: 100,
```
paula-stacho, line 562: this is a nit, but if you remove the second number, this test will not overlap with the `type occurring multiple times`. I noticed this because I was first confused reading this test, because I counted 2 numbers :D

packages/compass-schema/src/modules/schema-analysis.ts
Anemy: Removed the returning so it's now just top level.

packages/compass-schema/src/stores/schema-analysis-reducer.ts
Anemy: Good thinking, added a new action for cleaning up the analysis. Removed the reason checking and added the tracking in the stop analysis action. Moved the start time to the store 👍

packages/compass-schema/src/modules/schema-analysis.ts
paula-stacho: Thank you!

packages/compass-schema/src/stores/schema-analysis-reducer.ts
```diff
@@ -233,6 +302,10 @@ export const startAnalysis = (): SchemaThunkAction<
         driverOptions,
         logger
       );
+      if (abortSignal?.aborted) {
+        throw new Error(abortSignal?.reason || new Error('Operation aborted'));
+      }
+
```
paula-stacho, line 308: Do we want this to be an error even though it was an user action? Especially since now it's difficult to get to a re-try, the user basically needs to close&reopen the tab:
Screenshot 2025-02-19 at 10 56 01

paula-stacho: Sorry that I didn't notice this sooner

paula-stacho: Note: I'll be changing the error handling in https://github.com/mongodb-js/compass/pull/6733, I can include check for this case and show a warning toast, if we some notification for the user.

paula-stacho: Toast vs in-tab error TBD. For the question on whether intentional abort is an error - I checked what we have in other places. Looks like aggregation goes to "no results" state if you cancel.

paula-stacho: Looks like both this and the errors in general don't have a clear strategy :/

PR #6727, opened 2025-02-17
chore(smoke-tests): dispatch workflow run on GitHub Actions from Evergreen COMPASS-8751

kraenhansen:
## Description

Merging this PR will:
- Arm Evergreen to run smoke tests by dispatching a workflow run on GitHub actions and awaiting a successful completion.

Ideally, we would simply run `gh workflow run --watch` but that's not supported and blocked by https://github.com/cli/cli/issues/4001.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/.github/workflows/test-installers.yml
+++ b/.github/workflows/test-installers.yml
@@ -6,10 +6,6 @@ permissions:
 on:
   workflow_dispatch:
     inputs:
-      version:
-        type: string
-        description: 'Version of the installer to download'
-        required: true
       bucket_name:
         type: string
         description: 'S3 bucket to download installers from'
@@ -18,11 +14,14 @@ on:
         type: string
         description: 'S3 bucket key prefix to download installers from'
         required: true
+      dev_version:
+        type: string
+        description: 'Dev version of the installer to download'
       nonce:
         type: string
         description: 'A random string to track the run from dispatch to watching'
 
-run-name: Test Installers ${{ github.event.inputs.version }} / (nonce = ${{ github.event.inputs.nonce || 'not set' }})
+run-name: Test Installers ${{ github.event.inputs.dev_version || github.ref_name }} / (nonce = ${{ github.event.inputs.nonce || 'not set' }})
 
 jobs:
   test:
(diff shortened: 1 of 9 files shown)


packages/compass-smoke-tests/src/dispatch.ts
```diff
@@ -0,0 +1,188 @@
...
+      );
+
+      await new Promise((resolve) => setTimeout(resolve, sanityDelayMs));
+      debug('Getting latest run again - expecting it to be the same');
+      const latestRun = await getLatestRun(octokit);
+      assert.equal(
+        latestRun.run_number,
+        expectedRunNumber,
+        'Runs are being dispatched too quickly, for the heuristics to be accurate.'
+      );
```
kraenhansen, line 83: Doing a sanity check by waiting some time (default is 10s) from the workflow run has been identified to re-fetch and assert this is still the latest run. This will make the command fail if there's too many runs being dispatched for the heuristics (i.e. the latest run is the one we just dispatched) to be accurate.

.github/workflows/test-installers.yml
```diff
@@ -6,6 +6,10 @@ permissions:
 on:
   workflow_dispatch:
     inputs:
+      version:
+        type: string
+        description: 'Version of the installer to download'
+        required: true
```
lerouxb, line 12: Will it work with a required version? Do we know the version for beta and stable, for example? Wasn't that only really used for the dev version?

kraenhansen: 😬 you're probably right. This would happen with a run off `main` or one of the release branches, right?

lerouxb: If I remember correctly it is optional so that you specify it for dev, but then for beta/stable it gets read out of package.json. Probably indirectly in hadron build info somewhere.

packages/compass-smoke-tests/src/cli.ts
```diff
@@ -67,51 +80,116 @@ const argv = yargs(hideBin(process.argv))
...
+        })
+        .option('ref', {
+          type: 'string',
+          description: 'Git reference to dispatch the workflow from',
+          default: getDefaultRef(),
+        })
+        .option('version', {
+          type: 'string',
+          description: 'Compass version to run tests for',
+          default: process.env.DEV_VERSION_IDENTIFIER,
```
lerouxb, line 150: You're going to have to get version from hadron-build info, I think. So it will work for beta and stable releases and not just PRs or dev builds.

lerouxb: Currently when smoke testing it forms part of the filename that is in type PackageDetails. So we don't have an explicit version. You seem to only be using it for the task name so you can probably come up with something that just means beta or stable in that case? There might even be an env or evergreen var already 🤔

kraenhansen:
Moved this back into draft, since I just realized I likely need some changes to make it compatible with non-dev builds 🤔

kraenhansen:
I want to simulate a beta release too, ideally before merging this 🤞

lerouxb (approved):

PR #6738, opened 2025-02-20
test(sandbox): fix sandbox tests after protocol change COMPASS-8976

nvs119:
## Description
Updating sandbox to follow websocket protocol changes in MMS

### Checklist
- [x] Working locally with sandbox
- [x] e2e patch: https://spruce.mongodb.com/version/67b7763121b2e60007432cd2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-web/polyfills/net/index.ts
+++ b/packages/compass-web/polyfills/net/index.ts
@@ -145,7 +145,7 @@ class Socket extends Duplex {
       return message.subarray(1);
     } else {
       // eslint-disable-next-line no-console
-      console.error('message does not have valid type byte "%s":', message);
+      console.error('message does not have valid type byte "%s"', message);
     }
   }
 }
--- a/packages/compass-web/polyfills/net/index.ts
+++ b/packages/compass-web/polyfills/net/index.ts
@@ -144,6 +144,7 @@ class Socket extends Duplex {
     } else if (typeByte === MESSAGE_TYPE.BINARY) {
       return message.subarray(1);
     } else {
+      // eslint-disable-next-line no-console
       console.error('message does not have valid type byte "%s":', message);
     }
   }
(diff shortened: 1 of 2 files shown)


packages/compass-web/polyfills/net/index.ts
```diff
@@ -110,6 +116,37 @@ class Socket extends Duplex {
...
+    const typeByte = message[0];
+    if (typeByte === MESSAGE_TYPE.JSON) {
+      const jsonBytes = message.subarray(1);
+      const textDecoder = new TextDecoder('utf-8');
+      const jsonStr = textDecoder.decode(jsonBytes);
+      return JSON.parse(jsonStr);
+    } else if (typeByte === MESSAGE_TYPE.BINARY) {
+      return message.subarray(1);
+    } else {
+      console.error('message does not have valid type byte "%s":', message);
```
Anemy, line 147: The `check` script for this package is failing because of the no-console lint rule:
```
error  Unexpected console statement  no-console
```
To try it locally you can run `npm run check` in the `packages/compass-web` folder. 
If these are useful we could add eslint ignores. We also use the `debug` package in a few places in Compass, I'm not sure if that would be a better thing to use here though.

Anemy (approved):
lgtm, the check is failing though so we should fix that first

packages/compass-web/polyfills/net/index.ts
nvs119: Yeah I think it's important enough to leave since it would mean the protocol really screwed up OR some very unexpected type of message is coming through the pipeline. I'll add `// eslint-disable-next-line no-console`. And thanks for the `:` catch! Removing

PR #6740, opened 2025-02-21
feat(schema): File download COMPASS-8704

paula-stacho:
## Description
Doing the file download with blob + createObjectUrl solution. Tested in electron and chrome. Including e2e

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-e2e-tests/helpers/compass.ts
+++ b/packages/compass-e2e-tests/helpers/compass.ts
@@ -2,7 +2,6 @@ import { inspect } from 'util';
 import { ObjectId, EJSON } from 'bson';
 import { promises as fs, rmdirSync } from 'fs';
 import type Mocha from 'mocha';
-import path from 'path';
 import os from 'os';
 import { execFile } from 'child_process';
 import type { ExecFileOptions, ExecFileException } from 'child_process';
@@ -42,6 +41,8 @@ import {
   ELECTRON_PATH,
 } from './test-runner-paths';
 import treeKill from 'tree-kill';
+import { downloadPath } from './downloads';
+import path from 'path';
 
 const killAsync = async (pid: number, signal?: string) => {
   return new Promise<void>((resolve, reject) => {
@@ -677,6 +678,13 @@ async function startCompassElectron(
 
   try {
     browser = (await remote(options)) as CompassBrowser;
+    // https://webdriver.io/docs/best-practices/file-download/#configuring-chromium-browser-downloads
+    const page = await browser.getPuppeteer();
+    const cdpSession = await page.target().createCDPSession();
+    await cdpSession.send('Browser.setDownloadBehavior', {
+      behavior: 'allow',
+      downloadPath: downloadPath,
+    });
   } catch (err) {
     debug('Failed to start remote webdriver session', {
       error: (err as Error).stack,
@@ -755,6 +763,26 @@ export async function startBrowser(
   runCounter++;
   const { webdriverOptions, wdioOptions } = await processCommonOpts();
 
+  const browserCapabilities: Record<string, Record<string, unknown>> = {
+    chrome: {
+      'goog:chromeOptions': {
+        prefs: {
+          'download.default_directory': downloadPath,
+        },
+      },
+    },
+    firefox: {
+      'moz:firefoxOptions': {
+        prefs: {
+          'browser.download.dir': downloadPath,
+          'browser.download.folderList': 2,
+          'browser.download.manager.showWhenStarting': false,
+          'browser.helperApps.neverAsk.saveToDisk': '*/*',
+        },
+      },
+    },
+  };
+
   // webdriverio removed RemoteOptions. It is now
   // Capabilities.WebdriverIOConfig, but Capabilities is not exported
   const options = {
@@ -763,6 +791,7 @@ export async function startBrowser(
       ...(context.browserVersion && {
         browserVersion: context.browserVersion,
       }),
+      ...browserCapabilities[context.browserName],
 
       // from https://github.com/webdriverio-community/wdio-electron-service/blob/32457f60382cb4970c37c7f0a19f2907aaa32443/packages/wdio-electron-service/src/launcher.ts#L102
       'wdio:enforceWebDriverClassic': true,
--- a/packages/compass-schema/src/components/export-schema-modal.tsx
+++ b/packages/compass-schema/src/components/export-schema-modal.tsx
@@ -206,6 +206,7 @@ const ExportSchemaModal: React.FunctionComponent<{
         </Button>
         <Button
           variant="primary"
+          isLoading={exportStatus === 'inprogress'}
           loadingIndicator={<SpinLoader />}
           disabled={!exportedSchema}
           onClick={handleSchemaDownload}
--- a/packages/compass-schema/src/components/export-schema-modal.tsx
+++ b/packages/compass-schema/src/components/export-schema-modal.tsx
@@ -195,6 +195,7 @@ const ExportSchemaModal: React.FunctionComponent<{
             variant="primary"
             isLoading={!downloadUrl}
             loadingIndicator={<SpinLoader />}
+            disabled={!exportedSchema}
           >
             Export
           </Button>
--- a/packages/compass-schema/src/components/export-schema-modal.tsx
+++ b/packages/compass-schema/src/components/export-schema-modal.tsx
@@ -14,7 +14,6 @@ import {
   ErrorSummary,
   Label,
   CancelLoader,
-  Link,
   SpinLoader,
 } from '@mongodb-js/compass-components';
 import { CodemirrorMultilineEditor } from '@mongodb-js/compass-editor';
@@ -190,17 +189,18 @@ const ExportSchemaModal: React.FunctionComponent<{
         <Button onClick={onClose} variant="default">
           Cancel
         </Button>
-          <Button
-            variant="primary"
-            isLoading={!downloadUrl}
-            loadingIndicator={<SpinLoader />}
-            disabled={!exportedSchema}
-             download={filename}
-             href={downloadUrl}
-             onClick={onExportedSchema}
-          >
-            Export
-          </Button>
+        <Button
+          variant="primary"
+          isLoading={!downloadUrl}
+          loadingIndicator={<SpinLoader />}
+          disabled={!exportedSchema}
+          download={filename}
+          href={downloadUrl}
+          onClick={onExportedSchema}
+          data-testid="schema-export-download-button"
+        >
+          Export
+        </Button>
       </ModalFooter>
     </Modal>
   );
--- a/packages/compass-schema/src/components/export-schema-modal.tsx
+++ b/packages/compass-schema/src/components/export-schema-modal.tsx
@@ -83,6 +83,7 @@ const ExportSchemaModal: React.FunctionComponent<{
   exportFormat: SchemaFormat;
   exportedSchema?: string;
   blobToDownload?: Blob;
+  filename?: string;
   onCancelSchemaExport: () => void;
   onChangeSchemaExportFormat: (format: SchemaFormat) => Promise<void>;
   onClose: () => void;
@@ -95,6 +96,7 @@ const ExportSchemaModal: React.FunctionComponent<{
   exportFormat,
   exportedSchema,
   blobToDownload,
+  filename,
   onCancelSchemaExport,
   onChangeSchemaExportFormat,
   onClose,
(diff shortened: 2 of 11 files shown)


packages/compass-schema/src/stores/schema-export-reducer.ts
```diff
@@ -201,6 +226,27 @@ export const trackSchemaExported = (): SchemaThunkAction<void> => {
...
+        type: 'application/json',
+      });
+      const filename = `schema-${exportFormat}-${namespace}.json`;
+      dispatch({
+        type: SchemaExportActions.schemaDownloadReady,
+        blob,
+        filename,
+      });
+    } catch (error) {
+      _trackSchemaExportFailed('prepareDownload');
```
gribnoysup, line 245: This is not wrapped in a dispatch and so does nothing (well, it returns a function that does nothing), as I mentioned in another PR, don't wrap this logic into "fake" actions if possible, if you want to generalize it into a function, make sure that it's a normal function with required "services" passed directly as function arguments, otherwise it's really easy to miss cases where you call something, but it does nothing like that

packages/compass-schema/src/components/export-schema-modal.tsx
```diff
@@ -177,14 +190,16 @@ const ExportSchemaModal: React.FunctionComponent<{
...
+        <Link download={filename} href={downloadUrl} onClick={onExportedSchema}>
+          <Button
+            variant="primary"
+            isLoading={!downloadUrl}
+            loadingIndicator={<SpinLoader />}
+            disabled={!exportedSchema}
+          >
+            Export
+          </Button>
+        </Link>
```
gribnoysup, line 202: Leafygreen buttons support being rendered as links out of the box

```suggestion
          <Button
            variant="primary"
            isLoading={!downloadUrl}
            loadingIndicator={<SpinLoader />}
            disabled={!exportedSchema}
             download={filename}
             href={downloadUrl}
             onClick={onExportedSchema}
          >
            Export
          </Button>
```

paula-stacho: Ah thanks! ❤️

packages/compass-schema/src/stores/schema-export-reducer.ts
```diff
@@ -27,6 +27,8 @@ export type SchemaExportState = {
   exportFormat: SchemaFormat;
   errorMessage?: string;
   exportStatus: ExportStatus;
+  blobToDownload?: Blob;
```
gribnoysup, line 30: Non-serializeable values don't belong in the state, if we have to keep the url in the state, it's probably better to make it the serialized object url itself, not the blob. It should also help with reducing clogging the memory when people are clicking download in the app as every new object URL [will be kept in memory](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static#memory_management) while the page exists (so basically until you close the browser tab / compass app) and so generating it once (and disposing the old one when the new one is generated) should make it a bit less memory intensive

gribnoysup:
> but I've run into some security issues with link.clicking programmatically.

I'm curious what was the case when it happened. Generally browsers allow for this if the click is triggered by another user action, so if this programmatic link clicking is triggered by a user button click for example, browsers should not prevent that

packages/compass-schema/src/stores/schema-export-reducer.ts
paula-stacho: Thank you for the catch 🙈

paula-stacho:
> > but I've run into some security issues with link.clicking programmatically.
> 
> I'm curious what was the case when it happened. Generally browsers allow for this if the click is triggered by another user action, so if this programmatic link clicking is triggered by a user button click for example, browsers should not prevent that
> 
> (I tried naively replacing the href with a button that creates a link and clicks it on click event and it seems to work both in local sandbox and when using browser extension to run this code in cloud)


Aaaah 🤦 So now I spent some more time on it and I think the issue was appending the temporary <a> element to document.body. Appending it inside the modal appears to work (at least in electron). I don't hate the current solution though, so I'm thinking to keep it.

packages/compass-schema/src/stores/schema-export-reducer.ts
paula-stacho: ah thank you! I completely forgot since Friday that I need to do a cleanup for this

paula-stacho:
> > So now I spent some more time on it and I think the issue was appending the temporary element to document.body
> 
> If you're doing this from inside the modal, the modal logic is probably conflicting with the click (either the focus trap or a pointer-events directive somewhere I'd assume). FWIW you don't need to add it in the DOM at all to click, you can just call the method on the element. Semantics are good, but I'd consider to think about what other things this affects: keeping blob in state in addition to the schema basically means we're keeping schema in memory two times, then you create the object URL and it makes it three times, I'm not sure how much it's worth it, especially as this is not a "real" URL anyway

Okay, reverting to the temporary link solution, doing the cleanup wasn't too bad but you're right that the objects can get big and we shouldn't need to store two (the url itself is small so I'm not counting that)

packages/compass-e2e-tests/helpers/compass.ts
```diff
@@ -677,6 +678,13 @@ async function startCompassElectron(
 
   try {
     browser = (await remote(options)) as CompassBrowser;
+    // https://webdriver.io/docs/best-practices/file-download/#configuring-chromium-browser-downloads
+    const page = await browser.getPuppeteer();
+    const cdpSession = await page.target().createCDPSession();
+    await cdpSession.send('Browser.setDownloadBehavior', {
+      behavior: 'allow',
+      downloadPath: downloadPath,
+    });
```
gribnoysup, line 687: Just curious why it's different from what we do for the browser below where it's configured through chromeOptions. Asking because we do pass `goog:chromeOptions` to electron already and they seem to be applied correctly 🤔

packages/compass-schema/src/components/export-schema-modal.tsx
```diff
@@ -104,6 +112,25 @@ const ExportSchemaModal: React.FunctionComponent<{
...
+      link.href = url;
+      link.download = filename || 'export.json';
+      link.click();
+      window.URL.revokeObjectURL(url);
+      onExportedSchema();
+    } catch (error) {
+      onSchemaExportFailed('download button clicked');
+      throw error;
+    }
+  }, [exportedSchema, filename, onSchemaExportFailed, onExportedSchema]);
```
gribnoysup, line 132: This whole thing looks like a store action to me that doesn't necessarily needs to be part of the UI: all this state is coming from the store, the tracking "stages" feel like a bit too much info to give UI control over, and we probably don't want to throw in the click handler in the UI even if it failed, this should probably either dispatch an action of its own if we're planning to store and map it to some UI or just logged and ignored if not

packages/compass-e2e-tests/helpers/compass.ts
paula-stacho: Hmm.. this is how I understood the best practices docs, but let me try it out with just the options

paula-stacho: Not out of the box, perhaps if we found an option to disable the 'save as' dialog

packages/compass-schema/src/components/export-schema-modal.tsx
```diff
@@ -104,6 +112,25 @@ const ExportSchemaModal: React.FunctionComponent<{
...
+      if (!exportedSchema) return;
+      const blob = new Blob([exportedSchema], {
+        type: 'application/json',
+      });
+      const url = window.URL.createObjectURL(blob);
+      const link = document.createElement('a');
+      link.href = url;
+      link.download = filename || 'export.json';
+      link.click();
+      window.URL.revokeObjectURL(url);
```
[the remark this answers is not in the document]
paula-stacho: But the resource is the blob, I suppose the download prevents it being cleaned up, we're just unblocking the url path to it.

packages/compass-schema/src/components/export-schema-modal.tsx
```diff
@@ -178,10 +184,12 @@ const ExportSchemaModal: React.FunctionComponent<{
...
         </Button>
         <Button
-          onClick={() => {
-            /* TODO(COMPASS-8704): download and track with trackSchemaExported */
-          }}
           variant="primary"
+          isLoading={exportStatus === 'inprogress'}
+          loadingIndicator={<SpinLoader />}
+          disabled={!exportedSchema}
+          onClick={onSchemaDownload}
```
Anemy, line 191: Are we planning on adding a visual indicator that the file is saving & saved? Right now the save dialogue closes and they're back to the export page which they would then cancel out of. Looking at the designs there is an `in progress` toast included:
https://www.figma.com/design/CBHJriBQZ2qxMxtnXFgDxJ/EXPO-3174-%3A-Export-Schema-%2B-Validation-Detection?node-id=1298-55009&t=MBxcSDens4RhwtPM-1

packages/compass-schema/src/components/export-schema-modal.tsx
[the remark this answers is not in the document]
gribnoysup: Yeah, that's probably the case! I think I'd sleep better if we schedule revoke after the click instead of immediately calling it syncronously after the click (so like `setTimeout(() => URL.revokeObjectURL(...), 0)`), but that's a super unimportant nit and might be already what happens in the background (I don't want to check), so feel free to ignore it 🙂

packages/compass-schema/src/stores/schema-export-reducer.ts
```diff
@@ -150,6 +153,48 @@ async function getSchemaByFormat({
...
+      const blob = new Blob([exportedSchema], {
+        type: 'application/json',
+      });
+      const url = window.URL.createObjectURL(blob);
+      const link = document.createElement('a');
+      link.href = url;
+      link.download = filename || 'export.json';
+      link.click();
+      window.URL.revokeObjectURL(url);
+      dispatch(trackSchemaExported);
```
gribnoysup, line 172: You need to call it so that it actually tracks, passing a function to the dispatch technically "works" here and redux will call it for you, but it's not a thunk action that this function returns it will be calling, it will be just a function that returns another. I am once again strongly suggesting not to do tracking through "fake" actions and just define clear functions that you can explicitly call inside the actions and pass all required arguments

```
function trackExported(trackFn, schema, source, fmt) {
  trackFn(...)
}

// ...

function downloadSchema(): ThunkAction = () => {
  return (dispatch, state, { track }) => {
    // ...
    trackExported(track, ...)
  }
}
```

it's a bit more code, yes, you'd kinda have to pick the state and the thunk arguments every time you need to use it, but it's also way harder to mess up in that case.

I'd also recommend to explicitly define types of actions that you can dispatch from the thunk in the second argument of `SchemaThunkAction`, this will also help a lot with not noticing issues like that (but will not help with other cases like forgetting to wrap a function call in a dispatch)

packages/compass-web/sandbox/index.tsx
```diff
@@ -127,6 +127,7 @@ const App = () => {
             onActiveWorkspaceTabChange={updateCurrentTab}
             initialWorkspace={currentTab ?? undefined}
             initialPreferences={{
+              enableExportSchema: true,
```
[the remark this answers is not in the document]
paula-stacho: I'm actually thinking to enable this + the tests for web already, what do you think?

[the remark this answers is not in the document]
Anemy: Sounds good to me, let's remember to unset this after we enable it so that it starts using the default when we merge the feature flag pr.

packages/compass-schema/src/components/export-schema-modal.tsx
Anemy: I think when we have toasts that show that the file was saved, closing the modal makes sense.

packages/compass-schema/src/stores/schema-export-reducer.ts
```diff
@@ -150,54 +155,118 @@ async function getSchemaByFormat({
...
+
+      const link = document.createElement('a');
+      link.href = url;
+      link.download = filename || 'export.json';
+      link.click();
+      setTimeout(() => {
+        window.URL.revokeObjectURL(url);
+        link.remove();
+      }, 0);
+      dispatch(trackSchemaExported);
```
gribnoysup, line 178: Still not doing anything because you are passing the thunk action creator to dispatch, not thunk action itself

```suggestion
      dispatch(trackSchemaExported());
```

paula-stacho: oh my! thank you for being so thorough with the reviews, I really need to slow down and pay more attention

packages/compass-e2e-tests/helpers/compass.ts
gribnoysup: Got it, interesting, thanks for checking!

gribnoysup (approved):

PR #6744, opened 2025-02-24
chore(schema): add tooltip showing folks the export button when they legacy export schema COMPASS-8971

Anemy:
COMPASS-8971

This'll go away eventually. Didn't add any tests, I can if folks would like.
Drive by renamed the `banner` to `modal` as it is a modal.


paula-stacho:
Thank you! No tests needed. Just one thing - if it's not too complicated, it would be nice to show also if they switch to the new export. Since they came in the legacy way it might be they haven't noticed the button.

packages/compass-schema/src/stores/schema-export-reducer.ts
```diff
@@ -238,7 +238,9 @@ export const changeExportSchemaFormat = (
     try {
       const schemaAccessor = schemaAccessorRef.current;
       if (!schemaAccessor) {
-        throw new Error('No schema analysis available');
+        throw new Error(
+          "No schema analysis available. Please analyze the collection's schema before exporting."
```
Anemy, line 242: Drive-by, updated this error message to be a bit more descriptive as folks do see it if they try exporting and they have not yet run analysis.

Anemy:
@paula-stacho Pushed a commit to show the tooltip after the new export modal has closed after using the legacy menu entry. Looks like there was an simple way to do it after all.

PR #6748, opened 2025-02-25
chore(compass-shell): update mongosh logging

gagik:
Updates and applies the breaking changes from `@mongosh/logging`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43549,7 +43549,7 @@
         "mongodb-connection-string-url": "^3.0.1",
         "mongodb-data-service": "^22.25.2",
         "mongodb-instance-model": "^12.26.2",
-        "mongodb-log-writer": "^2.1.0",
+        "mongodb-log-writer": "^2.3.4",
         "mongodb-ns": "^2.4.2",
         "react": "^17.0.2",
         "react-dom": "^17.0.2",
@@ -44362,7 +44362,7 @@
         "mongodb": "^6.12.0",
         "mongodb-build-info": "^1.7.2",
         "mongodb-connection-string-url": "^3.0.1",
-        "mongodb-log-writer": "^2.1.0",
+        "mongodb-log-writer": "^2.3.4",
         "mongodb-runner": "^5.6.3",
         "node-fetch": "^2.7.0",
         "nyc": "^15.1.0",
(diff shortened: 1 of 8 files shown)


packages/compass-shell/src/modules/worker-runtime.ts
```diff
@@ -52,30 +52,36 @@ export function createWorkerRuntime(
...
     },
-    {
+    userTraits: {
       platform: process.platform,
       arch: process.arch,
     },
     // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-var-requires
-    require('../../package.json').version
-  );
+    mongoshVersion: require('../../package.json').version,
```
gagik, line 81: this actually seems weird. We're passing compass shell's version as something which gets ultimately reported as `mongosh_version` in telemetry. Is that expected? @addaleax

addaleax: Yeah, it's obviously not great ... not sure what a better version would be here, though? We don't get the version reported by shell-api until we are ready to evaluate code

packages/compass-shell/package.json
```diff
@@ -58,7 +58,7 @@
     "@mongodb-js/compass-utils": "^0.8.2",
     "@mongodb-js/compass-workspaces": "^0.31.2",
     "@mongosh/browser-repl": "^3.2.0",
```
addaleax, line 60: Let's stick to always updating all the `@mongosh` packages if we update one of them, we don't want to end up in a situation where we use mongosh packages that refer to genuinely different mongosh source tree versions

gagik: yeah that's fair, the other two ended up being bumped in https://github.com/mongodb-js/compass/pull/6747 but good to generally do that as principle

packages/compass-shell/src/modules/worker-runtime.ts
gagik: Hm yeah, I guess the solution would be to not call this `mongosh_version` though it'd be too destructive to change it now and not worth it unless there's explicit need from product/analytics

package-lock.json
```diff
@@ -44623,6 +44582,19 @@
...
+      "integrity": "sha512-Z2rXUotJX3DVBlRN8y2Dzv26/SnEYlD7gh3uzvxL2myFKfyB2zQjY8DeQWb11WEam9v9LF9VwhndapLOTkJ+yA==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "heap-js": "^2.3.0"
+      },
+      "peerDependencies": {
+        "bson": "6.x"
+      }
+    },
```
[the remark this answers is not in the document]
addaleax: Same thing would have applied in https://github.com/mongodb-js/compass/pull/6747, btw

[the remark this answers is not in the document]
gagik: would having this as a dependency at root force that to happen? I know we don't do this at all right now but seems weird for the fix to be something so manual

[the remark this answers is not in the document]
addaleax: Yeah, so, _ideally_ the way updating dependencies would work is that you add them as root dependencies, run `npm install` and then remove them again as root dependencies, because `npm` should keep them pre-hoisted

I don't know why that doesn't properly work right now, though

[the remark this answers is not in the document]
gagik: Gotcha, did that now 👍 but why do we not keep them as root dependencies just to make the hoisting explicit?

addaleax (approved):
LGTM if we want to fix the package-lock.json in a follow-up

gagik:
Since this is breaking CI and needs some more adjustments across the code-base, going to separate the `package-lock` patch into https://github.com/mongodb-js/compass/pull/6754 (will follow-up with the fix for the type errors)

PR #6750, opened 2025-02-25
feat(connection-form): add connection and anonymous id to app name for atlas connections COMPASS-8591

gagik:
See [COMPASS-8591](https://jira.mongodb.org/browse/COMPASS-8591)

### DIFF
--- a/packages/compass-web/src/entrypoint.spec.tsx
+++ b/packages/compass-web/src/entrypoint.spec.tsx
@@ -57,6 +57,8 @@ describe('CompassWeb', function () {
     Sinon.resetHistory();
   });
 
+  const testAnonymousId = 'test-anonymous-id';
+
   async function renderCompassWebAndConnect(
     props: Partial<React.ComponentProps<typeof CompassWeb>> = {},
     connectFn = mockConnectFn
@@ -72,6 +74,7 @@ describe('CompassWeb', function () {
           {...props}
           initialPreferences={{
             enableCreatingNewConnections: true,
+            telemetryAnonymousId: testAnonymousId,
             ...props.initialPreferences,
           }}
           onFailToLoadConnections={() => {}}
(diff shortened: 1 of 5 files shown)


packages/compass-web/src/entrypoint.spec.tsx
```diff
@@ -97,16 +100,20 @@ describe('CompassWeb', function () {
...
-      'mongodb://localhost:27017/?appName=Compass+Web'
-    );
-
     await waitFor(() => {
       screen.getByText('Connected to localhost:27017');
     });
 
     expect(onTrackSpy).to.have.been.calledWith('New Connection');
+
+    const connectionId = onTrackSpy.firstCall.args[1][
```
gagik, line 109: a bit weird to get it from there though has added benefit of making sure tracking + the URL are consistent.

packages/connection-form/src/utils/set-app-name-if-missing.ts
```diff
@@ -14,7 +20,11 @@ export function setAppNameParamIfMissing(
     const searchParams =
       connectionStringUrl.typedSearchParams<MongoClientOptions>();
     if (!searchParams.has('appName') && defaultAppName !== undefined) {
-      searchParams.set('appName', defaultAppName);
+      const appName = `${defaultAppName}${
```
gagik, line 23: I used `-` as a separator, open to suggestions if this can be problem causing in ways

gagik:
Should this skip release notes as this pertains to telemetry change?

packages/connection-form/src/utils/set-app-name-if-missing.ts
```diff
@@ -14,7 +20,11 @@ export function setAppNameParamIfMissing(
...
+        telemetryAnonymousId ? `-${telemetryAnonymousId}` : ''
+      }-${connectionId}`;
+
+      searchParams.set('appName', appName);
```
gagik, line 27: not related to this line but generally: Should this skip release notes as this pertains to telemetry change?

addaleax: Hm yeah, it's a telemetry change but it's one that could definitely be visible to users, especially non-Atlas users who do on-prem deployments.

Have we considered scoping this change to Atlas only?

addaleax (approved):
LGTM with a question

packages/connection-form/src/utils/set-app-name-if-missing.ts
gagik: that's a good idea

PR #6755, opened 2025-02-26
feat: show error details on insert COMPASS-8864

paula-stacho:
## Description
Side changes:
- Updated InsertDocumentDialog from Class to FC.
- add initial focus on another modal which was focusing on the copy button unexpectedly

The new dialog lives in the document tab. Even though the button says back, it is opened on top of the previous tab, not instead of. This is so that we can easily reuse it for other operations, whether they originate in a modal or directly from the documents list. Let me know if you see problems with this.
Added e2e test.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


paula-stacho:
Why I'm setting initialFocus: https://mongodb.slack.com/archives/G2L10JAV7/p1740755117364199

packages/compass-components/src/components/modals/error-details-modal.tsx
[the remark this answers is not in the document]
paula-stacho: Yes, these errors can happen in aggregation ($out, $merge) or the failed validation preview in validation tab.

packages/compass-crud/src/components/editable-document.tsx
```diff
@@ -278,6 +279,12 @@ class EditableDocument extends React.Component<
         onCancel={() => {
           this.handleCancel();
         }}
+        // onOpenErrorDetails{() => {
```
Anemy, line 282: Leftover?

packages/compass-crud/src/components/insert-document-dialog.tsx
```diff
@@ -61,119 +67,181 @@ export type InsertDocumentDialogProps = InsertCSFLEWarningBannerProps & {
...
-    if (this.props.isOpen && !this.hasManyDocuments()) {
-      if (prevProps.jsonView && !this.props.jsonView) {
+  useEffect(() => {
+    if (isOpen && !hasManyDocuments()) {
+      if (!jsonView) {
         // When switching to Hadron Document View.
         // Reset the invalid elements list, which contains the
         // uuids of each element that has BSON type cast errors.
-        this.invalidElements = [];
+        setInvalidElements([]);
```
Anemy, line 209: Getting an error when swapping views in the insert document editor without editing:
Screenshot 2025-03-04 at 12 44 50 PM

Not sure if it's this line, but it's likely related to the updated component.

packages/compass-import-export/src/components/export-modal.tsx
```diff
@@ -246,7 +246,12 @@ function ExportModal({
   }, [isOpen, resetExportFormState]);
 
   return (
-    <Modal open={isOpen} setOpen={closeExport} data-testid="export-modal">
+    <Modal
+      open={isOpen}
+      setOpen={closeExport}
+      data-testid="export-modal"
+      initialFocus="#export-collection-code-preview-wrapper"
```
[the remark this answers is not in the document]
paula-stacho: good idea, will do!

packages/compass-crud/src/components/insert-document-dialog.tsx
paula-stacho: Ah thanks! Fixed

packages/compass-crud/src/components/document-list.tsx
```diff
@@ -581,8 +589,20 @@ const DocumentList: React.FunctionComponent<DocumentListProps> = (props) => {
...
+                details: insert.error.info || {},
+                closeAction: 'back',
+              })
+            }
             {...insert}
           />
+          <ErrorDetailsModal
+            open={!!errorDetailsOpen}
+            onClose={closeErrorDetailsDialog}
+            details={errorDetailsOpen?.details}
```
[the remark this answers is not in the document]
paula-stacho: Thank you, this feels much better!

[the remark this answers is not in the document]
paula-stacho: The change in store alone didn't prevent the flickering, I resolved this by only updating the code content when the modal is open. I also played around with setTimeout for cleaning up the details, but 0 is not enough and I didn't want to risk a race condition with next modal opening. I don't see any flicker if the modal opens with a different content.

packages/compass-components/src/components/modals/error-details-modal.tsx
```diff
@@ -0,0 +1,72 @@
...
+  open,
+  ...modalProps
+}: ErrorDetailsModalProps) {
+  const [stringDetails, setStringDetails] = useState<string>('');
+
+  useEffect(() => {
+    if (open) {
+      setStringDetails(JSON.stringify(details, undefined, 2));
+    }
+  }, [details, open]);
```
gribnoysup, line 38: This really doesn't belong in the effect, this is a derived value that, if we're trying to optimise the rendering here a bit and avoid stringifying too often, should be a memo.

packages/compass-components/src/components/modals/error-details-modal.tsx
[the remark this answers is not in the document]
gribnoysup: Aha, got it!

packages/compass-components/src/components/modals/error-details-modal.tsx
paula-stacho: It was a memo, I changed it now because we saw a flicker (see conversation above). we set open -> close, details -> no details at the same time, but it renders sequentially - the modal closes a bit later than the details disappear. I see it as a problem with rendering, my best solution was to keep the details when the modal is closing. 
I could rewrite this with memo + previousDetails ref, but this made more sense to me. But as I'm explaining it I see we need a comment at least. Would you prefer the memo + ref? Or do you have a better idea? I do admit it feels a bit hackish, maybe I'm missing something weird that we do that causes the issue.

paula-stacho: Insert document modal has the same flicker. Bulk update does not.

Anemy (approved):
lgtm!

packages/compass-components/src/components/modals/error-details-modal.tsx
Anemy: Would an alternative be to not un-set the error details when we close it? Then we can fully use the prop with useMemo. Not a blocker for me, might be nice to mention why we are doing it that way in a comment for future folks though.

gribnoysup: > my best solution was to keep the details when the modal is closing.

This is already a store state, if you want to not reset it when the modal is closed, just don't do it in the store itself and reset it on open in the store logic instead of splitting this logic between store and component in a hard to control and explain way.

> Would an alternative be to not un-set the error details when we close it?

Yes, we already do this in a couple of places

paula-stacho: Alright, I will change it in the store then. It seemed off to me as the store shouldn't be concerned with the rendering timing. But if it's a common pattern already, it'll at least lower the confusion

PR #6766, opened 2025-03-04
feat(schema-validation): add confirmation modal when applying rules, add explicit editing state COMPASS-8861

Anemy:
This is most of COMPASS-8861, I'd like to do the toasts separately as they can be in a separate pr from this work without any conflicts.

Designs: https://www.figma.com/design/CBHJriBQZ2qxMxtnXFgDxJ/EXPO-3174-%3A-Export-Schema-%2B-Validation-Detection?node-id=1309-40813&t=ouDByoHIQallPTXr-1 

<details open>
<br>

Screenshot 2025-03-03 at 3 49 42 PM
Screenshot 2025-03-03 at 3 49 48 PM
Screenshot 2025-03-03 at 3 49 55 PM

</details>

We could refactor the store here to have fewer reducers to simplify some of the actions here. Probably not worth making part of this work, but something that is on the mind with these changes. Removing some of the debouncing and cleaning up the fetching of existing validation would be nice as well.

### DIFF
--- a/packages/compass-e2e-tests/helpers/commands/set-validation.ts
+++ b/packages/compass-e2e-tests/helpers/commands/set-validation.ts
@@ -15,6 +15,9 @@ export async function setValidation(
 
   await browser.clickVisible(Selectors.UpdateValidationButton);
 
+  // Confirm in the confirmation modal.
+  await browser.clickVisible(Selectors.confirmationModalConfirmButton());
+
   // both buttons should become hidden if it succeeds
   await validationActionMessageElement.waitForDisplayed({
     reverse: true,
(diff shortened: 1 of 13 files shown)


paula-stacho:
I'm seeing a different copy + option to not show the modal again in the designs, are we looking at the same version?

Screenshot 2025-03-04 at 14 05 19

Anemy:
@paula-stacho I just with Misba on the designs. We might use this wording in the pr instead of whats in the designs to be more inline with other confirmation modals in Compass (save aggregation for instance). I did miss that option not to show the modal again though thanks! I'll push up a commit today.

Anemy:
@paula-stacho I created a new ticket for the "don't show this message again" checkbox: https://jira.mongodb.org/browse/COMPASS-9076 
I think we probably want something generic there that's on the `useConfirmation` hook and didn't want to hack in a confirmation modal into this pr as there are already a lot of changes.

paula-stacho (approved):
Look great! Thanks for fixing additional details!

PR #6767, opened 2025-03-04
chore(schema-validation): add progress toast COMPASS-8861

Anemy:
Last part of COMPASS-8861

Opening as a draft as I want to clear up if we want all 3 of these toasts, some of them, or none at all. For errors it can be loud, while we also show the error in the UI:

Screenshot 2025-03-03 at 8 41 46 PM

The success toast is a bit outside of the user's flow, as the `updateCollection` command that's run is almost always instant. I did find it to be a nice affirmation though.
There isn't currently any way that visually indicates that the validation rule update is in progress, so I do think that toast could be nice. That being said we should also disable the update button until the response is in. I started refactoring


Anemy:
Oops meant to open as a draft and hit enter, closing and will re-open.

PR #6770, opened 2025-03-04
chore(validation): add success toast on validation update COMPASS-8861

Anemy:
Last part of COMPASS-8861

Chatting with Misba, we don't want to show an in-progress toast or an error toast. We do want to add an in-progress state to the apply button, that'll land in a new ticket.


packages/compass-schema-validation/src/modules/validation.ts
```diff
@@ -1,7 +1,8 @@
 import type { RootAction, RootState, SchemaValidationThunkAction } from '.';
-import { EJSON } from 'bson';
+import { EJSON, ObjectId } from 'bson';
```
paula-stacho, line 2:
```suggestion
import { EJSON } from 'bson';
```

paula-stacho: not used

paula-stacho (approved):

PR #6774, opened 2025-03-05
feat(deps): bump mongosh, driver to latest COMPASS-9056

addaleax:
This includes adjusting the CSFLE/QE collection tracker for the Node.js driver changes related to `$lookup` support, where filters for listCollections can now list multiple collections in an `$in` clause.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/data-service/src/run-command.ts
```diff
@@ -193,6 +193,7 @@ export type CollectionInfoNameOnly = {
 export type CollectionInfo = CollectionInfoNameOnly & {
   /** @see https://docs.mongodb.com/manual/reference/method/db.createCollection/#mongodb-method-db.createCollection */
   options: Document;
+  name: string;
```
nirinchev, line 196: With a risk to embarrass myself with my poor typescript knowledge, why do we need this here? 😅 My expectation is that since we're merging this with the `CollectionInfoNameOnly`, we already have a `name: string` field.

nirinchev (approved):
Looks reasonable to me, but obviously would be a good idea to get a review from someone more knowledgeable.

packages/data-service/src/run-command.ts
addaleax: Oh duh. Yeah, just missed that this was already part of the type, that's all 🙈

packages/data-service/src/csfle-collection-tracker.ts
```diff
@@ -437,53 +446,70 @@ export class CSFLECollectionTrackerImpl implements CSFLECollectionTracker {
...
+        ))
+    ) {
       // This is an assertion more than an actual error condition.
       // It ensures that we're only getting listCollections requests
       // in the format that we expect them to come in.
       return new Error(
-        `[Compass] Unexpected listCollections request on '${dbName}' with name: '${
-          filter?.name as string
-        }'`
+        `[Compass] Unexpected listCollections request on '${dbName}' with filter: ${EJSON.stringify(
```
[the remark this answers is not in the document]
addaleax: We could, but honestly I don't think that would be worth it, if this actually gets thrown at some point, that means we're going to have to do some digging into the root cause anyway

packages/data-service/src/csfle-collection-tracker.ts
```diff
@@ -437,53 +446,70 @@ export class CSFLECollectionTrackerImpl implements CSFLECollectionTracker {
     filter: Document,
     collectionInfos: CollectionInfo[]
   ): Error | undefined {
-    if (typeof filter?.name !== 'string' || collectionInfos.length > 1) {
+    // filter is either { name: string } or { name: { $in: string[] } }
```
gagik, line 449: can this be documented as a type?

gagik (approved):

packages/data-service/src/csfle-collection-tracker.ts
addaleax: Yeah, but the point here is that the input isn't of a known type ... we could split this out into a separate validation function if you prefer, that takes a generic `unknown` or `Document` input and returns this type

gagik: I see, I don't have strong feelings about it; leaving it as is is fine also 👍

PR #6797, opened 2025-03-14
feat: show validation error details COMPASS-8864 COMPASS-8867 COMPASS-8865

paula-stacho:
## Description
Combines the error details on insert, import and single update (the rest has been de-prioritized, but we might bring it in if we have extra time this week).

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-e2e-tests/tests/collection-import.test.ts
```diff
@@ -509,6 +509,129 @@ describe('Collection import', function () {
...
+        connectionName: DEFAULT_CONNECTION_NAME_1,
+        database: 'test',
+        collection: 'extended-json-file',
+        validator: FAILING_VALIDATOR,
+      });
+    });
+
+    afterEach(async function () {
+      await browser.navigateWithinCurrentCollectionTabs('Validation');
+      await browser.setValidationWithinValidationTab('{}');
```
Anemy, line 526: Would it make sense to instead run this resetting validation operation against the db directly and not use the UI for the cleanup? I'm thinking if the validation setting does fail this won't clean up and then more tests may start failing.
Or maybe the collection gets dropped altogether and we don't need this?

packages/compass-e2e-tests/helpers/selectors.ts
```diff
@@ -1311,11 +1319,11 @@ export const ConfirmationModalInput = `${ConfirmationModal} input`;
 
 export const confirmationModalConfirmButton = (
   modalSelector = ConfirmationModal
-) => `${modalSelector} [role=dialog] button:nth-of-type(1)`;
+) => `${modalSelector} [role=dialog] [data-testid*="confirm_button"]`;
```
Anemy, line 1322: What does the `*` next to `data-testid` here do?

packages/compass-crud/src/components/insert-document-dialog.tsx
```diff
@@ -63,117 +70,180 @@ export type InsertDocumentDialogProps = InsertCSFLEWarningBannerProps & {
...
 
-  componentWillUnount() {
-    if (!this.hasManyDocuments()) {
+  useEffect(() => {
+    if (!hasManyDocuments() && doc) {
       // When closing the modal.
       // Remove the listeners to the BSON type validation errors in order to clean up properly.
-      this.props.doc.removeListener(Element.Events.Invalid, this.handleInvalid);
-      this.props.doc.removeListener(Element.Events.Valid, this.handleValid);
+      doc.removeListener(Element.Events.Invalid, handleInvalid);
```
Anemy, line 234: Should these be removing the listeners in a returned function since we want this on unmount?
Also should it be using `isOpen` to manage the cleanup then as well since we're now keeping this modal around when `isOpen` is set to false?

packages/compass-crud/src/components/insert-document-dialog.tsx
```diff
@@ -63,117 +70,180 @@ export type InsertDocumentDialogProps = InsertCSFLEWarningBannerProps & {
...
+}> = ({
+  jsonView,
+  doc,
+  hasManyDocuments,
+  updateJsonDoc,
+  jsonDoc,
+  isCommentNeeded,
+  updateComment,
+}) => {
+  if (!jsonView) {
```
Anemy, line 90: nit: we could swap this if around to return the `InsertJsonDocument` with if(jsonView) to keep things a bit less nested and quicker to read.

Anemy (commented):
Nice! The insert document modal has been a place that has a lot of edge conditions, happy to see it moving to a functional component and becoming a bit easier to work with. Nice test adds and the error details are a good feature. Left a couple small questions and one or two suggestions.

packages/compass-e2e-tests/tests/collection-import.test.ts
Anemy: Part of me things it's better to do both directly that way we don't have to change things around when we change things that aren't related to this test. That way there are less cross dependencies and we're only focusing on the test target. That being said I also think fine it's fine as is, more of a nit. It might also speed up tests to avoid extra UI things.

packages/compass-crud/src/components/insert-document-dialog.tsx
paula-stacho: swapped

paula-stacho:
side fix: catching the error on array insert

how to reproduce:
try inserting
```
[{
  "_id": {
    "$oid": "67e1b5546aad49afb8f46e4a09"
  }
},
{
  "_id": {
    "$oid": "67e1b5546aad49afb8f46e4a09"
  }
}]
```

Anemy (approved):

PR #6801, opened 2025-03-17
feat(release): dual write release artifacts to new bucket COMPASS-9160

drichmdb:
This commit adjusts Compass to write release artifacts to a new S3 bucket.
Eventually the old bucket will be removed after we have confirmed that we
can serve artifacts from the new bucket. We also use assume role to access
this bucket instead of static credentials.

### DIFF
--- a/.evergreen/functions.yml
+++ b/.evergreen/functions.yml
@@ -504,12 +504,18 @@ functions:
         content_type: application/x-gzip
 
   publish:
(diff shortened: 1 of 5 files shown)


lerouxb (approved):
Looks good. Like I mentioned this code might only run when doing a beta/stable release, though. But if a beta fails we should have enough context to fix it.

PR #6804, opened 2025-03-18
fix(compass-schema): sticky copy button COMPASS-9141

paula-stacho:
## Description
Fixing the styles for code editor on export schema modal

Before:
The whole editor scrolls, so the actions get hidden down the line.
Screenshot 2025-03-18 at 18 38 23

After:
Only content of the editor scrolls.
Screenshot 2025-03-18 at 18 38 12

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-schema/src/components/export-schema-modal.tsx
+++ b/packages/compass-schema/src/components/export-schema-modal.tsx
@@ -45,13 +45,17 @@ const contentContainerStyles = css({
 });
 
 const codeEditorContainerStyles = css({
-  maxHeight: `${spacing[1600] * 4 - spacing[800]}px`,
-  overflow: 'auto',
+  height: `${spacing[1600] * 4 - spacing[400]}px`,
+  padding: spacing[100],
 });
 
 const codeStyles = css({
   '& .cm-editor': {
-    paddingLeft: spacing[2],
+    paddingLeft: spacing[200],
+    maxHeight: `${spacing[1600] * 4 - spacing[800]}px`,
+  },
+  '& .multiline-editor-actions': {
+    marginRight: spacing[300],
   },
 });
 


packages/compass-schema/src/components/export-schema-modal.tsx
```diff
@@ -45,13 +45,17 @@ const contentContainerStyles = css({
...
 
 const codeEditorContainerStyles = css({
-  maxHeight: `${spacing[1600] * 4 - spacing[800]}px`,
-  overflow: 'auto',
+  height: `${spacing[1600] * 4 - spacing[400]}px`,
+  padding: spacing[100],
 });
 
 const codeStyles = css({
   '& .cm-editor': {
```
Anemy, line 53: It does strike me as not the best practice to be overriding the internal styles of the editor here. Ideally compass-schema shouldn't know we're even using code mirror under the hood. Since we're already doing it, I think fine as is, but something we should probably avoid.

Anemy (approved):

packages/compass-schema/src/components/export-schema-modal.tsx
Anemy: I was the one who introduced this too 🤦 
Looks like we do it in a few other places in Compass also.

paula-stacho: You're right, we could've just added some parameters that the compass component would translate into styles. We don't and we even named the component <CodeMirrorMultilineEditor />, so I think the ship has sailed 😅 . And by sailed I mean I don't think we have time to fix that atm 🙈

PR #6821, opened 2025-03-26
fix: use mongodb version 8.0.5 on macOS versions < 14

lerouxb:
mongodb 8.0.6 does not work on macOS version < 14.

### DIFF
--- a/.github/workflows/test-installers.yml
+++ b/.github/workflows/test-installers.yml
@@ -229,6 +229,10 @@ jobs:
           DISTRO_ID: ${{ matrix.distro-id }}
           # Exposing token to prevent update server from being rate limited
           GITHUB_TOKEN: ${{ github.token }}
+          # mongodb 8.0.6 requires macOS 14+. Just use an older version for now
+          # so it will run on GitHub's mac runners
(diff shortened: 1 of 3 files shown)


.github/workflows/test-installers.yml
```diff
@@ -229,6 +229,10 @@ jobs:
           DISTRO_ID: ${{ matrix.distro-id }}
           # Exposing token to prevent update server from being rate limited
           GITHUB_TOKEN: ${{ github.token }}
+          # mongodb 8.0.6 requires macOS 14+. Just use an older version for now
+          # so it will run on GitHub's mac runners
+          MONGODB_VERSION: "8.0.5"
```
lerouxb, line 234: I thought of doing lots of work to only set this on macos 13, but then realised that the database version is irrelevant to testing that compass installs, starts up and can auto-update.

lerouxb: (I'm just hoping these env vars get passed down from the smoke tests process to the e2e tests process without further changes required.)

addaleax (approved):

PR #6831, opened 2025-04-01
fix: remove logging for currentOp COMPASS-9203

paula-stacho:
## Description
Let me know if there is something that is safe to log and we do need - we could filter that

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


Anemy (approved):
Changes look good. I also have the same question, do we want to keep the log and just not log the result here? I could see it being useful to know we're running operations.

PR #6832, opened 2025-04-01
feat(release): write artifacts to new release bucket with private acls COMPASS-9238

drichmdb:
This commit adjusts our release process to upload artifacts to the new release bucket using private ACLs. Because users fetch artifacts from our CDN, we are not allowing direct access to this bucket. If we do not set the ACLs to private, our uploads will fail. In order to support this, we had to update to the latest devtools-shared dl-center package.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8120,9 +8120,9 @@
       }
     },
     "node_modules/@mongodb-js/dl-center": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/@mongodb-js/dl-center/-/dl-center-1.2.0.tgz",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -59575,9 +59575,9 @@
       }
     },
     "@mongodb-js/dl-center": {
-      "version": "1.2.0",
```
Anemy, line 59578: Should we also be updating this dependency in the `package.json`?
https://github.com/mongodb-js/compass/blob/5c8b2ac42c3f524646c3f57e8e49f7bb781c9a8c/packages/hadron-build/package.json#L24

drichmdb: Yes! I just realized that didn't happen. I've been struggling to get this to work without changing the scope of the package in the top level `package-lock.json`. I'll try again.

drichmdb: I believe I got this to work this time. Thanks for pointing that out.

lerouxb (approved):

Anemy (approved):

PR #6834, opened 2025-04-02
feat: add special handling of vector sub-type COMPASS-8257

kraenhansen:
### As is

Screenshot 2025-04-02 at 10 42 31

### Suggested to be

Screenshot 2025-04-02 at 16 05 46

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Open Questions

The table view for this is not awesome, but we would have to revisit either the principle of showing copy-pastable code in the cells or the driver API to improve that.
 
Screenshot 2025-04-02 at 16 09 19
](url)

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-components/src/components/bson-value.tsx
+++ b/packages/compass-components/src/components/bson-value.tsx
@@ -157,6 +157,37 @@ const BinaryValue: React.FunctionComponent<PropsByValueType<'Binary'>> = ({
 
       return { stringifiedValue: `UUID('${uuid}')` };
     }
+    if (value.sub_type === Binary.SUBTYPE_VECTOR) {
+      const vectorType = value.buffer[0];
+      if (vectorType === Binary.VECTOR_TYPE.Int8) {
+        const truncatedSerializedBuffer = truncate(
+          value.toInt8Array().slice(0, 100).join(', '),
+          100
+        );
+        return {
+          stringifiedValue: `Binary.fromInt8Array(new Int8Array([${truncatedSerializedBuffer}]))`,
+        };
+      } else if (vectorType === Binary.VECTOR_TYPE.Float32) {
+        const truncatedSerializedBuffer = truncate(
+          [...value.toFloat32Array().slice(0, 100)]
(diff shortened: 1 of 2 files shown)


packages/compass-components/src/components/bson-value.tsx
```diff
@@ -157,6 +157,37 @@ const BinaryValue: React.FunctionComponent<PropsByValueType<'Binary'>> = ({
...
+          100
+        );
+        return {
+          stringifiedValue: `Binary.fromInt8Array(new Int8Array([${truncatedSerializedBuffer}]))`,
+        };
+      } else if (vectorType === Binary.VECTOR_TYPE.Float32) {
+        const truncatedSerializedBuffer = truncate(
+          [...value.toFloat32Array().slice(0, 100)]
+            // Using a limited precision and removing trailing zeros for better displaying
+            .map((num) => num.toPrecision(8).replace(/\.?0+$/, ''))
```
kraenhansen, line 174: This is the alternative of not adding this

Screenshot 2025-04-02 at 11 17 03

gagik: I see the value but not sure. In mongosh we're not going to apply this transformation as we'd be using inspect on the array directly, so maybe the disconnect can be confusing

gagik:
Going to take over to get this finalized.

packages/compass-components/src/components/bson-value.tsx
gagik: We did not end up doing the precision thing in mongosh: https://github.com/mongodb-js/mongosh/commit/999208bd397d69a90ec34ceea0fa7af66fc7b4d0 so there would hypothetically be some discrepancy. That said mongosh is more interactive as far as exploring data goes so maybe it would be more justified to have this adjustment here.

addaleax (approved):

gagik (approved):

PR #6837, opened 2025-04-02
fix(compass): disable spellcheck before creating window when no network traffic COMPASS-8166

Anemy:
COMPASS-8166

We already were disabling the spellcheck when network traffic was disabled, however it looks like it was being disabled after the request for the dictionary would happen. This pr updates when we do the override to happen earlier. Not sure if something changed in electron or on our end making this change necessary, looking into that currently, could be something else like how the [sessions](https://www.electronjs.org/docs/latest/api/session#sessiondefaultsession) work too.

Still looking into the other DNS request to google that chromium makes, so I'll likely have another pr for this ticket soon (maybe just to update the test).


Anemy:
Closing in favor of https://github.com/mongodb-js/compass/pull/6853

PR #6838, opened 2025-04-03
fix: semver sort the releases latest first before picking the first non-draft of the relevant channel

lerouxb:
GitHub's REST API returns releases sorted incorrectly, so `1.45.5-beta.9` appears before `1.45.5-beta.10`. We then pick the first non-draft of the relevant channel as the latest one for the manifest on the download centre and end up with .9 and not .10.

This sorts the releases with semver first.

I tested this by just putting the code in a separate file and running it there, printing to the console. Once this is merged we can just re-run the most recent action that used this code and that should update the download centre correctly.

### DIFF
--- a/packages/hadron-build/commands/upload.js
+++ b/packages/hadron-build/commands/upload.js
@@ -10,6 +10,7 @@ const path = require('path');
 const os = require('os');
 const { promises: fs } = require('fs');
 const { deepStrictEqual } = require('assert');
+const semver = require('semver');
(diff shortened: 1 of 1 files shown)


paula-stacho (approved):
Such a classic 😅 
Is this a known issue with github, or it's a non issue because it doesn't promise a particular order?

lerouxb:
Prereleases aren't very well specified by semver and there are just loads of issues like this even within one implementation, let alone across multiple for different languages.

PR #6844, opened 2025-04-08
fix(compass-indexes): fix tooltip triggers in indexes plugin COMPASS-8449

lerouxb:
Missed this when we last worked on the indexes plugin.


gribnoysup (approved):

addaleax (approved):
Should LG forbid passing a React element as the `trigger` property entirely, to avoid this type of issue?

PR #6846, opened 2025-04-08
fix: fix the shortcuts on windows to be the app name, not executable name COMPASS-8367

lerouxb:
* add shortcutName to MSI Creator so that the app installed by the .msi will have the expected space
* also fix the shortcut that's made by Squirrel for Windows

We had a dependency on https://github.com/mongodb-js/electron-squirrel-startup/blob/master/index.js and I think that package is absurd. I copied the code to Compass, fixed it up and it now renames both the Desktop and Start Menu shortcuts to be correct.
Screenshot 2025-04-11 at 15 38 10
Screenshot 2025-04-11 at 15 38 04
Screenshot 2025-04-11 at 15 37 58


lerouxb:
The shortcut created by windows setup .exe is also wrong and I think I'll try and fix it here too.

packages/compass/src/main/application.ts
```diff
@@ -121,8 +122,8 @@ class CompassApplication {
     );
 
     // needs to happen after setupProtocolHandlers
-    if ((await import('electron-squirrel-startup')).default) {
-      debug('electron-squirrel-startup event handled sucessfully\n');
+    if (await handleSquirrelWindowsStartup()) {
+      app.quit();
```
lerouxb, line 126: I really think it is much clearer what's going on when a library function isn't the one quitting the app. Reads better IMHO if that happens with the rest of the code that does that sort of thing. ie. this file.

packages/compass/src/main/squirrel-startup.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  const shortcutName = getShortcutName();
+
+  const cmd = process.argv[1];
+  debug('processing squirrel command `%s`', cmd);
+
+  /*
+  For more detailed info on these commands, so Electron and Squirrel Windows'
+  documentation:
```
lerouxb, line 52: It is a pity that electron-userland/electron-installer-squirrel-windows [links back to electron-squirrel-startup](https://github.com/electron-userland/electron-installer-squirrel-windows?tab=readme-ov-file#integration).

I think that README would be a lot clearer if they just put this code there as an example, linked to Squirrel's documentation and maybe explained what these three or four CLI args do.

packages/hadron-build/lib/target.js
```diff
@@ -450,6 +450,7 @@ class Target {
         outputDirectory: this.packagerOptions.out,
         exe: this.packagerOptions.name,
         name: this.productName,
+        shortcutName: this.productName,
```
lerouxb, line 453: This is the fix for the shortcut made by .msi files.

packages/compass/src/main/squirrel-startup.ts
Anemy: IIRC the https://github.com/mongodb-js/electron-squirrel-startup is used by some other folks. If we're no longer using it we might want to mark it unmaintained/archived or something. Could be worth bringing up in the compass channel.
FWIW I'm all for removing small packages and also having more typescript!

packages/hadron-build/lib/target.js
lerouxb: I am now unsure that the msi was ever doing the wrong thing, but this doesn't hurt - it at least makes it explicit.

Anemy (approved):
lgtm! Nice work, it's nice to have these cleaned up. We have a lot of windows users.

PR #6852, opened 2025-04-09
feat: disable db and coll stats via preferences COMPASS-5387

paula-stacho:
https://jira.mongodb.org/browse/COMPASS-5387

## Description
The stats are enabled by default and can be turned off in the configuration.

This affects views: 

### Menu
Screenshot 2025-04-14 at 14 08 21

### Databases list & grid
Screenshot 2025-04-14 at 14 07 27

### Collections list & grid
Screenshot 2025-04-14 at 14 13 29

### Collection -> Documents tab
Screenshot 2025-04-14 at 14 16 06

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/databases-collections-list/src/collections.tsx
gribnoysup, line 1: This doesn't seem to be matching [the designs](https://www.figma.com/design/XNdBIIGRDpRU2N2zwC2xHp/EXPO-4439--Compass-DB-Coll-stats?node-id=1-33&p=f&t=4SKv15kzYZT2PHSm-0), were there any changes to how we're supposed to handle this, but figma wasn't updated?

packages/compass-aggregations/src/modules/aggregation.ts
```diff
@@ -252,6 +260,7 @@ const confirmWriteOperationIfNeeded = async (
         dataService,
         database,
         collection,
+        preferences,
```
gribnoysup, line 263: Doesn't look like this is used by anything

packages/compass-app-stores/src/stores/instance-store.ts
```diff
@@ -102,7 +105,11 @@ export function createInstancesStore(
         instancesManager.getMongoDBInstanceForConnection(connectionId);
       const dataService = connections.getDataServiceForConnection(connectionId);
       isFirstRun = instance.status === 'initial';
-      await instance.refresh({ dataService, ...refreshOptions });
+      await instance.refresh({
+        dataService,
+        ...refreshOptions,
+        preferences,
```
gribnoysup, line 111: Not used

packages/compass-preferences-model/src/preferences-schema.tsx
```diff
@@ -13,10 +14,27 @@ import {
   proxyOptionsToProxyPreference,
   proxyPreferenceToProxyOptions,
 } from './utils';
+import { Link } from '@mongodb-js/compass-components';
 
 export const THEMES_VALUES = ['DARK', 'LIGHT', 'OS_THEME'] as const;
 export type THEMES = typeof THEMES_VALUES[number];
 
+const EnableDbAndCollStatsDescription: React.ReactNode = (
```
gribnoysup, line 22: This is element, not Component, doesn't need to be capitalized (capitalization is misleading, you can't construct it)

packages/compass-settings/src/components/settings/settings-list.tsx
```diff
@@ -82,7 +88,8 @@ function SettingLabel({ name }: { name: SupportedPreferences }) {
           </span>
         )}
       </Label>
-      {long && <Description>{long}</Description>}
+      {long && !longReact && <Description>{long}</Description>}
+      {longReact && <div className={inputDescriptionStyles}>{longReact}</div>}
```
gribnoysup, line 92: Why does this need a completely separate component and not a Description we're already using?

```suggestion
      <Description>{longReact ?? long}</Description>
```

packages/compass-workspaces/src/stores/workspaces.ts
```diff
@@ -724,6 +724,7 @@ const fetchCollectionInfo = (
         dataService,
         database,
         collection,
+        preferences,
```
gribnoysup, line 727: Not used

packages/compass-saved-aggregations-queries/src/stores/open-item.ts
```diff
@@ -494,6 +499,7 @@ export const openSavedItem =
           dataService,
           database,
           collection,
+          preferences,
```
gribnoysup, line 502: Not used

packages/databases-collections-list/src/collections.tsx
gribnoysup, line 1: This doesn't match the design, too much free space compared to what figma shows, can we tune it so that empty space is not so prominent?

image

image

packages/databases-collections-list/src/databases.tsx
gribnoysup, line 1: Better than collections, but also doesn't match the design, too much free space compared to what figma shows

image

image

packages/compass-settings/src/components/settings/settings-list.tsx
paula-stacho: I've had some problems with the component, but I've tried it again and it works good now, so I'll simplify this, thanks!

packages/collection-model/lib/model.js
```diff
@@ -232,18 +242,28 @@ const CollectionModel = AmpersandModel.extend(debounceActions(['fetch']), {
...
+   * }} options
    * @returns
    */
   async fetch({ dataService, fetchInfo = true, force = false }) {
     if (!shouldFetch(this.status, force)) {
       return;
     }
+
+    const instanceModel = getParentByType(this, 'Instance');
+    const { enableDbAndCollStats } = instanceModel.preferences.getPreferences();
```
gribnoysup, line 258: Instead of exposing all the preferences directly from the instance model, can we limit the exposed interface to only what we need here? Something like:

```suggestion
    const enableDbAndCollStats = instanceModel.shouldAllowStatsFetching();
```

As a rule of thumb I'd suggest to try to stay minimal when introducing these sort of cross boudary interfaces so that the intent of the exposed interface is clearer from just the name and the actual implementation can be changed easier

packages/compass-preferences-model/src/preferences-schema.tsx
```diff
@@ -203,6 +222,7 @@ type PreferenceDefinition<K extends keyof AllPreferences> = {
     : {
         short: string;
         long?: string;
+        longReact?: React.ReactNode;
```
mcasimir, line 225: Ended up scanning this PR for the ticket this jumped to my eye:
from the way we are using this, is there a point in having both `long` and `longReact`? Do we really need to add a new property here? We can just make "long" a `React.ReactNode | string` and everything would work.

paula-stacho: We need the plain string version for non-react envs (shell)

packages/databases-collections-list/src/index.spec.tsx
```diff
@@ -143,5 +230,51 @@ describe('databases and collections list', function () {
...
+      expect(screen.queryByText(/Storage size/)).not.to.exist;
+      // expect(screen.queryByText('1.50 kB')).not.to.exist;
+      // expect(screen.queryByText(/Documents/)).not.to.exist;
+      // expect(screen.queryByText('10')).not.to.exist;
+      // expect(screen.queryByText(/Avg. document size/)).not.to.exist;
+      // expect(screen.queryByText('150.00 B')).not.to.exist;
+      // expect(screen.queryByText(/Indexes/)).not.to.exist;
+      // expect(screen.queryByText('15')).not.to.exist;
+      // expect(screen.queryByText(/Total index size/)).not.to.exist;
+      // expect(screen.queryByText('16.00 B')).not.to.exist;
```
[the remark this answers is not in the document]
paula-stacho: Forgot to uncomment these when I fixed the test, thanks for noticing!

packages/collection-model/lib/model.js
paula-stacho: that's a good idea, I'll do that!

paula-stacho: so I ended up using a property instead of a function, as I wanted to make sure we listen to updates on the preferences.

packages/collection-model/index.d.ts
```diff
@@ -64,18 +64,18 @@ interface CollectionProps {
...
-  free_storage_size: number;
-  index_count: number;
-  index_size: number;
+  is_capped: boolean | undefined;
+  document_count: number | undefined;
+  document_size: number | undefined;
+  avg_document_size: number | undefined;
+  storage_size: number | undefined;
+  free_storage_size: number | undefined;
+  index_count: number | undefined;
```
Anemy, line 77: Instead of showing the `N/A` here in the indexes `TabTitle` should we hide the `N/A` altogether when this isn't available? 
https://github.com/mongodb-js/compass/blob/d661a15f39a4bbcf037cb9841cbb1fb6a9d0a972/packages/compass-indexes/src/plugin-title.tsx#L102

Screenshot 2025-04-28 at 5 58 56 PM

Anemy (commented):
Changes look good, one question on a possible improvement

packages/collection-model/index.d.ts
paula-stacho: Good catch, I've done this for documents but missed the indexes. Thanks!

Anemy (approved):
lgtm! Nice work on this, big preference add.

PR #6853, opened 2025-04-09
fix(compass): disable spellcheck before creating window when no network traffic COMPASS-8166

Anemy:
COMPASS-8166

We already were disabling the spellcheck when network traffic was disabled, however it looks like it was being disabled after the request for the dictionary would happen. This pr updates when we do the override to happen earlier.

The other change here is ignoring Google's public DNS ip in our no network test as its an expected functionality of chromium:
https://source.chromium.org/chromium/chromium/src/+/main:net/dns/README.md;l=351

### DIFF
--- a/packages/compass-e2e-tests/tests/no-network-traffic.test.ts
+++ b/packages/compass-e2e-tests/tests/no-network-traffic.test.ts
@@ -65,9 +65,7 @@ describe('networkTraffic: false / Isolated Edition', function () {
     const compass = await init(this.test?.fullTitle(), {
       extraSpawnArgs: ['--no-network-traffic'],
       wrapBinary,
-      // TODO(COMPASS-8166): firstRun: true seems to result in network traffic.
-      // Probably the welcome modal.
-      firstRun: false,
+      firstRun: true,
     });
     const browser = compass.browser;
 
@@ -117,7 +115,11 @@ describe('networkTraffic: false / Isolated Edition', function () {
 
     if (
       [...connectTargets].some(
-        (target) => !/^127.0.0.1:|^\[::1\]:/.test(target)
+        // Chromium heuristically detects IPv4-only networks by attempting a UDP connection
+        // to 2001:4860:4860::8888 (the IPv6 address for Google Public DNS).
+        (target) =>
+          !/^127.0.0.1:|^\[::1\]:/.test(target) &&
+          !/^\[2001:4860:4860::8888\]:443$/.test(target)
       )
     ) {
       throw new Error(`Connected to unexpected host! ${[...connectTargets]}`);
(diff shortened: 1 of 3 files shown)


packages/compass-e2e-tests/tests/no-network-traffic.test.ts
```diff
@@ -117,7 +115,11 @@ describe('networkTraffic: false / Isolated Edition', function () {
 
     if (
       [...connectTargets].some(
-        (target) => !/^127.0.0.1:|^\[::1\]:/.test(target)
+        // Chromium heuristically detects IPv4-only networks by attempting a UDP connection
+        // to 2001:4860:4860::8888 (the IPv6 address for Google Public DNS).
+        (target) =>
+          !/^127.0.0.1:|^\[::1\]:/.test(target) &&
+          !/^\[2001:4860:4860::8888\]:443$/.test(target)
```
addaleax, line 122: And I guess this is something we can't control at all?

addaleax (approved):

packages/compass/src/main/application.ts
```diff
@@ -129,6 +129,20 @@ class CompassApplication {
...
+
+    const { networkTraffic } = this.preferences.getPreferences();
+
+    if (!networkTraffic) {
+      // Electron fetches spellcheck dictionaries from a CDN
+      // on all OSs expect mac (it provides a built-in spell check).
+      // Passing a non-resolving URL prevents it from fetching
+      // as there aren't any options to disable it provided.
+      // https://github.com/electron/electron/issues/22995
+      session.defaultSession.setSpellCheckerDictionaryDownloadURL(
```
lerouxb, line 141: oh wow. well caught.

lerouxb (approved):
nice! tricky one.

packages/compass-e2e-tests/tests/no-network-traffic.test.ts
Anemy: I haven't found one yet. 
It seems like electron command line switches redirect it with the host rules / host resolver rules.
https://www.electronjs.org/docs/latest/api/command-line-switches#--host-rulesrules 
I'll spend a bit of time looking into that and circle back before merging.

Anemy: Early on in our application code to override the host resolution it looks like we can:
```ts
app.commandLine.appendSwitch('host-resolver-rules', 'MAP 2001:4860:4860::8888 some_address');
```
That will cause all of the network requests to start failing if it doesn't succeed though as I think more of the stack relies on this request for knowing where to send requests. I'm thinking as a fix we could maybe provide users a way to specify this themselves.

I'm going to merge this pr for the other fix and create a follow up ticket and link a todo in the code. I think it still needs a bit of investigation and thought

PR #6865, opened 2025-04-14
chore(crud): update query bar padding to align with other tabs

Anemy:
Other tabs (aggregation and schema) have their padding at 16px, crud was 12px which caused a slight jump when hopping between tabs. Looks like this was caused from updating the legacy number to the new number, might be worth a pr updating all of those legacy padding numbers to their new ones.


paula-stacho (approved):
good eye, I never noticed!

PR #6871, opened 2025-04-17
chore: use new spacing leafygreen numbers

Anemy:
The single digits are deprecated.


lerouxb (approved):
Nice! (I totally forgot that I was going to do this months ago..)

PR #6878, opened 2025-04-22
feat: Add analytics for Create Index Button Clicked CLOUDP-314818

rubydong:
## Description
￼
￼
￼

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
+++ b/packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
@@ -34,7 +34,7 @@ type CreateIndexModalProps = React.ComponentProps<typeof CreateIndexForm> & {
   isVisible: boolean;
   namespace: string;
   error: string | null;
-  currentTab: Tab;
+  currentTab: Tab | null;
   onErrorBannerCloseClick: () => void;
   onCreateIndexClick: () => void;
   onCancelCreateIndexClick: () => void;
@@ -83,6 +83,10 @@ function CreateIndexModal({
   const enableInIndexesGuidanceExp = usePreference('enableIndexesGuidanceExp');
   const showIndexesGuidanceVariant =
     usePreference('showIndexesGuidanceVariant') && enableInIndexesGuidanceExp;
+  if (showIndexesGuidanceVariant && !currentTab) {
+    // If user is in the variant and there has not been a currentTab set, default it to IndexFlow
+    props.onTabClick('IndexFlow');
+  }
 
   useFireExperimentViewed({
     testName: TestName.earlyJourneyIndexesGuidance,
--- a/packages/compass-telemetry/src/telemetry-events.ts
+++ b/packages/compass-telemetry/src/telemetry-events.ts
@@ -2678,6 +2678,14 @@ type ExperimentViewedEvent = CommonEvent<{
   payload: { test_name: string };
 }>;
 
+type CreateIndexButtonClickedEvent = CommonEvent<{
+  name: 'Create Index Button Clicked';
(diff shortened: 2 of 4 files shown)


packages/compass-telemetry/src/telemetry-events.ts
```diff
@@ -2678,6 +2678,14 @@ type ExperimentViewedEvent = CommonEvent<{
   payload: { test_name: string };
 }>;
 
+type CreateIndexButtonClickedEvent = CommonEvent<{
+  name: 'Create Index Button Clicked';
+  payload: {
+    flow: 'Start with Query' | 'Start with Index' | undefined;
```
Anemy, line 2684: We already have a `Index Created` event, can we add this flow field to that and avoid adding a new event?

packages/compass-telemetry/src/telemetry-events.ts
```diff
@@ -2678,6 +2678,14 @@ type ExperimentViewedEvent = CommonEvent<{
...
+    context: string;
```
Anemy, line 2685: Any reason we don't want this as an enum? `'Create Index Modal'` and then we can add more later when we have them.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
```diff
@@ -83,6 +83,10 @@ function CreateIndexModal({
   const enableInIndexesGuidanceExp = usePreference('enableIndexesGuidanceExp');
   const showIndexesGuidanceVariant =
     usePreference('showIndexesGuidanceVariant') && enableInIndexesGuidanceExp;
+  if (showIndexesGuidanceVariant && !currentTab) {
+    // If user is in the variant and there has not been a currentTab set, default it to IndexFlow
+    props.onTabClick('IndexFlow');
```
Anemy, line 88: I think we should avoid setting something like this in the react flow. We're only doing this for telemetry purposes right? Can we instead set that telemetry event metadata based on the preferences? And then we can go back to always having the currentTab set.

packages/compass-indexes/src/components/create-index-modal/create-index-modal.tsx
```diff
@@ -83,6 +83,10 @@ function CreateIndexModal({
...
+  }
```
gribnoysup, line 89: Dispatching actions from a render method is an antipattern, so is dispatching actions to set initial state. If you need your state to have tree distinct values, you should be using redux ability to set initial state to define it (this is already done for the indexes plugin, you can see some initial state being passed to the `createStore` method in `activateIndexesPlugin`). Looking at the logic though, I think you only need to check the same preference flag when tracking the event instead of introducing a third state to the `currentTab` value and needing to account for this change everywhere

packages/compass-telemetry/src/telemetry-events.ts
rubydong: My team's analyst wants to track the clicks regardless of whether the creation was successful or not so unfortunately I can't combine it

Anemy: There is a separate `Index Create Failed` event that happens when the creation wasn't successful, so that can be used together with the `Index Created` to capture all of the clicks

rubydong: from analyst: image

gribnoysup (approved):

Anemy (approved):

PR #6883, opened 2025-04-28
feat: Add covered and optimal query examples CLOUDP-311782

rubydong:
## Description

https://jira.mongodb.org/browse/CLOUDP-311782

Best viewed with 'hide whitespace' 

Added a tooltip for when the type is not supported:
image

Default state:
image

One field:
image

Two fields:
image

Three fields:
image

### Checklist

## Motivation and Context

- [ ] Bugfix
- [x] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-indexes/src/components/create-index-form/index-flow-section.tsx
+++ b/packages/compass-indexes/src/components/create-index-form/index-flow-section.tsx
@@ -10,6 +10,7 @@ import {
   Toggle,
   fontFamilies,
   InfoSprinkle,
+  Tooltip,
 } from '@mongodb-js/compass-components';
 import React, { useState } from 'react';
 import type { Field } from '../../modules/create-index';
@@ -81,15 +82,17 @@ export type IndexFlowSectionProps = {
 const generateCoveredQueries = (
   coveredQueriesArr: Array<Record<string, number>>
 ) => {
-  const examples = [];
-  let i = 0;
-  while (i < coveredQueriesArr.length) {
+  const rows = [];
+  for (let i = 0; i < coveredQueriesArr.length; i++) {
     const currentRow = Object.assign({}, ...coveredQueriesArr.slice(0, i + 1));
-    examples.push(currentRow);
-    i++;
+    rows.push(
+      <>
+        {JSON.stringify(currentRow)} <br />
+      </>
+    );
   }
 
-  return examples;
+  return <>{rows}</>;
 };
 
 const generateOptimalQueries = (
@@ -99,7 +102,7 @@ const generateOptimalQueries = (
 
   // Do not show for 1 field or less
   if (numOfFields < 2) {
-    return;
+    return '';
   }
 
   const lastField = coveredQueriesArr[numOfFields - 1];
@@ -153,11 +156,12 @@ const IndexFlowSection = ({
     return field.name && field.type;
   });
 
+  const hasUnsupportedQueryTypes = fields.some((field) => {
+    return field.type === '2dsphere' || field.type === 'text';
+  });
+
   const isCoveredQueriesButtonDisabled =
-    !areAllFieldsFilledIn ||
-    fields.some((field) => {
-      return field.type === '2dsphere' || field.type === 'text';
-    });
+    !areAllFieldsFilledIn || hasUnsupportedQueryTypes;
 
   const indexNameTypeMap = fields.reduce<Record<string, string>>(
     (accumulator, currentValue) => {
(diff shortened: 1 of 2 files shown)


packages/compass-indexes/src/components/create-index-form/index-flow-section.tsx
```diff
@@ -107,6 +173,26 @@ const IndexFlowSection = ({
...
+  });
+
+  const [coveredQueriesExamples, setCoveredQueriesExamples] =
+    useState<JSX.Element>(<></>);
+  const [optimalQueriesExamples, setOptimalQueriesExamples] = useState<
+    string | JSX.Element
+  >('');
+  const [showCoveredQueries, setShowCoveredQueries] = useState(false);
+
+  const onCoveredQueriesButtonClick = () => {
```
Anemy, line 187: We should use `useCallback` here so that we don't re-create this method on every render. https://react.dev/reference/react/useCallback

Anemy: Looks like we could also be creating the `coveredQueriesArr` in this function instead of every render as well.

packages/compass-indexes/src/components/create-index-form/index-flow-section.tsx
```diff
@@ -107,6 +173,26 @@ const IndexFlowSection = ({
...
+    generateCoveredQueries(coveredQueriesArr);
+    generateOptimalQueries(coveredQueriesArr);
+
+    setCoveredQueriesExamples(generateCoveredQueries(coveredQueriesArr));
+    setOptimalQueriesExamples(generateOptimalQueries(coveredQueriesArr));
+    setShowCoveredQueries(true);
```
Anemy, line 193: Looks like we're only ever updating these three together. Should we instead have just one state object for the covered queries state?

I wonder if it might be good to have on the store's state instead of here in the component's state. That way if we ever move out of a modal we don't have to do any additional work to keep this around between renders. How do you feel? Having it on the store might make the state updating here a bit cleaner as we would set all three of their values in the reducer when a show covered queries action happens.

packages/compass-indexes/src/components/create-index-form/index-flow-section.tsx
```diff
@@ -78,6 +79,70 @@ export type IndexFlowSectionProps = {
   collectionName: string;
 };
 
+const generateCoveredQueries = (
+  coveredQueriesArr: Array<Record<string, number>>
+) => {
+  const rows = [];
+  for (let i = 0; i < coveredQueriesArr.length; i++) {
+    const currentRow = Object.assign({}, ...coveredQueriesArr.slice(0, i + 1));
```
Anemy, line 87: What is this object assign and slicing deconstruction do? I'm wondering if there's a simpler way to write this.

Anemy (commented):
Looking good, left a couple questions/comments

packages/compass-indexes/src/components/create-index-form/index-flow-section.tsx
rubydong: basically the current row takes the first thing in the array to the current index so each row has one more than the previous row but i'm not sure how to make it easier to read

packages/compass-indexes/src/components/create-index-form/index-flow-section.tsx
```diff
@@ -107,6 +173,26 @@ const IndexFlowSection = ({
...
+    useState<JSX.Element>(<></>);
+  const [optimalQueriesExamples, setOptimalQueriesExamples] = useState<
+    string | JSX.Element
+  >('');
+  const [showCoveredQueries, setShowCoveredQueries] = useState(false);
+
+  const onCoveredQueriesButtonClick = () => {
+    generateCoveredQueries(coveredQueriesArr);
+    generateOptimalQueries(coveredQueriesArr);
+
```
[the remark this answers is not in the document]
rubydong: they're not extra -- they're called when the button is clicked to get the latest queries

packages/compass-indexes/src/components/create-index-form/index-flow-section.tsx
rubydong: i could add a state object for the 3 things! originally i think i may have written to have these in the store but i think Sergey thinks that it can be derived so we kept it out out there

packages/compass-indexes/src/components/create-index-form/index-flow-section.tsx
```diff
@@ -173,25 +173,30 @@ const IndexFlowSection = ({
     {}
   );
 
-  const coveredQueriesArr = fields.map((field, index) => {
-    return { [field.name]: index + 1 };
+  const [coveredQueriesObj, setCoveredQueriesObj] = useState<{
```
Anemy, line 176: When the user changes the fields should we reset this so that they have to re-click the show covered queries button? That way we don't show them something out of date.

rubydong: https://jira.mongodb.org/browse/CLOUDP-311787 i have this ticket for polishing up all the UX for further disabled states, error states etc for everything across the modal

Anemy (approved):

PR #6894, opened 2025-05-05
feat(schema, data-modeling): use iterable cursor in schema analysis COMPASS-9150 COMPASS-9315

Anemy:
COMPASS-9150 COMPASS-9315

Marked for release notes as this could impact how folks analyze their schema, they should be able to abort more predictably now as we check for [signal abort between each document being analyzed](https://github.com/mongodb-js/mongodb-schema/blob/6f22cb72aaaf97a0bbbb6ae92a4fdadc4290db67/src/schema-analyzer.ts#L671).

A bit of additional context, we have COMPASS-8925 for moving the rest of Compass to pass the abort signal to the driver's methods and get rid of some of our session workarounds in the data-service. This passing of an abort signal to the driver's aggregate method with the `Abortable` type that we have in this pr is the first we're doing it in Compass.


packages/compass-schema/src/modules/schema-analysis.ts
```diff
@@ -72,7 +72,10 @@ export const analyzeSchema = async (
       : {
           signal: abortSignal,
         };
-    const schemaAccessor = await analyzeDocuments(docs, schemaParseOptions);
+    const schemaAccessor = await analyzeDocuments(
+      sampleCursor,
```
Anemy, line 76: This could impact performance. I have not tested that.

packages/compass-schema-validation/src/stores/store.spec.ts
```diff
@@ -42,7 +42,12 @@ const fakeDataService = {
       /* never resolves */
     }),
   isCancelError: () => false,
-  sample: () => [{ prop1: 'abc' }],
+  sampleCursor: () =>
+    ({
+      *[Symbol.asyncIterator]() {
+        yield* [{ prop1: 'abc' }];
+      },
```
nbbeeken, line 49:
```suggestion
      async *[Symbol.asyncIterator]() {
        yield* [{ prop1: 'abc' }];
      },
```

Just so that this is actually the correct type, "next()" returns a promise instead of `{value, done}` directly.

packages/compass-schema/src/modules/schema-analysis.spec.ts
```diff
@@ -81,13 +81,17 @@ describe('schema-analysis', function () {
...
       const dataService = {
-        sample: () =>
-          Promise.resolve([
-            { x: 1 },
-            { y: 2, __safeContent__: [bson.Binary.createFromBase64('aaaa')] },
-          ]),
-        isCancelError: () => false,
+        sampleCursor: () =>
+          ({
+            *[Symbol.asyncIterator]() {
```
nbbeeken, line 91:
```suggestion
            async *[Symbol.asyncIterator]() {
```

packages/compass-schema/src/modules/schema-analysis.spec.ts
```diff
@@ -179,10 +183,16 @@ describe('schema-analysis', function () {
 
     it('adds promoteValues: false so the analyzer can report more accurate types', async function () {
       const dataService = {
-        sample: () => Promise.resolve([]),
-        isCancelError: () => false,
+        sampleCursor: () =>
+          ({
+            *[Symbol.asyncIterator]() {
```
nbbeeken, line 188:
```suggestion
            async *[Symbol.asyncIterator]() {
```

packages/compass-schema/src/modules/schema-analysis.spec.ts
```diff
@@ -199,19 +209,30 @@ describe('schema-analysis', function () {
...
-        isCancelError: () => true,
-      };
-
       const abortController = new AbortController();
       const abortSignal = abortController.signal;
 
+      const dataService = {
+        sampleCursor: () =>
+          ({
+            *[Symbol.asyncIterator]() {
```
nbbeeken, line 224:
```suggestion
            async *[Symbol.asyncIterator]() {
```

packages/compass-schema/src/modules/schema-analysis.spec.ts
```diff
@@ -228,13 +249,18 @@ describe('schema-analysis', function () {
...
+      } = new Error('pineapple');
       error.name = 'MongoError';
       error.code = new bson.Int32(1000);
 
       const dataService = {
-        sample: () => Promise.reject(error),
-        isCancelError: () => false,
+        sampleCursor: () =>
+          ({
+            *[Symbol.asyncIterator]() {
```
nbbeeken, line 259:
```suggestion
            async *[Symbol.asyncIterator]() {
```

packages/compass-schema/src/stores/store.spec.ts
```diff
@@ -30,26 +30,31 @@ const mockQueryBar = {
   },
 };
 
+function createMockCursor(data: any[]) {
+  return {
+    *[Symbol.asyncIterator]() {
```
nbbeeken, line 35:
```suggestion
    async *[Symbol.asyncIterator]() {
```

packages/compass-schema/src/modules/schema-analysis.ts
```diff
@@ -81,8 +84,11 @@ export const analyzeSchema = async (
     log.error(mongoLogId(1001000091), 'Schema', 'Schema analysis failed', {
       ns,
       error: err.message,
+      aborted: abortSignal.aborted,
     });
-    if (dataService.isCancelError(err)) {
+
+    if (abortSignal.aborted) {
+      // The operation was aborted, so we don't throw an error.
```
nbbeeken, line 91: Do we ever give abort() a `reason`? Should it be logged/captured here either way?

packages/compass-schema/src/modules/schema-analysis.ts
nbbeeken: Any specific concerns or predictions for where the biggest impact might be? I def defer to another team member to give you an answer here, but as a newbie what would it entail to validate at least the biggest bottlenecks we think may come up here?

nbbeeken (changes requested):

packages/compass-schema/src/modules/schema-analysis.ts
Anemy: Added to the log! We typically do pass a reason, not sure if it is all that actionable or useful, but maybe down the line it will be.

packages/compass-schema/src/modules/schema-analysis.ts
```diff
@@ -85,6 +85,7 @@ export const analyzeSchema = async (
       ns,
       error: err.message,
       aborted: abortSignal.aborted,
+      ...(abortSignal.aborted ? { abortReason: abortSignal.reason } : {}),
```
nbbeeken, line 88: Quite often, in my experience, a `signal.reason` is set to an Error to trace to the cause of the interruption, do we need to pull the instance apart to actually see the message here? (same as `err`?)

Anemy: Updated!

nbbeeken (approved):

packages/compass-schema/src/modules/schema-analysis.ts
nbbeeken: Sorry didn't get back sooner. I agree, we've shifted the cost but in a way that should leave the app a bit more responsive as it acts on a document at a time. Plus this takes advantage of the driver's lazy deserialize. In the end the time spent should be the equivalent, broken up over more event loop cycles

PR #6896, opened 2025-05-06
fix: Make create form keyboard accessible COMPASS-9335

paula-stacho:
## Description
Changes:
- form steps can be submitted by enter
- improved tab navigation
- refactored SelectTable to SelectList. the table wasn't very accessible and there were no usages where we actually had more than one column (it is only used here and in connections import/export). so there was no reason for it to be a table instead of a list of classic checkboxes with labels.

### Checklist

## Motivation and Context

- [ ] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*


packages/compass-components/src/components/select-table.tsx
```diff
@@ -55,51 +77,40 @@ export function SelectTable<T extends SelectItem>(
...
+            key={`select-table-item-${item.id}`}
+            data-testid={`select-table-item-${item.id}`}
+          >
+            <Checkbox
+              className={checkboxStyles}
+              key={`select-${item.id}`}
+              name={`select-${item.id}`}
+              data-testid={`select-${item.id}`}
+              label={item[label[0]]}
+              aria-labelledby="select-table-item-label" // I don't know why LG requires this, it seems ignored
```
Anemy, line 106: This should reference another element that labels this checkbox right? Should we go back to the `aria-label`?

paula-stacho: Ah! I tried to only include aria-label before but turns out it wasn't recognized and the labelledby was still required because the type of `item[label[0]]` isn't clear. this made me realise that actually it sometimes (in case of import) isn't a string but an element, so it makes sense to take a separate key for aria-label.
Unfortunately it's still not clear for TS hence the typecasting as string. I did make an attempt to type it so that a) either `item[label.displayLabelKey]` is a string, or b) `label.ariaLabelKey` is to be required and `item[label.ariaLabelKey]` be a string. But I couldn't get it working and it was getting overly complicated 🙈

packages/compass-components/src/components/select-list.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+export function SelectList<T extends SelectItem>(
+  props: SelectListProps<T>
+): React.ReactElement {
+  const { items, label, disabled, onChange } = props;
+
+  const isDarkMode = useDarkMode();
+  const evenRowStyles = isDarkMode
+    ? css({ backgroundColor: palette.gray.dark3 })
+    : css({ backgroundColor: palette.gray.light3 });
```
Anemy, line 57: We try to avoid calling `css(...)` in the react and instead use variables assigned at the start. Can we make a styles const for these?

Anemy (approved):
Looks good, one css usage we should move to a const at the top.

packages/compass-components/src/components/select-list.tsx
paula-stacho: Sure! Can I ask why? Is it to keep the render method leaner?

Anemy (approved):

PR #6907, opened 2025-05-08
fix(web): Provide an actionable error message on DE errors CLOUDP-313461

nvs119:
## Description

Matching on internal code patterns (CAPS_UNDERSCORE_WITH_COLON:) and HTTP codes 400-599 anywhere in the string. Wanted to make this more general but the Indexes page was where I saw the 403 error from filter, so followed it up to the regular-indexes modules. 

Screenshot 2025-05-08 at 12 01 03 PM

error without sanitization, in console:
Screenshot 2025-05-08 at 12 08 56 PM

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*

### DIFF
--- a/packages/compass-indexes/src/modules/regular-indexes.ts
+++ b/packages/compass-indexes/src/modules/regular-indexes.ts
@@ -187,7 +187,7 @@ function processError(error: string): string {
   const httpCodePattern = /\b(4\d{2}|5\d{2})\b/; // Matches HTTP codes 400-599 anywhere in the string
 
   if (internalCodePattern.test(error) || httpCodePattern.test(error)) {
-    return 'Sorry, an unexpected error has occurred. Try refreshing the page or reporting the issue if it persists.';
+    return "We're sorry, an unexpected error has occurred. Please try again.";
   }
 
   return error; // Return original error if it doesn't match the patterns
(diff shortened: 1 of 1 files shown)


packages/compass-indexes/src/modules/regular-indexes.ts
```diff
@@ -182,6 +182,17 @@ export const INITIAL_STATE: State = {
   error: undefined,
 };
 
+function processError(error: string): string {
+  const internalCodePattern = /[A-Z_]+:/; // Matches all caps with underscores followed by a colon anywhere in the string
+  const httpCodePattern = /\b(4\d{2}|5\d{2})\b/; // Matches HTTP codes 400-599 anywhere in the string
+
+  if (internalCodePattern.test(error) || httpCodePattern.test(error)) {
+    return 'Sorry, an unexpected error has occurred. Try refreshing the page or reporting the issue if it persists.';
```
johnjackweir, line 190: Nitpick but let's keep it as consistent with ye olde DE as possible

"We're sorry, an unexpected error has occurred. Please try again."

johnjackweir (approved):
Generally we should have a unit test, especially for tricky things like regexes

I'm ok with this as is while we're trying to get MDE out the door

PR #6909, opened 2025-05-09
chore: wait for text to include the expected result

lerouxb:
I saw some flakes in the nightly. Not sure if all of these are strictly necessary since maybe in some cases the moment the element appears the text is immediately stable. But I don't think this will hurt. Also not 100% sure if these are the only affected cases.

Example failure: https://spruce.mongodb.com/task/10gen_compass_main_test_packaged_app_macos_14_arm_test_packaged_app_3_681d5391e27c11000732a3de_25_05_09_01_00_01?execution=0&sortBy=STATUS&sortDir=ASC


paula-stacho (approved):
Thank you! Let's try it out

